PR #8, opened 2020-06-07
Feature: scheduled runs

samkovaly:
Code for scheduled runs API endpoint namespace_scheduled_runs
along with changes to namespace_runs to account for the new schema module.
Some bad UI to show the new api working.

Also slight changes to orchest-ctl to include docker binds for orchest-api and celery-worker
new dev files and changes to previous dev scripts to account for poor windows developers


orchest.sh
```diff
@@ -6,4 +6,5 @@ HOST_USER_DIR=$PWD/orchest/userdir/
 # create config dir if it doesn't exist
 mkdir -p $HOST_CONFIG_DIR
 
-docker run -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=$HOST_CONFIG_DIR -e HOST_PWD=$PWD -e HOST_USER_DIR=$HOST_USER_DIR orchestsoftware/orchest-ctl "$@"
\ No newline at end of file
+docker run --name orchest-ctl -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=$HOST_CONFIG_DIR -e HOST_PWD=$PWD -e HOST_USER_DIR=$HOST_USER_DIR orchestsoftware/orchest-ctl "$@"
```
yannickperrenet, line 9: Is the `orchest-ctl` container running on the `orchest` Docker network?

orchest/orchest-api/app/app/apis/namespace_runs.py
```diff
@@ -79,6 +24,14 @@
 })
 
 
+api.models[step_status.name] = step_status
+api.models[status_update.name] = status_update
+api.models[run_configuration.name] = run_configuration
+api.models[run.name] = run
+
+
```
yannickperrenet, line 32: Please review all the white lines in this file according to [PEP 8](https://www.python.org/dev/peps/pep-0008/)

orchest/orchest-api/app/app/apis/namespace_runs.py
```diff
@@ -91,13 +44,17 @@ def get(self):
...
     @api.doc('start_run')
     @api.expect(run_configuration)
     @api.marshal_with(run, code=201, description='Run started')
     def post(self):
         """Start a new run."""
         post_data = request.get_json()
 
+
+        post_data['run_config']['run_endpoint'] = RUN_ENDPOINT
```
yannickperrenet, line 56: I like the general solution of writing `run_endpoint` to the `run_config` to resolve whether `scheduled-runs` or `runs` was used. However, using the extra `RUN_ENDPOINT` seems overkill.

`post_data['run_config']['run_endpoint'] = 'runs'`

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,253 @@
...
+        post_data = request.get_json()
+
+        # https://github.com/celery/celery/issues/1813
+        # could use https://docs.celeryproject.org/en/latest/userguide/signals.html#before-task-publish
+
+
+
+        # TODO: make DB persistent (on file, not state)
+
+        task_id = uuid()
```
yannickperrenet, line 80: We want celery to manage the creation of `UUID`s for us. You can just obtain the id from the Celery task.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,253 @@
...
+        required=True,
+        description='Time at which the run is scheduled to start'),
+})
+
+
+scheduled_runs = api.model('Scheduled Runs', {
+    'scheduled_runs': fields.List(
+        fields.Nested(scheduled_run),
+        description='past, present and running scheduled_runs')
+})
```
yannickperrenet, line 49: Why are these models not also in the `schema.py`?

orchest/orchest-ctl/app/main.py
```diff
@@ -51,7 +51,11 @@ def slash_sub(path):
                 "source": "/var/run/docker.sock",
                 "target": "/var/run/docker.sock"
             },
-        ]
+        ],
+        "ports": {
+            "80/tcp": 9000
+        },
+
```
yannickperrenet, line 62: These changes are to make for your own testing and development, but they should not be pushed to the `master` branch since that alters the version that users are running on their machines.

orchest/orchest-api/app/app/core/runners.py
```diff
@@ -10,6 +10,9 @@
 from config import CONFIG_CLASS
 
 
+# TODO: create_app is called twice, meaning create_all (create databases) is 
+# called twice, which means celery-worker needs the /userdir bind to access the DB
+# which is probably not a good idea. create_all should only be called once per app right?
```
yannickperrenet, line 15: They both run inside different containers, the `celery-worker` and `orchest-api`. `create_app` is called instead of using the Flask's `current_app` context to avoid circular imports. The `celery-worker` needs an app context and thus `create_app` is called here.

orchest/orchest-ctl/app/main.py
```diff
@@ -85,11 +93,15 @@ def slash_sub(path):
             {
                 "source": "/var/run/docker.sock",
                 "target": "/var/run/docker.sock"
-            }
+            },
+            {
```
ricklamers, line 97: Why are we mounting HOST_USER_DIR in the proxy?

orchest.bat
```diff
@@ -6,4 +6,5 @@ set HOST_USER_DIR=%cd%\orchest\userdir\
 REM create config dir if it doesn't exist
 if not exist %HOST_CONFIG_DIR% mkdir %HOST_CONFIG_DIR%
 
-docker run -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=%HOST_CONFIG_DIR% -e HOST_PWD=%cd% -e HOST_USER_DIR=%HOST_USER_DIR% orchestsoftware/orchest-ctl %*
\ No newline at end of file
+docker run --name orchest-ctl -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=%HOST_CONFIG_DIR% -e HOST_PWD=%cd% -e HOST_USER_DIR=%HOST_USER_DIR% orchestsoftware/orchest-ctl %*
+docker rm orchest-ctl
```
yannickperrenet: Comments for `orchest.sh` apply here as well.

orchest.sh
```diff
@@ -6,4 +6,5 @@ HOST_USER_DIR=$PWD/orchest/userdir/
...
+docker rm orchest-ctl
```
yannickperrenet: Would love to hear you thoughts on why you added the `--name orchest-ctl` and added the `docker rm` here. In my opinion it is better to go with the old approach, because then the user is left with "history" and can clean it after if desired.

orchest/orchest-api/app/app/apis/namespace_launches.py
```diff
@@ -13,73 +13,18 @@
 from app.core.managers import JupyterDockerManager
 import app.models as models
 
+from app.schema import server, launch, launches, pipeline
+
```
yannickperrenet, line 17: PEP8

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,253 @@
...
+import app.models as models
+from app.celery_app import make_celery
+from app.utils import construct_pipeline
+
+from shutil import copytree, ignore_patterns
+import os
+
+from celery import uuid
+
+from app.schema import step_status, status_update, run_configuration, run
```
yannickperrenet, line 17: PEP8

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,253 @@
...
+        # could use https://docs.celeryproject.org/en/latest/userguide/signals.html#before-task-publish
+
+
+
+        # TODO: make DB persistent (on file, not state)
+
+        task_id = uuid()
+
+        # post_data['run_config']['pipeline_dir'] = 
+        #   /host_mnt/c/Users/ASUS/Documents/Orchest.io/orchest/orchest/userdir/pipelines/9ac1246c-b607-4ee3-b0af-bfa92f9299e6
```
yannickperrenet, line 83: Comments like these are fine for development and on a feature branch, but not to merge into master.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,253 @@
...
+        if not os.path.exists(scheduled_runs_path):
+            os.mkdir(scheduled_runs_path)
+
+        scheduled_run_pipeline_dir = os.path.join(scheduled_runs_path, task_id)
+        copytree(pipeline_dir, scheduled_run_pipeline_dir, ignore=ignore_patterns(scheduled_runs_dir))
+
+
+        # second, prepare new mount directory for runnable-images to run
+        scheduled_run_dir = os.path.join(scheduled_runs_dir, task_id)
+        pipeline_dir_mount = os.path.join(post_data['run_config']['pipeline_dir'], scheduled_run_dir)
```
yannickperrenet, line 100: Might be a little premature, but I think later this would be great to put into a function so it can be reused by experiments.

orchest/orchest-api/app/config.py
```diff
@@ -14,8 +26,28 @@ class Config:
...
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#task-track-started
+
+    # task_publish_retry
+    #   True by default
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#task-publish-retry
+
+    # task_acks_late
+    #   a configuration to keep an eye on. Could be useful
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#task-acks-late
+    #   https://stackoverflow.com/questions/54415515/celery-active-tasks-persistence
```
yannickperrenet, line 50: Note sure whether we want this option https://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late

orchest/orchest-api/app/config.py
```diff
@@ -14,8 +26,28 @@ class Config:
...
     broker_url = 'amqp://guest:guest@rabbitmq-server:5672//'
     imports = ('app.core.runners',)
+
+    # result_persistent = True
+    #   This is message result persistency, not queue persistency
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#rpc-backend-settings
+
     # result_backend = 'rpc://'
+    #   use with our scheduled_runs.db?
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#task-result-backend-settings
```
yannickperrenet, line 36: Why?

orchest/orchest-api/app/config.py
```diff
@@ -14,8 +26,28 @@ class Config:
     # config class will be loaded directly by the Celery instance.
     broker_url = 'amqp://guest:guest@rabbitmq-server:5672//'
     imports = ('app.core.runners',)
+
+    # result_persistent = True
+    #   This is message result persistency, not queue persistency
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#rpc-backend-settings
```
yannickperrenet, line 32: For now I don't think we need result persistency, or why do you think we do?

orchest/orchest-api/app/app/models.py
```diff
@@ -16,8 +16,10 @@ def __repr__(self):
         return f'<Launch {self.pipeline_uuid}>'
 
 
+
 class Run(BaseModel, db.Model):
     __tablename__ = 'runs'
+    __bind_key__ = "runs_db"
```
yannickperrenet, line 22: PEP8, go with `'runs_db'` because single quotes where already used in this file.

orchest/orchest-api/app/config.py
```diff
@@ -2,7 +2,19 @@ class Config:
     DEBUG = False
     TESTING = False
 
+    # must be uppercase
+    # https://flask-appbuilder.readthedocs.io/en/latest/multipledbs.html
     SQLALCHEMY_DATABASE_URI = 'sqlite:///resources.db'
+    SQLALCHEMY_BINDS = {
+        'runs_db': SQLALCHEMY_DATABASE_URI,
```
yannickperrenet, line 9: I don't think the `resources.db` should be repeated here. Using `SQLALCHEMY_DATABASE_URI` already sets it as default and therefore the db can be used without specifying a `__bind_key__` for the models. 

At least that is what I deduce from the link in the comment above.

orchest/orchest-api/app/app/models.py
```diff
@@ -56,3 +61,60 @@ class StepStatus(BaseModel, db.Model):
...
         return f'<StepStatus {self.run_uuid}.{self.step_uuid}>'
+
+
+
+
+
+# TODO: make DRY, need to:
+#   have ScheduledRun inherit from Run so we don't repeat fields but this inheritance
+#   must be implemented so that ScheduledRun is it's own table and does not refer to Run in any way
+#   since ScheduledRun is on it's own database
```
yannickperrenet, line 72: Agreed. Inheritance is hard to do in this case?

orchest/orchest-webserver/Dockerfile
```diff
@@ -27,6 +27,17 @@ RUN npx webpack
 
 # Compile style
 WORKDIR /app/static/css/
+
+# Windows needs this to run the script correctly
+# https://stackoverflow.com/questions/37419042/container-command-start-sh-not-found-or-does-not-exist-entrypoint-to-contain    
+#RUN sed -i 's/\r$//' ./compile-sass.sh  && \
+#chmod +x ./compile-sass.sh
```
yannickperrenet, line 34: I think it is indeed good to keep it here as comments. Even though all automatics builds will be happening on Linux machines.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
ricklamers: We should think about whether the fact that this is blocking should be changed. In theory data sources should alleviate the need for large files in the pipeline directory.

orchest.sh
yannickperrenet: We decided not to have it run on the `orchest` Docker network, because there is no direct reason to. It is not interacting over the network with any of the other containers.

orchest.sh
yannickperrenet: We decided to keep the history clean for the `orchest-ctl`. Instead of `docker rm orchest-ctl` we will add the `--rm` option to the `docker run` command.

orchest/orchest-api/app/config.py
```diff
@@ -2,7 +2,19 @@ class Config:
...
+        # /userdir works because it's mounted via docker
+        'scheduled_runs_db': 'sqlite:////userdir/persistent.db',
```
yannickperrenet, line 11: To fix the persistence issue that raises an error for the `celery-worker`, I think you can just catch the error with a `try except` block inside [create_app](https://github.com/orchest/orchest/blob/4cdb1e34bd6f48841673f294e06d41761c42ee5e/orchest/orchest-api/app/app/__init__.py#L17)

ricklamers: Not sure if I'm keen on having a deliberate failing code block be part of the codebase. That smells like bad design. I'll help review the Flask + Celery setup if necessary.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Well 'run_config' needs 'pipeline_dir' in order to work and celery.send_task needs 'run_config' in its celery_job_kwargs. pipeline_dir needs the celery task id (userdir/scheduled_runs/{pipeline_uuid}/{task_uuid}/) so I need to define the task_id early as opposed to waiting for send_task to return it.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: I was thinking that those schema are specific to namespace_scheduled_runs and no other file, but maybe its better to put those in schema anyway like all the other schema.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: I think a function is a good idea, maybe copy_pipeline_to_dir(....) or something. For this function, would it be in utils.py like construct_pipeline() is?

orchest/orchest-api/app/config.py
samkovaly: Just was thinking about it but we probably don't need it.

orchest/orchest-api/app/config.py
samkovaly: Yeah for now we don't but it could be useful in the future.

orchest/orchest-api/app/app/models.py
samkovaly: Yes I looked at the 3 types of sqlalchemy inheritance but all seemed to not do what we want. https://docs.sqlalchemy.org/en/13/orm/inheritance.html

orchest/orchest-api/app/config.py
yannickperrenet: @ricklamers and I agreed that this would be a temporary fix until we separate the `celery-worker` code from the `orchest-api` code.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
yannickperrenet: I would move them so it is always clear where to look for models.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
yannickperrenet: You can make a decision here. It can be in `utils.py`, inside the namespace module of the API, whatever you seem fit (with good reasoning of course)

samkovaly: I think its better to wait, its hard to see where the generalization is without knowing what experiments does.

orchest.bat
```diff
@@ -6,4 +6,4 @@ set HOST_USER_DIR=%cd%\orchest\userdir\
 REM create config dir if it doesn't exist
 if not exist %HOST_CONFIG_DIR% mkdir %HOST_CONFIG_DIR%
 
-docker run -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=%HOST_CONFIG_DIR% -e HOST_PWD=%cd% -e HOST_USER_DIR=%HOST_USER_DIR% orchestsoftware/orchest-ctl %*
\ No newline at end of file
+docker run --name orchest-ctl --rm --mount type=bind,source=%HOST_USER_DIR%,target=/userdir -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=%HOST_CONFIG_DIR% -e HOST_PWD=%cd% -e HOST_USER_DIR=%HOST_USER_DIR% orchestsoftware/orchest-ctl %*
```
yannickperrenet, line 9: What is the additional mount for?

orchest.sh
```diff
@@ -1,9 +1,9 @@
...
\ No newline at end of file
+#!/bin/bash
+
+HOST_CONFIG_DIR=$HOME/.config/orchest
+HOST_USER_DIR=$PWD/orchest/userdir/
+
+# create config dir if it doesn't exist
+mkdir -p $HOST_CONFIG_DIR
+
+docker run --name orchest-ctl --rm --mount type=bind,source=$HOST_USER_DIR,target=/userdir -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=$HOST_CONFIG_DIR -e HOST_PWD=$PWD -e HOST_USER_DIR=$HOST_USER_DIR orchestsoftware/orchest-ctl "$@"
```
yannickperrenet, line 9: Again, what is the additional mount for?

orchest/orchest-api/app/app/__init__.py
```diff
@@ -1,33 +1,34 @@
...
+
+    # Cross-origin resource sharing. Allow API to be requested from the
+    # different microservices such as the webserver.
+    CORS(app, resources={r'/*': {'origins': '*'}})
+
+    # Initialize the database and create the database file.
+    db.init_app(app)
+    with app.app_context():
+        if use_db:
+            db.create_all()
```
yannickperrenet, line 29: I like adding `use_db` but why did you not put the entire block inside the if-statement instead of just the database creation?

orchest/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -1,31 +1,25 @@
...
-@api.response(404, 'Experiment not found')
-class Experiment(Resource):
-    @api.doc('get_experiment')
-    def get(self, uuid):
-        """Fetch an experiment given its UUID."""
-        pass
+from flask_restplus import Namespace, Resource, fields
+from app.schema import experiment
+
+api = Namespace('experiments', description='Managing experiments')
```
yannickperrenet, line 4: Not completely PEP8.

But aren't you forgetting the `api.models[experiment.name] = experiment`?

orchest/orchest-api/app/app/apis/namespace_launches.py
```diff
@@ -1,193 +1,133 @@
...
-        response = jdm.shutdown_pipeline(pipeline_uuid)
-
-        if response is not None:
-            api.abort(404)
-
-        db.session.delete(launch)
-        db.session.commit()
-
-        return {'message': 'Server shutdown was successful'}, 200
+import logging
```
yannickperrenet, line 1: Just as a general comment.

Why does it say that more or less the entire file is changed, instead of only the lines that changed? This makes it rather hard to see the exact changes.

orchest/orchest-api/app/app/apis/namespace_launches.py
```diff
@@ -1,193 +1,133 @@
...
+        launch = {
+            'pipeline_uuid': post_data['pipeline_uuid'],
+            'server_ip': IP.server,
+            'server_info': r.json()
+        }
+        db.session.add(models.Launch(**launch))
+        db.session.commit()
+
+        return launch, 201
+
```
yannickperrenet, line 87: Should be extra space according to PEP8

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # https://github.com/celery/celery/issues/1813
+        # could use https://docs.celeryproject.org/en/latest/userguide/signals.html#before-task-publish
+
+        # TODO: remove bad UI
+        # TODO: figure out how rabbitmq/celery persists while in a docker container
+        # TODO figure out how not to call db.create_al() twice (from create_app twice)
+
+        pipeline_uuid = post_data['pipeline_description']['uuid']
+        run_uuid = uuid()
```
yannickperrenet, line 53: As discussed, please add comment as to why you need to create the UUID here yourself.

orchest/orchest-api/app/app/utils.py
```diff
@@ -1,583 +1,595 @@
-import asyncio
```
yannickperrenet, line 1: Please fix the line termination. I cannot possibly inspect this entire file. Additionally, this is confusing to review changes in the future.

orchest/orchest-api/app/config.py
```diff
@@ -1,37 +1,50 @@
...
+    # config class will be loaded directly by the Celery instance.
+    broker_url = 'amqp://guest:guest@rabbitmq-server:5672//'
+    imports = ('app.core.runners',)
+
+    # TODO: consider these
+    # task_acks_late
+    #   https://docs.celeryproject.org/en/stable/userguide/configuration.html#task-acks-late
+    #   https://stackoverflow.com/questions/54415515/celery-active-tasks-persistence
+    # task_track_started
+    #   tracks 'STARTED' state,
```
yannickperrenet, line 30: I remember discussed this option. Don't think we want it, because we manage the states ourselves anyways.

orchest/orchest-api/app/config.py
```diff
@@ -1,37 +1,50 @@
...
+
+    # ---- Celery configurations ----
+    # NOTE: the configurations have to be lowercase.
+    # NOTE: Flask will not configure lowercase variables. Therefore the
+    # config class will be loaded directly by the Celery instance.
+    broker_url = 'amqp://guest:guest@rabbitmq-server:5672//'
+    imports = ('app.core.runners',)
+
+    # TODO: consider these
+    # task_acks_late
```
yannickperrenet, line 26: Don't see why we would need this. Why do you think we do?

orchest/orchest-api/app/main.py
```diff
@@ -1,9 +1,9 @@
...
-app = create_app(config_class=CONFIG_CLASS)
-
-
-if __name__ == '__main__':
-    app.run(host='0.0.0.0')
+from app import create_app
+from config import CONFIG_CLASS
+
+
+app = create_app(config_class=CONFIG_CLASS, use_db = True)
```
yannickperrenet, line 5: `use_db=True` is the default, so no need to specify it here. Also if you still want to explicitly state, then no spaces around the `=` operator according to PEP8.

orchest/orchest-webserver/app/main.py
```diff
@@ -1,6 +1,6 @@
...
-app = create_app()
-
-if __name__ == '__main__':
-    app.run(host='0.0.0.0', port=80)
+from app import create_app
+
+app = create_app()
+
+if __name__ == '__main__':
+    app.run(host='0.0.0.0', port=80)
```
yannickperrenet, line 6: Please fix line termination. I don't see a single change here.

orchest.bat
samkovaly: in orchest-ctl main.py, rabbitmq has a new mount:
        "mounts": [
            {
                "source": os.path.join(HOST_USER_DIR, DURABLE_QUEUES_DIR),
                "target": "/var/lib/rabbitmq/mnesia",
            }
which is so that rabbittmq can save it's queue to '/userdir/.durable_queues' during scheduled runs.
In order to mount this, '.durable_queues' must already exist or else an error is thrown. So we can either  manually create .durable_queues and upload it to github or the orchest-ctl must create this folder dynamically if it's not created yet. In order to create a folder, the orchest-ctl must have a mount of it's own connected to 'userdir'. I can remove this code and manually create the folder if that's a better option.

orchest/orchest-ctl/app/main.py
```diff
@@ -125,6 +129,11 @@ def slash_sub(path):
 def start():
     logging.info("Stub: starting Orchest")
 
+    # need '.durable_queues' to put the rabbitmq queue backups into while tasks are scheduled
+    durable_queues_path = os.path.join('/userdir', DURABLE_QUEUES_DIR)
```
samkovaly, line 133: This is why the orchest-ctl container needs to mount to /userdir, so that it can create the durable queues folder here.

orchest/orchest-webserver/Dockerfile
```diff
@@ -28,16 +28,6 @@ RUN npx webpack
 # Compile style
 WORKDIR /app/static/css/
 
-# Windows needs this to run the script correctly
-# https://stackoverflow.com/questions/37419042/container-command-start-sh-not-found-or-does-not-exist-entrypoint-to-contain    
-#RUN sed -i 's/\r$//' ./compile-sass.sh  && \
-#chmod +x ./compile-sass.sh
```
samkovaly: no longer need these. Instead those .sh scripts need LF line endings on the developer's machine. Instead, developers need the global setting "git config --global core.autocrlf input" or * text eol=lf in the .gitattributes.

orchest.bat
yannickperrenet: I guess your current approach is more in line with how we are doing it for other containers. 

Could you split this one long line over multiple lines using `\` and additionally add a comment as to why the path is mounted. This way we don't forget in the future.

orchest.sh
yannickperrenet: Same as for the `.bat` file. Could make it into multi-line and add comment for the mount?

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,221 @@
...
+from flask_restplus import Namespace, Resource, fields
+
+from app.connections import db
+import app.models as models
+from app.celery_app import make_celery
+from app.utils import construct_pipeline
+from app.schema import step_status, status_update, scheduled_run_configuration, scheduled_run, scheduled_runs
+
+import os
+from shutil import copytree
```
yannickperrenet, line 15: Could you add these to imports to the top, so they are in the same paragraph as the `from datetime...` This is according to PEP8 and makes it easier to read and find imports used.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,221 @@
...
+        # TODO: possible set status of steps and Run to "REVOKED"
+
+        # Stop the run, whether it is in the queue or whether it is
+        # actually running.
+
+        # TODO: https://stackoverflow.com/questions/39191238/revoke-a-task-from-celery
+        # TODO: delete new pipeline files that were created for this specific run?
+        revoke(run_uuid, terminate=True)
+
+        return {'message': 'Run termination was successful'}, 200
```
yannickperrenet, line 171: Have you tested this endpoint?

Since we are managing the step statuses ourselves, I think we would need to but here that they are aborted or some other status. Otherwise they will most likely be kept at `PENDING` which makes the user believe they will still execute in the future.

orchest/orchest-api/app/app/utils.py
```diff
@@ -100,13 +101,18 @@ async def update_status(status: str,
     elif data['status'] in ['SUCCESS', 'FAILURE']:
         data['ended_time'] = datetime.utcnow().isoformat()
 
-    base_url = f'{CONFIG_CLASS.ORCHEST_API_ADDRESS}/runs/{task_id}'
+    base_url = f'{CONFIG_CLASS.ORCHEST_API_ADDRESS}/{run_endpoint}/{task_id}'
+
+
+
+
```
yannickperrenet, line 108: Could you remove these unnecessary white lines?

orchest/orchest-ctl/app/main.py
```diff
@@ -51,7 +55,12 @@ def slash_sub(path):
                 "source": "/var/run/docker.sock",
                 "target": "/var/run/docker.sock"
             },
-        ]
+            {
+                "source": HOST_USER_DIR,
+                "target": "/userdir"
+            },
```
yannickperrenet, line 61: Could you add a comment to explain why we need this mount? Because if we move the copying to a celery task, then this mount would no longer be needed.

orchest/orchest-webserver/Dockerfile
yannickperrenet: It might now be a bad idea to put this in "Contributer guides" in the Docs.

@ricklamers What do you think?

orchest/orchest-api/app/config.py
yannickperrenet: @samkovaly proposed a better fix to add it as a key-word argument to the `create_app` function.

orchest/orchest-webserver/Dockerfile
yannickperrenet: We will remove this and potentially put the approach to fix it on the developers machine through `.gitattributes` in the docs.

orchest.bat
samkovaly: Yes, ^ is what windows uses.

orchest.sh
samkovaly: Yes

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Ok

orchest/orchest-api/app/app/utils.py
samkovaly: yes

orchest/orchest-ctl/app/main.py
samkovaly: Ok, I've added a NOTE since we might not need the mount in the future. (if celery takes care of it)

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Yes that's true, I will add an aborted/revoked status to the database with this call

samkovaly: I also need to fix this: https://stackoverflow.com/questions/39191238/revoke-a-task-from-celery which is happening

orchest.sh
```diff
@@ -6,4 +6,7 @@ HOST_USER_DIR=$PWD/orchest/userdir/
 # create config dir if it doesn't exist
 mkdir -p $HOST_CONFIG_DIR
 
-docker run --name orchest-ctl --rm --mount type=bind,source=$HOST_USER_DIR,target=/userdir -v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=$HOST_CONFIG_DIR -e HOST_PWD=$PWD -e HOST_USER_DIR=$HOST_USER_DIR orchestsoftware/orchest-ctl "$@"
\ No newline at end of file
+# orchest-ctl needs to bind to /userdir to make the duable queue directory automatically
+docker run --name orchest-ctl --rm --mount type=bind,source=$HOST_USER_DIR,target=/userdir \
+-v /var/run/docker.sock:/var/run/docker.sock -e HOST_CONFIG_DIR=$HOST_CONFIG_DIR \
+-e HOST_PWD=$PWD -e HOST_USER_DIR=$HOST_USER_DIR orchestsoftware/orchest-ctl "$@"
```
yannickperrenet, line 12: I would indent here to make it visually clear that it is a line continuation.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -165,9 +161,15 @@ def delete(self, run_uuid):
...
         # TODO: https://stackoverflow.com/questions/39191238/revoke-a-task-from-celery
-        # TODO: delete new pipeline files that were created for this specific run?
+        # NOTE: delete new pipeline files that were created for this specific run?
         revoke(run_uuid, terminate=True)
 
+        res = models.ScheduledRun.query.filter_by(run_uuid=run_uuid).update({
+            'status': 'REVOKED'
+        })
+        if res:
+            db.session.commit()
```
yannickperrenet, line 171: This is now only done for the pipeline level run. Should also be done for individual `StepStatus`

orchest.sh
samkovaly: Ok

samkovaly: Ok

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Yes, good catch.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,224 @@
...
+
+        # Stop the run, whether it is in the queue or whether it is
+        # actually running.
+        revoke(run_uuid, terminate=True)
+
+        run = models.ScheduledRun.query.filter_by(run_uuid=run_uuid).first()
+        run.status = 'REVOKED'
+
+        for status in run.step_statuses:
+            status.status = 'REVOKED'
```
yannickperrenet, line 170: I think it would be cleaner to do

```python
res = models. ScheduledStepStatus.query.filter_by(
    run_uuid=run_uuid
).update(data)
```
where `data` contains the new status.

For [this reason](https://stackoverflow.com/questions/37738574/why-doesnt-flask-sql-alchemy-update-an-attribute-for-a-record) I would like to change the lines above 166-167 to use a `.update()`  as well.

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
```diff
@@ -0,0 +1,224 @@
...
+
+        post_data['run_config']['run_endpoint'] = 'scheduled_runs'
+
+        scheduled_date_time_string = post_data['scheduled_date_time']
+        scheduled_date_time = datetime.fromisoformat(scheduled_date_time_string.replace('Z', '+00:00'))
+            
+        # Construct pipeline.
+        pipeline = construct_pipeline(**post_data)
+
+        print(current_app.name)
```
yannickperrenet, line 81: Could you remove this `print`?

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
yannickperrenet: @samkovaly Where do you stand on this?

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: yeah

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Ok and I'll add this:
    if run_res and step_res:
            db.session.commit()

orchest/orchest-api/app/app/apis/namespace_scheduled_runs.py
samkovaly: Hey, I don't yet have a solution but I will.

PR #45, opened 2020-10-06
Add new feature: file manager

ricklamers:
File manager service is based on the excellent https://github.com/filebrowser/filebrowser project.

### DIFF
--- a/services/orchest-api/app/app/core/pipelines.py
+++ b/services/orchest-api/app/app/core/pipelines.py
@@ -136,7 +136,13 @@ def get_dynamic_mounts(run_config, task_id):
                 continue
 
             source = f'{datasource["connection_details"]["absolute_host_path"]}'
-            target = f'/data/{datasource["name"]}'
+
+            # the default (host) /userdata/data should be mounted in /data
+            if datasource['connection_details']['absolute_host_path'].endswith("/userdir/data"):
+                target = '/data'
+            else:
+                target = f'/mounts/{datasource["name"]}'
+
             mounts.append(f'{source}:{target}')
 
     # Determine the appropriate name for the volume that shares
--- a/services/orchest-ctl/app/cmdline.py
+++ b/services/orchest-ctl/app/cmdline.py
@@ -122,30 +122,30 @@ def help():
 
 def stop(skip_names=[]):
     
-    running_containers = docker_client.containers.list(all=True)
+    containers = docker_client.containers.list(all=True)
 
-    for running_container in running_containers:
+    for container in containers:
 
         # don't kill orchest-ctl itself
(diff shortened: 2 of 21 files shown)


services/orchest-api/app/app/core/pipelines.py
```diff
@@ -136,7 +136,13 @@ def get_dynamic_mounts(run_config, task_id):
                 continue
 
             source = f'{datasource["connection_details"]["absolute_host_path"]}'
-            target = f'/data/{datasource["name"]}'
+
+            # the default (host) /userdata/data should be mounted in /data
+            if datasource['connection_details']['absolute_host_path'].endswith("/userdir/data"):
```
yannickperrenet, line 141: Given that the `connection_details` state `'$HOST_USER_DIR/data'` and we know that `HOST_USER_DIR` will always end with `/userdir` since we chose the name of that directory, this should be fine indeed.

services/orchest-ctl/app/cmdline.py
```diff
@@ -122,7 +122,7 @@ def help():
 
 def stop(skip_names=[]):
     
-    running_containers = docker_client.containers.list()
+    running_containers = docker_client.containers.list(all=True)
```
yannickperrenet, line 125: I don't like the name `running_containers` since the `all=True` kwarg makes it work like `docker ps -a` thus also listing non-running containers.

services/orchest-webserver/app/app/res/kernels/launch_docker.py
```diff
@@ -55,9 +55,15 @@ def get_orchest_mounts(param_env):
...
+            
+            # the default (host) /userdata/data should be mounted in /data
+            if datasource['connection_details']['absolute_host_path'].endswith("/userdir/data"):
+                target_path = '/data'
+            else:
+                target_path = '/mounts/%s' % datasource['name']
 
             mount = Mount(
-                target='/data/%s' % datasource['name'],
+                target=target_path,
```
yannickperrenet, line 66: Now that `target_path` is set to have the `/mounts/` prefix the [HostDirectory.path](https://github.com/orchest/orchest/blob/feature-file-manager/orchest-sdk/python/orchest/datasources.py#L158) in the SDK has to be updated as well.

yannickperrenet (changes requested):
The `orchest-sdk` has to be updated at [this line](https://github.com/orchest/orchest/blob/feature-file-manager/orchest-sdk/python/orchest/datasources.py#L158) to be `/mounts/` instead of `/data/`.

Otherwise looking good!

yannickperrenet (approved):
Looks good!

PR #47, opened 2020-10-10
Feature small fixes

ricklamers:


services/jupyter-server/app/app/core/start_server.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any, Dict
 
 from jupyterlab.labapp import LabApp
+from flask import current_app
```
yannickperrenet, line 7: I don't see why this addition is needed.

services/jupyter-server/app/app/core/start_server.py
```diff
@@ -41,6 +42,7 @@ def _write_server_info_to_file(server_info: Dict[str, Any],
 
 
 def main():
+
```
yannickperrenet, line 45: Also this addition does not add anything.

services/update-server/app/app/utils.py
```diff
@@ -3,7 +3,7 @@
 
 def orchest_ctl(client, command):
 
-    return client.containers.run("orchestsoftware/orchest-ctl:latest", command, detach=True,
+    return client.containers.run("orchestsoftware/orchest-ctl:latest", command, name='orchest-ctl', detach=True, auto_remove=True,
```
yannickperrenet, line 6: I see this function is repeated about three times in different services, should we move this into a module in our internal package?

yannickperrenet (changes requested):
Ready to merge after the `start_server` fixes 👍

services/update-server/app/app/utils.py
ricklamers: It was two times, not three. And I was still not sure whether I wanted to put it in the 'lib' hence postponing for a bit. Aware of the duplication. Let's discuss the "what do we put in lib" question. I felt this was too specific to the web apps (update-server and orchest-webserver) to put in lib.

PR #48, opened 2020-10-15
Streaming cell output, Xterm.js for logs, orchest-ctl refactor, tests for pipeline step runner and various small changes

ricklamers:


lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -1,9 +1,10 @@
 import docker
 import os
+import uuid
 
 def orchest_ctl(client, command):
```
yannickperrenet, line 5: I would like to see a different name used here. I don't think `orchest_ctl` is clear to what it does.

services/custom-images/runnable-shared/runner/runner/__init__.py
```diff
@@ -0,0 +1,110 @@
+import nbformat
```
yannickperrenet, line 1: To follow convention it would be better to restructure the application code so the `__init__.py` file is empty.

services/orchest-webserver/app/app/__init__.py
```diff
@@ -112,7 +114,7 @@ def create_app():
 
     logging.info("Flask CONFIG: %s" % app.config)
 
-    db.init_app(app)
+    db.init_app(app)
```
yannickperrenet, line 117: We should start using formatters that strip any whitespace characters from the end of lines. I remember this happening to multiple files due to the differences in editors. This only makes the git history unclear.

So not perse a fix for now, but something we should consider in the future.

services/orchest-webserver/app/app/views.py
```diff
@@ -426,7 +427,7 @@ def get(self):
                 if show_internal:
                     datasources = DataSource.query.all()
                 else:
-                    datasources = DataSource.query.filter(not_(DataSource.name.startswith('_'))).all()
+                    datasources = DataSource.query.filter(~DataSource.name.like("\_%", escape="\\")).all()
```
yannickperrenet, line 430: :+1:

yannickperrenet (changes requested):
Great work!

lib/python/orchest-internals/_orchest/internals/utils.py
ricklamers: Any suggestions? Maybe something like `run_orchest_ctl`?

services/custom-images/runnable-shared/runner/runner/__init__.py
ricklamers: Will refactor

services/orchest-webserver/app/app/__init__.py
ricklamers: Agree, maybe linting in CI/CD?

yannickperrenet (approved):

PR #55, opened 2020-10-30
adding result backend (sqllite)

fruttasecca:


services/orchest-api/app/config.py
```diff
@@ -20,6 +20,18 @@ class Config:
...
+    # note: the database might require trimming from time to time,
+    # to enable having the db trimmed automatically use:
+    # https://docs.celeryproject.org/en/master/userguide/configuration.html#std:setting-result_expires
+    # note that "When using the database backend, celery beat must be running for the results to be expired."
+    # The db reaching a large size is probably an unreasonable edge case, given that currently our celery tasks do
+    # not produce output.
+    # For this reason no solution is provided for this, taking into account that we will eventually
+    # implement cronjobs and such trimming might be an internal cronjob, or automatically managed
+    # by celery if we end using "celery beat".
+    result_backend = "db+sqlite:////tmp/celery_result_backend.sqlite"
```
yannickperrenet, line 33: Shouldn't we be storing the `result_backend` in the `userdir`, making it persistent?

Now we would lose the ability to abort a task when the `orchest-api` has to be rebooted (for whatever reason).

fruttasecca: Agree on that, will fix.

yannickperrenet (approved):

PR #59, opened 2020-11-16
Update to JupyterLab 3.0.0

ricklamers:
Update to support JupyterLab 3.0.0 (rc.9).

It gets rid of the need to patch the reconnect procedure for the enterprise gateway which previously resulted in significant compile times for the jupyter-server service image.

This PR also removes the need for token based authorization of the JupyterLab instance since we authenticate all requests through the nginx proxy. This simplifies the jupyter-server spawning by just starting JupyterLab through the CLI instead of directly calling into it's Python classes (they provide a more stable interface on the CLI compared to direct Python initialization).

It also includes some minor changes to the front end (ctrl click for multi-step selection, and visibility checking for the fitAddon of XTerm)


services/jupyter-server/Dockerfile
```diff
@@ -5,43 +5,26 @@ RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificat
 # install nodejs for jupyterextension install support
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - && apt-get install -y nodejs
 
-RUN pip3 install git+https://github.com/jupyter/notebook.git@b44fc3b01dd1ae059170ab9ea657e34d44f5e36c
-RUN pip3 install flask_socketio \
+RUN pip3 install pip --upgrade
+
+RUN pip3 install jupyterlab==3.0.0rc9 \
+    flask_socketio \
```
fruttasecca, line 11:
```suggestion
    flask_socketio \
    requests \
```
Should we pin down the versions?

fruttasecca, line 11: Should we pin down the versions?

services/jupyter-server/app/requirements.txt
```diff
@@ -1,5 +1,5 @@
 flask-restplus==0.13.0
 Flask-SocketIO==4.2.1
-jupyterlab==2.1.0
 requests==2.23.0
 Werkzeug==0.16.1
+jupyterlab
```
fruttasecca, line 5: Should we pin down the version?

fruttasecca (commented):
lgtm, apart from the fact that we should pin down versions more often when it comes to the requirements

yannickperrenet:
As discussed we will try a larger rewrite where we no longer require the JupyterLab to be wrapped in a Flask API. Basically converting the into a Dockerfile with an `ENTRYPOINT`.

services/jupyter-server/app/requirements.txt
ricklamers: No longer exists.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -263,27 +271,27 @@ def launch(
...
-            "gateway-url": f"http://{IP.jupyter_EG}:8888",
-            "ServerApp.base_url": f'/jupyter_{IP.jupyter_server.replace(".", "_")}/',
+
+        self._notebook_server_info = {
+            "port": 8888,
+            "base_url": "/%s"
+            % _config.JUPYTER_SERVER_NAME.format(
+                project_uuid=project_uuid[: _config.TRUNCATED_UUID_LENGTH],
+                pipeline_uuid=pipeline_uuid[: _config.TRUNCATED_UUID_LENGTH],
+            ),
```
yannickperrenet, line 281: I think this would be clearer:
```python
"base_url": "/" + _config.JUPYTER_SERVER_NAME.format(
                project_uuid=project_uuid[: _config.TRUNCATED_UUID_LENGTH],
                pipeline_uuid=pipeline_uuid[: _config.TRUNCATED_UUID_LENGTH],
            ),
```

services/jupyter-server/Dockerfile
```diff
@@ -1,16 +1,19 @@
...
+# Install Jupyter Lab
+RUN pip3 install pip --upgrade
 RUN pip3 install jupyterlab==3.0.0rc9 \
-    flask_socketio \
-    flask_restplus \
-    requests \
+    flask_socketio==4.3.1 \
+    flask_restplus==0.13.0 \
+    requests==2.25.0 \
     Werkzeug==0.16.1
```
yannickperrenet: Can we use a `requirements.txt` instead just like with other services.

services/jupyter-server/Dockerfile
```diff
@@ -21,10 +24,8 @@ RUN git clone --single-branch --branch jupyterlab3 https://github.com/orchest/or
 RUN jupyter labextension install /jupyter-extensions/visual-tags --no-build
 RUN jupyter labextension install /jupyter-extensions/orchest-integration
 
-# Swap dash for normal bash to get more full fledged terminal session in Jupyter
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
-
 RUN mkdir /root/.jupyter/
```
yannickperrenet, line 27: Don't think this line is needed as the `COPY` command should needed paths.

services/jupyter-server/Dockerfile
```diff
@@ -1,16 +1,19 @@
-FROM tiangolo/uwsgi-nginx-flask:python3.7
+FROM python:3.7-slim
```
yannickperrenet, line 2: Should probably also add the line
```
LABEL maintainer="Orchest B.V. https://www.orchest.io"
```

services/jupyter-server/Dockerfile
```diff
@@ -21,10 +24,8 @@ RUN git clone --single-branch --branch jupyterlab3 https://github.com/orchest/or
 RUN jupyter labextension install /jupyter-extensions/visual-tags --no-build
 RUN jupyter labextension install /jupyter-extensions/orchest-integration
```
yannickperrenet, line 25: Should probably make this into a one-line `RUN` command with `&&` as otherwise it would create two Docker layers.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -306,7 +314,10 @@ def shutdown(self) -> None:
         # TODO: make sure a graceful shutdown is instantiated via a
         #       DELETE request to the flask API inside the jupyter-server
```
yannickperrenet, line 315: The `shutdown_jupyter_server` used below does it gracefully, so we should remove this comment.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -263,27 +271,27 @@ def launch(
             "Starting Jupyter Server on %s with Enterprise "
             "Gateway on %s" % (IP.jupyter_server, IP.jupyter_EG)
         )
-        payload = {
-            "gateway-url": f"http://{IP.jupyter_EG}:8888",
-            "ServerApp.base_url": f'/jupyter_{IP.jupyter_server.replace(".", "_")}/',
```
yannickperrenet: Sadly, GitHub did not allow me to put the comment there. But I would like to also remove the comments

```python
# The launched jupyter-server container is only running the API
# and waits for instructions before the Jupyter server is
# started. Tries to start the Jupyter server, by waiting for the
# API to be running after container launch.
```
a few lines above.

yannickperrenet (changes requested):
I love the reduced complexity! Just need some small changes.

services/orchest-api/app/app/core/sessions.py
ricklamers: Anything in particular about the docstring?

```
"""Launches the interactive session.

Additionally connects the launched `jupyter-server` with the
`jupyter-enterprise-gateway` (shot `jupyter-EG`).

Args:
    See `Args` section in parent class :class:`Session`.

"""
```

Seems to cover it?

lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -15,6 +15,14 @@
...
 # Containers
 PIPELINE_STEP_CONTAINER_NAME = "orchest-step-{run_uuid}-{step_uuid}"
+JUPYTER_SERVER_NAME = "jupyter-server-{project_uuid}-{pipeline_uuid}"
+JUPYTER_EG_SERVER_NAME = "jupyter-EG-{project_uuid}-{pipeline_uuid}"
+
+# Whenever UUIDs take up too much space in an identifier the UUIDs are
+# truncated to this length. This typically only happens when multiple UUIDs are
+# concatenated.
+# E.g. hostname length is limited to 63 characters per label (parts enclosed in dots)
+TRUNCATED_UUID_LENGTH = 18
```
fruttasecca, line 25: I think it could be wise to explain why only some lengths are valid, e.g. otherwise you might get hostnames
ending with an hyphen.

services/jupyter-server/Dockerfile
```diff
@@ -1,48 +1,28 @@
...
-RUN git config user.email "ricklamers@gmail.com" ; \
-    git config user.name "Rick Lamers"
-RUN git am < 0001-Reconnect-a-websocket-when-a-kernel-is-restarted.patch
-RUN pip3 install -e .
-RUN mkdir /jupyter-extensions/
-
-RUN git clone https://github.com/orchest/visual-tags.git /jupyter-extensions/visual-tags
-RUN jlpm run add:sibling /jupyter-extensions/visual-tags
+# Swap dash for normal bash to get more full fledged terminal session in Jupyter
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
```
fruttasecca, line 11: Can these be squashed into a single RUN command?

fruttasecca (approved):
lgtm, might have to change a couple of things if you agree with it

services/jupyter-server/Dockerfile
ricklamers: These RUNs are pretty much unrelated, I like the separation between them and the overhead of multiple layers is quite minimal in this case I believe.

lib/python/orchest-internals/_orchest/internals/config.py
ricklamers: Good one! Added to the comment.

PR #60, opened 2020-11-17
docker artifacts cleanup

fruttasecca:
Make orchest cleanup collateral docker artifacts, e.g. images, volumes, etc.

This PR makes it so that a session will also take care of removing the temporary volume which is created with it.
Moreover, an environment build will delete the old environment image, instead of untagging it, and it gets rid of unnecessary
images in case of failure or abortion.
Deleting a project will now result in the related environment images being removed.
Orchest will now cleanup its temporary volumes on shutdown.


services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -37,3 +38,45 @@ def delete(self, project_uuid, environment_uuid):
...
+            for img in docker_client.images.list()
+            if img.attrs["RepoTags"]
+            and isinstance(img.attrs["RepoTags"][0], str)
+            and img.attrs["RepoTags"][0].startswith(image_name)
+        ]
+
+        try:
+            for image_name in image_names_to_remove:
+                # using force true will actually remove the image instead of simply untagging it
+                docker_client.images.remove(image_name, force=True)
```
yannickperrenet, line 68: Would it be better to collect all images for which an exception is raised instead of having it fail the first time an image cannot be removed. Maybe another image cannot be removed either, which will raise an Exception the next time the endpoint is called, repeating the same process.

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -16,6 +17,51 @@
...
+        docker_client.containers.prune(filters=filters)
+        containers_to_prune = docker_client.containers.list(filters=filters, all=True)
+        # be as responsive as possible, only sleep at the first iteration if necessary
+        if containers_to_prune:
+            tries += 1
+            if tries > 100:
+                break
+            time.sleep(1)
+        else:
+            break
```
yannickperrenet, line 42: These lines can be removed.

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -16,6 +17,51 @@
...
+        docker_client.images.prune(filters=filters)
+        images_to_prune = docker_client.images.list(filters=filters)
+        # be as responsive as possible, only sleep at the first iteration if necessary
+        if images_to_prune:
+            tries += 1
+            if tries > 100:
+                break
+            time.sleep(1)
+        else:
+            break
```
yannickperrenet, line 62: Removable.

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -110,24 +164,36 @@ def build_docker_image(
...
+                "building script had a non 0 exit code, build failed\n"
             )
             user_logs_file_object.write(msg)
             complete_logs_file_object.write(msg)
             complete_logs_file_object.flush()
+
             return "FAILURE"
 
+        # note: we only attempt to cleanup if the environment was correctly created
+        if previous_image:
```
yannickperrenet, line 176: Should be explicitly compared to `None` with `if previous_image is not None`

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -110,24 +164,36 @@ def build_docker_image(
...
+            try:
+                # using force true will actually remove the image instead of simply untagging it
+                docker_client.images.remove(previous_image.id, force=True)
+            except Exception:
+                pass
+
         return "SUCCESS"
```
yannickperrenet, line 183: Is it possible that the removing is unsuccessful (and thus hits the Exception)? Because in that case we would still return `"SUCCESS"` without the image being removed.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -153,18 +153,46 @@ def shutdown(self) -> None:
...
+
+            # we are relying on the fact
+            # that the session_identity_uuid and project_uuid are consistent among
+            # these containers, i.e. there is 1 of each
+            if not session_identity_uuid:
+                session_identity_uuid = container.labels.get(
+                    "session_identity_uuid", None
+                )
+            if not project_uuid:
+                project_uuid = container.labels.get("project_uuid", None)
```
yannickperrenet, line 179: Should explicitly be compared to `None`. And specifying a default value of `None` for the get is unnecessary as it is the default value.

yannickperrenet (changes requested):
I like the general structure in which clean up is done. Just some minor tweaks.

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: agree, will fix

services/orchest-api/app/app/core/environment_builds.py
fruttasecca: agree, will fix

services/orchest-api/app/app/core/environment_builds.py
fruttasecca: agree, will fix

services/orchest-api/app/app/core/environment_builds.py
fruttasecca: agree, will fix

services/orchest-api/app/app/core/sessions.py
fruttasecca: agree, will fix

services/orchest-api/app/app/core/environment_builds.py
fruttasecca: It is. 
There is still the case in which a container is running and it is based on the given image, that will lead to an exception even if force=True.
In the context of orchest, it may happen if an environment build request gets through while an interactive run or an experiment is running.
What will happen is that the newly created image will be the one actually having the name
orchest-env-{proj_uuid}-{env_uuid}, while the old one will become a nameless and tagless image.

In a previous discussion with Rick we have decided that an environment (re)build should not be allowed during a pipeline run, because it may otherwise happen that step A is using a given version of the environment, while step B will use the new version. 
I think this also implies not allowing to start an experiment or an interactive run while an environment is building.

Note that these 2 points are not implemented yet, assuming they will be, catching the exception
is more of a safety net to not tie the success of a build to the success of the removal of the old image, if you feel like those 2 should be tied we can go for it, but I think such a change should come once the first point is implemented.
Nevertheless, I think I should add some comments in there to explain what I've explained here.

ricklamers: After discussing Jacopo and I thought it would be nice if we could allow the user to build the environments while the existing environment is being used by a pipeline run (noninteractive and interactive) or even by a kernel.

This would require a bit of complex cleanup logic, but we think it's worth the trade-off. This will be part of a new separate PR.

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -37,3 +38,52 @@ def delete(self, project_uuid, environment_uuid):
...
+
+
+@api.route(
+    "/<string:project_uuid>",
+)
+@api.param("project_uuid", "UUID of the project")
+class ProjectEnvironmentImages(Resource):
+    @api.doc("delete-project_environment-images")
+    def delete(self, project_uuid):
+        """Removes all environment images a project."""
```
ricklamers, line 50: Small typo "Removes all environment images from a project."

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -16,6 +17,47 @@
...
+    containers_to_prune = docker_client.containers.list(filters=filters, all=True)
+    tries = 0
+    while containers_to_prune:
+        docker_client.containers.prune(filters=filters)
+        containers_to_prune = docker_client.containers.list(filters=filters, all=True)
+        # be as responsive as possible, only sleep at the first iteration if necessary
+        if containers_to_prune:
+            tries += 1
+            if tries > 100:
+                break
```
ricklamers, line 39: I think we should log an error here

services/orchest-webserver/app/app/views/core.py
```diff
@@ -521,6 +509,36 @@ def remove_experiment_directory(experiment_uuid, pipeline_uuid, project_uuid):
...
+
+        Returns:
+
+        """
+        for removed_proj in fs_removed_projects:
+            # remove experiments related to the project
+            # exs_to_remove = Experiment.query.filter(Experiment.project_uuid == removed_proj.uuid).all()
+            # for ex in exs_to_remove:
+            #     remove_experiment_directory(ex.uuid, ex.pipeline_uuid, ex.project_uuid)
+            #     db.session.delete(ex)
```
ricklamers, line 527: Add #TODO so we know it's going to be used/finished/refactored soon TM.

ricklamers (changes requested):

services/orchest-api/app/app/core/environment_builds.py
yannickperrenet: I think it indeed makes sense to include that explanation as comments.

yannickperrenet (approved):

PR #63, opened 2020-11-18
Refactor gate to validation

ricklamers:
This PR by @yannickperrenet changes the gate endpoint on the `orchest-api` to a simplified validation endpoint specifically focused on checking the state of environment images.

Accompanying the change is a front-end refactor that is updated to match the changes introduced in the orchest-api endpoint.


services/orchest-api/app/app/apis/namespace_validations.py
```diff
@@ -0,0 +1,108 @@
...
+api = register_schema(api)
+
+
+def validate_environment(project_uuid: str, env_uuid: str) -> Tuple[str, Optional[str]]:
+    """Validates whether the environments exist on the system.
+
+    Only passes if all of the conditions below are satisfied:
+        * The `project_uuid` and `env_uuid` combination exists in the
+          persistent database `EnvironmentBuild` model.
+        * There is no "PENDING" or "STARTED" build for the environment.
```
fruttasecca, line 26: Is this still a requirement in the light of the fact that we are going to allow environment builds while
running pipelines?

services/orchest-api/app/app/apis/namespace_validations.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+        for env_uuid in environment_uuids:
+            # Check will be either "fail" or "pass".
+            validation, action = validate_environment(project_uuid, env_uuid)
+            res[validation].append(env_uuid)
+
+            if validation == "fail":
+                res["actions"].append(action)
+
+        res["validation"] = "fail" if len(res["fail"]) != 0 else "pass"
```
fruttasecca, line 107: I think using validation both as a variable (variable name) and as a string to access the dictionary can be kinda confusing to read.

services/orchest-api/app/app/apis/namespace_validations.py
yannickperrenet: "going to" as in in the future? Because then it should still be part of this PR.

I think we should go over a few user cases, since there are pros and cons for both approaches, before we make the decision to remove that particular check from the environment validation.

fruttasecca (changes requested):
Part of the logic of the gate might need to change.

services/orchest-api/app/app/apis/namespace_validations.py
yannickperrenet: @ricklamers What do you think? I think it is actually pretty nice that those are the same.

PR #64, opened 2020-11-19
Environment consistency across steps and runs on environment rebuilds

fruttasecca:
This PR introduces the concept of a dangling image from the point of view of orchest.
A dangling image is an image that is nameless, tag-less and that no further pipeline run (PENDING or STARTED) is going to use. An environment image might become nameless and tag-less on an environment rebuild, where it essentially becomes outdated.

Upon queuing an interactive run or an experiment, a reference to all images used by every step/run is maintained so that consistency is preserved in the case of an environment rebuild that happens while the run/experiment is being executed. 
From the point of view of the user, queuing a run or an experiment takes a snapshot of the current state of the environment images. Upon terminating a run, each environment image that was used is removed if it is now dangling, which means this run was effectively the last run using it.

The implementation mostly consist in new models, new api endpoints and an update on the endpoints that are actually launching pipelines runs. The entire PR affects only the orchest-api service.

In a previous discussion with Rick we got to the conclusion that we would need to add 2 new models (NonInteractiveRunImageMapping, InteractiveRunImageMapping) to maintain the mapping between the run and what images it uses, I wonder if we can instead get rid of them and add the needed columns to the PipelineStep model.


yannickperrenet:
Please correct me if I am wrong. 

This PR creates a new endpoint in `namespace_environment_images.py` and `namespace_runs.py` that removes dangling images w.r.t. Orchest, i.e. Orchest is not using the image currently and will not use the image in the future. And then we will also need to update database entries outside of these endpoints for database consistency.

fruttasecca:
> Please correct me if I am wrong.
> 
> This PR creates a new endpoint in `namespace_environment_images.py` and `namespace_runs.py` that removes dangling images w.r.t. Orchest, i.e. Orchest is not using the image currently and will not use the image in the future. And then we will also need to update database entries outside of these endpoints for database consistency.

Which entries are you referring to?

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -486,6 +485,16 @@ def to_dict(self) -> PipelineDescription:
...
 
+    @property
+    def environments(self) -> List[str]:
+        """Returns the list of UUIDs of the used environments
+
+        Returns:
+            Unique list of environments uuids used among the pipeline steps.
+
+        """
+        return list(set([step.properties["environment"] for step in self.steps]))
```
yannickperrenet, line 496: I don't think this should be a `property`. Probably `.get_environments(self)` is better because the property does not support assignment.

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -57,45 +65,114 @@ def post(self):
...
+                )
+                for env_uuid, docker_id in env_uuid_docker_id_mappings.items()
+            ]
+            db.session.bulk_save_objects(non_interactive_run_image_mappings)
+            db.session.commit()
+
+            # this is necessary because between the read of the images docker ids and the commit to the db
+            # of the mappings a new environment could have been built, an image could have become nameless
+            # and subsequently removed because the image mappings where not in the db yet, and we would
+            # end up with mappings that are pointing to an image that does not exist
```
yannickperrenet, line 117: So this means that the environments of the pipeline at the moment that the experiment is started by the user, do not  have to be the environments that are used during the experiment?

services/orchest-api/app/app/utils.py
```diff
@@ -92,3 +94,46 @@ def update_status_db(
         db.session.commit()
 
     return
+
+
+def get_environment_image_docker_id(name_or_id: str):
+    return docker_client.images.get(name_or_id).id
```
yannickperrenet, line 100: The `return` should probably be wrapped in a try-except block.

services/orchest-api/app/app/apis/namespace_runs.py
```diff
@@ -185,3 +279,39 @@ def put(self, run_uuid, step_uuid):
         )
 
         return {"message": "Status was updated successfully"}, 200
+
+
+@api.route(
+    "/dangling-images/<string:run_uuid>",
+)
+@api.param("run_uuid", "UUID of the run")
+class ProjectDanglingEnvironmentImages(Resource):
```
yannickperrenet, line 288: What is the need of this endpoint given that there is a similar one in `namespace_environment_images.py`?

yannickperrenet (changes requested):
Overall I think there are too many things happening in too many places, causing me to somewhat lose the overview of what is happening. It should be clear exactly what happens when, for example, an interactive run is started.

Secondly, we should start moving code into functions as certain blocks are repeated between endpoints.

Lastly, could you also wrap your comments and docstrings at a maximum line length of 72 (for PEP8). It seems like black does not format comments, but currently some are too long to fit on my (split) screen.

services/orchest-api/app/app/core/pipelines.py
fruttasecca: Agree

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: >So this means that the environments of the pipeline at the moment that the experiment is started by the user, do not have to be the environments that are used during the experiment?

This happens when the experiment is queued. About them potentially not being the exact same environment as when the user queues an experiment in the GUI: given the speed at which the request is resolved, if an environment finishes building in between a request getting resolved (actually, between the moment we read the docker ids and commit to db), the environment might very well have finished building right before the moment the user had clicked to queue an experiment through the GUI. 

If we want to get to this level of snapshotting then the client would have to know the docker ids beforehand, queue an experiment along with the docker ids to use, which might still lead to race conditions (e.g. an image getting rebuilt, becoming nameless and then deleted), except that now we have to maintain consistent docker ids in the client.

>Secondly, what is the benefit of explicitly checking whether they have become "nameless" because the moment this check passes they could become nameless again and thus ending up in the same situation. In other words, I don't see what this block solves or am I missing something obvious?

It resolves a race condition.
If an image has become nameless then a call to delete it as a dangling image might be on its way to delete the image. See `remove_if_dangling` in `services/orchest-api/app/app/utils.py`. 

Essentially, if an image has not become nameless after we have committed to db, then for sure this image won't be deleted, since `remove_if_dangling` will check for two things: if the image is nameless and if the image is/will be used (there are commits to the db showing that this image is used by a run that is PENDING or STARTING). 

If we discover that an image has become nameless after we have committed, then it could be that `remove_if_dangling`
has actually passed its checks before we could commit to db, and it is about to delete the image. For this reason we reference to the most up to date image, and try to "lock" it by committing.

services/orchest-api/app/app/utils.py
fruttasecca: I actually wanted the request to fail if a run was trying to use an image that did not exist, I'll handle the request failure better.

services/orchest-api/app/app/apis/namespace_runs.py
fruttasecca: Different levels of granularity. That endpoint will check for all images used in the run if they are dangling, the other tries to remove a single image. I did not want to give to the task in celery the responsibility of querying the more granular endpoint for each used image. It is possible and I would be ok with it if we are trying to trim down end points.

fruttasecca:
Agree on the fact that the PR is rather immature, the PR was mostly for feedback about what to change and where to trim.

> Lastly, could you also wrap your comments and docstrings at a maximum line length of 72 (for PEP8). It seems like black does not format comments, but currently some are too long to fit on my (split) screen.

Will do!

fruttasecca:
As the outcome of a discussion, the number of endpoints to take care of dangling images has been trimmed down to 1.
Attempting to deleting potentially dangling images now happens only after an environment build.

yannickperrenet (approved):
Nice work! Not sure whether the discussion cleared things up or the improved structure of the code did the talking :)

services/orchest-api/app/app/utils.py
```diff
@@ -92,3 +95,189 @@ def update_status_db(
...
+    As a collateral effect, new entries for interactive or non
+    interactive image mappings will be added, which is at the same
+    time the mechanism through which we "lock" the images, or, protect
+    them from deletion as long as they are needed.
+    About the race condition:
+        between the read of the images docker ids and the commit
+        to the db of the mappings a new environment could have been
+        built, an image could have become nameless and be
+        subsequently removed because the image mappings where not
+        in the db yet, and we would end up with  mappings that are
```
ricklamers, line 163: Small typo: were*

ricklamers:
So much cleaner! Love how we got it to be smaller and simpler yet still achieve the same underlying goals.

LGTM!

We'll definitely need to do some proper testing (although the race condition sounds tricky to trigger).

ricklamers (approved):

PR #66, opened 2020-11-20
Morph NotebookViewer into generic FileViewer powered by CodeMirror

ricklamers:
This PR makes it easy to view files in pipelines. It enables viewing both regular text files and notebooks (that continues to use the server side HTML renderer from nbconvert).

Previously, one could only view files in the read only view of the pipeline editor. However, that means that in the pipeline editor in edit mode you could not see the step file contents without having a session active.

Now, the edit mode also allows you to view pipeline step files using the "View file" button which works regardless of the session state.


fruttasecca (commented):
lgtm

fruttasecca (approved):
lgtm

yannickperrenet (approved):

PR #68, opened 2020-11-23
add help menu to the client

fruttasecca:
This PR adds a new menu, "help", to the client, through which the user will be able
to get to the docs, Slack, GitHub and website of Orchest.

help_menu


services/orchest-webserver/app/app/config.py
```diff
@@ -23,6 +23,16 @@ class Config:
     DEFAULT_DATASOURCES = _config.DEFAULT_DATASOURCES
     ORCHEST_API_ADDRESS = _config.ORCHEST_API_ADDRESS
 
+    # TODO: point readthedocs to stable instead of latest once stable
+    #  is up
+    HELP_MENU_CONTENT = {
```
ricklamers, line 28: Refactor the name to be something like "ORCHEST_WEB_URLS" so the generic purpose of these URLs in the webserver config becomes more clear.

services/orchest-webserver/app/static/js/views/HelpView.js
```diff
@@ -0,0 +1,69 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      slack: orchest.config.HELP_MENU_CONTENT.slack,
+      readthedocs: orchest.config.HELP_MENU_CONTENT.readthedocs,
+      readthedocks_quickstart:
+        orchest.config.HELP_MENU_CONTENT.readthedocks_quickstart,
+      github: orchest.config.HELP_MENU_CONTENT.github,
+      website: orchest.config.HELP_MENU_CONTENT.website,
```
ricklamers, line 13: I'd probably prefer the state to be a bit more explicit about the variable referring to a URL. slack_url, readthedocs_url, etc...

ricklamers (changes requested):

yannickperrenet:
Can we change "https://www.orchest.io" to something like "orchest.io" or "Website", as I now think it is too different from the other items.

I would also like to change the text "doubt or suggestion" to "questions or suggestions". To me "doubts" sound negative.

ricklamers (approved):
LGTM apart from Yannick's comments! I might have some feedback on how the view is rendered, but if need be we'll address it in a small hotfix commit.

ricklamers:
> Can we change "https://www.orchest.io" to something like "orchest.io" or "Website", as I now think it is too different from the other items.
> 
> I would also like to change the text "doubt or suggestion" to "questions or suggestions". To me "doubts" sound negative.

I think for consistency we should use www.orchest.io since we use it everywhere over just orchest.io. We can drop the https:// though. Or say "Website".

fruttasecca:
> > Can we change "https://www.orchest.io" to something like "orchest.io" or "Website", as I now think it is too different from the other items.
> > I would also like to change the text "doubt or suggestion" to "questions or suggestions". To me "doubts" sound negative.
> 
> I think for consistency we should use [www.orchest.io](http://www.orchest.io) since we use it everywhere over just orchest.io. We can drop the https:// though. Or say "Website".

I'll go with Website since it seems more consistent with the rest of the help menu (no wwws).

PR #69, opened 2020-11-23
PostHog based event tracking

ricklamers:
This PR replaces the old telemetry system and instead makes use of a self-hosted instance of PostHog to track anonymized user events in Orchest.

The tracking can still be disabled using the config.json user settings (which is also accessible in the settings pane).

Which events are tracked?
- Loading a view (the name of the view).
- Running a pipeline (through an experiment or through an interactive run).
- Alert/confirm dialog popups.
- Saving a pipeline (number of pipeline steps, and number of environments used).


services/orchest-webserver/app/app/analytics.py
```diff
@@ -3,31 +3,104 @@
...
+
+    for _, step in steps.items():
+        step.pop("title", None)
+        step.pop("parameters", None)
+        step.pop("file_path", None)
+
+        # Capture environments for count
+        env = step.get("environment", "")
+        if env not in environments and len(env) > 0:
+            environments.append(env)
```
fruttasecca, line 34: I think environments could be a set.

services/orchest-webserver/app/app/analytics.py
```diff
@@ -3,31 +3,104 @@
...
+            "project_size": project_size,
+            "run_type": run_type,
+        },
+    )
 
-        data = {"user_uuid": telemetry_uuid, "active": check_active(app)}
 
-        requests.post("https://analytics.orchest.io", json=data, timeout=1)
+def get_telemetry_uuid(app):
+    telemetry_uuid = ""
```
fruttasecca, line 65: minor thing, but I don't think the variable actually needs to be declared

services/orchest-webserver/app/requirements.txt
```diff
@@ -14,4 +14,5 @@ nbformat
 docker>=4.3.0
 Werkzeug==0.16.0
 python-socketio[client]
+posthog
```
fruttasecca, line 17: should we fix the version?

fruttasecca (commented):
lgtm

Since Orchest is project centric, should we also track project level information? Number of pipelines, number of environments, number of run experiments,  maybe also number of projects.

ricklamers:
> lgtm
> 
> Since Orchest is project centric, should we also track project level information? Number of pipelines, number of environments, number of run experiments, maybe also number of projects.

Let's keep this for future expansion if necessary. I don't want to track too many things right now.

yannickperrenet:
Looks good! Although it is probably a good idea to add specific comments in the code explaining why we track certain characteristics.

yannickperrenet (approved):

PR #70, opened 2020-11-24
Add ability to specify the memory size for data passing

yannickperrenet:
It is now possible to specify the memory to allocate for data passing per pipeline. 

Specifying the size can be done through the settings of a pipeline in the UI or by adding  `"data_passing_memory_size"` (together with a value) to the `"settings"` property in the pipeline definition.


ricklamers:
LGTM!

PR #71, opened 2020-11-24
Add `jupyter` syntax highlighting in CodeMirror for consistency

ricklamers:


yannickperrenet:
It looks reasonably similar, although highlighting of functions, classes and other keywords is different.

ricklamers:
> It looks reasonably similar, although highlighting of functions, classes and some keywords (did not check all keywords) is different.

As discussed you were looking at the nbconvert html output. Which indeed is not exactly identical, but that's an issue of the current state of `nbconvert`.

I guess we can merge this PR as the CodeMirror syntax highlighting works as expected.

PR #72, opened 2020-11-25
Better cleanup on removal of pipelines, experiments, projects 

fruttasecca:
This PR is about housekeeping when a deletion of a pipeline, experiment or project is issued. This needs to happen at the file system level, database(s) level and docker environment level.

Upon the deletion of a pipeline, its session is closed, any interactive run is aborted, and the related data is removed from the db.

Upon the deletion of an experiment, the experiment is aborted, removed from disk, related data is removed from the db.

Upon the deletion of a project, any running session, interactive run, experiment, environment build are closed/aborted, related data is removed from disk and the db.


services/orchest-api/app/app/apis/namespace_environment_builds.py
```diff
@@ -290,3 +290,84 @@ def get(self, project_uuid, environment_uuid):
...
+
+
+@api.route("/cleanup/<string:project_uuid>/<string:environment_uuid>")
+@api.param("project_uuid", "UUID of the project.")
+@api.param("environment_uuid", "UUID of the environment.")
+class ProjectEnvironmentBuildsCleanup(Resource):
+    """Used when an environment is cleaned up"""
+
+    @staticmethod
+    def cleanup(project_uuid, environment_uuid):
```
yannickperrenet, line 302: I am not really a fan of these `staticmethod`s. They are even referenced in different endpoints as well and used outside of the class. I would much rather have them inside of the `utils.py.`

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -19,20 +26,38 @@
...
+        Will stop any run or experiment making use of this environment.
+        """
         image_name = _config.ENVIRONMENT_IMAGE_NAME.format(
             project_uuid=project_uuid, environment_uuid=environment_uuid
         )
 
+        # stop all interactive runs making use of the environment
+        int_runs = interactive_runs_using_environment(project_uuid, environment_uuid)
+        for run in int_runs:
+            Run.stop(run.run_uuid)
```
yannickperrenet, line 40: Essentially we are invoking (not calling) another endpoint of the API here. Is this "RESTful"?

And on another note, why does the deletion of an environment delete the pipeline runs that use it? Should the environment not at least be locked during (or possibly instantly removed after) a pipeline run? Due to complexity I think locking makes more sense, it just checks whether the environment is used anywhere and states it can therefore not be deleted. If the user however chooses to delete it through the filesystem, then it is up to them.

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -191,45 +191,50 @@ def get(self, experiment_uuid):
...
+        """Stops an experiment given its UUID.
+
+        However, it will not delete any corresponding database entries,
+        it will update the status of corresponding objects to "REVOKED".
+        """
+        experiment = models.Experiment.query.get_or_404(
+            experiment_uuid,
+            description="Experiment not found",
+        )
+        Experiment.stop(experiment)
```
yannickperrenet, line 236: Should probably be refactor because it uses staticmethods, but also in the refactor this would still apply. I think all the function that do the cleanup should receive a similar object, e.g. always the API parameters instead of sometimes the API parameters and sometimes a DB model.

services/orchest-api/app/app/models.py
```diff
@@ -103,18 +103,34 @@ class InteractiveRun(PipelineRun):
 
     run_uuid = db.Column(db.String(36), primary_key=True)
 
-    pipeline_steps = db.relationship("InteractiveRunPipelineStep", lazy="joined")
-    image_mappings = db.relationship("InteractiveRunImageMapping", lazy="joined")
+    pipeline_steps = db.relationship(
+        "InteractiveRunPipelineStep",
+        lazy="joined",
+        passive_deletes=False,
+        cascade="all, delete",
```
yannickperrenet, line 110: Not sure whether it would be overkill, but I think it is good to add comments explaining what behavior we are solving by adding these kwargs.

services/orchest-api/app/app/utils.py
```diff
@@ -245,6 +246,83 @@ def lock_environment_images_for_run(
...
+        models.InteractiveRun.status.in_(["PENDING", "STARTED"]),
+    ).all()
+
+
+def experiments_using_environment(project_uuid: str, env_uuid: str):
+    """Get the list of experiments using a given environment.
+
+    Args:
+        project_uuid (str):
+        env_uuid (str):
```
yannickperrenet, line 270: These type annotations in the docstring are not needed since we are already specifying their type in the function definition.

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -324,3 +329,39 @@ def put(self, experiment_uuid, run_uuid, step_uuid):
         )
 
         return {"message": "Status was updated successfully"}, 200
+
+
+@api.route("/cleanup/<string:experiment_uuid>")
```
yannickperrenet, line 334: Why is experiment cleanup not run on deletion of an experiment by default instead of having to call a dedicated endpoint? This question applies to all cleanup that has dedicated endpoints.

yannickperrenet (changes requested):
I think the general implementation structure needs some work, i.e. usage of endpoints and staticmethods. I would like to minimize the complexity to achieve the cleanup, although I do see that the architecture of the application requires more "complex" cleanup.

services/orchest-api/app/app/models.py
fruttasecca: At line 144 I've added a link to the docs thinking that the first instance of a model using such args was 
at that line, looks like I was wrong. I'll move that up and improve the explanation.

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: Because we also support abortion without deletion to allow the state to be read, e.g. by the orchest webserver.

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: >And on another note, why does the deletion of an environment delete the pipeline runs that use it? Should the environment not at least be locked during (or possibly instantly removed after) a pipeline run? Due to complexity I think locking makes more sense, it just checks whether the environment is used anywhere and states it can therefore not be deleted. If the user however chooses to delete it through the filesystem, then it is up to them.

As of now, there is an endpoint to get if an environment is in use, which can be used to tell the user that an environment
deletion will lead to stopping runs/pipelines or to simply reject the user action.

>Due to complexity I think locking makes more sense
I think it might make things slightly more complex because currently its a "fire and forget' behavior, which locking
will not allow.

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: good point

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: I went with these kind of `staticmethods`s to avoid, essentially, moving a chunk of the actual code of the endpoints
into `utils.py`. 

Since the actual content of these methods is tied to the namespace and the endpoint, an alternative would be
to have, for each `namespace_*.,py`, another file which contains the actual code of the `staticmethods`s, freeing
them from being part of a Resource. I did not do that because I found the code to be more readable (~ less files to traverse) and because I found it handy to have the pattern `Resource.stop` et similia.  

Restructuring from `staticmethod`s to another solution should not be an issue, given that each `staticmethod` is, in a sense, easy to move out of a calss.

services/orchest-api/app/app/apis/namespace_environment_images.py
yannickperrenet: > As of now, there is an endpoint to get if an environment is in use, which can be used to tell the user that an environment
> deletion will lead to stopping runs/pipelines or to simply reject the user action.

Since it is the webserver that is then responsible for first checking whether it will lead to stopping of runs it is just a shift of responsibility. Either we lock an environment or not, and I am obviously in favor of locking.

> I think it might make things slightly more complex because currently its a "fire and forget' behavior, which locking
> will not allow.

I don't agree. How I see the implementation for locking is also a "fire and forget". Just try to delete the environment, if its in use, then deleting won't work. Once the pipeline run is complete, don't do anything just wait for the user to try to delete the environment again.

yannickperrenet: After our discussion on the locking, we together decided to provide the user with explicit warning messages in case the user wants to delete an environment that is currently being used and that that will stop all runs that depend on it.

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@api.route("/<string:project_uuid>/<string:pipeline_uuid>")
+@api.param("project_uuid", "UUID of the project")
+@api.param("pipeline_uuid", "UUID of the pipeline")
+class Pipeline(Resource):
+    @api.doc("cleanup_pipeline")
+    @api.response(200, "Pipeline cleaned up")
+    def delete(self, project_uuid, pipeline_uuid):
+        """Cleanup a pipeline.
```
yannickperrenet, line 25: Could we change the naming here to "delete" instead of "cleanup"? I feel that the primary action is deleting the pipeline which will then also clean it up.

fruttasecca: Will do, and will do the same for other cleanup related functions where applicable

services/orchest-api/app/app/models.py
```diff
@@ -125,19 +150,41 @@ class NonInteractiveRun(PipelineRun):
...
+    )
 
 
 class NonInteractiveRunPipelineStep(PipelineRunPipelineStep):
     __tablename__ = "non_interactive_run_pipeline_steps"
     __bind_key__ = "persistent_db"
 
+    # TODO: verify why we have the exp uuid as a column, seems to be
+    # redundant info since we already have the run_uuid
```
ricklamers, line 172: experiment_uuid is used to update NonInteractiveRunPipelineStep's status.

fruttasecca: my bad, that was actually meant to say "why we have the exp uuid as part of the primary key"

fruttasecca: indeed, but it is not needed because of the relationship between `NonInteractiveRun` and `NonInteractiveRunPipelineStep`, I think

yannickperrenet (approved):

PR #74, opened 2020-11-26
Wrap docker sdk images.list with internal util function

ricklamers:
By default the Docker SDK for Python suffers from a race condition when calling client.images.list(). As one of the images retrieved by the loop could be missing, in which case it throws an exception:

https://github.com/docker/docker-py/blob/master/docker/models/images.py

This PR wraps the list function to avoid the issue by retrying on failure.

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -139,3 +139,14 @@ def get_orchest_mounts(project_dir, host_project_dir, mount_form="docker-sdk"):
                 mounts.append(mount)
 
     return mounts
+
+
+def docker_images_list_safe(docker_client, *args, attempt_count=10, **kwargs):
+    while True:
+        attempt_count -= 1
+        try:
+            return docker_client.images.list(*args, **kwargs)
+        except Exception as e:
+            logging.debug("Failed to call docker_client.images.list(): %s" % e)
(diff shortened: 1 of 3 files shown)


lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -139,3 +139,14 @@ def get_orchest_mounts(project_dir, host_project_dir, mount_form="docker-sdk"):
                 mounts.append(mount)
 
     return mounts
+
+
+def docker_images_list_safe(docker_client, *args, attempt_count=10, **kwargs):
+    while True:
+        attempt_count -= 1
+        try:
```
fruttasecca, line 147: probably better to just catch the ImageNotFound exception and let it fail otherwise?

fruttasecca (approved):

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -139,3 +139,14 @@ def get_orchest_mounts(project_dir, host_project_dir, mount_form="docker-sdk"):
                 mounts.append(mount)
 
     return mounts
+
+
+def docker_images_list_safe(docker_client, *args, attempt_count=10, **kwargs):
+    while True:
```
yannickperrenet, line 145: I would just do a `for _ in range(attempt_count)`

PR #75, opened 2020-11-27
Improve InteractiveRun and Experiment POST endpoints

fruttasecca:
This PR is a minor refactor of the API endpoints for POSTing interactive runs and experiments. 

It improves logging and storing to db in the case of a failure that made it impossible to start a run (interactive or belonging to an experiment) through the request.

Moreover, it delivers an improved experience in the case a user is actually able to start an interactive run through the GUI when a docker image which is a dependency of the pipeline has been deleted. The GUI will now correctly display a failure and the right button(s), e.g. the run won't be stuck in a permanent pending (and cancelable) state anymore.

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_experiments.py
+++ b/services/orchest-api/app/app/apis/namespace_experiments.py
@@ -183,6 +183,8 @@ class ExperimentList(Resource):
                 # storing and transmitting results) associated to the task.
                 # Uncomment the line below if applicable.
                 res.forget()
+
+            return experiment, 201
         else:
             logging.error("\n".join(experiment_creation_error_messages))
 
@@ -201,7 +203,9 @@ class ExperimentList(Resource):
             ).update({"status": "FAILURE"})
             db.session.commit()
 
-        return experiment, 201
+            return {
+                "message": "Failed to create experiment because not all referenced environments are available."
+            }, 500
 
 
 @api.route("/<string:experiment_uuid>")
--- a/services/orchest-api/app/app/apis/namespace_experiments.py
+++ b/services/orchest-api/app/app/apis/namespace_experiments.py
@@ -190,12 +190,12 @@ class ExperimentList(Resource):
             # and the db while avoiding multiple update statements
             # (1 for each object)
             for pipeline_run in experiment["pipeline_runs"]:
-                pipeline_run.status = "FAILURE"
+                pipeline_run.status = "SUCCESS"
                 for step in pipeline_run["pipeline_steps"]:
                     step.status = "FAILURE"
             models.NonInteractiveRun.query.filter_by(
                 experiment_uuid=post_data["experiment_uuid"]
-            ).update({"status": "FAILURE"})
+            ).update({"status": "SUCCESS"})
             models.NonInteractiveRunPipelineStep.query.filter_by(
                 experiment_uuid=post_data["experiment_uuid"]
             ).update({"status": "FAILURE"})
--- a/services/orchest-api/app/app/apis/namespace_experiments.py
+++ b/services/orchest-api/app/app/apis/namespace_experiments.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import logging
 import uuid
 
 from celery.contrib.abortable import AbortableAsyncResult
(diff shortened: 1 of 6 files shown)


services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -131,49 +148,59 @@ def post(self):
...
+            models.NonInteractiveRun.query.filter_by(
+                experiment_uuid=post_data["experiment_uuid"]
+            ).update({"status": "SUCCESS"})
+            models.NonInteractiveRunPipelineStep.query.filter_by(
+                experiment_uuid=post_data["experiment_uuid"]
+            ).update({"status": "FAILURE"})
+            db.session.commit()
 
-        experiment["pipeline_runs"] = pipeline_runs
         return experiment, 201
```
yannickperrenet, line 204: Should the return code not depend on whether the `experiment_creation_error_messages` list is empty or not?

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -112,10 +130,9 @@ def post(self):
                         is_interactive=False,
                     )
                 except errors.ImageNotFound as e:
-                    return (
+                    experiment_creation_error_messages.append(
```
yannickperrenet, line 133: This comment is actually about the try block. 

Since the set of images will be the same for all pipeline runs of an experiment, I do not think it makes sense to run `lock_environment_images_for_run` inside the for-loop. This also makes this PR easier as we no longer need to maintain a list for `experiment_creation_error_messages` and can instead return here (with an unsuccessful error code). Additionally, the locking of environments should ideally happen before writing to the DB so that the written entries do not have to be updated.

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -131,49 +148,59 @@ def post(self):
...
+                pipeline_run.status = "SUCCESS"
+                for step in pipeline_run["pipeline_steps"]:
+                    step.status = "FAILURE"
+            models.NonInteractiveRun.query.filter_by(
+                experiment_uuid=post_data["experiment_uuid"]
+            ).update({"status": "SUCCESS"})
+            models.NonInteractiveRunPipelineStep.query.filter_by(
+                experiment_uuid=post_data["experiment_uuid"]
+            ).update({"status": "FAILURE"})
+            db.session.commit()
```
yannickperrenet, line 202: I think we should reduce the number of times we call `db.session.commit()` to a minimal. This way we don't have to fix the DB entries on errors (if `experiment_creation_error_messages` is not empty) as we simply don't commit to the DB and disregard the changes.

yannickperrenet (changes requested):
The experiments POST request needs to be refactored.

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: I think the entries need to be there, even if failed, so that the client or webserver can make use of that data to inform the user, this is actually a feature, imo.
About the number of commits, not sure how much it can be improved, considering we need the image mapping entries in the db to "lock" an image. I think it's preferable to have all runs of an experiment in the db as FAILED than having only the one needed to lock the required images.

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: > Since the set of images will be the same for all pipeline runs of an experiment, I do not think it makes sense to run `lock_environment_images_for_run` inside the for-loop. This also makes this PR easier as we no longer need to maintain a list for `experiment_creation_error_messages` and can instead return here (with an unsuccessful error code). 

The experiment POST endpoint needs a bigger refactoring overall, see the TODO around line 55. I did not want to make the PR bigger than it needed to be for its purpose given that we are getting nearer the stable release.

> Additionally, the locking of environments should ideally happen before writing to the DB so that the written entries do not have to be updated.
Locking the environment needs to have entries written to the DB as a requirement, since we are using the DB to know if an environment is or will be used, this is out of scope of the PR.

services/orchest-api/app/app/apis/namespace_experiments.py
fruttasecca: It did, initially. The `orchest-webserver` is acting as a proxy between the `client` and the `orchest-api`, returning error codes such as 500 or 404 will lead to an error. I think this is parallel to the fact that a run needs to have the status of `SUCCESS` to be considered done by the client (and perhaps other moving parts).

yannickperrenet: Then the `orchest-webserver` needs to be updated as well. Possibly also an error code different from 500 or 404 that better covers this case.

services/orchest-api/app/app/apis/namespace_experiments.py
yannickperrenet: In my opinion this should be the time that we do that refactor then. I do not want to keep building on something that is not done well, if that means this PR cannot be included in the stable release then I am fine with that.

fruttasecca: I see the PR more as a small refactoring to solve an issue rather than building upon something. I completely agree on the experiments POST endpoint requiring a facelift.

>if that means this PR cannot be included in the stable release then I am fine with that
The issue described in the first comment of the PR should be covered in the stable release, I think.

ricklamers (approved):

yannickperrenet (approved):

PR #80, opened 2020-12-04
Fix memory server allocation

ricklamers:
Overallocating by a fraction instead of a fixed size to increase the probability with which the user gets exactly the requested size for the memory server.

### DIFF
--- a/services/orchest-api/app/app/utils.py
+++ b/services/orchest-api/app/app/utils.py
@@ -362,7 +362,7 @@ def remove_if_dangling(img) -> bool:
 
 
 def parse_string_memory_size(memory_size: Union[str, int]) -> int:
-    """Returns string value of memory in bytes"""
+    """Returns value of memory in bytes specified by input string"""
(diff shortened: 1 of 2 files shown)


services/orchest-api/app/app/utils.py
```diff
@@ -361,7 +361,21 @@ def remove_if_dangling(img) -> bool:
     return False
 
 
-def calculate_shm_size(data_passing_memory_size: Union[str, int]) -> int:
+def parse_string_memory_size(memory_size: Union[str, int]) -> int:
+    """Returns value of memory in bytes specified by input string"""
```
yannickperrenet, line 365: The docstring should reflect that its just a conversion from string to int.

ricklamers: """Returns value of memory in bytes specified by input string""" is not clear enough?

fruttasecca (approved):
lgtm

PR #84, opened 2020-12-07
Remove deprecated pyarrow.serialize/deserialize calls, make memory passing more efficient.

fruttasecca:
This PR aims to remove references to `pyarrow.serialize` and `pyarrow.deserialize` since they are
deprecated.

It improves data serialization by avoiding an unnecessary serialization that was previously
done to pickles.

Getting data from memory and disk has been made more efficient by avoiding unnecessary
copies of in-memory data.

### DIFF
--- a/orchest-sdk/python/orchest/transfer.py
+++ b/orchest-sdk/python/orchest/transfer.py
@@ -144,7 +144,6 @@ def _output_to_disk(
     if isinstance(serialization, Serialization):
         with pa.OSFile(f"{full_path}.{serialization}", "wb") as f:
             f.write(obj)
-            f.close()
     else:
         raise ValueError("Function not defined for specified 'serialization'")
 
--- a/orchest-sdk/python/orchest/transfer.py
+++ b/orchest-sdk/python/orchest/transfer.py
@@ -142,7 +142,7 @@ def _output_to_disk(
         ValueError: If the specified serialization is not valid.
     """
     if isinstance(serialization, Serialization):
-        with pa.OSFile(f"{full_path}.{serialization}", "wb") as f:
+        with pa.OSFile(f"{full_path}.{serialization.name}", "wb") as f:
             f.write(obj)
     else:
         raise ValueError("Function not defined for specified 'serialization'")
@@ -202,7 +202,7 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) -> 
     head_file = os.path.join(step_data_dir, "HEAD")
     with open(head_file, "w") as f:
         current_time = datetime.utcnow()
-        f.write(f'{current_time.isoformat(timespec="seconds")}, {serialization}')
+        f.write(f'{current_time.isoformat(timespec="seconds")}, {serialization.name}')
 
     # Full path to write the actual data to.
     full_path = os.path.join(step_data_dir, step_uuid)
@@ -210,25 +210,25 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) -> 
     return _output_to_disk(data, full_path, serialization=serialization)
 
 
-def _get_output_disk(full_path: str, serialization: Serialization) -> Any:
+def _get_output_disk(full_path: str, serialization: str) -> Any:
     """Gets data from disk.
 
     Raises:
         ValueError: If the serialization argument is invalid.
     """
     file_path = f"{full_path}.{serialization}"
-    if serialization == str(Serialization.ARROW_TABLE):
+    if serialization == Serialization.ARROW_TABLE.name:
         # pa.memory_map is for reading (zero-copy)
         with pa.memory_map(file_path, "rb") as input_file:
             # read all batches as a table
             stream = pa.ipc.open_stream(input_file)
             return stream.read_all()
-    elif serialization == str(Serialization.ARROW_BATCH):
+    elif serialization == Serialization.ARROW_BATCH.name:
         with pa.memory_map(file_path, "rb") as input_file:
             # return the first batch (the only one)
             stream = pa.ipc.open_stream(input_file)
             return [b for b in stream][0]
-    elif serialization == str(Serialization.PICKLE):
+    elif serialization == Serialization.PICKLE.name:
         # https://docs.python.org/3/library/pickle.html
         # The argument file must have three methods,
         # a read() method that takes an integer argument,
@@ -449,7 +449,7 @@ def output_to_memory(data: Any, disk_fallback: bool = True) -> None:
 
     # Try to output to memory.
     obj_id = _convert_uuid_to_object_id(step_uuid)
-    metadata = bytes(f"{Config.IDENTIFIER_SERIALIZATION};{serialization}", "utf-8")
+    metadata = bytes(f"{Config.IDENTIFIER_SERIALIZATION};{serialization.name}", "utf-8")
 
     try:
         obj_id = _output_to_memory(obj, client, obj_id=obj_id, metadata=metadata)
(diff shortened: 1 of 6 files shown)


orchest-sdk/python/orchest/transfer.py
```diff
@@ -21,6 +22,21 @@
 from orchest.utils import get_step_uuid
 
 
+class Serialization(Enum):
```
yannickperrenet, line 25: I like the use of `Enum` here :+1:

orchest-sdk/python/requirements.txt
```diff
@@ -1,4 +1,4 @@
 SQLAlchemy
 boto3
-pyarrow==1.0.1
+pyarrow==2.0.0
```
yannickperrenet, line 3: I think it makes more sense to remove the hardcoded version altogether, making it just `pyarrow`.

orchest-sdk/python/setup.py
```diff
@@ -10,7 +10,7 @@
     version="0.2.0",
     packages=setuptools.find_packages(),
     install_requires=[
-        "pyarrow==1.0.1",
+        "pyarrow==2.0.0",
```
yannickperrenet, line 13: Removing the hardcoded version would apply here as well, although I since it is the `setup.py` it might make more sense to make it e.g. `pyarrow>=2.0.0` instead.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -92,62 +98,60 @@ def serialize(
...
+        ValueError: If the specified serialization is not valid.
     """
-    if serialization in ["arrow", "arrowpickle"]:
-        with open(f"{full_path}.{serialization}", "wb") as f:
-            obj.write_to(f)
-
+    if isinstance(serialization, Serialization):
+        with pa.OSFile(f"{full_path}.{serialization}", "wb") as f:
+            f.write(obj)
+            f.close()
```
yannickperrenet, line 147: Is this `f.close()` necessary, since it is in a `with` statement anyways?

orchest-sdk/python/orchest/transfer.py
```diff
@@ -92,62 +98,60 @@ def serialize(
...
     Raises:
-        ValueError: If the specified `type` is not one of ``['pickle']``
+        ValueError: If the specified serialization is not valid.
     """
-    if serialization in ["arrow", "arrowpickle"]:
-        with open(f"{full_path}.{serialization}", "wb") as f:
-            obj.write_to(f)
-
+    if isinstance(serialization, Serialization):
+        with pa.OSFile(f"{full_path}.{serialization}", "wb") as f:
```
yannickperrenet, line 145: The file-extension would now become, e.g. `Serialization.ARROW_TABLE` instead of `.ARROW_TABLE`, is that intended?

orchest-sdk/python/orchest/transfer.py
```diff
@@ -212,24 +211,50 @@ def output_to_disk(
...
+def _get_output_disk(full_path: str, serialization: Serialization) -> Any:
+    """Gets data from disk.
 
-    return data
+    Raises:
+        ValueError: If the serialization argument is invalid.
+    """
+    file_path = f"{full_path}.{serialization}"
+    if serialization == str(Serialization.ARROW_TABLE):
+        # pa.memory_map is for reading (zero-copy)
```
yannickperrenet, line 222: If it becomes zero-copy, does this mean that for pyarrow related datastructures the user would operate on zero-copy and not have the object copied into the memory of the respective pipeline step? If this is indeed the case, what does this mean for the reference style counting eviction of the plasma store?

orchest-sdk/python/requirements.txt
fruttasecca: I don't think it makes that much sense.  In the general case, it is good practise to pin
down versions, and I think that for this particular case it is even more important.
This PR started as a fix to move away from deprecated stuff, and while at it we discovered
that pyarrow might deprecate more things. I think it's just safer to pin down the version, but if you feel strongly about it
then it can go.

orchest-sdk/python/orchest/transfer.py
fruttasecca: I think the `f.close()` can go.

orchest-sdk/python/tests/test_transfer.py
```diff
@@ -52,32 +65,21 @@ def plasma_store(monkeypatch):
...
-
 @pytest.mark.parametrize(
     "data_1",
     [
         generate_data(KILOBYTE),
-        np.array([UnserializableByPyarrowObject(1) for _ in range(3)]),
+        np.random.rand(10, 5, 2),
+        np.array([CustomClass(1) for _ in range(3)]),
+        get_test_record_batch(),
+        get_test_table(),
```
yannickperrenet, line 75: Should we add a test for other commonly passed data types such as a pandas dataframe?

orchest-sdk/python/orchest/transfer.py
fruttasecca: Let's say that I saw no reason to make it the opposite, it's more explicit and does not affect anything else.

orchest-sdk/python/tests/test_transfer.py
fruttasecca: Indeed, we should.

yannickperrenet (changes requested):
Great work! Should definitely make the data passing a lot more efficient. Just think we need to discuss little details first.

orchest-sdk/python/orchest/transfer.py
yannickperrenet: It is on the filesystem of the user and it is common practice to have a certain format of file extensions.

orchest-sdk/python/requirements.txt
yannickperrenet: I agree it is good to pin down versions, but I also want to make sure we move with the latest version. We should catch deprecation errors anyways before we release our new versions. No strong feelings though.

orchest-sdk/python/orchest/transfer.py
fruttasecca: > If this is indeed the case, what does this mean for the reference style counting eviction of the plasma store?

This part of the code is about getting data from disk, not sure that is related.

>If it becomes zero-copy, does this mean that for pyarrow related datastructures the user would operate on zero-copy and not have the object copied into the memory of the respective pipeline step?

`pa.memory_map` does not allocate new memory on read, while using `pa.OSFile` would. Getting
a `pa.RecordBatch` or a `pa.Table` from the stream will get you your own "copy" of said data types, with less overhead, at least  that is what I could infer from the docs and testing.

```python
with pa.memory_map(file_path, "rb") as diskf:
        stream = pa.ipc.open_stream(diskf)
        // getting a RecordBatch
        data = [b for b in stream][0]
        os.remove(file_path)
        diskf.close()
        print(data.nbytes)
        print(data.num_rows)
```

orchest-sdk/python/orchest/transfer.py
fruttasecca: I'll go ahead and remove the Serialization part.

orchest-sdk/python/orchest/transfer.py
yannickperrenet: Alright that makes sense. Just wanted to make sure that multiple processes would not alter the same file in memory.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -480,15 +502,26 @@ def _get_output_memory(obj_id: plasma.ObjectID, client: plasma.PlasmaClient) ->
...
+        return [b for b in stream][0]
 
-    return obj
+    elif metadata == bytes(serialization.format(Serialization.PICKLE.name), "utf-8"):
+        # Can load the buffer directly because its a bytes-like-object:
+        # https://docs.python.org/3/library/pickle.html#pickle.loads
+        return pickle.loads(buffer)
+
+    else:
+        raise ValueError("Object was serialized with an invalid serialization")
```
yannickperrenet, line 524: Lets go with "unsupported" instead of "invalid" here.

yannickperrenet (approved):

PR #85, opened 2020-12-08
Add Julia support

ricklamers:
This commit adds Julia support to Orchest. The SDK for Julia (similar to the one for R) still needs be created.

With this PR users will be able to build Julia environments, run Julia scripts, and (interactively) run Julia based Notebooks.


services/base-images/base-kernel-julia/README.md
```diff
@@ -0,0 +1,5 @@
+# base-kernel-r
+
+This is the base image provided by default with the Orchest installation that provides a R
+focused Linux environment based off
+[`r-notebook`](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-r-notebook).
```
yannickperrenet, line 5: Guess this should be for Julia instead of R

ricklamers: Good catch!

yannickperrenet (changes requested):
Great work on adding Julia support. I do think there are a few things missing:

* The `base-kernel-julia` needs to be added to the GitHub Actions.

* If we want to make Julia install possible without any other dependencies, then it needs to be added to the [`Language` `Enum` in the ctl](https://github.com/orchest/orchest/blob/750e0d8b2082e6abd131aa4d1131e25a274f305a/services/orchest-ctl/app/main.py#L49).

Lastly, it would be easier to review if the respective changes were divided over multiple commits. Now it is hard to see what code change is for which end result.

fruttasecca (commented):
lgtm

PR #86, opened 2020-12-08
Fix race condition when multiple parents attempt starting their children

ricklamers:


services/orchest-api/app/app/core/pipelines.py
```diff
@@ -264,40 +284,27 @@ async def run_on_docker(
...
-            self._status,
-            task_id,
-            session,
-            type="step",
-            run_endpoint=run_config["run_endpoint"],
-            uuid=self.properties["uuid"],
-        )
+        except Exception as e:
+            print("Exception", e)
+            self._status = "FAILURE"
```
yannickperrenet, line 297: Can we reduce the amount of code in this `try` block? Now it is unclear to me what exactly we are trying to catch here.

Additionally, I want to propose to change the `print` into use of `logging`.

PR #87, opened 2020-12-10
Named data passing

fruttasecca:
This PR changes the semantics of the Orchest SDK, in particular, when it comes to data passing.

Outputting data now requires a name. Passing `None` as the name will make the data nameless, while a string will actually become the name of the data. This affects how the data will be later accessed by any child step.

Getting input data (`get_inputs()`) now returns a dictionary instead of a list. This dictionary effectively
acts as a namespace of the received data. Named data is accessed through its key, unnamed data
is aggregated in a list accessed through a reserved key.

The number of public functions in the `transfer` module of the Orchest SDK has been reduced.


orchest-sdk/python/orchest/__init__.py
```diff
@@ -4,7 +4,4 @@
     get_params,
     update_params,
 )
-from orchest.transfer import (
-    get_inputs,
-    output,
-)
+from orchest.transfer import get_inputs, output, output_to_disk, output_to_memory
```
yannickperrenet, line 7: I do not see the need to add `output_to_disk` and `output_to_memory` here. I want users to use the `get_inputs` and `output` only, however, if they explicitly want to call the methods then they will have to use the `transfer` module.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -17,10 +18,32 @@
...
+        return
+
+    if name == _RESERVED_UNNAMED_OUTPUTS_STR:
+        raise ValueError(
+            "'%s' is a reserved output data name." % _RESERVED_UNNAMED_OUTPUTS_STR
+        )
+
+    if __METADATA_SEPARATOR__ in name:
+        raise ValueError(
+            "'%s' cannot be part of the name of the data" % __METADATA_SEPARATOR__
```
yannickperrenet, line 44: `.I would like to use f-strings here instead of the old string formatting. Could you also add a `.` to the end of the sentence? These are errors to the users so I want to be strict here.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -151,7 +174,9 @@ def _output_to_disk(
     return
 
 
-def output_to_disk(data: Any, serialization: Optional[Serialization] = None) -> None:
+def output_to_disk(
+    data: Any, name: Union[str, None], serialization: Optional[Serialization] = None
```
yannickperrenet, line 178: I think using `Optional[str]` over `Union[str, None]` is easier to understand and would follow the same typing as other parts in this module.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -163,6 +188,10 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) ->
 
     Args:
         data: Data to output to disk.
+        name: Name of the output data. As a string, it becomes
+        the name of the data, when `None`, the data is considered
+        nameless. This affects the way the data can be later retrieved
+        using :func:`get_inputs`.
```
yannickperrenet, line 194: The docstring will be formatted incorrectly by sphinx, it has to be like
```python
name: Name of the output data. As a string, it becomes
    the name of the data, when `None`, the data is considered
    nameless. This affects the way the data can be later retrieved
    using :func:`get_inputs`.
```
And please wrap at 72 instead of using arbitrary line wrapping.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -180,6 +209,10 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) ->
         to be only calling the function once.
 
     """
+    _check_data_name_validity(name)
```
yannickperrenet, line 212: Since this function raises errors when checking the validity, these errors should be documented in this method's docstring.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -202,16 +235,21 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) ->
...
+        f.write(metadata)
 
     # Full path to write the actual data to.
     full_path = os.path.join(step_data_dir, step_uuid)
 
     return _output_to_disk(data, full_path, serialization=serialization)
 
 
-def _get_output_disk(full_path: str, serialization: str) -> Any:
+def _get_output_disk_local(full_path: str, serialization: str) -> Any:
```
yannickperrenet, line 252: Not sure whether I like the `_local` suffix, because to me it works confusion: "What if it is a network mounted disk? What is different from this function and the one without `_local`?".

In Python we can use a double underscore, giving `__get_output_disk` although this would somewhat (I would argue not entirely) a misuse of the double underscore as a prefix.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -202,16 +235,21 @@ def output_to_disk(data: Any, serialization: Optional[Serialization] = None) ->
...
     head_file = os.path.join(step_data_dir, "HEAD")
     with open(head_file, "w") as f:
-        current_time = datetime.utcnow()
-        f.write(f'{current_time.isoformat(timespec="seconds")}, {serialization.name}')
+        metadata = [
+            datetime.utcnow().isoformat(timespec="seconds"),
+            serialization.name,
+            name,
+        ]
+        metadata = __METADATA_SEPARATOR__.join(metadata)
```
yannickperrenet, line 243: This would yield a line such as:
```
2020-12-10T17:03:08__ORCHEST_METADATA_SEP__PICKLE__ORCHEST_METADATA_SEP__myname
```
which I think is unreadable. So I would rather set the separator to something like `;`.

orchest-sdk/python/orchest/transfer.py
fruttasecca: I think it would indeed be a misuse of the double underscore. I've used `_local` for lack of better names, I am open to suggestions.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -248,7 +286,7 @@ def _get_output_disk(full_path: str, serialization: str) -> Any:
         raise ValueError("The specified serialization is unsupported: %s")
 
 
-def get_output_disk(step_uuid: str, serialization: str) -> Any:
+def _get_output_disk(step_uuid: str, serialization: Serialization) -> Any:
```
yannickperrenet, line 289: This is a wrong merge of master, the type annotation should be `str`. This argument to this function is read from the `HEAD` file and is a string (since it is never parsed to the `Serialization` class afterwards).

orchest-sdk/python/orchest/transfer.py
fruttasecca: The separator can indeed become easier to read, perhaps we can use something slightly more complex
than `;`, I would prefer not making the `;` character unavailable. `;;;` or something similar could be fine.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -389,14 +432,20 @@ def _output_to_memory(
     return obj_id
 
 
-def output_to_memory(data: Any, disk_fallback: bool = True) -> None:
+def output_to_memory(
+    data: Any, name: Union[str, None], disk_fallback: bool = True
```
yannickperrenet, line 436: Here Id also like to suggest going with `Optional[str]`

orchest-sdk/python/orchest/transfer.py
```diff
@@ -17,10 +18,32 @@
     ObjectNotFoundError,
     OrchestNetworkError,
     StepUUIDResolveError,
+    InputNameCollisionError,
 )
 from orchest.pipeline import Pipeline
 from orchest.utils import get_step_uuid
 
+_RESERVED_UNNAMED_OUTPUTS_STR = "unnamed_inputs"
+__METADATA_SEPARATOR__ = "__ORCHEST_METADATA_SEP__"
```
yannickperrenet, line 27: Should we not put these in the `config.py`?

orchest-sdk/python/orchest/transfer.py
```diff
@@ -311,10 +350,14 @@ def resolve_disk(step_uuid: str) -> Dict[str, Any]:
...
 
     res = {
-        "timestamp": timestamp,
-        "method_to_call": get_output_disk,
+        "method_to_call": _get_output_disk,
         "method_args": (step_uuid,),
         "method_kwargs": {"serialization": serialization},
+        "metadata": {
+            "timestamp": timestamp,
+            "serialization": serialization,
```
yannickperrenet, line 358: This `["metadata"]["serialization"]` does not seem to be used anywhere.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -389,14 +432,20 @@ def _output_to_memory(
...
 
     To manage outputing the data to memory for the user, this function
     uses metadata to add info to objects inside the plasma store.
 
     Args:
         data: Data to output.
+        name: Name of the output data. As a string, it becomes
+        the name of the data, when `None`, the data is considered
+        nameless. This affects the way the data can be later retrieved
+        using :func:`get_inputs`.
```
yannickperrenet, line 448: This needs more indentation for sphinx. Additionally, for sphinx we need to use double backticks instead of single to get it to be highlighted as code and not italic:
```python
"""
``None``
"""
```

orchest-sdk/python/orchest/transfer.py
```diff
@@ -422,6 +471,8 @@ def output_to_memory(data: Any, disk_fallback: bool = True) -> None:
         therefore want to be only calling the function once.
 
     """
+    _check_data_name_validity(name)
```
yannickperrenet, line 474: Same applies here regarding the exception raising documentation.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -446,11 +497,26 @@ def output_to_memory(data: Any, disk_fallback: bool = True) -> None:
...
     # Try to output to memory.
     obj_id = _convert_uuid_to_object_id(step_uuid)
-    metadata = bytes(f"{Config.IDENTIFIER_SERIALIZATION};{serialization.name}", "utf-8")
+    metadata = [
+        str(Config.IDENTIFIER_SERIALIZATION),
+        # the plasma store allows to get the creation timestamp, but
+        # creating it this way makes the process more consistent
+        # with the metadata we are writing when outputting to disk,
+        # moreover, it makes the code less dependent on the
+        # plasma store API
```
yannickperrenet, line 510: I would like to suggest line wrapping here also and using regular punctuation and capitalization as comments should be written like regular sentences. (Yes, I know I am being overly strict here but I think saying it once is enough and helpful down the line.)

orchest-sdk/python/orchest/transfer.py
fruttasecca: Indeed it is not, as of now. I think it is more consistent since this way it is the complete metadata.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -462,12 +528,14 @@ def output_to_memory(data: Any, disk_fallback: bool = True) -> None:
...
         #       Therefore we will only support metadata added by the
         #       user, once disk also supports passing metadata.
-        return output_to_disk(obj, serialization=serialization)
+        return output_to_disk(obj, name, serialization=serialization)
 
     return
 
 
-def _get_output_memory(obj_id: plasma.ObjectID, client: plasma.PlasmaClient) -> Any:
+def _get_output_memory_from_plasma(
```
yannickperrenet, line 536: I do like this name :+1:

yannickperrenet (changes requested):
My `mypy` is complaining about a few type annotations, which I think is important to get fixed as these type annotations are in the docs and should be correct (nonetheless). Also I get a warnings when compiling the docs.

orchest-sdk/python/orchest/transfer.py
yannickperrenet: Hardest question ever. `_deserialize_output_disk`?

orchest-sdk/python/orchest/transfer.py
yannickperrenet: The only problem with `;` I can think of is that the user can no longer use it in their name when passing the data. However, you do very nicely check whether `__METADATA_SEPARATOR__ in name` so I think its fine to go with `;` (as this is not really a character someone want to include in the name), keeping it more elegant in the `HEAD` file.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -737,17 +828,49 @@ def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> List[Any]
...
+
+        # keep it for later
+        get_output_methods.append((parent, get_output_method, args, kwargs, metadata))
+
+        if metadata["name"] != _RESERVED_UNNAMED_OUTPUTS_STR:
+            collisions_dict[metadata["name"]].append(parent.properties["title"])
+
+    # if there are collisions raise an error
+    collisions_dict = {k: v for k, v in collisions_dict.items() if len(v) > 1}
+    if len(collisions_dict) > 0:
```
yannickperrenet, line 852: Here you can just do `if collisions_dict:`

fruttasecca: Wouldn't this, in a way, be inconsistent with the fact that when checking for `None` we actually do `if X is None`?

orchest-sdk/python/orchest/transfer.py
```diff
@@ -766,19 +889,29 @@ def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> List[Any]
...
+        if name == _RESERVED_UNNAMED_OUTPUTS_STR:
+            data[_RESERVED_UNNAMED_OUTPUTS_STR].append(incoming_step_data)
+        else:
+            data[name] = incoming_step_data
 
     return data
 
 
-def output(data: Any) -> None:
+def output(data: Any, name: Union[str, None]) -> None:
```
yannickperrenet, line 903: `Optional[str]` (just mentioning so we don't miss this one)

orchest-sdk/python/orchest/transfer.py
fruttasecca: >The only problem with ; I can think of is that the user can no longer use it in their name when passing the data.

Yes, that's what I meant.

>I think its fine to go with `;` or `;`

Did you mean `;;;`?

yannickperrenet: I meant `;<space>` but I see the literal space was removed.

orchest-sdk/python/orchest/transfer.py
yannickperrenet: Checking for `is None` is a special case that should be handled separately. Please refer to https://google.github.io/styleguide/pyguide.html#2144-decision

orchest-sdk/python/orchest/transfer.py
fruttasecca: `;<space>` it is then

services/orchest-api/app/tests/input_execution_order/pipeline-7.json
```diff
@@ -16,7 +16,7 @@
             "name": "step-1",
             "uuid": "uuid-1",
             "file_path": "",
-            "environment": 0.3
+            "environment": 0.5
```
yannickperrenet, line 19: I don't understand why the tests for the orchest-api are changed here? Also, should it not add a new pipeline definition instead of changing an old one?

yannickperrenet (changes requested):
Everything looks good to me, except for the changes orchest-api tests. It seems to me they are part of other changes and were accidentally included in these commits.

services/orchest-api/app/tests/input_execution_order/pipeline-7.json
fruttasecca: The issue with those tests is that they are dependant on race conditions going in a certain way.
The changes to the `transfer` module seem to have made the overall process slightly slower, could be because of the changes to `_resolve_memory`, without that change, those tests would fail so often it was not feasible to make the PR without having failing tests.
Since we decided to not make changes to those tests in this PR, I had to leave in those changes.

fruttasecca: As discussed, this PR will include a fix for the aforementioned problem.

PR #91, opened 2020-12-15
Move Orchest database backend to postgres

fruttasecca:
This PR changes the database backend of Orchest from `sqllite` to `postgres`. A new service, `orchest-database`, is launched on start.

The PR affects `orchest-ctl` and services that were persisting their data: the `auth-server`, `orchest-webserver`, `orchest-api`, `celery-worker`.

### DIFF
--- a/services/auth-server/app/app/__init__.py
+++ b/services/auth-server/app/app/__init__.py
@@ -17,11 +17,15 @@ def create_app(config_class=None):
     app = Flask(__name__)
     app.config.from_object(config_class)
 
-    # Initialize the database and create the database file.
+    # Create the database if it does not exist yet. Roughly equal to
+    # a "CREATE DATABASE IF NOT EXISTS <db_name>" call.
     if not database_exists(app.config["SQLALCHEMY_DATABASE_URI"]):
         create_database(app.config["SQLALCHEMY_DATABASE_URI"])
+
     db.init_app(app)
     with app.app_context():
+        # This call will create tables if needed (the ones which do not
+        # exist in the database yet).
         db.create_all()
 
     register_views(app)
--- a/services/auth-server/app/app/__init__.py
+++ b/services/auth-server/app/app/__init__.py
@@ -7,6 +7,8 @@ Additinal note:
         https://docs.pytest.org/en/latest/goodpractices.html
 """
 from flask import Flask
+from sqlalchemy_utils import create_database, database_exists
+
 from app.views import register_views
 from app.connections import db
 
@@ -16,6 +18,8 @@ def create_app(config_class=None):
     app.config.from_object(config_class)
 
     # Initialize the database and create the database file.
+    if not database_exists(app.config["SQLALCHEMY_DATABASE_URI"]):
+        create_database(app.config["SQLALCHEMY_DATABASE_URI"])
     db.init_app(app)
     with app.app_context():
         db.create_all()
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -61,12 +61,15 @@ def create_app(config_class=None, use_db=True):
             # atomic operation that will result in an error if the
             # directory is already there, this cleanup operation will
             # run only once per container.
-            if os.system("mkdir /tmp/interactive_cleanup_done > /dev/null 2>&1") == 0:
+            try:
+                os.mkdir("/tmp/interactive_cleanup_done")
                 InteractiveSession.query.delete()
                 InteractivePipelineRun.query.filter(
                     InteractivePipelineRun.status.in_(["PENDING", "STARTED"])
                 ).delete(synchronize_session="fetch")
                 db.session.commit()
+            except FileExistsError:
+                pass
 
     # Register blueprints.
     app.register_blueprint(api, url_prefix="/api")
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -12,16 +12,24 @@ from sqlalchemy_utils import create_database, database_exists
 
 from app.apis import blueprint as api
 from app.connections import db
-from app.models import (
-    InteractiveSession,
-    InteractiveRun,
-    InteractiveRunPipelineStep,
-    InteractiveRunImageMapping,
-)
 from config import CONFIG_CLASS
 
 
 def create_app(config_class=None, use_db=True):
+    """Create the Flask app and return it.
+
+    Args:
+        config_class: Configuration class. See orchest-api/app/config.
+        use_db: If true, associate a database to the Flask app instance,
+            which implies connecting to a given database and possibly
+            creating such database and/or tables if they do not exist
+            already. The reason to differentiate instancing the app
+            through this argument is that the celery worker does not
+            need to connect to the db that "belongs" to the orchest-api.
+
+    Returns:
+        Flask.app
+    """
     app = Flask(__name__)
     app.config.from_object(config_class)
 
@@ -43,9 +51,5 @@ def create_app(config_class=None, use_db=True):
 
     # Register blueprints.
     app.register_blueprint(api, url_prefix="/api")
-
-    # create celery database if needed
-    if not database_exists(CONFIG_CLASS.result_backend_sqlalchemy_uri):
-        create_database(CONFIG_CLASS.result_backend_sqlalchemy_uri)
 
     return app
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -30,11 +30,15 @@ def create_app(config_class=None, use_db=True):
     CORS(app, resources={r"/*": {"origins": "*"}})
 
     if use_db:
-        # Initialize the database and create the database file.
+        # Create the database if it does not exist yet. Roughly equal to
+        # a "CREATE DATABASE IF NOT EXISTS <db_name>" call.
         if not database_exists(app.config["SQLALCHEMY_DATABASE_URI"]):
             create_database(app.config["SQLALCHEMY_DATABASE_URI"])
+
         db.init_app(app)
         with app.app_context():
+            # This call will create tables if needed (the ones which do
+            # not exist in the database yet).
             db.create_all()
 
     # Register blueprints.
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -36,14 +36,6 @@ def create_app(config_class=None, use_db=True):
         db.init_app(app)
         with app.app_context():
             db.create_all()
-            # cleanup stuff that is temporary/ephemireal, the cleanup
-            # happens on startup so that it's still possible to inspect
-            # the db on shutdown
-            InteractiveSession.query.delete()
-            InteractiveRun.query.delete()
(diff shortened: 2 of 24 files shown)


services/auth-server/app/app/__init__.py
```diff
@@ -16,6 +18,8 @@ def create_app(config_class=None):
     app.config.from_object(config_class)
 
     # Initialize the database and create the database file.
+    if not database_exists(app.config["SQLALCHEMY_DATABASE_URI"]):
+        create_database(app.config["SQLALCHEMY_DATABASE_URI"])
     db.init_app(app)
     with app.app_context():
         db.create_all()
```
yannickperrenet, line 25: I think this works fine, but it seems that `create_database` and `db.create_all()` would both try to create the database. Although the latter won't if the first succeeds (because then the db already exists).

In my opinion, the code should be clearer to what exactly is happening here.

services/orchest-api/app/app/__init__.py
```diff
@@ -23,11 +31,25 @@ def create_app(config_class=None, use_db=True):
...
     if use_db:
         # Initialize the database and create the database file.
+        if not database_exists(app.config["SQLALCHEMY_DATABASE_URI"]):
+            create_database(app.config["SQLALCHEMY_DATABASE_URI"])
         db.init_app(app)
         with app.app_context():
             db.create_all()
+            # cleanup stuff that is temporary/ephemireal, the cleanup
+            # happens on startup so that it's still possible to inspect
+            # the db on shutdown
```
yannickperrenet, line 41: Could you elaborate? I want to make sure I understand it better, because now it seems like a reboot of the `orchest-api` (maybe on failure) could end up deleting its previously issued db changes.

services/orchest-api/app/app/__init__.py
```diff
@@ -23,11 +31,25 @@ def create_app(config_class=None, use_db=True):
...
+            InteractiveRunPipelineStep.query.delete()
+            InteractiveRunImageMapping.query.delete()
+            db.session.commit()
 
     # Register blueprints.
     app.register_blueprint(api, url_prefix="/api")
 
+    # create celery database if needed
+    if not database_exists(CONFIG_CLASS.result_backend_sqlalchemy_uri):
+        create_database(CONFIG_CLASS.result_backend_sqlalchemy_uri)
```
yannickperrenet, line 53: To me the `use_db` kwarg is somewhat confusing if this database creating is put outside of it regardless. Should probably also have a docstring to explain why the kwarg is there in the first place. Seems that its only used for [the celery tasks](https://github.com/orchest/orchest/blob/master/services/orchest-api/app/app/core/tasks.py#L26).

services/orchest-ctl/app/config.py
```diff
@@ -149,6 +153,19 @@
...
+    "postgres:13.1": {
+        "name": "orchest-database",
+        "environment": {
+            "PGDATA": "/userdir/.orchest/database/data",
+            "POSTGRES_HOST_AUTH_METHOD": "trust",
+        },
+        "mounts": [
+            {
+                "source": os.path.join(ENVS["HOST_USER_DIR"], ".orchest"),
+                "target": "/userdir/.orchest",
```
yannickperrenet, line 165: The `"source"` should be more precise, e.g. `".orchest/database"`, similarly to how `rabbitmq` is setup. Now the `postgres` container can see the entire `.orchest` directory including `kernels` and `rabbitmq-mnesia`.

Depending on the interaction between the `"target"` , `"PGDATA"` and the `postgres` container this should be as precise as possible.

services/orchest-ctl/app/cmdline.py
```diff
@@ -124,7 +125,17 @@ def start():
         run_config = utils.convert_to_run_config(container_image, container_spec)
 
         logging.info("Starting image %s" % container_image)
-        docker_client.containers.run(**run_config)
+        container = docker_client.containers.run(**run_config)
+
+        # wait for the db to be accepting connections before launching
+        # other containers, this will likely take 1 try or two.
+        # TODO: should we have a generic abstraction when it comes to
+        # dependencies among the services? I don't think it's needed.
```
yannickperrenet, line 133: Its good that you note it here, because we do have a similar dependency between the `jupyter-enterprise-gateway` and the `jupyter-server`. So good to keep in mind that if this keeps reoccurring then we should have a more robust fix.

services/auth-server/app/app/__init__.py
fruttasecca: Note that `db.create_all()` will only attempt to create tables, not the database itself. Agree with the fact that I should add some comments in there to better explain what's happening.

yannickperrenet (changes requested):
The models in the `orchest-api` should, in my opinion, be reworked because we now have one database and therefore the models don't have to be completely isolated anymore.

The schema can be simplified as now there can be inheritance between, e.g. interactive and non-interactive runs. This is a reasonably big schema change and should be done before we ship with the new db implementation (as otherwise we are looking at a difficult migration).

services/auth-server/app/app/__init__.py
yannickperrenet: Okay good, because the `db.create_all()` was creating the databases before. Or am I missing something?

services/orchest-api/app/app/__init__.py
fruttasecca: Essentially, I have tried to maintain the behaviour where `interactive_*` entities would not really be persisted between different Orchest instances, e.g. restarts. Since those tables are now persisted in the `orchest-database` service and not in a `sqlite` file within the `orchest-api` container, those tables need to be emptied "manually".
I preferred to do this on start so that the `interactive_*` tables would still be inspectable when shutting down Orchest, for easier debugging. 

This might indeed cause those tables to be cleared when the `orchest-api` is running in dev mode and restarts, so clearing only in non dev mode is probably preferable.

services/orchest-api/app/app/__init__.py
fruttasecca: Indeed.

services/orchest-ctl/app/config.py
fruttasecca: Agreed.

services/orchest-ctl/app/cmdline.py
fruttasecca: Yep.

fruttasecca:
Agreed that this is the right moment (and PR) to change schemas to avoid issues down the line, will do.

services/auth-server/app/app/__init__.py
fruttasecca: It did, with `sqlite`. But this is not true for all db flavors. I think it's, actually, only true for `sqlite`, but I can't find the exact specification in the docs.

services/orchest-api/app/app/__init__.py
yannickperrenet: And why do we need to clean those entries? Now that we are moving to a different database I would say we can change that behavior as well. I think we should simply store everything.

fruttasecca: Fine with me.

services/orchest-api/app/app/celery_app.py
```diff
@@ -1,9 +1,15 @@
 from celery import Celery
+from sqlalchemy_utils import create_database, database_exists
 
 from config import CONFIG_CLASS
 
 
-def make_celery(app):
+def make_celery(app, create_backend_db=False):
```
yannickperrenet, line 7: Maybe too much nutpicker but I would propose `use_backed_db` instead of `create_backend_db` since it is not always creating a backend db, the caller of the function just wants to use it.

services/orchest-api/app/app/apis/namespace_experiments.py
```diff
@@ -262,10 +266,12 @@ class PipelineRun(Resource):
...
-        non_interactive_run = models.NonInteractiveRun.query.get_or_404(
-            ident=(experiment_uuid, run_uuid),
-            description="Given experiment has no run with given run_uuid",
-        )
+        non_interactive_run = models.NonInteractivePipelineRun.query.filter_by(
+            run_uuid=run_uuid,
+        ).one_or_none()
+        if non_interactive_run is None:
+            abort(404, "Given experiment has no run with given run_uuid")
+        print(non_interactive_run.__dict__)
```
yannickperrenet, line 274: Could you delete this `print` statement?

fruttasecca: Yep, totally missed that, my bad.

yannickperrenet (approved):
Love it!

services/orchest-api/app/app/models.py
```diff
@@ -59,248 +88,188 @@ def __repr__(self):
...
+        # True usually implies an ON DELETE <CASCADE|SET NULL> rule is
+        # in place which will handle updating/deleting child rows on the
+        # database side.
+        passive_deletes=True,
+        # https://docs.sqlalchemy.org/en/14/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships
+        # In order to use ON DELETE foreign key cascades in conjunction
+        # with relationship(), it’s important to note first and foremost
+        # that the relationship.cascade setting must still be configured
+        # to match the desired “delete” or “set null” behavior
+        # Essentially, the specifed behaviour in the FK column
```
ricklamers, line 130: specified*

services/orchest-api/app/app/models.py
```diff
@@ -20,6 +21,34 @@ def as_dict(self):
...
+
+    # https://stackoverflow.com/questions/63164261/celery-task-id-max-length
+    build_uuid = db.Column(db.String(36), primary_key=True, unique=True, nullable=False)
+    project_uuid = db.Column(db.String(36), nullable=False, index=True)
+    environment_uuid = db.Column(db.String(36), nullable=False, index=True)
+    project_path = db.Column(db.String(4096), nullable=False, index=True)
+    requested_time = db.Column(db.DateTime, unique=False, nullable=False)
+    started_time = db.Column(db.DateTime, unique=False, nullable=True)
+    finished_time = db.Column(db.DateTime, unique=False, nullable=True)
+    status = db.Column(db.String(15), unique=False, nullable=True)
```
ricklamers, line 46: Wasn't this supposed to be more than 15?

ricklamers (approved):

yannickperrenet (approved):

services/orchest-api/app/app/__init__.py
```diff
@@ -50,13 +52,21 @@ def create_app(config_class=None, use_db=True):
...
+            # In case of an ungraceful shutdown, these entities could be
+            # in an invalid state, so they are deleted, since for sure
+            # they are not running anymore.
+            # To avoid the issue of entities being deleted because of a
+            # flask app reload triggered by a dev mode code change, we
+            # attempt to create a directory first. Since this is an
+            # atomic operation that will result in an error if the
+            # directory is already there, this cleanup operation will
+            # run only once per container.
+            if os.system("mkdir /tmp/interactive_cleanup_done > /dev/null 2>&1") == 0:
```
yannickperrenet, line 64: Can it not be the `os.mkdir` function instead of issues an os command?

fruttasecca: Will do

PR #97, opened 2020-12-18
`userdir/` permission rework

ricklamers:
This PR changes how permissions are handled for the `userdir/`

Instead of relying on a daemon to monitor and correct permissions (slow, doesn't scale). We've unified the permissions of the `userdir/` to allow for seamless sharing of files between the host and the containers that need access to the `userdir/` contents.

The `userdir/` contains projects, experiments, and user persisted state of the Orchest application (in `userdir/.orchest`).

Using Docker's `group_add` mechanism in conjunction with the `setgid` flag on the `userdir/` and the appropriate umask (`002`) makes sure that files/directories created in the `userdir/` from either a container or the host will be readable/writeable (and executable in the case of directories).

You should notice a significant decrease in resource consumption of Orchest on your machine as the permission daemon has been purged from the codebase.


services/base-images/base-kernel-julia/Dockerfile
```diff
@@ -24,4 +24,9 @@ RUN pip install -r requirements.txt
 COPY ./runnable-shared/bootscript.sh /orchest/bootscript.sh
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+USER $NB_USER
+ENV HOME=/home/$NB_USER
```
yannickperrenet, line 30: I think these lines should move up in the Dockerfile because the above `ARG` is invalidated almost always (on every new commit).

services/base-images/base-kernel-py-gpu/Dockerfile
```diff
@@ -22,4 +22,9 @@ RUN pip install -r requirements.txt
 COPY ./runnable-shared/bootscript.sh /orchest/bootscript.sh
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+USER $NB_USER
+ENV HOME=/home/$NB_USER
```
yannickperrenet, line 28: Same goes for these lines.

services/base-images/base-kernel-py/Dockerfile
```diff
@@ -22,4 +22,9 @@ RUN pip install -r requirements.txt
 COPY ./runnable-shared/bootscript.sh /orchest/bootscript.sh
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+USER $NB_USER
+ENV HOME=/home/$NB_USER
```
yannickperrenet, line 28: Just stating it again for completeness that these lines should move up.

services/base-images/base-kernel-r/Dockerfile
```diff
@@ -24,4 +24,9 @@ RUN pip install -r requirements.txt
 COPY ./runnable-shared/bootscript.sh /orchest/bootscript.sh
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+USER $NB_USER
+ENV HOME=/home/$NB_USER
```
yannickperrenet, line 30: .

services/orchest-api/Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM tiangolo/meinheld-gunicorn-flask:python3.7
 LABEL maintainer="Orchest B.V. https://www.orchest.io"
 
 # Get all requirements in place.
-COPY ./requirements.txt ./start.sh /orchest/services/orchest-api/
+COPY ./requirements.txt /orchest/services/orchest-api/
+COPY ./start.sh /
```
yannickperrenet, line 6: Since the `start.sh` is no longer copied to the same directory as the `requirements.txt` (and thus combining the COPY command to reduce the number of Docker layers), it should be moved below the installation of the requirements as it could otherwise end up invalidating the cache leading to installation of the requirements (increasing rebuild time).

services/orchest-api/app/main.py
```diff
@@ -1,9 +1,15 @@
...
+import subprocess
+import sys
+import logging
 
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
 app = create_app(config_class=CONFIG_CLASS)
-
+logging.info(
+    "Running orchest-api as %s" % subprocess.check_output("whoami", shell=True).decode()
```
yannickperrenet, line 11: What is the benefit of running this in a subprocess over `os.command`?

services/orchest-ctl/app/utils.py
```diff
@@ -194,6 +194,16 @@ def clean_containers():
             container.remove()
 
 
+def fix_userdir_permissions():
+    # setgid on all directories in userdir to make sure new files
+    # created by containers are read/write for sibling containers
+    # and the host user
```
yannickperrenet, line 200: I think this comment should be a docstring instead.

services/orchest-webserver/app/app/res/kernels/launch_docker.py
```diff
@@ -59,83 +57,48 @@ def launch_docker_kernel(kernel_id, response_addr, spark_context_init_mode):
...
-        orchest_mounts = get_orchest_mounts(
-            project_dir=_config.PROJECT_DIR,
-            host_project_dir=param_env.get("ORCHEST_HOST_PROJECT_DIR"),
-        )
-        volume_source, volume_spec = get_volume_mount(
-            param_env.get("ORCHEST_PIPELINE_UUID"),
-            param_env.get("ORCHEST_PROJECT_UUID"),
-        )
-        orchest_mounts[volume_source] = volume_spec
+    print("Started Jupyter kernel in normal docker mode")
```
yannickperrenet, line 66: Just so that I understand. Bu the case of `if swarm_mode` is removed because we do not run Swarm anyways?

Removing the section might make the `launch_docker.py` less portable to newer version of the EG or do you feel that the file is such an integrated part of Orchest we should make it is minimal as possible (as with regular code)? No real preference on my part, just trying to understand the change.

services/base-images/base-kernel-julia/Dockerfile
```diff
@@ -24,4 +24,9 @@ RUN pip install -r requirements.txt
 COPY ./runnable-shared/bootscript.sh /orchest/bootscript.sh
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
```
yannickperrenet, line 28: Should we use the `visudo` command instead to change the `sudoers.d` file? According to `/etc/sudoers.d/README`:
> Finally, please note that using the visudo command is the recommended way
to update sudoers content, since it protects against many failure modes.
See the man page for visudo for more information.

services/base-images/runnable-shared/bootscript.sh
```diff
@@ -3,11 +3,10 @@
 # 1) the kernel for Jupyter or
 # 2) the run.py to execute the container as a pipeline step.
 
+umask 002
```
yannickperrenet, line 6: Please correct me if I am wrong, but explicitly setting `umask 002` sets the "write" permission for other? 

Moreover, what exactly does this solve?

services/file-manager/Dockerfile
```diff
@@ -4,8 +4,12 @@ LABEL maintainer="Orchest B.V. https://www.orchest.io"
...
-RUN ["/filebrowser", "users", "add", "--perm.admin", "admin", "admin"]
+RUN /filebrowser config init --branding.name "File Manager" --auth.method noauth --branding.files /static && \
+    /filebrowser users add --perm.admin admin admin && \
+    chmod o+rw /database.db
 
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
+
+ENTRYPOINT ["/bin/sh"]
+CMD ["-c", "umask 002 && /filebrowser"]
```
yannickperrenet, line 15: Why is the command split between the `ENTRYPOINT` and `CMD` instead of putting it all inside the `ENTRYPOINT`? We are not issuing commands to the container that should overwrite the `CMD`, right?

services/jupyter-server/Dockerfile
```diff
@@ -30,4 +30,6 @@ WORKDIR /project-dir
 
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
-ENTRYPOINT [ "jupyter", "lab" ]
+
+COPY start.sh /
+ENTRYPOINT [ "/start.sh" ]
```
yannickperrenet, line 35: Can these statements be changed similar to how the Dockerfile for the `file-manager` is done? (After that discussion is resolved of course.) I like that approach better than having a separate shell script.

services/base-images/runnable-shared/bootscript.sh
ricklamers: It sets read permission for other. The default is 022, and we want the group to have `r/w` and `r/w/x` for directories. This happens with umask 002 because by default directories are 0777 and files are 0666 (and then the umask is applied).

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -231,7 +231,7 @@ def write_environment_dockerfile(
     # and cause the docker build to fail, as it should.
     # the bash script is removed so that the user won't be able to see it after the build is done
     statements.append(
-        f"RUN cd \"{os.path.join('/', work_dir)}\" && chmod +x {bash_script} && echo \"{flag}\" && bash {bash_script} && echo \"{flag}\" && rm {bash_script}"
+        f"RUN cd \"{os.path.join('/', work_dir)}\" && sudo chmod +x {bash_script} && echo \"{flag}\" && bash {bash_script} && echo \"{flag}\" && sudo rm {bash_script}"
```
yannickperrenet, line 234: I would suggest writing it like:
```python
f'RUN cd "{os.path.join("/", work_dir)}" '
f'&& sudo chmod +x {bash_script} '
f'&& echo "{flag}" '
f'&& bash {bash_script} '
f'&& echo "{flag}" '
f'&& sudo rm {bash_script}'
```

services/jupyter-server/Dockerfile
ricklamers: That doesn't work in this case because of escaping since we need to insert the umask in the same bash context as the jupyter lab command (requiring -c) but we also need to capture the Docker commands (with "$@").

services/orchest-webserver/app/static/js/components/PipelineDetails.js
```diff
@@ -54,11 +54,7 @@ class PipelineDetails extends React.Component {
         return {
           paneWidth: Math.max(
             0,
-            Math.max(
-              50,
-              state.paneWidth -
-                e.originalEvent.movementX / window.devicePixelRatio
-            )
+            Math.max(50, state.paneWidth - e.originalEvent.movementX)
```
yannickperrenet, line 57: Is this change intended here? It probably is, but just checking.

services/orchest-webserver/app/main.py
```diff
@@ -2,11 +2,17 @@
...
 import sys
+import subprocess
 
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
 from app import create_app, create_app_managed
 
+logging.info(
+    "Running orchest-webserver as %s"
+    % subprocess.check_output("whoami", shell=True).decode()
```
yannickperrenet, line 13: Just refering to https://github.com/orchest/orchest/pull/97/files#r545794522

services/orchest-webserver/app/static/js/components/PipelineDetails.js
ricklamers: Good catch, yes was there on purpose. Noticed this was wrong after reading up on the movementX spec (it's a numeric value in CSS pixels - meaning devicePixelRatio corrected) and seeing the existing solution misbehave on a retina display (non 1 window.devicePixelRatio).

services/orchest-api/app/main.py
ricklamers: os.system doesn't allow you to capture the output

services/base-images/base-kernel-julia/Dockerfile
ricklamers: I've put the RUN at the top, but the USER/ENV HOME actually need to be at the bottom to make sure the previous commands are executed as root. I've put those two above ARG/ENV though.

services/orchest-webserver/app/app/res/kernels/launch_docker.py
ricklamers: The launch_docker.py is highly custom to Orchest already, and indeed I removed swarm_mode because we don't support it yet. This way the file becomes a lot cleaner.

services/base-images/base-kernel-julia/Dockerfile
ricklamers: visudo actually starts the interactive vim editor AFAIK which isn't desirable for programmatic.

I pretty much saw appending to the sudoers.d/<user> file being recommended in multiple places: https://stackoverflow.com/questions/8784761/adding-users-to-sudoers-through-shell-script

Do you know how `visudo` can be invoked to append to the sudoers.d?

services/base-images/runnable-shared/bootscript.sh
yannickperrenet: Given that the default is `022` it makes sense now. Thanks :)

services/orchest-api/app/main.py
yannickperrenet: My bad, should have looked closer.

services/orchest-webserver/app/static/js/components/PipelineDetails.js
ricklamers: Moved this to hotfix on master.

orchest
```diff
@@ -44,9 +44,12 @@ if [ "$1" == "update" ] ; then
 fi
 # End of warnings
 
-
 HOST_CONFIG_DIR=$HOME/.config/orchest
 HOST_USER_DIR=$DIR/userdir
+HOST_OS="linux"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    HOST_OS="darwin"
+fi
```
yannickperrenet, line 52: Could you add as a comment why the `HOST_OS` is needed?

ricklamers: On macOS we still want to set the GID to a default (of 100) because we can't get the mac host's userdir/ GID because of how Docker for Mac mounts.

services/base-images/base-kernel-py/Dockerfile
```diff
@@ -2,6 +2,7 @@
 FROM elyra/kernel-py:2.3.0
 
 USER root
+RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
```
fruttasecca, line 5: I see some comments explaining what and why is happening in the docker file might help in the future.

services/orchest-ctl/app/cmdline.py
```diff
@@ -100,6 +100,9 @@ def start():
     # Clean up lingering, old images from previous starts.
     utils.clean_containers()
 
+    # Make sure userdir/ permissions are correct
+    utils.fix_userdir_permissions()
```
fruttasecca, line 104: Does this make it so that users coming from an old release (and an old repo) will experience no issues?

services/jupyter-server/Dockerfile
yannickperrenet: Isn't that solved when running `ENTRYPOINT ["/bin/sh", "-c", "umask 002 && /start.sh"]`?

orchest
yannickperrenet: Yea I got that, I meant to add it as a comment in the script ;)

Since users might very well inspect the shell script before running it (as that is considered best practice) and it is good to have a comment there explaining why the `HOST_OS` is needed.

services/base-images/base-kernel-julia/Dockerfile
yannickperrenet: I agree with you and agree with leaving it like you implemented. Just wanted to state what was inside the `REAMDE`, but running `vi` doesn't make sense running it "programmatically".

services/jupyter-server/Dockerfile
ricklamers: I don't think that's better. `start.sh` is still there and I prefer umask to be in the bash script.

orchest
ricklamers: Ah, misread. Fixing it.

services/jupyter-server/Dockerfile
yannickperrenet: Since the `orchest-api` and `orchest-webserver` also use a script, I am fine with leaving it the way it is now.

services/orchest-ctl/app/cmdline.py
ricklamers: Not really, it's there for setting the correct GID when orchest is first cloned (`git clone` doesn't set the group bit).

ricklamers: Although it will fix the userdir/ for existing users, not the contents of that directory though (the projects and the files in there).

ricklamers: I think for this change it's actually okay as per the migrations we will have breaking changes in the userdir/ anyway right?

ricklamers: (because the alternative would be to set permissions for all file on startup which can be expensive and I'd prefer `orchest start` to be as cheap as possible)

fruttasecca: I understand, lgtm.

PR #98, opened 2020-12-18
Enable schema migrations

fruttasecca:
This PR enables the management of schema migrations, so that, from now on, outdated databases will be brought to an up to date schema automatically.

From a developer perspective, managing migrations will require running `scripts/migration_manager.sh` to issue migration commands. Generally speaking, the only needed command will be the one to generate new revision files:
`scripts/migration_manager.sh <service-name> migrate`


scripts/migration_manager.sh
```diff
@@ -0,0 +1,60 @@
...
+  usage
+  exit 1
+fi
+
+# the rest of the commands will be passed to the migration script of the
+# service
+COMMANDS="${@}"
+docker exec ${SERVICE} python migration_manager.py db ${COMMANDS}
+code=$?
+if [ $code -eq 0 ]; then
```
ricklamers, line 48: I guess the command will output its errors on the terminal that invoked this bash file, hence you're ignoring the != 0 exit code case?

services/orchest-webserver/app/app/models.py
```diff
@@ -54,16 +56,24 @@ class Environment(db.Model):
     __tablename__ = "environments"
 
     uuid = db.Column(
-        db.String(255), unique=True, nullable=False, primary_key=True, default=str_uuid4
+        db.String(255),
+        unique=True,
+        nullable=False,
+        primary_key=True,
+        # using a server_default would require to install a postgres
+        # plugin, leaving it as an ORM-only default
```
ricklamers, line 64: What will the actual default value in the DB be in this case?

ricklamers (commented):
The env.py, script.py.mako, and migration_manager.py seem to be duplicated for each service. Can we be smarter about this?

ricklamers:
Overall I think this is an _excellent_ PR that will add lots of value to users by not breaking the database state when we have to make schema changes!

services/orchest-webserver/app/app/models.py
fruttasecca: No default value at the database schema level. Since it is non nullable, it simply will not allow to input records with no uuids, e.g. (note the missing `uuid` column):
`insert into environments (name, project_uuid, language, setup_script, base_image, gpu_support) values ('name', 'proj', 'english', '1', 'my_image', '0');`
Will result in a error. The default (generating a uuid) is of course still present at the ORM level. It makes for a less strict schema but installing a plugin did not seem like a better trade-off.

scripts/migration_manager.sh
fruttasecca: Yes, that's the case.

fruttasecca:
> The env.py, script.py.mako, and migration_manager.py seem to be duplicated for each service. Can we be smarter about this?

`migration_manager.py` contains a minimal amount of lines of code and cannot be exactly the same for each service. Note, for example, that the one in the `orchest-webserver` directory is slightly different because of what `init_app_` is returning. I see it as a sibling of `main.py`.

The `migrations` directory of each service that requires migrations needs to be there for `flask-migrate`/`alembic`. When issuing a `migrate` command you actually pass the `migrations` directory, not the `revisions` one. It's going to pick up what's in there, like `env.py`, and save to `revisions`.

We might be able to avoid duplication by having a single copy of those files in the repo and `COPY`ing them in the dockerfile, but that would mean not having them in place in `dev` mode I believe, and possibly not being able to customize them for each service (not that I think we are going to need to do that).

services/orchest-api/app/app/__init__.py
```diff
@@ -47,10 +47,13 @@ def create_app(config_class=None, use_db=True):
             create_database(app.config["SQLALCHEMY_DATABASE_URI"])
 
         db.init_app(app)
+        # necessary for migration related stuff
```
ricklamers, line 50: I'd prefer not to use the word 'stuff' >.<
Can we just say `# necessary for migration`

services/auth-server/app/app/__init__.py
```diff
@@ -23,10 +24,13 @@ def create_app(config_class=None):
         create_database(app.config["SQLALCHEMY_DATABASE_URI"])
 
     db.init_app(app)
+    # necessary for migration related stuff
```
ricklamers, line 27: Same as before

services/orchest-webserver/app/app/__init__.py
```diff
@@ -124,18 +125,13 @@ def create_app():
         create_database(app.config["SQLALCHEMY_DATABASE_URI"])
     db.init_app(app)
     ma.init_app(app)
+    # necessary for migration related stuff
```
ricklamers, line 128: Same as before

ricklamers: https://user-images.githubusercontent.com/1309307/102649977-6f9f9400-416a-11eb-81a1-20830c2612aa.mov

services/auth-server/app/app/connections.py
```diff
@@ -1,3 +1,9 @@
 from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy import MetaData
 
-db = SQLAlchemy()
+from _orchest.internals import config
+
+# this will make it so that constraints and indexes follow a certain
+# naming pattern
+metadata = MetaData(naming_convention=config.database_naming_convention)
```
yannickperrenet, line 8: For consistency and explicitness that the `config` is actually the internal library config.
```suggestion
from _orchest.internals import config as _config

# this will make it so that constraints and indexes follow a certain
# naming pattern
metadata = MetaData(naming_convention=_config.database_naming_convention)
```

services/auth-server/Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - && apt-get install -
 
 # Get all Python requirements in place and install them.
 COPY ./requirements.txt /orchest/services/auth-server/
+COPY ./lib /orchest/lib
```
yannickperrenet, line 12: The javascript library is already copied in in a later step and should not invalidate the cache for installing the Python requirements.

```suggestion
COPY ./lib/python /orchest/lib/python
```

services/auth-server/app/migration_manager.py
```diff
@@ -0,0 +1,15 @@
...
+
+# Used to issue migration commands to a running service (container),
+# e.g. docker exec commands. See /scripts/migration_manager.sh for more
+# info.
+
+if __name__ == "__main__":
+    app = create_app(config_class=CONFIG_CLASS)
+    manager = Manager(app)
+    manager.add_command("db", MigrateCommand)
+    manager.run()
```
yannickperrenet, line 15: The `Manager` seems to be started when running the `scripts/migration_manager.sh` but it is never shut down? Is that on purpose and if so could you explain why?

I think it can be beneficial to not shut down the manager afterwards, but want to better understand your reasoning behind it (if it is indeed still running).

services/orchest-api/app/app/connections.py
```diff
@@ -1,15 +1,20 @@
 from docker.client import DockerClient
+from sqlalchemy import MetaData
 from flask_sqlalchemy import SQLAlchemy
 
+from _orchest.internals import config
```
yannickperrenet, line 5: Would like to refer to the same suggestion for the auth-server adding `as _config`.

services/orchest-webserver/app/app/connections.py
```diff
@@ -1,5 +1,11 @@
...
+from sqlalchemy import MetaData
 from flask_sqlalchemy import SQLAlchemy
 
-db = SQLAlchemy()
+from _orchest.internals import config
+
+# this will make it so that constraints and indexes follow a certain
+# naming pattern
+metadata = MetaData(naming_convention=config.database_naming_convention)
+db = SQLAlchemy(metadata=metadata)
```
yannickperrenet, line 10: For consistency and explicitness that the `config` is actually the internal library config.
```suggestion
from _orchest.internals import config as _config

# this will make it so that constraints and indexes follow a certain
# naming pattern
metadata = MetaData(naming_convention=_config.database_naming_convention)
```

yannickperrenet (changes requested):
The migrations are a great addition, improving the lives of the ("updating") user.

I only had a few minor suggestions and would like to first better understand the choice behind the `Manager` that seems to keep running.

services/auth-server/app/migration_manager.py
ricklamers: Doesn’t this script only run when you perform the migration?

fruttasecca: The script does indeed only run when performing a migration, it's pretty much just parsing the arguments and passing them to the related function.
https://github.com/smurfix/flask-script/blob/master/flask_script/__init__.py#L396

yannickperrenet (approved):

PR #99, opened 2020-12-21
Feature improve SDK robustness

fruttasecca:
This PR aims to improve the robustness of the Orchest SDK. It adds some `try/except` blocks and some customs Exceptions so that errors are more explicit.


ricklamers:
LGTM!

ricklamers (approved):

orchest-sdk/python/orchest/errors.py
```diff
@@ -2,6 +2,42 @@ class Error(Exception):
...
+class DiskDeserializationError(DeserializationError):
+    """DeserializationError for disk objects (files)."""
+
+    pass
+
+
+class MemoryDeserializationError(DeserializationError):
+    """DeserializationError for in-memory objects."""
+
+    pass
```
yannickperrenet, line 38: To me these classes are a bit overkill. I would not go more granular than `DeserializationError` or simply `SerializationError` (and also removing `DeserializationError`).

orchest-sdk/python/orchest/errors.py
```diff
@@ -57,3 +93,13 @@ class DataInvalidNameError(Error):
     """Data has an invalid name"""
```
yannickperrenet, line 93:
```suggestion
"""Data has an invalid name."""
```

orchest-sdk/python/orchest/errors.py
```diff
@@ -57,3 +93,13 @@ class DataInvalidNameError(Error):
...
 
     pass
+
+
+class InvalidMetaDataError(Error):
+    """Metadata is invalid
+
+    For metadata that is missing the separator, that does not contain
+    the right amount of elements, or that has invalid content.
+    """
```
yannickperrenet, line 103:
```suggestion
    """Metadata is invalid."""
```
Not sure the second paragraph of the error has an added benefit since when using the `InvalidMetaDataError` the message already explain why the metadata is valid, e.g. it is missing the separator.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -12,19 +12,39 @@
...
     MemoryOutputNotFoundError,
+    OutputNotFoundError,
     ObjectNotFoundError,
     OrchestNetworkError,
+    PipelineDefinitionNotFoundError,
+    SerializationError,
     StepUUIDResolveError,
-    InputNameCollisionError,
-    DataInvalidNameError,
 )
```
yannickperrenet, line 28: Now that we are importing so many error it would make more sense if we do `from orchest import errors` and use it like `errors.OrchestNetworkError` instead.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -43,19 +63,67 @@ def _check_data_name_validity(name: Optional[str]):
...
+
+    Raises:
+        InvalidMetaDataError: If the input string is invalid.
 
+    Returns:
+        A tuple of 3 strings. The first is the timestamp of when the
+            data related to the metadata was produced (utc, ISO format).
+            The second string is the type of serialization used (See the
+            Serialization enum). The third string is the name with which
+            the data was output.
```
yannickperrenet, line 88: This indentation level can be reduced.
```suggestion
        A tuple of 3 strings. The first is the timestamp of when the
        data related to the metadata was produced (utc, ISO format).
        The second string is the type of serialization used (See the
        Serialization enum). The third string is the name with which the
        data was output.
```

orchest-sdk/python/orchest/transfer.py
```diff
@@ -43,19 +63,67 @@ def _check_data_name_validity(name: Optional[str]):
         )
 
 
-class Serialization(Enum):
-    """Possible types of serialization.
+def _interpret_metadata(metadata: str) -> Tuple[str, str, str]:
```
yannickperrenet, line 66: Even though I love this function, I am not sure I want output messages such as: "Metadata has an invalid timestamp". I don't see why a user should know this, the user never even wrote it in the first place and it is therefore not a helpful message to the user. Even though this is indeed the cause, we should give a different message to the user, e.g.: "Metadata is invalid. The project directory might have cached the old metadata when upgrading to the new version of Orchest."

orchest-sdk/python/orchest/transfer.py
```diff
@@ -43,19 +63,67 @@ def _check_data_name_validity(name: Optional[str]):
...
+    # Metadata that was stored on disk has 3 elements.
+    if len(metadata) in [3, 4]:
+        timestamp, serialization, name = metadata[-3:]
+
+        # check timestamp for correctness
+        try:
+            datetime.fromisoformat(timestamp)
+        except ValueError:
+            raise InvalidMetaDataError(
+                f"Metadata {metadata} has an" f"invalid timestamp ({timestamp})."
```
yannickperrenet, line 108:
```suggestion
                f"Metadata {metadata} has an invalid timestamp ({timestamp})."
```

yannickperrenet (changes requested):
Overall I do really like the idea of adding robustness. However, I think to the user many of the error messages are not "helping". Yes they state exactly what the problem is, but the user does not know how to fix it and did not directly cause the issue, we did.

ricklamers:
> Overall I do really like the idea of adding robustness. However, I think to the user many of the error messages are not "helping". Yes they state exactly what the problem is, but the user does not know how to fix it and did not directly cause the issue, we did.

I think it's good to keep in mind that even though the user might not fully understand all error messages - often it can be helpful if the user "just wants to fix it" that they know in which direction to look. Obviously many of these errors _should_ never be triggered at all. But when the ball is dropped, at least the user is empowered to find a solution.

yannickperrenet:
> I think it's good to keep in mind that even though the user might not fully understand all error messages - often it can be helpful if the user "just wants to fix it" that they know in which direction to look. 

I strongly disagree! The user has no influence over the metadata and therefore has no idea where to look (let alone how to fix, the user cannot even fix it in some of the cases). We need to tell the user where to look. 

For example (like I proposed in one of the other comments), instead of "Metadata has an invalid timestamp" we can do "Metadata is invalid. The project directory might have cached the old metadata when upgrading to the new version of Orchest." Although even better might even be: "Metadata is invalid, it has an invalid timestamp. The issue might be caused due to cached metadata when upgrading to the new version of Orchest, try to delete the `.orchest/pipelines` directory inside your pipeline directory."

I do think an error message should include what went wrong (so adding the "timestamp" in the example makes sense), but it should also steer the user in the right direction to fix it.

orchest-sdk/python/orchest/errors.py
fruttasecca: Agree

ricklamers:
I think we actually agree then. You just want to also add more detailed messages with the errors, which I think is a good idea.

fruttasecca:
> I strongly disagree! The user has no influence over the metadata and therefore has no idea where to look (let alone how to fix, the user cannot even fix it in some of the cases). We need to tell the user where to look.

I agree that the messages should be improved to steer the user towards fixing its issue.  I'll improve the messages to contain both the cause and a suggestion to fix. From my POV, the cause is there so that we can quickly find the reason an occurrence that should never happen has taken place.

orchest-sdk/python/orchest/transfer.py
fruttasecca: Will do.

orchest-sdk/python/orchest/errors.py
fruttasecca: I can do without having both `Disk/Memory DeserializationError`, but I think that at least `DeserializationError` should stay.

orchest-sdk/python/orchest/error.py
```diff
@@ -2,6 +2,30 @@ class Error(Exception):
...
+    """Error when an object cannot be serialized.
+
+    Essentially a wrapper around pickle.PicklingError.
+    """
+
+    pass
+
+
+class DeserializationError(Error):
+    """Error when an object cannot be serialized."""
```
yannickperrenet, line 24: Should probably be "deserialized"

orchest-sdk/python/orchest/error.py
```diff
@@ -54,6 +78,21 @@ class InputNameCollisionError(Error):
...
+    helper_message = (
+        "The issue might be caused due to cached metadata when upgrading to "
+        'the new version of Orchest, try to delete the ".orchest/pipelines" '
+        "directory inside your pipeline directory or try rerunning the steps "
+        "that produced data."
+    )
+
+    def __init__(self, msg="", *args, **kwargs):
+        msg = " ".join([msg, self.helper_message])
+        super().__init__(msg, *args, **kwargs)
```
yannickperrenet, line 98: Love it! :100:

orchest-sdk/python/orchest/error.py
fruttasecca: Indeed.

yannickperrenet (approved):

PR #100, opened 2020-12-21
Invalidate user environment on update

fruttasecca:
This PR makes it so that user environments are invalidated when Orchest updates, to avoid issues related to mismatching versions of the `Orchest SDK` in different environments.

### DIFF
--- a/services/orchest-ctl/app/cmdline.py
+++ b/services/orchest-ctl/app/cmdline.py
@@ -236,6 +236,21 @@ def update(language):
         # during the update to support _updateserver
         stop(skip_names=["nginx-proxy", "update-server"])
 
+    # Delete all user built environments. This is to avoid the issue of
+    # having environments with mismatching Orchest SDK versions.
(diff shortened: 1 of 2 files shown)


ricklamers (approved):

services/orchest-ctl/app/cmdline.py
```diff
@@ -220,6 +220,22 @@ def _updateserver():
     docker_client.containers.run(**run_config)
 
 
+def clear_environment_images():
```
yannickperrenet, line 223: I think this functions should be defined in `utils.py` instead.

yannickperrenet (changes requested):
Although in general I think the orchest-ctl could need some touch-ups, the new functions should in my opinion be defined in `utils.py`

Other than that a great addition!

services/orchest-ctl/app/cmdline.py
fruttasecca: Indeed.

yannickperrenet (approved):

PR #101, opened 2020-12-22
Fix orchest-webserver race conditions 1

fruttasecca:
This PR aims to fix some race conditions that were present in the `orchest-webserver`, to do so it requires a schema change for the `project` table of said service.


services/orchest-webserver/app/app/views/core.py
```diff
@@ -991,20 +1006,36 @@ def projects():
...
+        # created through the FS.
+        # By checking the FS after the db, we will avoid the race
+        # condition since deleting a project involves first deleting the
+        # directory, then deleting the db entries. If no db entries
+        # refer to the path and the path is there, then this is actually
+        # a project which has been created through the FS.
+        project_paths = [
+            name
+            for name in os.listdir(projects_dir)
+            if os.path.isdir(os.path.join(projects_dir, name))
```
yannickperrenet, line 1035: Even though I don't think it is necessary here, since the `projects_dir` is unlikely to contain many directories (indeed a good idea to defensibly check), but using [`os.scandir`](https://docs.python.org/3/library/os.html#os.scandir) can increase performance here. It does not issue another call (like `listdir` does) to check whether it is indeed a directory or not.

Whilst the difference in performance here is small, it is a small change and I figured it was good to note for future reference :)

yannickperrenet (approved):

ricklamers (approved):

PR #102, opened 2020-12-23
Add logging configuration to the Flask services

yannickperrenet:
This PR improves log messages by no longer logging directly to the `root` logger for Flask applications.  Additionally, logging can now be configured through the `dictConfig` in the top-level `__init__.py` of the services. For example, to set the log-level of a specific package such as `engineio`. 

For code that is directly called as part of the Flask application:
```python
from flask import current_app

current_app.logger.info("message")
```
In other cases, we still log directly to the `root` logger as we have a complex logging system with our concurrency settings (think subprocesses and Celery):
```python
import logging

logging.info("message")
````

### CODE REVIEW

File: services/auth-server/app/app/__init__.py [hunk 1]
```diff
@@ -15,9 +19,14 @@
 
 
 def create_app(config_class=None):
-    app = Flask(__name__)
+    app = Flask("auth-server")
```
fruttasecca, line 22: Seems like the app name has some implications related to templates and static files. In this particular case, the `auth-server` will not be able to find templates.

Not sure if we want to simply add the `template_folder` arg or actually rename the `app` directory since the app name has also other implications, [docs](https://flask.palletsprojects.com/en/1.1.x/api/#application-object) for reference.

```suggestion
    app = Flask("auth-server", template_folder="app/templates")
```

fruttasecca (changes requested):
The way we initialize the Flask application might require some changes since it has implications on the way Flask and its extensions might look for resources, log, etc. Depending on the solution we might be able to avoid repeating most if not all the content of the `logging_config` dictionary.

I think we have at least 3 alternatives:

- passing initialization arguments to the Flask app, which might also require making sure if and how extensions depend on the app name
- changing the name of the `app` directory to be an ad-hoc name for each service
- keep the name of the `app` directory, use `__name__` in the `logging_config` dictionary.

ricklamers:
This logging cleanup is great! I’d love to be able to re-use a single definition of the logging dict such that it’s consistent everywhere (my best guess is to have it defined in the internal orchest lib).

Jacopo makes a valid point about needing to understand the side effects of initializing the Flask apps with a different name. If we can achieve the above I think we’re ready to merge.

yannickperrenet:
You guys are making great points! I quickly want to touch upon the following two (which you are both making):

> my best guess is to have it defined in the internal orchest lib

This had crossed my mind also, however, since the different services have different dependencies for which we want to set logging configurations, I think we cannot move the logging configuration to the internal library. Logging seems like something you want to be able to control on a per-service basis and therefore moving it to the internal library might not make sense.

> The way we initialize the Flask application might require some changes since it has implications on the way Flask and its extensions might look for resources, log, etc.

We do need a better answer for the `__name__` of the Flask application. Indeed currently a simple rename from `__name__` to e.g. `"orchest-webserver"` would break the templates. When it comes to the logs it probably makes more sense to see `orchest-webserver` instead of `app`, especially if we end up writing a script or whatever to get an overview of all the logs of the different services (then we would see `app` for all the Flask applications).

services/orchest-webserver/app/app/__init__.py
```diff
@@ -183,7 +184,81 @@ def send_files(path):
...
+            },
+        },
+        "loggers": {
+            # NOTE: this is the name of the Flask app, since we use
+            # ``__name__``. Using ``__name__`` is required for the app
+            # to function correctly.
+            "app": {
+                "handlers": ["console"],
+                "level": os.getenv("ORCHEST_LOG_LEVEL", "INFO"),
+            },
```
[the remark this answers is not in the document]
yannickperrenet: > I am assuming that using __name__ is required because otherwise the orchest-webserver will not be able to find templates.

Yes, indeed!

> Whatever solution we come up with might allow us to have this logging_config dictionary in a single place, e.g. orchest-ctl or the orchest internal library.

Here I want to refer to [another reply in this PR](https://github.com/orchest/orchest/pull/102#issuecomment-751128423) that touches on this point

services/auth-server/app/app/__init__.py
yannickperrenet: I will look into the suggestions, right now I would say I really like the `template_folder` arg 👍

yannickperrenet:
Replying to the alternative that @fruttasecca proposed:

> keep the name of the app directory, use __name__ in the logging_config dictionary.

This is the approach I went with at this time. As the value `app` is descriptive enough since we are not bundling the logs of the different services together.

> changing the name of the app directory to be an ad-hoc name for each service

We could later go for this approach. I did not like it at this time as the `app` seems like a good default to have across all the services that run a Flask application and `orchest-webserver` would be an invalid name for a package (also according to PEP8: "Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.")

fruttasecca (commented):
lgtm.
I would like to understand if we can have the logging config in a single place. I assume that's currently not the case in case the logging options become more service specific.

yannickperrenet:
> I assume that's currently not the case in case the logging options become more service specific.

Indeed.

PR #105, opened 2021-01-07
Test Environment Building

fruttasecca:
This PR adds some tests to the environment building task of the `orchest-api`.


services/orchest-api/app/tests/test_environment_build.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+# String that should not appear in the logs.
+_NOT_TO_BE_LOGGED = "_NOT_TO_BE_LOGGED"
+
+
+@pytest.mark.parametrize(
+    "task_uuid",
+    ["task_uuid"],
+    ids=["basic"],
```
yannickperrenet, line 32: I think it makes more sense to delete this line `ids=["basic"]` as it does not add any valuable information to the test (yet is still printed in the output).
```bash
tests/test_environment_build.py::test_environment_build[[]-abort_task-image_in_env-basic-basic-basic-basic]
```

services/orchest-api/app/tests/test_environment_build.py
```diff
@@ -0,0 +1,273 @@
...
+    # monkeypatching docker.
+    monkeypatch.setattr(
+        app.core.environment_builds, "prepare_build_context", mock_prepare_build_context
+    )
+    # Logs will be written here.
+    monkeypatch.setattr(
+        app.core.environment_builds,
+        "__ENV_BUILD_FULL_LOGS_DIRECTORY",
+        "/tmp/output_environment_build",
+    )
```
yannickperrenet, line 195: Would it make sense to create a temporary directory (using [tempfile from standard library](https://docs.python.org/3/library/tempfile.html)) instead of hardcoding the path. What if the test does not complete successfully then the directory might not get cleaned.

services/orchest-api/app/tests/test_environment_build.py
```diff
@@ -0,0 +1,273 @@
...
+
+        expected_events = []
+        for event in build_events:
+            if event is None:
+                break
+            expected_events.append(event)
+        expected_events = "\n".join(expected_events)
+        assert expected_events in logged_events
+
+    # Successful tests can remove their log file.
```
yannickperrenet, line 267: Does this mean we want to keep the log files of unsuccessful tests? For inspection purposes? But what if it fails multiple times consecutively, can we still identify what log file was from which test run?

yannickperrenet:
The excessive need for mocking shows the complexity around environment builds. When the application is running, an unexpected error could be hiding in any of those calls.

I do like how the test makes it clearer what is actually going on inside the application. If you had to better understand environment builds, then the test could be a good start.

The test seems to cover what it should. So great job!

services/orchest-api/app/tests/test_environment_build.py
fruttasecca: The test not completing results indeed in the directory not being cleaned, which is intended, for inspection purposes.

services/orchest-api/app/tests/test_environment_build.py
fruttasecca: >Does this mean we want to keep the log files of unsuccessful tests?

Yep.

>But what if it fails multiple times consecutively, can we still identify what log file was from which test run?

That is indeed a mistake on my side, will fix it.

services/orchest-api/app/tests/test_environment_build.py
fruttasecca: Agree

services/orchest-api/app/tests/test_environment_build.py
fruttasecca: >Would it make sense to create a temporary directory (using tempfile from standard library) instead of hardcoding the path.

Don't think so, it would require more mocking, and if I understand the way the temporary directory would work we would end up not being able to inspect log files.

yannickperrenet: I agree that the current approach is preferred.

However, I think it is good to keep in mind that automated tests (for example on GitHub Actions) do not give us the same inspection options as when running them locally.

services/orchest-api/app/tests/test_environment_build.py
```diff
@@ -0,0 +1,265 @@
...
+            "events:",
+            str(build_events),
+            "-abort:",
+            str(abort),
+            "-image_in_local_environment:",
+            str(image_in_local_environment),
+        ]
+    )
+    environment_uuid = "environment_uuid"
+    project_path = "project_path"
```
yannickperrenet, line 64: Can this code be moved to under the functions and class definitions? I think it is a bit confusing this way.

PR #106, opened 2021-01-07
Improve Orchest SDK error messages.

fruttasecca:
This PR improves the error messages produced by the Orchest SDK when some input data cannot be found on disk and in the memory server. The resulting error message produced when an input cannot be found will contain the name of the pipeline step which supposedly produced the input.

### DIFF
--- a/orchest-sdk/python/orchest/transfer.py
+++ b/orchest-sdk/python/orchest/transfer.py
@@ -398,7 +398,7 @@ def _get_output_disk(step_uuid: str, serialization: str) -> Any:
         # TODO: Ideally we want to provide the user with the step's
         #       name instead of UUID.
         raise error.DiskOutputNotFoundError(
-            f'Output from incoming step "{step_uuid}" cannot be found on disk. '
+            f'Output from incoming step "{step_uuid}" cannot be found. '
             "Try rerunning it."
         )
     # IOError is to try to catch pyarrow failures on opening the file.
@@ -441,7 +441,7 @@ def _resolve_disk(step_uuid: str) -> Dict[str, Any]:
         # TODO: Ideally we want to provide the user with the step's
         #       name instead of UUID.
         raise error.DiskOutputNotFoundError(
-            f'Output from incoming step "{step_uuid}" cannot be found on disk. '
+            f'Output from incoming step "{step_uuid}" cannot be found. '
             "Try rerunning it."
         )
 
@@ -723,8 +723,8 @@ def _get_output_memory(step_uuid: str, consumer: Optional[str] = None) -> Any:
 
     except error.ObjectNotFoundError:
         raise error.MemoryOutputNotFoundError(
-            f'Output from incoming step "{step_uuid}" cannot be found in the '
-            "memory server. Try rerunning it."
+            f'Output from incoming step "{step_uuid}" cannot be found. '
+            "Try rerunning it."
         )
     # IOError is to try to catch pyarrow deserialization errors.
     except (pickle.UnpicklingError, IOError):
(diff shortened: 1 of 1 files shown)


orchest-sdk/python/orchest/transfer.py
```diff
@@ -398,7 +398,7 @@ def _get_output_disk(step_uuid: str, serialization: str) -> Any:
         # TODO: Ideally we want to provide the user with the step's
         #       name instead of UUID.
         raise error.DiskOutputNotFoundError(
-            f'Output from incoming step "{step_uuid}" cannot be found. '
+            f'Output from incoming step "{step_uuid}" cannot be found on disk. '
```
yannickperrenet, line 401: I am not so sure about the "on disk" and "in the memory server" additions as the type of the error, i.e. `DiskOutputNotFoundError` and `MemoryOutputNotFoundError` respectively, already tells the user where the error happened. I don't think the term "memory server" makes it any clearer and rather works confusing. Also the "on disk" is implied given the type of the error.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -866,11 +866,8 @@ def _resolve(
     # If no info could be collected, then the previous step has not yet
     # been executed.
     if not method_infos:
-        method_infos_exceptions = "\n".join(method_infos_exceptions)
-        raise error.OutputNotFoundError(
-            f'Output from incoming step "{step_uuid}" cannot be found. '
-            f"Try rerunning it. Info:\n{method_infos_exceptions}"
-        )
+        msg = "\n".join(method_infos_exceptions)
+        raise error.OutputNotFoundError(msg)
```
yannickperrenet, line 870: Wouldn't this always (since we (will) have `>=2` resolve methods) have the same error message (but different type) for every resolve method? Can't we just make it something like (or another message):
```python
raise error.OutputNotFoundError("Output could not be found in memory or on disk.")
```
to bundle it into a single message.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -963,9 +960,19 @@ def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> Dict[str,
...
+        try:
+            get_output_method, args, kwargs, metadata = _resolve(
+                parent_uuid, consumer=step_uuid
+            )
+        except error.OutputNotFoundError as e:
+            parent_title = parent.properties["title"]
+            msg = (
+                f'Output from incoming step "{step_uuid}" '
+                f'("{parent_title}") cannot be found. Try rerunning it. '
+                f"Info:\n{str(e)}"
```
yannickperrenet, line 973: I think we should do `title (step_uuid)` instead of the other way around.

Can we not remove `f"Info:\n{str(e)}"` since the error traceback should already show the previously raised error (which causes this one) together with the message. I don't think it is good practice to bubble up and join all messages together from throughout the traceback.

orchest-sdk/python/orchest/transfer.py
fruttasecca: If we go for this solution then we can definitively drop the "on disk" and "in the memory server" from the emitted exceptions. I'll go with this since it simplifies the code.

yannickperrenet (changes requested):
Great addition of the `parent_title` it was there for all along!

Only need some changes (or discussion) regarding the exact message for the exceptions.

orchest-sdk/python/orchest/transfer.py
fruttasecca: I am happy to remove the bubble up & join, felt like a code smell.

yannickperrenet (approved):

PR #107, opened 2021-01-08
Improve orchest api status updates race conditions

fruttasecca:
This PR makes it so that the status of an entity, such as a pipeline run or pipeline step, cannot be updated anymore once it has reached the state of ABORTED, FAILURE, or SUCCESS. This is to solve consistency and/or race conditions issues given by the fact that both the `orchest-api` and a celery task might be trying to PUT the status of an entity.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,6 +8,8 @@ ignore =
     # Allow breaks before/after binary operators
     W503,
     W504,
+    # Ignore invalid escape sequence, see services/orchest-api/app/app/core/environment_builds.py:262:55
+    W605
 
 # black is set to 88
 max-line-length = 88
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@ ignore =
     # Allow breaks before/after binary operators
     W503,
     W504,
-    # Ignore invalid escape sequence, see services/orchest-api/app/app/core/environment_builds.py:262:55
+    # Ignore invalid escape sequences.
     W605
 
 # black is set to 88
--- a/services/orchest-api/app/app/apis/namespace_environment_images.py
+++ b/services/orchest-api/app/app/apis/namespace_environment_images.py
@@ -1,9 +1,5 @@
-import logging
-
-from docker import errors
 from flask_restx import Namespace, Resource
 
-import app.models as models
 from _orchest.internals import config as _config
 from _orchest.internals.utils import docker_images_list_safe, docker_images_rm_safe
 from app.apis.namespace_environment_builds import (
@@ -33,7 +29,7 @@ api = register_schema(api)
 class EnvironmentImage(Resource):
     @api.doc("delete-environment-image")
     def delete(self, project_uuid, environment_uuid):
-        """Removes an environment image given project_uuid and image_uuid
+        """Removes an environment image given project and image uuids.
 
         Will stop any run or experiment making use of this environment.
         """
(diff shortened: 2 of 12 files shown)


.flake8
```diff
@@ -8,6 +8,8 @@ ignore =
     # Allow breaks before/after binary operators
     W503,
     W504,
+    # Ignore invalid escape sequence, see services/orchest-api/app/app/core/environment_builds.py:262:55
```
ricklamers, line 11: This exact line/character annotation will become outdated pretty quickly. I'd rather have an inline example of why it's a false alert.

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -33,7 +29,7 @@
 class EnvironmentImage(Resource):
     @api.doc("delete-environment-image")
     def delete(self, project_uuid, environment_uuid):
-        """Removes an environment image given project_uuid and image_uuid
+        """Removes an environment image given project and image uuids.
```
ricklamers, line 32: I think this should be `project_uuid` and `environment_uuid` or am I wrong?

.flake8
yannickperrenet: For me `# Ignore invalid escape sequence` would be enough also as you can read the flake8 docs for specifics on the listed error.

ricklamers (approved):
LGTM! Some minor comments about comments.

.flake8
fruttasecca: What about going for `# Ignore invalid escape sequence` and having a comment in the code reminding us to remove this error from the ignore list if/when we stop making use of the string that triggers the false positive?

ricklamers: I think I don't mind it indefinitely added to flake8 config if it triggers false positives.

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: Indeed, will fix.

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -827,9 +836,5 @@ async def run(
         for step in self.steps:
             step._status = "PENDING"
 
-        # Status will contain whether any failures occured during execution
-        # of the pipeline.
-        return status
```
yannickperrenet, line 832: Should this, instead of being removed, be
```python
return "SUCCESS"
```
similar to the environment builds?

.flake8
yannickperrenet: flake8 makes it rather easy to find the errors, does it not? For that reason I would just go with the "cleanest" approach of having as little about flake8 itself in comments anywhere in the code.

yannickperrenet (changes requested):
Got to love flake8 ;)

Just wanted to check about the final return call for pipeline runs. Other than that, looking good! How a minor change can have such an impact.

services/orchest-api/app/app/core/pipelines.py
fruttasecca: This result would not be used, since the celery task executing pipeline runs will return the result of `run_pipeline_async`, in `tasks.py` of the `orchest-api`. Also the celery task completion is not solely tied to the pipeline run completion, again, because of the body of  `run_pipeline_sync`.

ricklamers:
```
async def run_pipeline_async(run_config, pipeline, task_id):
    await asyncio.gather(
        *[
            asyncio.create_task(pipeline.run(task_id, run_config=run_config)),
            asyncio.create_task(
                check_pipeline_run_task_status(run_config, pipeline, task_id)
            ),
        ]
    )
    # any code that depends on the fact that both pipeline.run and
    # check_pipeline_run_task_status have terminated should be here.
    # for example, pipeline.run  PUTs the state of the run when it ends,
    # so any code dependant on the status being set cannot be run
    # in check_pipeline_run_task_status
    run_config["docker_client"] = docker_client
    pipeline.remove_containerization_resources(task_id, "docker", run_config)
    # The celery task has completed successfully. This is not related to
    # the success or failure of the pipeline itself.
    return "SUCCESS"
```

This code demonstrates why that return value is not used.

yannickperrenet: Yea my bad, I did not check out the branch locally when looking at the `tasks.py` file and missed the addition of the return statement there.

yannickperrenet (approved):

PR #109, opened 2021-01-14
Rework orchest-ctl

yannickperrenet:
This rework abstracts away the Docker calls to the `DockerWrapper` class, hiding whether the calls are made synchronously or asynchronously. Additionally, it is now a lot easier to mock out the Docker functionality to implement tests for the CLI and give the user more valuable responses.


ricklamers:
This looks like a nice cleanup to abstract out the Docker commands to help us with other runtime environments in the future.

I think the tests should be added to the GitHub Action - or is the already covered somehow?

ricklamers:
I still get TQDM alignment issues:

image

Perhaps we could fix this as part of this PR?

ricklamers:
The `./orchest update` command gives me:

```
From https://github.com/orchest/orchest
 * branch              HEAD       -> FETCH_HEAD
First, rewinding head to replay your work on top of it...

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'root@b608c642c029.(none)')
```

yannickperrenet:
> I still get TQDM alignment issues

Agree we can fix that in this PR. Would have to look into when exactly this is triggered, because sometimes it works and sometimes it does not.

> The ./orchest update command gives me

I wasn't sure whether this is caused by the rework, because the underlying git command was never changed. Could be because you are now on the feature branch and not master.

yannickperrenet:
> I still get TQDM alignment issues

Note sure we can fix the issue in your example, because it seems that the line wrapped in your terminal (because your width is so small). I will reduce the width of the progress bar to be equal to the `config.WRAP_LINES` constant, making it less likely that that edge case is hit.

Additionally, when you press enter whilst the bar is loading you can get something like:
```bash
Installing Orchest...
Pulling images: 0/2|                                                   |
Pulling images: 1/2|#########################5                         |

Pulling images: 2/2|###################################################|
```
Which I don't think we can solve.

fruttasecca:
Running `./orchest help` describes Orchest as `A tool for creating and running data science pipelines.`. Should that be changed?

Aside from that, lgtm.

yannickperrenet:
> The ./orchest update command gives me:

This is the case with the old update implementation as well as the new. Calling `./orchest update` will first pull the orchest-ctl image from DockerHub and the same error is thrown. Since the issue arises in both cases and the underlying `git-update.sh` file was not changed (apart from adding the `set -e` flag) I would guess the issue is due to: 

> Could be because you are now on the feature branch and not master.

@ricklamers @fruttasecca Closing thoughts on this matter? Otherwise I will merge.

ricklamers:
> > The ./orchest update command gives me:
> 
> This is the case with the old update implementation as well as the new. Calling `./orchest update` will first pull the orchest-ctl image from DockerHub and the same error is thrown. Since the issue arises in both cases and the underlying `git-update.sh` file was not changed (apart from adding the `set -e` flag) I would guess the issue is due to:
> 
> > Could be because you are now on the feature branch and not master.
> 
> @ricklamers @fruttasecca Closing thoughts on this matter? Otherwise I will merge.

After looking at it carefully I think the issue comes from the requirement of a name/email for running the `git-update.sh` script. Because the git repo (`.git`) could contain no project specific name and no global user/email is configured in the `orchest-ctl` image we sometimes get this error.

I went with setting a global user for the git configured in `orchest-ctl`.

ricklamers:
> > I still get TQDM alignment issues
> 
> Note sure we can fix the issue in your example, because it seems that the line wrapped in your terminal (because your width is so small). I will reduce the width of the progress bar to be equal to the `config.WRAP_LINES` constant, making it less likely that that edge case is hit.
> 
> Additionally, when you press enter whilst the bar is loading you can get something like:
> 
> ```shell
> Installing Orchest...
> Pulling images: 0/2|                                                   |
> Pulling images: 1/2|#########################5                         |
> 
> Pulling images: 2/2|###################################################|
> ```
> 
> Which I don't think we can solve.

The enter newline is ok imo. It seems to work well for me now. Since it's running through a container layer I'm not sure if we could actually go as far as detecting the the terminal size of the invoking terminal. 

This new WRAP_LINES value should be good enough for most people running terminals with that size or larger.

ricklamers:
Just as a note: I've created an `is_running(self, running_containers)` method that excludes `orchest-ctl` instead of that check being repeated in various methods.

ricklamers:
LGTM after the changes! Please check if you can all live with my changes.

Merge time 😃 .

yannickperrenet:
> Just as a note: I've created an is_running(self, running_containers) method that excludes orchest-ctl for the check, instead of that check-logic being repeated in various methods.

I think I will do a small refactor there, because now it adds an additional docker engine API call which is not necessary (and will slow down the CLI).

PR #111, opened 2021-01-14
Improve endpoints transactions

fruttasecca:
This PR aims to improve the readability, testability, and error handling of the `orchest-webserver` and `orchest-api`.

A common theme throughout a lot of endpoints is to have effects on the database and collateral effects that affect the filesystem, the docker environment, the celery worker queue, a background scheduler, etc. When those two things are intertwined it makes for code that is harder to debug and that can be more prone to leave the state inconsistent in case of error.

For this reason, the PR introduces a `TwoPhaseExecutor` and a `TwoPhaseFunction` class, to be used in the aforementioned endpoints. Without delving into implementation details, the execution of requests that affect both the database and have collateral effects is separated into two phases, a transactional one, and a collateral one. So that issues during the first can leave the state unadulterated. See `lib/python/orchest-internals/_orchest/internals/two_phase_executor.py` for more details.

Moreover, the PR pulls up database commits to higher levels of the call stack, as suggested by the good practises of `sqlalchemy`, introduces rollbacks where necessary and partially refactors the `orchest-webserver`. The PR also slightly modifies the schema of the database of the `orchest-webserver`.


lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,99 @@
...
+#             environment_build_uuid
+#         )
+# except Exception as e:
+#     current_app.logger.error(e)
+#     return {"message": str(e)}, 500
+#
+#
+# Note that you should not call commit during any implementation of
+# the transaction method, while you are free to do so if you need to
+# apply db changes during the collateral phase, and cannot do otherwise.
```
yannickperrenet, line 32: I would suggest using a module level docstring instead.

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __enter__(self):
+        return self
+
+    def revert(self, failed_idx):
+        for i in range(failed_idx, -1, -1):
+            try:
+                self.collateral_queue[i].revert()
+            except Exception as e:
+                logging.error(f"Error during revert call {i}: {e}")
```
yannickperrenet, line 50: I think this would end up using the root logger, instead we would want something like
```python
import logging

logger = logging.getLogger(__name__)

# And then do logging like:
logger.error(...)
```

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,99 @@
...
+        self.orig_transaction = self.transaction
+        self.transaction = self._transaction
+
+    def _transaction(self, *args, **kwargs):
+
+        self.tpe.collateral_queue.append(self)
+        return self.orig_transaction(*args, **kwargs)
+
+    def transaction(self, *args, **kwargs):  # pylint: disable=E0202
+        raise NotImplementedError()
```
yannickperrenet, line 93: I am a bit confused by this pattern. Can we make it into a decorator?

```python
def add_to_collateral_queue(func):
    def wrapper(*args, **kwargs):
        # Add to queue
        pass

        # This would then run the transaction
        return func(*args, **kwargs)
    return wrapper
```
Or is this not possible as we would not be apple to pass the `self.tpe` as an argument to the decorator?

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def collateral(self):
+        raise NotImplementedError()
```
yannickperrenet, line 96: Should we make this an abstract method?

services/orchest-api/app/app/apis/namespace_environment_builds.py
```diff
@@ -296,54 +221,139 @@ def get(self, project_uuid, environment_uuid):
...
+            kwargs=celery_job_kwargs,
+            task_id=self.task_id,
+        )
+
+
+class AbortEnvironmentBuild(TwoPhaseFunction):
+    def transaction(self, environment_build_uuid: str):
 
-    Environment builds that are in progress are stopped.
+        self.environment_build_uuid = environment_build_uuid
```
yannickperrenet, line 289: Just as a thought. Could the use of attributes on `self` backfire? Should we instead go with a `self.data = {}` that is defined in the `TwoPhaseFunction`?

Now every subclass of `TwoPhaseFunction` defines a different set of attributes. Generally speaking it is good practice to define all your attributes in the `__init__` function. Not sure whether that translates to this setting through.

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -27,36 +29,44 @@ def delete(self, project_uuid, pipeline_uuid):
...
         """
-        delete_pipeline(project_uuid, pipeline_uuid)
-        return {"message": "Pipeline deletion was successful"}, 200
+        try:
+            with TwoPhaseExecutor(db.session) as tpe:
+                DeletePipeline(tpe).transaction(project_uuid, pipeline_uuid)
 
+        except Exception as e:
+            current_app.logger.error(e)
+            return {"message": str(e)}, 500
```
yannickperrenet, line 38: This try block seems a bit of overkill as the exception is also caught and logged in by the `TwoPhaseExecutor`. Although I do understand the need for it to return a different message / status code based on failure.

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -40,51 +41,34 @@ def post(self):
...
+        except Exception as e:
+            current_app.logger.error(e)
+            # Error handling. If it does not succeed then the initial
+            # entry has to be removed from the database as otherwise no
+            # session can be started in the future due to the uniqueness
+            # constraint.
+            models.InteractiveSession.query.filter_by(
+                project_uuid=post_data["project_uuid"],
+                pipeline_uuid=post_data["pipeline_uuid"],
+            ).delete()
```
yannickperrenet, line 63: Is this something that we can add to the `CreateInteractiveSession` class by adding another notion to the `TwoPhaseFunction`, a certain method that should be run "on-failure"? Or is this case too rare to handle so generally?

fruttasecca: Indeed this should have been implemented by overriding `revert()` of the `TwoPhaseFunction` class.

yannickperrenet (changes requested):
@fruttasecca, you are an absolute machine! Great, great work :+1: 

The structure this PR brings to the code is invaluable in my opinion, so I am very glad to have it. 

I just had some minor comments to start a discussion.

services/orchest-api/app/app/apis/namespace_pipelines.py
fruttasecca: I think we can get away without further logging since the `TwoPhaseExecutor` will take care of that, but we need to have the try block there to catch the exception and have custom behaviour when there is a failure, unless we want to introduce more abstraction and functionality in the `TwoPhaseExecutor`.

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
fruttasecca: I'll look into this.

services/orchest-api/app/app/apis/namespace_pipelines.py
yannickperrenet: I will leave it up to you. I would be fine with having to put the tpe context manager (always) in a try-except block.

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -183,6 +175,16 @@ def collateral(self):
...
 
+    def revert(self):
+        # Error handling. If it does not succeed then the initial
+        # entry has to be removed from the database as otherwise no
+        # session can be started in the future due to the uniqueness
+        # constraint.
+        models.InteractiveSession.query.filter_by(
+            project_uuid=self.project_uuid, pipeline_uuid=self.pipeline_uuid
+        ).delete()
+        db.session.commit()
```
yannickperrenet, line 186: Could you add some information in the docstring of the `TwoPhaseFunction` stating whether or not one of its methods is allowed to commit? We want it to be consistent over the entire application I assume and otherwise have it clear that the commit can change on a case-by-case basis.

fruttasecca: Will do.

services/orchest-api/app/app/apis/namespace_pipelines.py
fruttasecca: Then I will leave it this way (except for the unnecessary logging, which has been removed).

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,138 @@
...
+rollback can be issued if any error arises during the transactional
+phase, leaving the state clean. Moreover, this pattern wants to
+encourage having collateral effecs into a second phase, the collateral
+one. The separation between DB changes and collateral effects should
+help in making the code clearer, easier to debug and to test, while
+improving the capacity of leaving a consistent state in case of error.
+The pattern works through the use of TwoPhaseFunction(s). Implementing
+a TwoPhaseFunction means implementing a transaction and a collateral
+function. The user of a TwoPhaseFunction is in charge of creating the
+context and calling 2phase_function.transaction(*args, **kwargs), while
```
ricklamers, line 15: Don't think `2phase_function` is a valid identifier. Maybe just `two_phase_function`

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,138 @@
...
+the TwoPhaseExecutor will take care of making it so that a commit will
+happen after all transactional code has execute, and will take core of
```
ricklamers, line 17: will take care of*

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,138 @@
...
+        could_abort = AbortEnvironmentBuild(tpe).transaction(
+            environment_build_uuid
+        )
+except Exception as e:
+    current_app.logger.error(e)
+    return {"message": str(e)}, 500
+
+
+Note that you should not call commit during any implementation of the
+transaction method, while you are free to do so if you need to apply db
```
ricklamers, line 33: Perhaps `Note that you should not call commit inside the transaction method,`

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: >Just as a thought. Could the use of attributes on self backfire? Should we instead go with a self.data = {} that is defined in the TwoPhaseFunction?

Not sure about this, would make accessing data more verbose.

>Now every subclass of TwoPhaseFunction defines a different set of attributes. Generally speaking it is good practice to define all your attributes in the __init__ function. Not sure whether that translates to this setting through.

This is something I would not mind, although as you say it might not translate to this setting. It's what felt more natural when we were prototyping it. @ricklamers ?

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -0,0 +1,138 @@
...
+    solved by issuing a rollback of the entire transaction. Similarly,
+    the transaction method does not have (nor need) to rollback, since
+    any uncaught exception during the transactional phase will bubble
+    up to the TwoPhaseExecutor, which will take care of that. The
+    collateral and revert functions are free to commit and rollback,
+    given that whatever happens in their body is considered a collateral
+    effect and not part of a transaction. This mean that they must take
+    care of their own commits. Taking care of rollbacks is optional
+    since the TwoPhaseExecutor will rollback if any exception raises
+    during a collateral() or revert call(). All in all, this means that:
```
ricklamers, line 108: the parents should be after revert instead: `revert()`

services/orchest-api/app/app/apis/namespace_environment_builds.py
ricklamers: Seems to me that self.data is more Pythonic (as all instance methods will be defined in the `__init__`). I prefer to have explicit data passing through `self.data` in case the collateral method depends on some of the data passed in the arguments of the `transaction(args, kwargs)`, instead of passing those arguments to the constructor of the class and having all of that available for both the `transaction()` and `collateral()`.

services/orchest-api/app/app/apis/namespace_environment_builds.py
```diff
@@ -296,54 +218,139 @@ def get(self, project_uuid, environment_uuid):
...
 
-    Args:
-        project_uuid:
-    """
-    builds = (
-        models.EnvironmentBuild.query.filter_by(project_uuid=project_uuid)
-        .with_entities(
-            models.EnvironmentBuild.project_uuid,
-            models.EnvironmentBuild.environment_uuid,
+class DeleteProjectBuilds(TwoPhaseFunction):
```
ricklamers, line 338: I'm a bit confused by `DeleteProjectBuilds`, wouldn't we not use a two phase executor if there is no collateral?

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -27,36 +28,43 @@ def delete(self, project_uuid, pipeline_uuid):
...
-        # interactive run pipeline step
-        # interactive run image mapping
-        db.session.delete(run)
-
-    # stop and delete any session if it exists
-    stop_interactive_session(project_uuid, pipeline_uuid)
-
-    db.session.commit()
+
+    def transaction(self, project_uuid: str, pipeline_uuid: str):
```
ricklamers, line 49: Love the types!

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -27,36 +28,43 @@ def delete(self, project_uuid, pipeline_uuid):
...
 
+        except Exception as e:
+            return {"message": str(e)}, 500
+
+        return {"message": "Pipeline deletion was successful."}, 200
 
-def delete_pipeline(project_uuid, pipeline_uuid):
-    """Delete a pipeline and all related entities.
 
+class DeletePipeline(TwoPhaseFunction):
```
ricklamers, line 41: Are you wrapping transaction only methods in a TwoPhaseFunction in order to cleanly compose them with other TwoPhaseFunctions that have them as a subroutine that _do_ have collateral effects?

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: This is because of nesting. `DeleteProjectBuilds` uses `DeleteProjectEnvironmentBuilds`, which is itself a `TwoPhaseFunction`. `DeleteProjectEnvironmentBuilds` makes use of `AbortEnvironmentBuild`, which is a `TwoPhaseFunction`, and needed to be because it has a collateral effect.  The alternative would be to make the first two into functions that receive a `tpe` as an argument, although I think this would be a change for the worse.

services/orchest-api/app/app/apis/namespace_pipelines.py
fruttasecca: Yes, I found it preferable and easier to expand later in case a collateral effect would need to take place.

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -137,38 +121,115 @@ def put(self, project_uuid, pipeline_uuid):
...
-    db.session.commit()
-    return True
+
+class StopInteractiveSession(TwoPhaseFunction):
+    def transaction(
+        self,
+        project_uuid: str,
+        pipeline_uuid: str,
+    ):
+        pass
```
ricklamers, line 193: I think this pass is a leftover?

fruttasecca: Indeed

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -137,38 +121,115 @@ def put(self, project_uuid, pipeline_uuid):
...
+            docker_client,
+            container_IDs=session.container_ids,
+            network="orchest",
+            notebook_server_info=session.notebook_server_info,
+        )
+
+        # TODO: error handling?
+        # TODO: If we can do this task in the background then the
+        # request can return. The session shutting down should not
+        # depend on the existence of the object in the DB. Need to make
```
ricklamers, line 227: I think session needs to be fully shutdown in order to start a new one because we name Docker containers by the project/pipeline uuid. We might be able to _not_ do this which would make stopping sessions instantaneous to the user. Probably a TODO if we notice that refactor is worthwhile.

fruttasecca: It should be possible to rename a container, although it will for sure lead to more changes.

services/orchest-webserver/app/app/core/pipelines.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # Will delete cascade experiment -> pipeline run.
+        Pipeline.query.filter_by(project_uuid=project_uuid, uuid=pipeline_uuid).delete()
+
+    def collateral(self):
+        """Remove a pipeline from the FS and the orchest-api"""
+
+        # CleanupPipelineFromOrchest can be used when deleting a
+        # pipeline through Orchest or when cleaning up a pipeline that
+        # was deleted through the filesystrem by the user, so the file
```
ricklamers, line 91: Small typo: filesystem

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: I'll go with using `self.data` and leave the argument passing in `transaction`.

services/orchest-webserver/app/app/core/pipelines.py
```diff
@@ -0,0 +1,162 @@
...
+            file_pipeline_uuid = pipeline_json.get("uuid")
+            # See if pipeline_uuid is taken or if the pipeline has no
+            # uuid, in both cases it's going to need a new uuid.
+            if (not file_pipeline_uuid) or Pipeline.query.filter(
+                Pipeline.uuid == file_pipeline_uuid
+            ).filter(Pipeline.project_uuid == project_uuid).count() > 0:
+                file_pipeline_uuid = str(uuid.uuid4())
+                # Only write to the file to update the uuid if it's
+                # necessary.
+                self.do_collateral = True
```
ricklamers, line 131: I'm seeing this `self.do_collateral` pattern emerge. In the sense that during the transaction one could find out the collateral should not happen. If we were to raise an error the collateral would also not happen (along with bubbling up such that other collaterals part of the TwoPhaseExecutor will not run).

What do you think is preferred? Just not executing this particular collateral? In a way _this_ TwoPhaseFunction does not know in what TwoPhaseExecutor it will be used (by encapsulation). But you could argue that the transaction should raise an exception in the case this particular cannot proceed as intended (giving the callee the responsibility to appropriately handle that).

services/orchest-webserver/app/app/core/pipelines.py
```diff
@@ -0,0 +1,162 @@
...
+            new_pipeline = Pipeline(
+                uuid=file_pipeline_uuid,
+                path=pipeline_path,
+                project_uuid=project_uuid,
+            )
+            db.session.add(new_pipeline)
+
+            # To be used by collateral and revert.
+            self.project_uuid = project_uuid
+            self.pipeline_uuidd = file_pipeline_uuid
```
ricklamers, line 143: Typoe: `pipeline_uuidd` should probably be `pipeline_uuid`.

services/orchest-webserver/app/app/core/pipelines.py
fruttasecca: I think we should leave the semantics of what happens in the `transaction` and `collateral` functions to the implementor (excluding commits, ofc). It might be the case that the `transaction` needs to take place while the `collateral` does not need to, I would not consider that an error.

services/orchest-webserver/app/app/core/pipelines.py
ricklamers: Oh the irony: a typo in my comment about a typoe!

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
ricklamers: I don't prefer this because we'd require the class implementer to decorate a function to add to the collateral queue. Which I don't want to implementer to be concerned with (properly decorating a specific function).

ricklamers: To make this less confusing we could add to the collateral queue in the class constructor. @fruttasecca would this work?

fruttasecca: Then you might have cases where the order in which the `transaction` functions are called does not dictate the order in which the `collateral` ones are since it would depend on the constructor.

ricklamers: I guess it's less natural to have the collateral ordering be determined by the class initialization order. It could work, but now that you phrase it like that I'm not sure that my proposal is an improvement. Transaction invocation = collateral order seems most natural to me.

fruttasecca: Indeed. From the user POV it would make things more complex, imo.

yannickperrenet: Yea indeed from the POV of the user it actually becomes more complex. Lets leave it the way it is now ;)

yannickperrenet (approved):

yannickperrenet (approved):

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -191,25 +200,25 @@ def transaction(
...
+        else:
+            session.status = "STOPPING"
+            self.collateral_kwargs["project_uuid"] = project_uuid
+            self.collateral_kwargs["pipeline_uuid"] = pipeline_uuid
         return True
 
-    def collateral(self):
-        if not self._do_shutdown:
+    def collateral(self, project_uuid: str, pipeline_uuid: str):
+        if not project_uuid or not pipeline_uuid:
```
ricklamers, line 217: Maybe a comment about why they could be `None`?

ricklamers: Other than that, latest changes LGTM!

services/orchest-webserver/app/app/core/pipelines.py
ricklamers: As discussed, we'll swap `do_collateral` for passing arguments that can dictate which parts of the collateral function (if any) execute.

services/orchest-api/app/app/apis/namespace_environment_builds.py
ricklamers: We decided to pass the data using `self.collateral_kwargs` to make it more explicit what data is being used by the `collateral` function and to avoid introducing new class attributes outside of the `__init__`.

lib/python/orchest-internals/_orchest/internals/two_phase_executor.py
```diff
@@ -129,21 +147,41 @@ class TwoPhaseFunction(ABC):
...
+    def revert(self):
+        if self._has_run_revert:
+            raise RevertHasBeenRunTwice()
+
+        self._has_run_revert = True
+        self._revert()
 
     @abstractmethod
-    def transaction(self, *args, **kwargs):  # pylint: disable=E0202
+    def _transaction(self, *args, **kwargs):  # pylint: disable=E0202
```
ricklamers, line 178: We should be able to get rid of the `# pylint: disable=E0202`

fruttasecca: Indeed, will do

PR #112, opened 2021-01-18
Refactor of experiments to jobs

ricklamers:
This PR refactors what was previously known as 'Experiments' to 'Jobs'.

Given the upcoming feature of being able to schedule pipeline runs using cron-like recurring schedules we think the new nomenclature of 'Jobs' does more justice to the scope of the features that are currently part of 'Experiments'.


ricklamers:
@fruttasecca @yannickperrenet would be great if you could check this PR for comprehensiveness and test locally to make sure we don't introduce errors.

Known TODO: update doc images that mention Experiments.

yannickperrenet:
What a rename! Will do some testing tomorrow.

yannickperrenet:
When searching the repository I found a few other occurrences of "experiment" and in the process of renaming I came across "job_json". What do we use this for? Seems to be partially legacy code.

yannickperrenet (approved):

fruttasecca (approved):
lgtm

ricklamers:
> EDIT: You can ignore this comment or see the commit that comes after.
> 
> When searching the repository I found a few other occurrences of "experiment" and in the process of renaming I came across "job_json". What do we use this for? Seems to be partially legacy code.
> 
> Apart from the orchest-webserver it should probably be renamed to "parameters" or be removed. What are your thoughts?

I agree, `job_json` should only occur once in the webserver when marshalling the return JSON for the next request.

ricklamers:
> which, in my opinion, don't need updating similar to the quickstart YT video which we won't update over such a minor change.

Ended up replacing the images for completeness sake. I agree the YT video goes to far.

PR #114, opened 2021-01-19
Minimally change pipeline definition for diffs

yannickperrenet:
Since the pipeline definition is stored in JSON and JSON is unordered, every change to the pipeline definition could end up in large diffs. Instead of relying on the implicit order of keys in dictionaries, especially since the pipeline definition is also interacted with in Javascript, explicitly sort the keys.

Caveat: The order given by the user of the defined parameters is now also automatically sorted.


yannickperrenet:
@ricklamers I wanted to get your opinion on automatically sorting the order of the parameters defined by the user. What do you think?

ricklamers:
I like this!

PR #115, opened 2021-01-21
Feature job scheduler

fruttasecca:
This PR brings a job scheduler into Orchest. When creating a job, the job can be scheduled to be run right now, at a given time, or recurringly through a cron string.

The scheduler will take care of running any job run that was missed while Orchest was not running. Meaning that if 10 runs of a recurring job were missed, 10 runs will be scheduled once Orchest is back running.


services/orchest-api/app/app/core/scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+was not running at that time, the scheduler will run the job and set the
+next_scheduled_time accordingly, in a way that all job runs that have
+been missed will be scheduled. For example, given a recurring job with
+cron string "* * * * *" (runs every minute), next_scheduled_time set to
+12:00, if Orchest was offline and it's started at 12:10 the scheduler
+will schedule the job run of 12:00, and will set the next_scheduled_time
+to 12:01, which will in turn trigger another scheduling, which will
+set the next_scheduled_time to 12:02, etc., all the way up to being on
+par with all the runs that should have been scheduled. This way all
+job runs that were missed will be scheduled and run.
```
yannickperrenet, line 27: If I am not mistaken the celery beat scheduler does not have this behavior, instead if a task was scheduled to run every minute and Orchest is down for 10 minutes then it would only schedule it once.

We should decide what exactly the behavior is we want. For example: a job that does some ETL task, then it might not make sense scheduling the task in quick succession because that would end up doing the exact same work.

services/orchest-api/app/app/core/scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+                )
+                .filter(
+                    # Filter out jobs that do not have to run anymore.
+                    Job.next_scheduled_time.isnot(None)
+                )
+                .filter(
+                    # Jobs which have next_scheduled_time before now
+                    # need to be scheduled.
+                    now
+                    > Job.next_scheduled_time
```
yannickperrenet, line 65: Seems like a strange auto-formatting to me, not sure why black would do this...

services/orchest-api/app/app/core/scheduler.py
ricklamers: I had a quick discussion about this with Jacopo earlier.

My preference is that jobs do not get ‘lost’ if Orchest is not running. So if my cron jon runs every hour and Orchest is down for 5, I don’t think it should run once and be done with it. But perhaps my preferred default is not the commonly accepted one. I’m open to changing opinion on this.

services/orchest-api/app/app/models.py
```diff
@@ -97,20 +97,56 @@ class Job(BaseModel):
...
+    # containing the parameters of that step for that particular run.
+    # [{ <step_uuid>: {"a": 1}, ...}, ...GG]
+    job_parameters = db.Column(
+        JSONB,
         nullable=False,
+        # This way migrated entries that did not have this column will
+        # still be valid. Note that the entries will be stored as a list
+        # of dicts.
+        server_default="[]",
     )
```
yannickperrenet, line 117: Does this mean that the parameters of a job should not be updated, making the job immutable? Or do we want to have mutable jobs?

The front-end reads the pipeline definitions (from the filesystem) to display the parameters of the job so this would still work in both cases. An argument against immutable jobs would be that you cannot create a job from another job (and thus changing the parameters on the same version of the pipeline is harder, although not impossible). An argument against mutable jobs is untraceability/inconsistency.

services/orchest-api/app/config.py
```diff
@@ -12,6 +12,10 @@ class Config:
     ORCHEST_API_ADDRESS = "http://orchest-api:80/api"
     ORCHEST_WEBSERVER_ADDRESS = "http://orchest-webserver:5000"
 
+    # How often to run the scheduling logic when the process is running
+    # as scheduler.
+    SCHEDULER_INTERVAL = 10
```
yannickperrenet, line 17: Could you add that the unit is seconds?

services/orchest-api/app/app/models.py
ricklamers: The front-end reads the pipeline definition from the `snapshot` that is copied at job creation time. So mutability doesn't fit naturally with how we've structured and IMO is not preferred. The more you can manually change the less reliable the system becomes. The overhead of creating a new job if you want to use a different state of the project seems perfectly fine with me.

services/orchest-api/app/app/core/scheduler.py
fruttasecca: I don't have a strong opinion on this either, and it could be a setting of Orchest.

services/orchest-api/app/app/core/scheduler.py
fruttasecca: No idea, wil fix.

services/orchest-api/app/app/models.py
fruttasecca: @yannickperrenet not sure I understand the question, but the highlighted code does not imply that the job is immutable, it's actually the opposite since you could update the parameters with new ones and the next time the job is scheduled the parameters will affect how many runs are launched and their parameters.

services/orchest-api/app/config.py
fruttasecca: Will do

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -311,33 +314,31 @@ def catch_api_proxy_jobs_get(job_uuid):
...
+                    pipeline_def = copy.deepcopy(json_return["pipeline_definition"])
+                    pipeline_def["parameters"] = run["pipeline_parameters"]
+                    run["parameters"] = run["pipeline_parameters"]
+
+                    # Set the steps parameters in the pipeline
+                    # definition.
+                    for step_uuid, step_parameters in pipeline_def[
+                        "parameters"
+                    ].items():
+                        pipeline_def["steps"][step_uuid]["parameters"] = step_parameters
```
yannickperrenet: The nested for-loop does not actually change the `json_return`, does it? Anyways I am a bit confused what exactly is being modified here with the deepcopy and pointing to the `json_return`

fruttasecca: I believe you were looking at an outdated version of the code, the for loop has been fixed.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -81,11 +129,63 @@ class Job(Resource):
...
+        job_update = request.get_json()
+
+        cron_schedule = job_update.get("cron_schedule", None)
+        parameters = job_update.get("parameters", None)
+
+        try:
+            job = models.Job.query.filter_by(job_uuid=job_uuid).one()
+
+            if job.schedule is None or job.status != "STARTED":
+                return {"message": "Failed update operation."}, 500
```
yannickperrenet, line 164: What exactly are the possibilities for a job status? If a job is a cron-like scheduled job, then its status will always be `STARTED` unless aborted?

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -81,11 +129,63 @@ class Job(Resource):
...
+        are launched. Only recurring ongoing jobs can be updated.
+
+        """
+        job_update = request.get_json()
+
+        cron_schedule = job_update.get("cron_schedule", None)
+        parameters = job_update.get("parameters", None)
+
+        try:
+            job = models.Job.query.filter_by(job_uuid=job_uuid).one()
```
yannickperrenet, line 161: Why is this block not contained in a tpe?

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: Job status can be any of `ABORTED, PENDING, STARTED, SUCCESS`. As you say, a cron-like job is either `STARTED` or `ABORTED`.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: Because the PUT request does not have any collateral effects, unless I am missing it.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -137,28 +237,63 @@ def put(self, job_uuid, run_uuid):
         """Set the status of a pipeline run."""
         status_update = request.get_json()
```
yannickperrenet, line 238: I think I am missing the general logic of the scheduling implementation, because this endpoint no longer seems to be updating the status of the pipeline run and instead changes the status of the job.

yannickperrenet (changes requested):
I think we need some discussions around the general implementation before merging it.

services/orchest-api/app/app/core/scheduler.py
yannickperrenet: > My preference is that jobs do not get ‘lost’ if Orchest is not running.

I don't think this makes sense if all the jobs are running with the exact same set of parameters. They would essentially all return the same result and just use compute (and therefore possibly lagging other tasks down the chain).

services/orchest-api/app/app/models.py
yannickperrenet: > but the highlighted code does not imply that the job is immutable

I fully agree, but when reviewing the code I wanted to put down my ideas on the subject before forgetting it whilst searching for the corresponding code that implements it.

> So mutability doesn't fit naturally with how we've structured and IMO is not preferred.

I agree on this point and would rather go with immutable jobs. This then also nicely introduces a concept around versioning and forking (which we can add later).

> The functionality to update a job parameters or its cron string is already there as a PUT endpoint, we can expose such functionality through the GUI or not.

Not sure whether I want to leave the endpoint there if we choose not to go with mutable jobs, since then we have dangling code. And since the `strategy_json` would have to be updated (like you said) the endpoint in and of itself would not work.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: The logic of this endpoint is unchanged, imo. On [master](https://github.com/orchest/orchest/blob/master/services/orchest-api/app/app/apis/namespace_jobs.py#L136), a run PUTs its state through this endpoint, and the endpoint also updates the job accordingly [by changing  one of its attributes](https://github.com/orchest/orchest/blob/master/services/orchest-api/app/app/apis/namespace_jobs.py#L147). 

The difference is that instead of using `completed_pipeline_runs` we are querying the db and based on that we decide if the job is completed or not, instead of updating `completed_pipeline_runs`, the end point is updating `status`.

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: Should we maintain a document or whatever to write this down? Possibly putting it in the development section somewhere in the [implementation details docs](https://orchest.readthedocs.io/en/latest/developer_guide/implementation_details.html)?

services/orchest-api/app/app/models.py
fruttasecca: The endpoint would actually still work for updating the cron string of a recurrent job without additional webserver work involved I believe. We can simply delete the end point if we do not make use of it, of course.

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: I recall a similar scenario (but I can definitely be mistaken) where there was also no collateral effect, yet it was still wrapped in a tpe for consistency and future extensibility. If not, then I am fine leaving it this way.

fruttasecca: Could it be that the reason was that the request was actualling making use of a `TwoPhaseFunction`, although it did not have any collateral effect itself? (but the 2pf did/could).

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: I think I'll go ahead and write a comment on top of the column declaration in `models`, we can then further extend the docs if we decide to do so.

services/orchest-api/app/app/core/scheduler.py
fruttasecca: I agree with @yannickperrenet. We have also to keep in mind that we are not providing a workflow management tool, so jobs getting lost while Orchest is offline could be part of the contract.

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: I am unsure and since you are more familiar then I am, its probably best to go with your judgement here :)

services/orchest-api/app/app/models.py
yannickperrenet: Since the option of having mutable jobs is more or less implemented already and offers the users more flexibility, we decided to go with mutable jobs instead of immutable jobs. This might change in the future if the flexibility is restored in other way (possibly integrated versioning of projects) and going with immutable jobs for reliability.

services/orchest-webserver/app/app/views/orchest_api.py
yannickperrenet: > I believe you were looking at an outdated version of the code

Indeed I was.

ricklamers:
I'll do some testing over the weekend, but I think most of it should be ready now.

Specifically I want to review whether we want to move the orchest-webserver `Job` state fully to the `orchest-api`. If the required refactoring is minimal I think I want to do it in this PR. We'll review that bit on Monday.

Great work everyone!

yannickperrenet:
After playing around a bit with the application I think this is a job well done! I do have some pointers though:

* After creating and starting a job, you can open that job to see its pipeline runs. However, the _ID_ column is no longer populated.
* I find _Creation status_ in the job list (so after clicking _Jobs_ in the left pane menu) somewhat confusing/useless as it only shows _Submitted_ or _Draft_, whilst I would argue that the _Job status_ you see when opening the job is what you actually want to see in the job list. Otherwise a user will have to click on all the jobs in order to find one that is still running. When adding the status in the job list, the status should also become searchable (isn't that right @ricklamers?) which I consider a value add.
* The job list is now ordered based on _Date created_ with the most recent date at the bottom, I would like to see that being reversed.

yannickperrenet:
I think we should add the the _refresh_ functionality from the pipeline runs list (after clicking on a job) to the job list as well. Otherwise you need to navigate away and back to see the job itself changing.

yannickperrenet:
> I think we should add the the refresh functionality from the pipeline runs list (after clicking on a job) to the job list as well. 

This is actually more in-line with how the projects and pipelines view work. And so we decided not to add this refresh button.

> Otherwise you need to navigate away and back to see the job itself changing.

You can just click on the "Jobs" in the left-pane menu, which is also just one click like the refresh button would have been.

yannickperrenet (approved):

Issue #116, opened 2021-01-23
Project detection failed


howie6879:
Hi there:

Sometimes I delete a project by using `filemanager` and then import it again.

This causes the orchest to fail to load the project, The operation process is as follows:

```shell
cd orchest/userdir
git clone https://github.com/orchest/quickstart
```

 orchest will load this project automatically, The next web operation is as follows:

- Click FileManage->projects
- Delete quickstart dir.

Finally  import `quickstart` project by using Git again. 

At this point, orchest can't load the  `quickstart` project.

You can check the screenshot that I provided for details:

![bd](https://gitee.com/howie6879/oss/raw/master/uPic/bd.gif)

howie6879:
I've found the cause, I'll fix this later

ricklamers:
Thank you @howie6879 for reporting this. Will gladly accept a PR. Either way we’ll make sure this gets fixed asap.

Great find!

howie6879:
> Thank you @howie6879 for reporting this. Will gladly accept a PR. Either way we’ll make sure this gets fixed asap.
> 
> Great find!

Done!

ricklamers:
This is fixed as per #117.

PR #117, opened 2021-01-23
Fix project detection failed: #116

howie6879:
This PR made the following changes:
  - About #116 , This bug is caused by not catching the FileNotFoundError error, see line 147 of the script services/orchest-webserver/app/app/core/projects.py. `  shutil.rmtree(full_project_path)`
  - Set `projects_dir` as a configuration property, such as: `app.config["USER_DIR"]`, because this variable is frequently combined in the code


ricklamers:
Great fix, LGTM!

PR #133, opened 2021-01-28
Memory size changes to take effect on memory-server restart

### LINKED ISSUE #127: Memory size increase of memory-server without requiring session reboot
When changing the size of the `memory-server` through the settings of a pipeline, the user is required to reboot the entire session and thus losing the state of all kernels.

It would be better if the user could only restart the memory-server itself for the changes to take effect. Or possibly dynamically, but don't think this is supported without losing the current objects in the plasma store.

yannickperrenet:
## Description

Due to Docker not supporting passing environment variables when starting a container (https://github.com/moby/moby/issues/7561), the `memory-server` now reads the `pipeline-definition` to determine the memory size for the internal Plasma store. As a side-effect only the `memory-server` has to be restarted for changes in the _Data passing memory size_ pipeline setting to take effect.

As an additional note, when starting the `memory-server` the `shm-size` is set to a large value to solely let the `memory-server` determine how much memory to consume (based on the setting). Mounting `/dev/shm` directly to the container's `/dev/shm` is not out of the box supported on Mac (and neither are [tmpfs mounts](https://docs.docker.com/storage/tmpfs/)).

Resolves #127 

### Checklist

### DIFF
--- a/services/memory-server/app/config.py
+++ b/services/memory-server/app/config.py
@@ -1,7 +1,5 @@
 import os
 
-import utils
-
 from _orchest.internals import config as _config
 
 # Default location where the socket is created.
@@ -11,5 +9,3 @@ STORE_SOCKET_NAME = os.path.join(_config.MEMORY_SERVER_SOCK_PATH, "plasma.sock")
 PIPELINE_FNAME = os.path.join(
     _config.PROJECT_DIR, os.environ.get("ORCHEST_PIPELINE_PATH", "")
 )
-
-STORE_MEMORY = utils.get_store_memory_size(PIPELINE_FNAME)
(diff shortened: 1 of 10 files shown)


fruttasecca (approved):
lgtm. This is not stricly related to the changes made by the PR, is there any way that we can improve the section between "Data passing"  and  "SAVE" ?  The sentence, followed by a checkbox, followed by another sentence doesn't feel nice, imo.

image

yannickperrenet:
> The sentence, followed by a checkbox, followed by another sentence doesn't feel nice, imo.

@ricklamers What do you think?

I think it makes sense to do these particular changes once we are doing a UI/UX rework.

ricklamers:
@fruttasecca I agree this should be cleaned up. But it can happen outside of this PR.

Issue #125, opened 2021-01-26
Pipeline level parameters


yannickperrenet:
Ability to define pipeline level parameters, besides the already implemented step parameters.

* Pipeline level parameters to be accessible in all steps. Both the pipeline level and step level parameters will be returned to the user (and so the user has to decide what to do on name collisions).

Implementation details:

* The parameters will be added as `"parameters"` to the pipeline definition.
* When starting a job the pipeline level parameters have to be passed together with the step parameters, but containing a special prefix. Otherwise the selection of pipeline runs in jobs will break.

fruttasecca:
Currently, retrieving (step) parameters is done through `orchest.get_params`, while updating is done through `orchest.update_params`.  Just scanning some alternatives we have:
- have `get_params` return a tuple `(step params, pipeline params)`
- split `get_params` in `get_step_params` and `get_pipeline_params`
- `get_params(pipeline_params=False)`, returns either step or pipeline params based on input

I am leaning towards the first since it's 1 less line of code for the user when retrieving parameters.

About `update_params`: I don't believe we want to allow updating pipeline parameters at run time, since it would be prone to race conditions and encourage bad practices. If we agree on that, `update_params` should be renamed to `update_step_params` (open to suggestions).

yannickperrenet:
> I am leaning towards the first

Agreed. The user can then always choose whether to use only one of them or both.

> If we agree on that, `update_params` should be renamed to `update_step_params`

I agree on the reasoning behind the change as well as the proposed rename.

ricklamers:
I think I might prefer update_step_params and update_pipeline_params.

For consistency I’d prefer to then also have get_step_params() and get_pipeline_params(). With the tuple approach it’s not clear from the method alone which tuple entry is which.

With regard to the updating, I feel like we’re arbitrarily handicapping by now allowing the user to update pipeline params. I can see legit use cases like in the first step determining some global param to be set for all remaining steps in the pipeline. It’s just writing to the JSON file, so it’s something they could do anyway.

Perhaps a warning about race conditions, but that’s possibly implied if the user sees steps running in parallel.

yannickperrenet:
> For consistency I’d prefer to then also have get_step_params() and get_pipeline_params(). With the tuple approach it’s not clear from the method alone which tuple entry is which.

[...]

PR #134, opened 2021-01-29
Feature pipeline parameters

fruttasecca:
This PR adds the concept of pipeline parameters. Pipeline parameters can be set up through the pipeline settings view when in the pipeline editor; moreover, pipeline parameters can be overridden for jobs.


yannickperrenet:
Resolves #125

orchest-sdk/python/orchest/parameters.py
```diff
@@ -31,34 +32,44 @@ def get_params() -> Dict[str, Any]:
...
 
-    return params
+    return params, pipeline.get_params()
 
 
-def update_params(params: Dict[str, Any]) -> None:
-    """Updates the parameters of the current step.
+def update_params(
+    step_params: Optional[dict] = None, pipeline_params: Optional[dict] = None
+) -> None:
```
yannickperrenet, line 40: Didn't we want to add a warning as updating `pipeline_params` could introduce a race condition unaware to the user?

yannickperrenet (changes requested):
I think we should add a warning somehow. Probably using the docstring:
```python
"""...
Warning:
    The warning message
"""
```
This will then also get rendered in the documentation as a warning using sphinx.

orchest-sdk/python/orchest/parameters.py
fruttasecca: Will  do

yannickperrenet:
When configuring the parameters of the job, you see something like:
```
Pipeline: pipe
    a:[5]
step-name
    b:[6]
```
but I am not sure whether this is clear enough to the user. The confusion to me is mostly introduced by the fact that in the top it already states:
```
Pipeline
pipe
```

Happy to hear your thoughts as maybe I am just imaging the confusion :)

yannickperrenet (approved):

Issue #131, opened 2021-01-26
Persisting user configurations of integrated IDEs


yannickperrenet:
The configurations of IDEs we integrate in Orchest:

* JupyterLab
* VS Code (through code-server #113)

should be persisted so that a user does not have to configure their IDEs every time. Otherwise (as it is currently), for example added extensions to JupyterLab would have to be reinstalled every time JupyterLab is started.

This feature should be easy to use and so purely defining the configuration programmatically (think dotfiles) is probably not the way to go. Additionally, the configurations should be portable between upgrades of Orchest (and possibly upgrades of the specific IDE service containers).

The current idea is to mount a new directory (from `userdir/.orchest/...`) to the appropriate location in the IDE service container to persist the configurations. For JupyterLab we have to make sure this does not require a rebuild and includes extensions (the `jupyter lab clean` command suggests this approach is possible since it allows for `--extensions`, `--settings` and `--static` flags).

yannickperrenet:
`jupyter lab path` could also be helpful to find the correct target path to mount.

ricklamers:
This as been added 🔥

PR #136, opened 2021-02-01
Feature jupyterlab config persistence

yannickperrenet:
## Description
This PR persists JupyterLab extensions and user configurations. 

See: #131 

### Implementation details
JupyterLab is build in a non-default directory (inside the `Dockerfile`), so that we can mount a directory containing the persisted user configurations into the default location. The `start.sh` script takes care of the logic to start JupyterLab and make sure it is started with the right application directory (containing the extensions and build files).

Whenever JupyterLab or the extensions Orchest uses to integrate JupyterLab update, then the `static/` files containing the JupyterLab build together with the extension `tar` files are copied from the non-default location. Because without those files JupyterLab does not start correctly as it has to be integrated within Orchest.

### Checklist


yannickperrenet:
More information about the application directory and other paths can be found in the [Jupyter docs](https://jupyterlab.readthedocs.io/en/latest/user/directories.html).

yannickperrenet:
As far as documentation goes, I don't think we need to add anything. Even though persistence of the configuration can be considered a feature w.r.t. Orchest, it is probably not seen as a feature in general (since configuration is usually always persisted). Secondly, it works out of the box and users don't have to set anything up to get started with persisting their JupyterLab configurations.

services/jupyter-server/start.sh
```diff
@@ -2,4 +2,79 @@
...
+# Add orchest extension tarballs to `extensions/`. This way the
+# extensions get automatically included in the build.
+cp -rfT "$build_path/extensions" "$userdir_path/extensions"
+
+# Overwrite the static files from the userdir with the static files
+# from the build. Otherwise JupyterLab cannot start as part of Orchest.
+rm -rf "$userdir_path/static" && cp -r "$build_path/static" "$userdir_path/static"
+
+# Trigger rebuild in background. When the user refreshes JupyterLab
+# once the build it finished, the extensions are installed.
```
fruttasecca, line 76:
```suggestion
# Trigger a rebuild in the background. When the user refreshes JupyterLab
# once the build is finished, the extensions are installed.
```

services/jupyter-server/start.sh
```diff
@@ -2,4 +2,79 @@
...
+# (jupyter lab build --dev-build=False --LabBuildApp.app_dir="$userdir_path" \
+#     && jupyter lab clean --LabCleanApp.app_dir="$userdir_path") &
```
fruttasecca, line 78: Is this to be kept?

yannickperrenet: I will remove it, since we will not be triggering the rebuild in the background and instead let the user trigger it. 

JupyterLab actually handles it pretty nicely with pop-ups and reloading, whereas a rebuild in the background is not handled at all.

fruttasecca (approved):
lgtm, love the config persistence :]

ricklamers:
PR looks great! And super happy we now have JupyterLab config/extension persistence in Orchest :)!

Issue #124, opened 2021-01-26
Environment variables 


yannickperrenet:
Environment variables will replace the current _Data sources_:

* _Data sources_ is currently nothing more than management of secrets, which can be better done using environment variables. Additionally, the concept of environment variables is understood by many.
* Arguably, a user wants full authority over their data source connectors and thus we should consider it as part of the code they want to (or have to) write themselves.

The current idea around adding environment variables to Orchest is as follows:

* ENVs should be treated as secrets and should thus be excluded from versioning.
* Defined at the project, put with the possibility of specifying pipeline level overwrites.
* When removing _Data sources_ the `/data` concept is kept. And so host system file mounting is no longer supported (symlinks from the `/data` directory won't work due to Docker), instead you need to directly put the data you want to use in the `/data` directory.

Implementation details:

* The values of the ENVs are stored in the `orchest-webserver` and persisted in the `orchest-api` whenever a job is started. This is similar to how we treat parameters.

PR #140, opened 2021-02-05
Feature env variables

fruttasecca:
## Description
This PR introduces the concept of environment variables. Defining environment variables can be useful when you want to avoid persisting sensible data in your versioning system. These environment variables will be injected in your pipeline steps, and can be retrieved using the native way of your language of choice, e.g. `os.environ["MY_VAR"]` in python. It is possible to define environment variables at the project, pipeline or job level.

Closes: #124 

### Checklist


docs/source/getting_started/overview.rst
```diff
@@ -6,16 +6,25 @@ Overview
...
-* :ref:`Integrate commonly used data sources <integrate data sources>`.
 * :ref:`Visually construct pipelines <visually construct>`.
 * :ref:`Write code using JupyterLab <edit with jupyterlab>`.
 * :ref:`Write code using any other editor of choice <edit with vim>`.
 * :ref:`Run any subset of a pipeline <run any subset>`.
 * :ref:`Skip certain cells when executing a notebook top-to-bottom <skip cells>`.
 * :ref:`Parametrize your data science pipelines to try out different modeling ideas
   <parametrize pipelines>`.
+* :ref:`Define environment variables at the project or pipeline level
+  <env variables>`.
```
yannickperrenet, line 17: I think it makes more sense to do
```suggestion
* :ref:`Define environment variables <env variables>`.
```
because the notion of a project/pipeline is not familiar to the user, since this is the overview.

Possibly we could add the word "granular" in there to indicate there is a certain granularity involved when defining environment variables (which I consider a feature).

docs/source/user_guide/environment_variables.rst
```diff
@@ -0,0 +1,57 @@
+.. _environment_variables:
+
+Environment variables
+=====================
```
yannickperrenet, line 4: We should probably state it somewhere or put it in the prettier configuration (if possible), but we wrap all lines in the `.rst` (documentation) files at 100.

docs/source/user_guide/environment_variables.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Defining environment variables can be useful when you want to avoid
+persiting sensible data in your versioning system. These environment
```
yannickperrenet, line 7:
```suggestion
persisting sensible data in your versioning system. These environment
```

docs/source/user_guide/environment_variables.rst
```diff
@@ -0,0 +1,57 @@
...
+variables will be injected in your pipeline steps, and can be retrieved
+using the native way of your language of choice, e.g. ``os.environ["MY_VAR"]``
+in python. It is possible to define environment variables at the project, pipeline or job level.
+
+.. warning::
+   Environment variables are persisted within Orchest. Make sure only
+   authorized people have access to your instance and sensible data.
```
yannickperrenet, line 14: Could link them to the sections:

* https://orchest.readthedocs.io/en/latest/user_guide/other.html#authentication
* https://orchest.readthedocs.io/en/latest/getting_started/installation.html?highlight=ssl#run-orchest-on-the-cloud

but of course using RTD style linking.

docs/source/user_guide/environment_variables.rst
yannickperrenet, line 4: In this section we should probably make clear to the user how the values and names should be input, i.e. are they converted to strings? Do they need to put quotes? Are they just set verbatim?

orchest-sdk/README.md
```diff
@@ -8,5 +8,4 @@ using the SDK:
 - Data passing between pipeline steps. It manages the target and source of the data, leaving the
   user only with the decision what data to pass. The target and source of the data are inferred
   through the defined pipeline definition in Orchest (the `pipeline.json` file).
```
yannickperrenet, line 10: Should probably also update the `pipeline.json` sentence since it is now a pipeline definition with a `.orchest` extension.

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -3,10 +3,14 @@
 Despite the fact that the orchest api has no model related to a
 pipeline, a good amount of other models depend on such a concept.
 """
+from flask import abort, request
+from flask.globals import current_app
```
yannickperrenet, line 7: You can actually import `current_app` from `flask` directly also.

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -248,11 +249,19 @@ async def run_on_docker(
         # directory will be my/project/path/.
         working_dir = os.path.split(run_config["pipeline_path"])[0]
 
+        user_env_variables = [
+            f"{key}={value}" for key, value in run_config["user_env_variables"].items()
+        ]
```
yannickperrenet, line 254: We could point the users (from the docs) to here or at least make clear how we do string interpolation.

On that note, we should probably make it:
```python
f"{key!r}={value!r}"

# Or following PEP 498: https://www.python.org/dev/peps/pep-0498/#s-r-and-a-are-redundant
f"{repr(key)}={repr(value)}"
```
This would mean we need to store it differently in the database in the first place (and make sure the front-end accounts for quotes as well).

services/orchest-api/app/app/core/tasks.py
```diff
@@ -234,10 +234,6 @@ def start_non_interactive_pipeline_run(
     # of a step.
     host_base_user_dir = os.path.split(run_config["host_user_dir"])[0]
 
-    # It gets passed to the run of the pipeline, however, it is never
-    # used. Thus we delete it to not cause confusion.
-    del run_config["host_user_dir"]
```
yannickperrenet, line 239: Why is this no longer needed? The `data/` mount?

fruttasecca: yep

yannickperrenet (changes requested):
I think we need to get the interpolation of environment variables exactly right and make it clear to the user what is expected of them. Otherwise environment variables will work more confusing then adding benefit (to a certain degree of course).

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: It seems to work fine already :)

yannickperrenet:
I do want to discuss a certain case we (at least I) have overlooked. 

The environment variables are injected for pipeline runs only. This means that the environment variables do not exist in your Jupyter kernels. I consider this a problem since it means that you cannot probably develop your pipeline, as running it is different from developing (in the kernels).

Then there is another point. Environment variables (as done by the enterprise gateway) are currently injected the moment kernels are launched (see `launch_docker.py`). This then has the effect that there can exist a discrepancy between defined environment variables and environment variables that are present in the kernel. Back when we had data sources support, we tackled this problem (although arguably suitable for data source and not for environment variables) by using the SDK to query an endpoint. For example, the SDK could query an internal endpoint to get the environment variables (which I don't think we want since environment variables are then no longer "native" environment variables).

ricklamers:
To reply to Yannick.

1. Project + pipeline environment variables should be passed when a session starts. Such that kernels receive them too as native environment variables.

2. Changes in the env variables then require a session restart. Which seems acceptable to me as the nature of env variables indicates some level of ‘staticness’ anyway.

In short, to me this was expected behavior. And we need to make sure the PR reflects the above.

yannickperrenet:
> as the nature of env variables indicates some level of ‘staticness’

I agree, the PR does not reflect this yet though. 

It should be clear from the docs and sections where you can specify environment variables (e.g. pipeline settings same as we indicate a restart of the memory-server) that a session reboot is required to get up to date environment variables.

ricklamers:
> I agree, the PR does not reflect this yet though.

I've implemented the kernel environment variables and added the description of this behavior to the docs.

docs/source/user_guide/environment_variables.rst
```diff
@@ -48,13 +48,21 @@ You can access your pipeline environment variables through the pipeline settings
...
+the job, moreover, cron jobs can have their environment variables edited after they have started.
 Every pipeline run belonging to the job will have these environment variables set.
+
+Environment variables for interactive runs
+------------------------------------------
+
+Environment variables will also be available in kernels that are launched in JupyterLab. To 
+refresh the environment variables for kernels a restart of the session is required. As per usual 
+the environment variables defined at the pipeline level override the project defined environment 
+variables (of the same name).
```
yannickperrenet, line 68: The section should probably have another title, since the section describes behavior in kernels and not during interactive runs. Maybe: "Environment variables inside Notebooks"

ricklamers: I agree, the title should be changed. Feel free to use your suggested title.

yannickperrenet: Will make the change now.

PR #147, opened 2021-02-10
Feature debug dump

fruttasecca:
This PR introduces a new CLI command, `debug`. The command will produce a dump of the logs and other information related to Orchest services. This will help in the debugging process since a user will have to run fewer commands to provide information.

The command also provides an `--ext` option, which outputs more extensive debug information. Note that some user data might be included, depending on the circumstances, e.g. depending on the current state of Orchest and/or one of its services. Moreover, note that when running Orchest in `dev` mode some user data might be included depending on the current state.

Content of the dump:
- Orchest config file
- Version of Orchest images
- Currently running Orchest containers
- Logs of Orchest containers
- Result of health checks 
-  Database schemas and other metadata
- Celery worker logs and the output of different `celery inspect` commands.

### Checklist
- [x] The documentation reflects the changes. (Will update once the name of the command `debug`, is agreed upon.)


yannickperrenet:
> The command also provides an --ext option, which outputs more extensive debug information. Note that some user data might be involuntarily included, depending on the circumstances, e.g. depending on the current state of Orchest and/or one of its services. Moreover, note that when running Orchest in dev mode some user data might also leak depending on the current state.

The "involuntarily" and "data might also leak" I do not like. Is this only present when the `--ext` flag is given? If so, then I would rather just say that the `--ext` could include data and so should only be used by users who are okay with this and drop the "involuntarily" as this sounds as if the software was designed incorrectly.

services/orchest-ctl/app/cli/main.py
```diff
@@ -112,11 +112,44 @@ def stop():
...
+    ext: bool = typer.Option(
+        False,
+        "--ext",
+        show_default=False,
+        help="Get extensive status information.",
+    ),
+):
     """
-    Get status of Orchest.
+    Get Orchest version.
```
yannickperrenet, line 124: This probably got copied along from the version.

```suggestion
    Get status of Orchest.
```

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+import os
+from functools import reduce
+from typing import List, Set
+
+from docker.errors import NotFound
+
+from app import utils
+from app.config import _on_start_images
+from app.docker_wrapper import DockerWrapper
+from app.orchest_resource_manager import OrchestResourceManager
```
yannickperrenet, line 11: I think putting both the `DockerWrapper` and `OrchestResourceManager` into their own module is a bit overkill. Mostly because the resource manager is just another layer of abstraction on top of the `DockerWrapper` (at least in its current state).

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+
+health_check_command = {
+    # Works locally, doesn't work here.
```
yannickperrenet, line 14: I do not understand this comment.

yannickperrenet: I guess its about the `nginx-proxy` line below? It should probably just be removed if it does not work.

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+def health_check() -> None:
+    """
+
+    Returns:
+        Dict mapping container names to health check exit codes.
+    """
+    (
+        running_containers_ids,
+        running_containers_names,
+    ) = OrchestResourceManager().get_containers(state="running")
```
yannickperrenet, line 36: Feels a bit like a code smell to create a `DockerWrapper` and `OrchestResourceManager` for every function in this module. Mostly because the functions in this module are all called from one parent function, so why have every function create its own instance of the class?

services/orchest-ctl/app/debug.py
fruttasecca: You guessed right, will remove.

services/orchest-ctl/app/debug.py
fruttasecca: I'll put both of them in a single module, cannot put them in `orchest.py` because it will lead to circular imports AFAIK.

services/orchest-ctl/app/debug.py
fruttasecca: We can have the parent function create 1 single instance and that's it, I went with making them as independent as possible but I have no problem with changing it.

yannickperrenet, line 36: Why is the health check only run on running containers? I do get it partially, but shouldn't we make sure the health check passes (i.e. just getting the output) when Orchest is fully running? What if the user already stopped Orchest?

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+
+    docker_wrapper = DockerWrapper()
+
+    try:
+        container = docker_wrapper.get_container("orchest-database")
+        container_id = container.id
+
+        if container.status != "running":
+            utils.echo("\tThe orchest-database is not running.")
+            return
```
yannickperrenet, line 67: Now that I am also seeing this part, shouldn't we make it clear to the user that Orchest has to be running before running the debug command?

We could just give them a help message. Possibly even detecting whether Orchest is running and displaying the message based on the state. "Please make sure to start Orchest first", or something like this (see `OrchestApp.status` for an example).

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+
+    docker_wrapper = DockerWrapper()
+    try:
+        container_id = docker_wrapper.get_container("orchest-webserver").id
+    except NotFound:
+        utils.echo(
+            (
+                "\tCould not find the orchest-webserver container. "
+                "That means that it is not running nor stopped, e.g. it "
+                "has never been run or it was removed when stopping Orchest."
```
yannickperrenet, line 245: I think we should keep in mind here that these messages might not make sense to the user. For example: "it was removed when stopping Orchest." makes a lot of sense to us, but I would argue that it does not have to make sense to a user.

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+def orchest_config_dump(path: str) -> None:
+    """Get the Orchest config file, with telemetry UUID removed"""
+
+    # Copy the config
+    with open("/config/config.json") as input_json_file:
+        config = json.load(input_json_file)
+        # Removed for privacy.
+        del config["TELEMETRY_UUID"]
```
yannickperrenet, line 325: Nice catch ;)

services/orchest-ctl/app/debug.py
fruttasecca: Not sure if I understand the question, but the health check is run only on running containers, so  if Orchest is not running it will not report any health issues , as expected, imo.

services/orchest-ctl/app/docker_wrapper.py
```diff
@@ -0,0 +1,383 @@
...
+    async def _copy_file_from_containers(self, files: List[Tuple[str, str, str]]):
+        await asyncio.gather(
+            *[
+                self._copy_file_from_container(id, from_path, to_path)
+                for (id, from_path, to_path) in files
+            ]
+        )
+        await self.close_aclient()
+
+    def copy_files_from_containers(self, files: List[Tuple[str, str, str]]):
```
yannickperrenet, line 374: Should have the `-> None` to indicate it does not actually return the files.

services/orchest-ctl/app/debug.py
fruttasecca: >Now that I am also seeing this part, shouldn't we make it clear to the user that Orchest has to be running before running the debug command?

I think that we can make it clearer that running the debug command while Orchest is running will give you more information, but Orchest does not have to be running for the debug command to go through. And it should not be a requirement imo. The command might be used in unforeseen cases where Orchest is not able to run, completely and/or partially.

services/orchest-ctl/app/docker_wrapper.py
```diff
@@ -0,0 +1,383 @@
...
+        """AI is creating summary for copy_files_from_containers
+
+        Args:
+            files: A list of triples where, for each triple, the first
+            element is the container id, the second element is a path
+            in the container (to copy from) and the third element is a
+            path in the fs (to copy to).
```
yannickperrenet, line 381: Why instead not just receive three different args or kwargs? It would be a lot clearer than taking a tuple.

services/orchest-ctl/app/debug.py
fruttasecca: Agree

services/orchest-ctl/app/docker_wrapper.py
```diff
@@ -0,0 +1,383 @@
...
+    ):
+        container = self.aclient.containers.container(container_id)
+
+        tar_file = await container.get_archive(from_path)
+        file_name = from_path if "/" not in from_path else from_path.split("/")[-1]
+        file = tar_file.extractfile(file_name)
+
+        with open(to_path, "wb") as dest_file:
+            dest_file.write(file.read())
+        tar_file.close()
```
yannickperrenet, line 363: How long does this section take? Otherwise we could make this async as well. But I don't think we should bother if its quick anyways.

services/orchest-ctl/app/docker_wrapper.py
fruttasecca: Mostly because it would make the fact that those 3 pieces of info (id, from ,to) are related more explicit.

services/orchest-ctl/app/docker_wrapper.py
fruttasecca: Takes pretty much no time, the bottleneces are `celery inspect` command and getting container versions.

services/orchest-ctl/app/utils.py
```diff
@@ -88,8 +88,8 @@ def wait_for_zero_exitcode(
     # This will likely take a maximum of 4 tries.
     exit_code = 1
     while exit_code != 0:
-        exit_code = docker_client.exec_run(container_id, cmd)
-        time.sleep(0.5)
+        exit_code = docker_client.exec_runs([(container_id, cmd)])[0]
+        time.sleep(0.25)
```
yannickperrenet, line 92: Now that the `time.sleep` is updated the comment should probably be `8` instead of `4` now. But it would just become a guess at this point and so going with `10` is also completely fine with me.

yannickperrenet (changes requested):
The dump includes the right information I would say, so good job on that.

I do think we should communicate better to the user that Orchest has to be running before executing the debug command.

services/orchest-ctl/app/docker_wrapper.py
yannickperrenet: I prefer going with args/kwargs over the tuple, but since its just a matter of opinion I am fine either way.

yannickperrenet: On another note, the indentation of the docstring has to be like (not sure about the linewrapping here since that is hard to check in a GitHub comment):
```python
"""
            files: A list of triples where, for each triple, the first
                element is the container id, the second element is a path
                in the container (to copy from) and the third element is a
                path in the fs (to copy to).
"""
```
You can refer to: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html if things are unclear about docstring formatting.

fruttasecca: I'll then opt to leave it as it is to save some time.

services/orchest-ctl/app/debug.py
```diff
@@ -0,0 +1,426 @@
...
+
+    docker_wrapper = resource_manager.docker_client
+
+    try:
+        container = docker_wrapper.get_container("celery-worker")
+        if container.status != "running":
+            utils.echo(
+                (
+                    "\tThe celery-worker container is not running, the debug "
+                    "data will be incomplete."
```
yannickperrenet, line 155: Should comments like these not be in the debug dump instead of outputting them to the user? The user might read this and think "oh the debug command is not working correctly". 

Another option could be to go use the logging module, which will then be outputted if the debug command is run with the `-vvv` option (depending on the log level of course).

Lastly, the `utils.echo` function should probably be changed to do the line wrapping by default (since all output should be consistent). Now it needs to be passed which is just unnecessary. Do you know what I mean or do you want me to fix it?

fruttasecca: >Should comments like these not be in the debug dump instead of outputting them to the user? The user might read this and think "oh the debug command is not working correctly".

>Another option could be to go use the logging module, which will then be outputted if the debug command is run with the -vvv option (depending on the log level of course).

I think using the logger for this kind of message is probably for the best.

>Lastly, the utils.echo function should probably be changed to do the line wrapping by default (since all output should be consistent). Now it needs to be passed which is just unnecessary. Do you know what I mean or do you want me to fix it?

I'll do that.

yannickperrenet (approved):

PR #151, opened 2021-02-15
Improved JupyterLab View

ricklamers:
This PR improves the JupyterLab View to have a cleaner loading indicator when a session is starting.

In addition, it fixes some minor issues related to the changes introduced by url-routing.

This PR branch extends `feature-url-routing` as these features were loosely coupled in their implementation.


yannickperrenet:
Before I start reviewing the code, I want to say that I think the swap makes a lot of sense especially for new users. Additionally, the URL routing is terrific as well, this is a huge quality of life improvement in my opinion.

Great work!

services/orchest-webserver/app/app/core/projects.py
```diff
@@ -98,6 +98,14 @@ def _collateral(self, project_uuid: str, project_path: str):
         elif not os.path.isdir(expected_env_dir):
             populate_default_environments(project_uuid)
 
+        # Initialize .git directory
+        expected_git_dir = os.path.join(full_project_path, ".git")
```
yannickperrenet, line 102: I think this assumption works in 99.9% (but might also be my lack of git knowledge) of the cases, especially when it comes to git repositories as projects inside Orchest. So looks good to me.

services/orchest-webserver/app/static/js/src/components/PipelineList.js
```diff
@@ -90,16 +90,17 @@ class PipelineList extends React.Component {
...
-      pipeline_path: pipeline.path,
+      queryArgs: {
+        pipeline_uuid: pipeline.uuid,
+        project_uuid: this.props.project_uuid,
+      },
     };
 
     if (readOnly) {
-      props.readOnly = true;
+      props.queryArgs.read_only = "true";
```
yannickperrenet, line 100: @ricklamers The parsing to strings here is for consistency in the query string (of the URL), like you mentioned the other day. Am I right?

yannickperrenet (changes requested):
Even though everything looks good, I think it makes sense to hold off a little with merging to make sure we have enough time for testing due to the sheer number of minor changes across all the different views.

fruttasecca:
Looks like there are a couple of issues when adding new users. Adding multiple users and then trying to go backward in the page means having to go back 1 time for each user that was added. Moreover, some previous state is still visible (see the picture).
image

Moreover, adding the same user multiple times correctly leads to an error, although the error does not seem to have been managed properly, since it leads to a blank view.

fruttasecca:
Routing does not always set the pane selection properly. For example, select `Projects` and go to the project settings. Then click on any link (pipeline, job, environment). You will be brought to the correct view and the correct pane will be selected, but going back does not select `Projects` again.

yannickperrenet:
Now that we have URL routing I figured I try to use Orchest using multiple tabs. Then I noticed, over all open tabs only one JupyterLab iframe will load. 

Is this intended? And if so, is this purely browser based meaning that using the same instance with multiple users does not run into the same issue?

fruttasecca:
Routing does not currently allow to move forward in the file manager when coming from another view. E.g. it's possible to select any pane but the file manager, the select the file manager and navigate through the file system. It is then possible to go backwards up to the views that was initially selected, while going forward will stop at the "root" of the file manager view. It's probably related to the file manager itself, just mentioning in case it is not.

services/orchest-webserver/app/static/js/src/components/PipelineList.js
ricklamers: That's correct! All queryArgs should be strings.

ricklamers:
> Routing does not currently allow to move forward in the file manager when coming from another view. E.g. it's possible to select any pane but the file manager, the select the file manager and navigate through the file system. It is then possible to go backwards up to the views that was initially selected, while going forward will stop at the "root" of the file manager view. It's probably related to the file manager itself, just mentioning in case it is not.

Deeper routing for the file manager is currently not implemented by choice. It's not clear whether we're sticking to it long term, so further effort for deeper integration is not something we're prioritizing now.

> Now that we have URL routing I figured I try to use Orchest using multiple tabs. Then I noticed, over all open tabs only one JupyterLab iframe will load.
> Is this intended? And if so, is this purely browser based meaning that using the same instance with multiple users does not run into the same issue?

JupyterLab should work in multiple tabs (multiple instances). Turns out Jupyter automatically forces you to use different workspaces if you open the same JupyterLab instance in multiple tabs. This behavior was broken because `pushState` calls were overriden (which was a fix to not introduce no-op history pushes JupyterLab redirects created as an iframe). It looks like we can either give up allowing JupyterLab to be open in multiple tabs _or_ clean back/forward history (no no-op entries). I'm voting to go for giving up the latter, thus supporting having open multiple JupyterLab pages in a single browser. This is because other side effects of disabling in `pushState` are difficult to predict.

> Routing does not always set the pane selection properly. For example, select Projects and go to the project settings. Then click on any link (pipeline, job, environment). You will be brought to the correct view and the correct pane will be selected, but going back does not select Projects again.

With explicit view relations (parent -> child) defined this should now be fixed.

> Looks like there are a couple of issues when adding new users. Adding multiple users and then trying to go backward in the page means having to go back 1 time for each user that was added. Moreover, some previous state is still visible (see the picture).

This is default browser navigation behavior for <form> interactions (POST submit requests). So we'll leave this as is.

>Moreover, adding the same user multiple times correctly leads to an error, although the error does not seem to have been managed properly, since it leads to a blank view.

Added server side validation and client side warnings for this.

services/orchest-webserver/app/app/core/projects.py
ricklamers: https://git-scm.com/docs/gitrepository-layout

I think we're good according to this. It's either a `.git` folder or file. 'bare' git repositories are not supported (as directories in `projects/`).

services/orchest-api/app/app/utils.py
```diff
@@ -327,7 +327,9 @@ def is_environment_in_use(project_uuid: str, env_uuid: str) -> bool:
     """
 
     int_runs = interactive_runs_using_environment(project_uuid, env_uuid)
+    print("int_runs", len(int_runs))
     exps = jobs_using_environment(project_uuid, env_uuid)
+    print("exp", len(exps))
```
yannickperrenet: I guess these `print` statements will be removed later on or be turned into logging calls?

PR #152, opened 2021-02-18
Expand api tests

fruttasecca:
## Description
The PR expands the tests of the `orchest-api` service. Some bugs that have been found during the implementation are also fixed.  

The `orchest-api` tests will now be run against a real database. Setup of the test database is done by invoking docker, which is a requirement for running those tests.

### Checklist


lib/python/orchest-internals/_orchest/internals/test_utils.py
```diff
@@ -0,0 +1,39 @@
+import uuid
+
+
+def uuid4():
```
yannickperrenet, line 4: I would probably just call it `uuid():` or `gen_uuid()`.

lib/python/orchest-internals/_orchest/internals/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    def forget(self):
+        pass
+
+    def revoke(self, *args, **kwargs):
+        self.revoked_tasks.append(args[0])
+
+
+class AbortableAsyncResultMock:
+    def __init__(self, *args, **kwargs):
+        self.task_id = args[0]
```
yannickperrenet, line 32: Should it have an additional check in case `args` has no elements?

```python
self.task_id = None if not args else args[0]
```
depending on what the default case should be of course (so I just took `None` as an example)

scripts/run_tests.sh
```diff
@@ -70,6 +70,49 @@ fi
 VENVS_DIR=$DIR/../.venvs
 mkdir -p $VENVS_DIR
 
+function setup_local_test_db() {
```
yannickperrenet, line 73: What has been the choice to setup the db here instead of in the tests (the Python tests) themselves?

services/orchest-api/app/app/apis/namespace_environment_builds.py
```diff
@@ -108,10 +110,12 @@ class EnvironmentBuild(Resource):
...
     def get(self, environment_build_uuid):
         """Fetch an environment build given its uuid."""
-        env_build = models.EnvironmentBuild.query.get_or_404(
-            ident=environment_build_uuid, description="EnvironmentBuild not found"
-        )
-        return env_build.as_dict()
+        env_build = models.EnvironmentBuild.query.filter_by(
+            uuid=environment_build_uuid
+        ).one_or_none()
+        if env_build:
```
yannickperrenet, line 116: Wouldn't it be better to explicitly check for `None` here?
```python
if env_build is not None:
```

lib/python/orchest-internals/_orchest/internals/test_utils.py
fruttasecca: I'll go with `gen_uuid`, did not go with `uuid` because it gets very annoying with collisions.

lib/python/orchest-internals/_orchest/internals/test_utils.py
fruttasecca: I'll add some checks. [This](https://docs.celeryproject.org/en/stable/_modules/celery/app/task.html#Task.AsyncResult) is what the real constructor expects.

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: Indeed

services/orchest-api/app/tests/namespaces/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+)
+from app.connections import db
+
+
+@pytest.fixture()
+def celery(monkeypatch):
+    """Mock celery and access added and revoked tasks."""
+    celery = CeleryMock()
+    for module in [namespace_environment_builds, namespace_runs, namespace_jobs]:
+        monkeypatch.setattr(module, "make_celery", lambda *args, **kwargs: celery)
```
yannickperrenet, line 34: Would it be cleaner to use `from unittest.mock import MagicMock` like:
```python
from app import celery_app import make_celery

celery_app.make_celery = MagicMock(return_value=celery)
```
Although it does introduce a new dependency.

I am fine either way. I just felt that having to specify `lambda *args, **kwargs: celery` is a bit "hacky" or not completely the purpose of the method.

services/orchest-api/app/tests/namespaces/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+    end of scope of the fixture.
+    """
+
+    config = copy.deepcopy(CONFIG_CLASS)
+
+    # Setup the DB URI.
+    db_host = os.environ.get("ORCHEST_TEST_DATABASE_HOST", "localhost")
+    db_port = os.environ.get("ORCHEST_TEST_DATABASE_PORT", "5432")
+    # Postgres does not accept "-" as part of a name.
+    db_name = str(uuid.uuid4()).replace("-", "_")
```
yannickperrenet, line 82: Maybe this can be incorporated in the `get_uuid()` function through a kwarg such as `use_underscores=False`.

scripts/run_tests.sh
yannickperrenet: It seems that the Docker setup on GitHub Actions runs without any changes to its configuration files. That is pretty nice! 

We probably also don't need to add any Caching to the Actions as all the `orchest-api` run within `1m` :+1: 

The original questions of "What has been the choice to setup the db here" still remains though.

fruttasecca: I would say it's a matter of separation of concerns, e.g. the tests (the fixtures providing connectivity to the db) do not care where the db is coming from and are not responsible for managing it. Giving such responsibility to the python tests means that we are not easily able to provide the same db to tests belonging to other services, whereas this implementation would be a matter of moving the aforementioned function call to another place.

services/orchest-api/app/tests/namespaces/conftest.py
fruttasecca: Not super familiar with `MagicMock`, but would we not be required to use `patch` to avoid the mock being visible out of the scope of the fixture? We would then have to mix `monkeypatch` and `patch`, which might be a code smell.

Not sure if you implied mocking the celery object only once with the snippet of code you posted, but the monkey patching happens in the fixture so that the test function which is making use of it has access to a new `celery` mock object which properties are updated by the tested code and read by the test function.

scripts/run_tests.sh
fruttasecca: >It seems that the Docker setup on GitHub Actions runs without any changes to its configuration files. That is pretty nice!

Indeed, the reason is that the action is not running inside a container, so that's why we have access to docker. I have also experimented with using `services` to have a `postgres` instance running, there were no improvements when it came to pulling the image, and the other tests part of the matrix strategy would end up starting their own service (so 1 postgres service started for each job of the matrix), so I ended up leaving as it is. But it's good to know that services are there if we need to run the action inside a container in the future.

yannickperrenet: > I would say it's a matter of separation of concerns, e.g. the tests (the fixtures providing connectivity to the db) do not care where the db is coming from and are not responsible for managing it.

Exactly this, is a reason (for me) to have the fixtures take control over the db. It could very well be that the database needs to be setup differently for different tests and thus the fixtures need to be able to manage/control it. Now the lifetime of the db is out of the control for the fixtures, which I don't think is a good thing (although for now we might be able to get away with it, but it is not good for extensibility).

Take for example the plasma store of the memory-server. I think its great that its managed as part of a fixture, since the store is part of the testing environment and should therefore also be manageable by the testing environment itself.

In case you do not agree with me we can have a more extensive discussion over a call ;)

services/orchest-api/app/tests/namespaces/test_namespace_environment_builds.py
```diff
@@ -0,0 +1,150 @@
...
+    assert len(data["failed_requests"]) == 1
+
+
+def test_environmentbuildlist_post_with_error2(client, project, monkeypatch):
+    celery = CeleryMock()
+    # Make it so that only the first request will go through.
+    monkeypatch.setattr(
+        namespace_environment_builds,
+        "make_celery",
+        lambda *args, **kwargs: 1 / 0 if len(celery.tasks) > 0 else celery,
```
yannickperrenet, line 57: The ZeroDivisionError is produced so that only (as stated by the comment) the first request succeeds? Is the order deterministic so that different runs of the tests succeed? Or is it just testing what happens in case of a failure (no matter if its the second or first)?

Also I think its better to be explicit here and use `raise` instead of `1 / 0`.

fruttasecca: >The ZeroDivisionError is produced so that only (as stated by the comment) the first request succeeds? 
yep

>Is the order deterministic so that different runs of the tests succeed?
Should be deterministic AFAIK

AFAIK you cannot raise in a lambda, since the body of the lambda should be an expression that reduces to a value. If we want to raise an exception we have to define a function that does so and use it in place of the lambda.

yannickperrenet:
Quite the test cases. What stood out to me it the amount of state that needs to be pre-configured before a particular functionality can be tested.

@fruttasecca How extensible do you think the current test suite is? Extensible when it comes to adding more tests and not having to change tests when changes (in particular additions) are made to the orchest-api.

fruttasecca:
>Quite the test cases. What stood out to me it the amount of state that needs to be pre-configured before a particular functionality can be tested.

Indeed some state needs to be there, e.g. the `project` or a `pipeline`, but I think fixtures are covering that properly.

>@fruttasecca How extensible do you think the current test suite is? Extensible when it comes to adding more tests and not having to change tests when changes (in particular additions) are made to the orchest-api.

The `project`, `pipeline` and other "setup" fixtures help with encapsulation to a nice extent. The fact that we make use of the concept of "two phase" functions should also help in that regard. Of course, a change in the semantics of an end point will probably lead to a change in the tests, that's not really avoidable I suppose. Adding features in the `orchest-api` should bring fewer issues than changing already existing functionality imo, but only time will tell.

services/orchest-api/app/tests/namespaces/test_namespace_environment_builds.py
yannickperrenet: Style wise (following Google's Python styleguide) the use of a nested function is preferred over complicated lambda functions. 

I don't want to be making an argument based on style, but I do think having a nested function together with a `raise` is a lot clearer. Seeing `1 / 0` I would have to think twice "Is it intended or not?". I will leave the decision up to you.

fruttasecca: I agree, I'll use a nested function.

ricklamers (approved):

yannickperrenet (approved):

Issue #135, opened 2021-02-01
Swapping the language of an existing notebook step does not work


ricklamers:
Currently, when an existing pipeline step contains a notebook with a certain language (e.g. Python) the Notebook is created with the metadata in the Notebook for this language.

Ideally, when you change the language (to e.g. R or Julia) the Notebook metadata is updated to reflect the changes in the pipeline definition. Currently, the language is not updated. So when a pipeline step with a notebook is assigned a different language after the first save, the notebook will incorrectly use the language used during the first save of this pipeline step.

Issue #146, opened 2021-02-09
Spacebar drag pipeline panning doesn't work when mousedown is fired on the connection element


ricklamers:
^^^

Solution will need proper event target checking.

PR #157, opened 2021-02-24
Fix notebook language swapping

ricklamers:
## Description
This PR fixes https://github.com/orchest/orchest/issues/135. In addition, some refactors have been made to support this fix.

Specifically, we've made pipeline changes auto save. Which in turn required upgrading the MDC framework to the latest version.

Credits mainly go to @fruttasecca for his work figuring out the right change kernel JupyterLab requests in order to facilitate this fix.

### Checklist


services/orchest-webserver/app/static/js/src/jupyter/Jupyter.js
```diff
@@ -127,6 +138,29 @@ class Jupyter {
...
         }
+      } else {
+        docManager.services.sessions
+          .findByPath(notebook)
+          .then((notebookSession) => {
+            if (notebookSession) {
+              if (notebookSession.kernel.name !== kernel) {
+                orchest.confirm(
+                  "Warning",
+                  "Do you want to change the active kernel of the opened Notebook? \n\nYou will lose the current kernel's state if no other Notebook is attached to it.",
```
yannickperrenet: I am not entirely sure about the context when this warning is thrown, but wouldn't it be better to say: "Are you sure you want to ..."?

fruttasecca: This is to cover the case of a notebook having a kernel running but the notebook not being open in JupyterLab, thus no widget can be found. About the exact wording, I have no strong preference, @ricklamers ?

ricklamers: What @fruttasecca is saying is correct.

In addition, in the 'regular case' in which JupyterLab is running choosing _not_ to change the active kernel does not revert changing of the kernel in the pipeline definition. As JupyterLab and the pipeline are allowed to diverge if the user explicitly chooses to do so.

ricklamers:
This PR also fixes https://github.com/orchest/orchest/issues/146.

PR #162, opened 2021-03-05
Feature adding users through CLI

fruttasecca:
## Description
This PR introduces the possibility of adding users to Orchest through the `orchest` CLI, Orchest needs to be running to do so. Moreover, users can now be an admin. Admin users cannot be deleted.

### Checklist


orchest
```diff
@@ -46,6 +46,7 @@ $SUDO_PREFIX chown $USER "${HOST_CONFIG_DIR}"
 
 $SUDO_PREFIX docker run \
     --rm \
+    -it \
```
yannickperrenet, line 49: Now that we are adding more and more flags to the `docker run` command, I think we should add a comment explaining why we use the options that we do. We don't want to mistakenly remove an option or have a dangling option once we change something in the application.

services/auth-server/app/add_user.py
```diff
@@ -0,0 +1,60 @@
+import argparse
+import uuid
+
+from config import CONFIG_CLASS
```
yannickperrenet, line 4: This import should actually be in the paragraph below, since it is part of our application. Guess this was done by `isort` automatically? I'm sure we can configure this.

services/auth-server/app/add_user.py
```diff
@@ -0,0 +1,60 @@
...
+from werkzeug.security import generate_password_hash
+
+from app import create_app
+from app.connections import db
+from app.models import User
+from app.utils import get_user_conf
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Start plasma store and manager.")
```
yannickperrenet, line 13: The `description` is wrong ;)

services/orchest-ctl/app/cli/main.py
```diff
@@ -232,3 +232,52 @@ def updateserver():
...
+):
+    """
+    Add user to Orchest.
+    """
+    if non_interactive:
+        password = non_interactive_password
+        token = non_interactive_token
+    else:
+        if non_interactive_password:
+            print("Cannot use non_interactive_password in interactive mode.")
```
yannickperrenet, line 264: For the CLI we make use of `typer`, which has the `typer.echo` function which we wrap in `utils.echo` (to allow for line wrapping). I think we should use [printing to standard error](https://typer.tiangolo.com/tutorial/printing/#printing-to-standard-error) instead of `print`.

services/orchest-ctl/app/cli/main.py
```diff
@@ -232,3 +232,52 @@ def updateserver():
...
+        False, "--is-admin", help="True if the user is an admin."
+    ),
+    non_interactive: bool = typer.Option(
+        False, hidden=True, help="Set non interactive mode."
+    ),
+    non_interactive_password: str = typer.Option(
+        "", hidden=True, help="Set non interactive mode."
+    ),
+    non_interactive_token: str = typer.Option(
+        None, hidden=True, help="Set non interactive mode."
```
yannickperrenet, line 253: The type should probably be `Optional[str]` since the default value is set to `None`

yannickperrenet, line 253: The `help` strings need to be updated.

And why is `non_interactive_password` not also set to `None` instead of an empty string `""`?

yannickperrenet, line 253: The `help` strings need to be updated.

And why is `non_interactive_password` not also set to `None` instead of an empty string `""`?

I know its a hidden CLI option, but I still want to make sure we use consistency across all options. So when it comes to `"True if the user is an admin."` for example I don't think this is useful to a user, since the option is specified as `--is-admin` (for true) or `--no-is-admin` (for false), for reference see `orchest install --help` and look for gpu.

services/orchest-ctl/app/orchest.py
```diff
@@ -342,6 +342,45 @@ def version(self, ext=False):
...
+        database_running = False
+        for id, container in zip(ids, running_containers):
+            if "postgres" in container:
+                database_running = True
+            if "auth-server" in container:
+                auth_server_id = id
+
+        if not database_running:
+            utils.echo("The orchest-database service needs to be running.")
+            exit(1)
```
yannickperrenet, line 366: Just for completeness, this `exit` should also use the `typer` alternative.

services/orchest-ctl/app/orchest.py
```diff
@@ -342,6 +342,45 @@ def version(self, ext=False):
...
+        cmd = f"python add_user.py {username} {password}"
+        if token:
+            cmd += f" --token {token}"
+        if is_admin:
+            cmd += " --is_admin"
+
+        exit_code = self.docker_client.exec_runs([(auth_server_id, cmd)])[0]
+
+        if exit_code != 0:
+            utils.echo(f"Non zero exit code: {exit_code}.")
```
yannickperrenet, line 381: I think this help message can be improved, e.g. "Non zero exit code whilst trying to add a user to the auth-server." And then use the logging module to log the exact `exit_code`.

yannickperrenet (changes requested):
Shouldn't the `postgres` and `auth-server` be started when trying to add a user, instead of requiring them to be started? (Wouldn't the `nginx-proxy` be required to be running as well?)

fruttasecca:
> Shouldn't the `postgres` and `auth-server` be started when trying to add a user, instead of requiring them to be started? 

The first reason not to do that was to save time and avoid different code paths, the second is that we would start supporting state changes while Orchest is not running. If in the future we were to expand on the CLI or implement an API I think this might lead to some unforeseen implications. With this said, it is something that can be implemented if we want to do so.

> Wouldn't the `nginx-proxy` be required to be running as well?

Does not seem to be the case, at least locally.

yannickperrenet:
> The first reason not to do that was to save time and avoid different code paths, the second is that we would start supporting state changes while Orchest is not running.

I agree with you here. For the time being we can just require Orchest to be running for CLI commands to be working. The only exception to the rule, would be the `debug` dump command (which I think is a reasonable exception).

If we were to support running a pipeline using the CLI, then it is also easier to require that Orchest is running (plus shutting down Orchest is probably not in the user's interest due to scheduled jobs).

yannickperrenet (approved):

PR #163, opened 2021-03-07
Feature JupyterLab server extensions

ricklamers:
This PR adds JupyterLab server extensions using an image build mechanism similar to environments.

### Checklist


services/jupyter-server/start.sh
```diff
@@ -58,15 +73,23 @@ fi
 # configuration from the userdir.
 build_version=$(jq .jupyterlab.version "$build_path/static/package.json")
 
+# Add new extension tarballs to `extensions/`. This way the
+# extensions get automatically included in the build.
+# Note: new extensions should trigger a build automatically.
+cp -rnT "$build_path/extensions" "$userdir_path/extensions"
```
yannickperrenet, line 79: Why is this moved upwards? This will always do the copy even if it is unnecessary.

services/orchest-api/app/app/apis/__init__.py
```diff
@@ -20,6 +21,7 @@
 )
 
 api.add_namespace(ns_env_builds)
+api.add_namespace(ns_jupyter_builds)
```
yannickperrenet, line 24: Could you make it such that these namespaces are alphabetically sorted (just like the imports).

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -137,6 +138,17 @@ def _transaction(
         project_dir: str,
         host_userdir: str,
     ):
+        # Gate check to see if there is a Jupyter lab build active
+        latest_jupyter_build = models.JupyterBuild.query.order_by(
+            desc(models.JupyterBuild.requested_time)
+        ).first()
```
yannickperrenet, line 144: Would it make sense to make this "gate check" to the `namespace_validations.py`? To make sure the orchest-api doesn't call its own API, the logic can be put in a module level function.

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -137,6 +138,17 @@ def _transaction(
...
+
+        if latest_jupyter_build is not None and latest_jupyter_build.status in [
+            "PENDING",
+            "STARTED",
+        ]:
+            raise Exception("JupyterLab build in progress. Cannot start session.")
```
yannickperrenet, line 150: I think we should have a custom error (probably best defined in the internal library) to trigger the TwoPhaseExecutor. Something like: `TriggerTPEError`. This is way more explicit than it is currently.

@fruttasecca What do you think?

services/orchest-api/app/app/schema.py
```diff
@@ -483,6 +483,48 @@
...
+    },
+)
+
+jupyter_builds = Model(
+    "JupyterBuilds",
+    {
+        "jupyter_builds": fields.List(
+            fields.Nested(jupyter_build),
+            description="Collection of jupyter_builds",
+        ),
```
yannickperrenet, line 514: Could you add `required=True` where necessary, if applicable and otherwise explicitly state `required=False`.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -675,9 +675,23 @@ def _get_container_specs(
         project_uuid=project_uuid[: _config.TRUNCATED_UUID_LENGTH],
         pipeline_uuid=uuid[: _config.TRUNCATED_UUID_LENGTH],
     )
+
+    jupyer_server_image = "orchest/jupyter-server:latest"
+
+    user_jupyer_server_image = (
+        _config.JUPYTER_IMAGE_NAME.format(
+            orchest_version=os.environ.get("ORCHEST_VERSION")
+        )
```
yannickperrenet, line 684: Why does the format need the `orchest_version`? I guess it is so that the up-to-date build is used when running Orchest.

Can't we then just make use of the `:latest` tag to ensure the latest build (user-configured) jupyterlab is used? I think this is cleaner from a Docker perspective. 

Please let me know if I am missing something.

services/orchest-api/app/app/apis/namespace_jupyter_builds.py
```diff
@@ -0,0 +1,248 @@
...
+            )
+            .limit(1)
+            .all()
+        )
+
+        return {"jupyter_builds": [build.as_dict() for build in jupyter_builds]}
+
+
+class SessionInProgressException(Exception):
+    pass
```
yannickperrenet, line 146: Should be in a `error.py` module. Similar to `orchest-sdk/python/orchest/error.py` and `services/orchest-ctl/app/error.py` (you can refer to the [Python docs on user defined exceptions](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions).

services/orchest-api/app/app/apis/namespace_jupyter_builds.py
```diff
@@ -0,0 +1,248 @@
...
+        """Queues a Jupyter build."""
+        try:
+            with TwoPhaseExecutor(db.session) as tpe:
+                jupyter_build = CreateJupyterBuild(tpe).transaction()
+        except SessionInProgressException:
+            return {"message": "SessionInProgressException"}, 500
+        except Exception:
+            jupyter_build = None
+
+        if jupyter_build:
```
yannickperrenet, line 54: We need to specifically check for `None` here, so it should be
```python
if jupyter_build is not None:
```

services/orchest-api/app/app/apis/namespace_jupyter_builds.py
```diff
@@ -0,0 +1,248 @@
...
+
+class SessionInProgressException(Exception):
+    pass
+
+
+class CreateJupyterBuild(TwoPhaseFunction):
+    def _transaction(self):
+
+        # Check if there are any active sessions
+
```
yannickperrenet, line 153: The lines can be removed.

services/orchest-api/app/app/apis/namespace_jupyter_builds.py
```diff
@@ -0,0 +1,248 @@
...
+        # already running, given by the status of PENDING/STARTED.
+        already_running_builds = models.JupyterBuild.query.filter(
+            or_(
+                models.JupyterBuild.status == "PENDING",
+                models.JupyterBuild.status == "STARTED",
+            ),
+        ).all()
+
+        for build in already_running_builds:
+            AbortJupyterBuild(self.tpe).transaction(build.uuid)
```
yannickperrenet, line 176: Does this mean that if you click the `BUILD` button, before the previous build is finished, that the previous build will be aborted? Should this not give a warning to the user instead, e.g.: "Cannot build, a build is already running".

Similarly, we also do not shut down sessions if they are in progress.

yannickperrenet (changes requested):
Great work! This should wrap up persisting user configurations for JupyterLab (see #131).

When it comes to the documentation, what do you think about putting it in [Other](https://orchest.readthedocs.io/en/latest/user_guide/other.html), or do we want to have a complete section about IDE configurations? It is quite the feature I think.

To me the most important questions regarding this PR are:

* Using `:latest` only instead of also using the Orchest version for the names of the Docker images (for jupyter builds).
* Should the user not be warned when building without the previous build being finished?

services/orchest-api/app/app/apis/namespace_sessions.py
fruttasecca: I am not really sure what the use case would be. Would the TwoPhaseExecutor act differently based on an exception being or not being a `TriggerTPEError`?

yannickperrenet: The TPE would not act differently, but raising a generic `Exception` is a bit vague to me. When raising a `TriggerTPEError` (or another name) then at least it is clear that the exception is specifically raised to trigger the TPE.

The fact this it is an `Exception` does not do anything extra besides triggering the TPE, and thus (in my opinion) it should be as clear as possible from the code.

fruttasecca: I think that having a specific exception related to "triggering" the TPE would in some sense imply that there is logic that acts differently based on that, although I don't think that should be the case.  
Maybe the type `Exception` could be replaced with one more specific, but having a TPE specific exception to show intent might not be that useful given that raising an exception is already telling that the code path cannot continue, in a certain sense the code in `TwoPhaseFunction` should not care about the type of exceptions it raises.

Maybe some more input on this would help, @ricklamers?

services/jupyter-server/start.sh
ricklamers: This is different, this just copies _new_ extensions that might have been created in the `$build_path` during the image build of JupyterLab. This is because of the `-n` flag for `cp`.

services/orchest-api/app/app/apis/namespace_sessions.py
ricklamers: I think this doesn't fit here. Because the particular check is never supposed to be called from outside the `orchest-api`. So adding it to `namespace_validations` feel wrong.

services/orchest-api/app/app/apis/namespace_sessions.py
ricklamers: I agree with @fruttasecca. Although I think Yannick correctly identified that raising a generic exception is problematic. It's now a custom exception that is handled in the TPE execution part.

services/orchest-api/app/app/core/sessions.py
ricklamers: It's an explicit choice to make the image name dependent on the Orchest version. Whenever Orchest is updated the JupyterLab image on which the setup_script builds can be changed in arbitrary ways. Thus we always want to rebuild the JupyterLab image whenever Orchest is updated (instead of doing any clever checking to see if it's necessary to rebuild the JupyterLab image).

Make the Orchest version part of the name of the JupyterLab image makes this simple. Orchest just checks whether this Orchest's version JupyterLab image exists, if it doesn't it uses the base JupyterLab image that's included with Orchest when starting a session.

During boot of Orchest a build of the JupyterLab image is triggered if it detects that the user has a JupyterLab setup_script but no JupyterLab image built for the current version of Orchest.

I don't see how this behavior would be possible if you use the `:latest` tag (I think you mean overwriting the JupyterLab base image that comes with Orchest, but that would make rebuilding the `setup_script` starting from the clean bundled base JupyterLab image impossible).

services/orchest-api/app/app/apis/namespace_jupyter_builds.py
ricklamers: You can't actually trigger this from the GUI, as the build button is hidden when a build is in progress. But in case the network request does make it through, the robust thing is to cancel the JupyterLab build. As there can only ever exist on JupyterLab user image at once anyway.

ricklamers:
Thank you :smiley:!

>When it comes to the documentation, what do you think about putting it in Other, or do we want to have a complete section about IDE configurations? It is quite the feature I think.

I think the location I used 'Installing additional packages' currently makes the most sense (given the table of contents of the docs).

services/jupyter-server/start.sh
yannickperrenet: > This is because of the -n flag for cp.

I know that ;)

Giving it a second look, I guess it was moved upwards so that the extensions the user installs (through the bash script) will be copied to the userdir path as well. Am I right?

If this is the case, then the comments should reflect this. To me this was not obvious from the current comments.

services/orchest-api/app/app/apis/namespace_sessions.py
yannickperrenet: I get what you are saying, but the current structure of the `orchest-api` allows utility functions to be added to `namespace_*` files that can be imported in other namespaces. Remember going with this approach over creating a `utils_*` file for every namespace. So due to that structure I would say it is fine putting it in the `namespace_validations.py`

@fruttasecca What do you think?

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -16,6 +17,10 @@
 api = register_schema(api)
 
 
+class JupyterBuildInProgressException(Exception):
+    pass
```
yannickperrenet, line 21: This should be moved to a `error.py` as well.

yannickperrenet:
> 'Installing additional packages' currently makes the most sense

I disagree here. "Installing additional packages" is about installing packages such as `tensorflow`, not jupyterlab extensions, which in my opinion is a totally different context.

I will go over the newly written section in the docs and see whether I can come up with another place to put it.

yannickperrenet:
When opening the pipeline editor, whilst doing a JupyterLab build the user is greeted with the error saying "Cannot start session. A JupyterLab build is still in progress.". The error is triggered because we automatically try to start the session (if none is active) when a user opens the pipeline editor.

I think, instead, we should not automatically try to start the sessions (since we can check whether the build is in progress). Then if a user wants to start a session, then it is fine to get the error. I would rather not give a user an error if we can prevent it.

To achieve this behavior we would have to discuss it in https://github.com/orchest/orchest/pull/163#discussion_r590046482 as well.

services/orchest-api/app/app/core/sessions.py
yannickperrenet: > Making the Orchest version part of the name of the JupyterLab image makes this simple. Orchest just checks whether this Orchest's version JupyterLab image exists, if it doesn't it uses the base JupyterLab image that's included with Orchest when starting a session.

This is a different approach from the environment images. Whenever Orchest is updated, we remove all environment build images (refer to [orchest-ctl update](https://github.com/orchest/orchest/blob/feature-jupyterlab-server-extensions/services/orchest-ctl/app/orchest.py#L295)). Similarly, whenever Orchest is updated all JupyterLab builds should be removed.

Using that approach, the Orchest version no longer needs to be included in the image name and we can just use the `:latest` tag to get the most up-to-date image whenever a user does multiple builds (without updating Orchest in between).

ricklamers:
Some thoughts on why I'd prefer to keep it the way it is currently:

- I think it's inconsistent to 'silently' disable autostart when JupyterLab is building in the background.
- When a JupyterLab build is triggered (which can happen from the back-end as well, for example during start, similar to how migrations work) it should also fail if there are sessions running. So this validation should be part of the build TPF.
- I think the code is too short to factor our into a utility function that is in namespace_validations. This isn't exact science imo. But my heuristic is roughly: if I don't know any other caller and it's a short piece of code, I don't abstract it out.
- The JupyterLab image name with the version is more robust, as it will not be possible to load an outdated JupyterLab after update (even if for example the removal of the JupyterLab use image did not succeed)

ricklamers:
> I will go over the newly written section in the docs and see whether I can come up with another place to put it.

I like your changes to the docs! It does it more justice like this and is more cleanly separated from the environment packages (to avoid confusion).

services/orchest-api/app/app/apis/namespace_sessions.py
fruttasecca: I think this might have to go in validations so that it can easily become an exposed endpoint/gate check in the future.

>I think, instead, we should not automatically try to start the sessions (since we can check whether the build is in progress). Then if a user wants to start a session, then it is fine to get the error. I would rather not give a user an error if we can prevent it.

In any case, I think another problem would be that from the pipeline editor I would have to try out starting a session to know if a session can be started or not without changing views.

yannickperrenet (approved):

fruttasecca (approved):
lgtm, great PR

PR #170, opened 2021-03-11
Feature UX cleanup

ricklamers:
This PR cleans up the user interface of Orchest. The changes are mainly aimed at consistency and better alignment with the core concepts in Orchest (projects, pipelines, jobs, and environments).

In addition, the front-end entry point is consolidated into a single main `App` React component to clean up the application code structure.

Screenshots for comparison:

Old
image

New
image

Old
image

New
image

### Checklist


yannickperrenet:
First things first, I love the UX cleanup. It looks way better in my opinion <3

I do have a few pointers:

* The left menu panel seems to be closed when Orchest is started. Should it not be open instead?
* The UI seems to be slower then before. When I move around I am often greeted with a loading indicator. Any idea why I might be experiencing this behavior?

Other than that I think it is a definite improvement!

ricklamers:
> The left menu panel seems to be closed when Orchest is started. Should it not be open instead?

Good catch! The logic initializing the App state was off for the default case.

In terms of performance the only thing that could cause a sense of reduced performance imo is the fact that we have to load 'projects' on initial page load.

The 'projects' endpoint does file system based discovery as a 'trigger'. Perhaps we should make this toggleable with query argument to have a 'faster' projects query for the top menu item.

fruttasecca:
Great UX cleanup, massive improvement imo.

Can "Pipeline path" be moved above the actual pipeline path or be changed in some way so that the pipeline path (the gray box actually) is vertically aligned with the text areas of the pipeline name and the memory size?
Moreover, could we just use "path"?
image

PR #176, opened 2021-03-15
Add cloud mode

fruttasecca:
## Description

This PR adds another mode in which Orchest can be run, `cloud`. Currently this only implies that some settings cannot be modified through the GUI. Cloud mode can be seen as a "regular" production mode, with some changes on top of it. This means that it does not have the interactivity of `dev` mode.

### Checklist

### DIFF
--- a/services/orchest-ctl/app/cli/main.py
+++ b/services/orchest-ctl/app/cli/main.py
@@ -242,11 +242,23 @@ def restart(
 
 
 @typer_app.command(hidden=True)
-def updateserver():
+def updateserver(
+    # Necessary to make it so that the update server restarts Orchest
+    # with the correct settings.
+    cloud: bool = typer.Option(
+        False,
+        show_default="--no-cloud",
+        help=cli_start.__CLOUD_HELP_MESSAGE,
+        hidden=True,
+    ),
+    dev: bool = typer.Option(
+        False, show_default="--no-dev", help=cli_start.__DEV_HELP_MESSAGE
+    ),
+):
     """
     Update Orchest through the update-server.
     """
-    app._updateserver()
+    app._updateserver(cloud, dev)
 
 
 @typer_app.command(hidden=True)
--- a/services/orchest-ctl/app/cli/main.py
+++ b/services/orchest-ctl/app/cli/main.py
@@ -72,11 +72,6 @@ typer_app.add_typer(cli_start.typer_app, name="start")
 app = OrchestApp()
 
 
-class Mode(str, Enum):
-    reg = "reg"
-    dev = "dev"
-
-
 class Language(str, Enum):
     python = "python"
     r = "r"
@@ -149,7 +144,7 @@ def debug(
     The command does not need Orchest to be running when called, but
     will produce a less inclusive dump if that is not the case.
 
-    Note: if Orchest was/has been running in dev mode, then there is the
+    Note: if Orchest was/has been running with --dev, then there is the
     possibility of some user data getting into the dump due to the log
     level being set to DEBUG.
 
@@ -203,17 +198,36 @@ def update(
 
 @typer_app.command()
 def restart(
-    mode: Mode = typer.Option(
-        Mode.reg, help="Mode in which to start Orchest afterwards."
+    # Support the old way of passing mode so that an old version of
+    # Orchest does not break when updating.
+    mode: str = typer.Option(
+        "reg",
+        help=(
+            "Mode in which to start Orchest afterwards. This option is "
+            "deprecated and will be removed in a future version. Use --dev "
+            "instead of --mode=dev."
+        ),
+        hidden=True,
     ),
     port: Optional[int] = typer.Option(
         8000, help="The port the Orchest webserver will listen on."
+    ),
+    cloud: bool = typer.Option(
+        False,
+        show_default="--no-cloud",
+        help=cli_start.__CLOUD_HELP_MESSAGE,
+        hidden=True,
+    ),
+    dev: bool = typer.Option(
+        False, show_default="--no-dev", help=cli_start.__DEV_HELP_MESSAGE
     ),
 ):
     """
     Restart Orchest.
     """
-    container_config = get_container_config(mode.value)
+
+    dev = dev or (mode == "dev")
+    container_config = get_container_config(cloud=cloud, dev=dev)
     port_bind: dict = {
         "nginx-proxy": {
             "HostConfig": {
--- a/services/orchest-ctl/app/cli/start.py
+++ b/services/orchest-ctl/app/cli/start.py
@@ -82,7 +82,7 @@ def cloud(
 
     Running Orchest in CLOUD mode changes GUI functionality. For example
     making it impossible to disable the authentication layer. Settings
-    that cannot be modified through the GUI because of cloud mode, as
+    that cannot be modified through the GUI because of CLOUD mode, as
     all settings, can still be modified by changing the config.json file
     directly, which can be found at ~/.config/orchest/config.json.
 
--- a/services/orchest-ctl/app/cli/start.py
+++ b/services/orchest-ctl/app/cli/start.py
@@ -81,7 +81,9 @@ def cloud(
     Start Orchest in CLOUD mode.
 
     Starting Orchest in cloud mode makes it so that the user uuid, auth
-    mode and telemetry_disabled are not modifiable through the GUI.
+    mode and telemetry_disabled are not modifiable through the GUI. To
+    modify them you will need to modify the config.json file directly,
+    which can be found at ~/.config/orchest/config.json.
     """
     container_config = get_container_config("cloud")
 
--- a/services/orchest-ctl/app/cli/start.py
+++ b/services/orchest-ctl/app/cli/start.py
@@ -6,6 +6,7 @@ import typer
 
 from app.orchest import OrchestApp
 from app.spec import get_container_config, inject_dict
+from app.utils import echo
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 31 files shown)


services/orchest-ctl/app/cli/start.py
```diff
@@ -70,6 +71,34 @@ def reg(
...
+def cloud(
+    port: Optional[int] = typer.Option(
+        8000, help="The port the Orchest webserver will listen on."
+    )
+):
+    """
+    Start Orchest in CLOUD mode.
+
+    Starting Orchest in cloud mode makes it so that the user uuid, auth
+    mode and telemetry_disabled are not modifiable through the GUI.
```
yannickperrenet, line 84: I suggest making it into:
```suggestion
    Running Orchest in CLOUD mode changes GUI functionality. For 
    example making it impossible to disable the authentication layer.
```
because otherwise the comment might quickly become outdated.

After additional functionality it should probably become a section in the docs (as you noted).

services/orchest-webserver/app/app/config.py
```diff
@@ -53,6 +53,13 @@ class Config:
     TELEMETRY_DISABLED = False
     TELEMETRY_INTERVAL = 15  # in minutes
 
+    CLOUD_MODE = _config.CLOUD_MODE
+    _CLOUD_UNMODIFIABLE_SETTINGS = [
```
yannickperrenet, line 57: `_CLOUD_UNMODIFIABLE_CONFIG_VALUES` makes more sense to me.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
```diff
@@ -189,6 +207,19 @@ class SettingsView extends React.Component {
...
               <span className="code">config.json</span>.
+              {(() => {
+                if (orchest.config["CLOUD_MODE"] === true) {
+                  return (
+                    <span>
+                      {" "}
+                      Note that <span className="code">AUTH_ENABLED</span>,{" "}
+                      <span className="code">TELEMETRY_DISABLED</span>,{" "}
+                      <span className="code">TELEMETRY_UUID</span> cannot be
+                      modified in cloud mode.
```
yannickperrenet, line 218: * `cloud` should also be `<span className="code">cloud</span>`.
* The version information in settings should now also contain `cloud`. Or we could just remove: "The application mode can either be: production or development", since for cloud users it will always be cloud and thanks to this PR we can finally make that distinction.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
```diff
@@ -97,9 +97,27 @@ class SettingsView extends React.Component {
...
+
+          try {
+            let configJSON = JSON.parse(data);
+
+            this.setState({
+              configJSON,
+            });
+
+            // Refresh the page when auth gets enabled in the config.
+            shouldReload = configJSON.AUTH_ENABLED && !authWasEnabled;
```
yannickperrenet, line 111: Why is this done? Doesn't really make sense to me, because for other users that connect to Orchest the authentication would be enabled. Now the user just gets "thrown" out of Orchest. 

On another note, enabling authentication should not be possible without having configured a single user.

@ricklamers Do you want to chip in here?

services/orchest-webserver/app/app/views/views.py
```diff
@@ -197,21 +199,24 @@ def version():
...
             config = request.form.get("config")
 
             try:
-                # only save if parseable JSON
-                json.loads(config)
-                save_user_conf_raw(config)
+                # Only save if parseable JSON.
+                config_update = json.loads(config)
+                config = update_orchest_config(current_config, config_update)
+                save_user_conf_raw(json.dumps(config))
```
yannickperrenet, line 212: To me it is a bit unclear what is being done here (including in the `update_orchest_config` function).

yannickperrenet (changes requested):
The way it is implemented now would require the `_CLOUD_UNMODIFIABLE_SETTINGS` to be set in the `config.json` beforehand. So instead of setting a default value in the code for the `cloud` mode, the default value is assumed to be set in the `config.json`. 

Not sure whether that is something we want.

Additionally, I think the updating mechanism is a little hard to follow from a code perspective.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
fruttasecca: >Or we could just remove: "The application mode can either be: production or development"

Would that also imply removing the `Application mode: production` on the right? If so, I would go for that since it's less text that the user has to look at.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
fruttasecca: >Now the user just gets "thrown" out of Orchest.

In that regard, the PR does not alter functionality.

>On another note, enabling authentication should not be possible without having configured a single user.

We would need to expand the API of the `auth-server` I think.

fruttasecca:
> The way it is implemented now would require the `_CLOUD_UNMODIFIABLE_SETTINGS` to be set in the `config.json` beforehand. 

I think this is untrue, each one of those settings has a default value set in the code if they are not set already, although said values are set in a somewhat scattered way.

services/orchest-ctl/app/cli/start.py
```diff
@@ -70,6 +71,38 @@ def reg(
...
+    )
+):
+    """
+    Start Orchest in CLOUD mode.
+
+    Running Orchest in CLOUD mode changes GUI functionality. For example
+    making it impossible to disable the authentication layer. Settings
+    that cannot be modified through the GUI because of CLOUD mode, as
+    all settings, can still be modified by changing the config.json file
+    directly, which can be found at ~/.config/orchest/config.json.
```
yannickperrenet, line 87: This path is not 100% true as the `$XDG_CONFIG_HOME`  variable takes precedence if set. Maybe make it "'config.json' configuration file".

services/orchest-webserver/app/static/js/src/views/SettingsView.js
ricklamers: If a user enables authentication, and the page is not reloaded, iframes and endpoint requests start to get redirected to the login page. I felt reloading the page when auth is enabled is nice, because it redirects to the login screen to create a session, or is a (almost, just page reloading) no-op in case an authenticated session is in place.

yannickperrenet: I see. In that case I understand why the page is refreshed, but from a user interaction I don't like that it is.

Take again my example where you could basically lock your Orchest installation as you can enable authentication without having a single user defined.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
```diff
@@ -261,11 +292,20 @@ class SettingsView extends React.Component {
...
         <div className="columns">
           <div className="column">
-            <p>
-              Find out Orchest's version and application mode. The application
-              mode can either be: <span className="code">production</span> or{" "}
-              <span className="code">development</span>.
-            </p>
+            {(() => {
+              if (orchest.config["CLOUD_MODE"] === true) {
+                return <p>Find out Orchest's version.</p>;
```
ricklamers, line 297: This should be changed before the PR merges I think.

This cloud|development|production is a bit of a modeling error perhaps. I think it should still just be development|production and the "cloud" flag should be an orthogonal concern.

yannickperrenet:
I just discussed with @ricklamers and we decided it would be better to remove the notion of a `mode` in Orchest, meaning that `cloud` and `dev` would become flags.

This then also affects the documentation and general wording in the CLI.

@fruttasecca Let me know if you strongly disagree.

fruttasecca:
@yannickperrenet @ricklamers 

I guess we are removing such notions at the GUI level as well, right?
image

yannickperrenet:
> I guess we are removing such notions at the GUI level as well, right?

For example Flask does call it "development mode". 

@ricklamers Was the main advocate for making it flags instead of separate commands, so I will let him make the decision here.

ricklamers:
> I guess we are removing such notions at the GUI level as well, right?

I think the best thing to do would be to indicate that the development flag is enabled. The notion of a 'mode' should I think be removed.

services/orchest-webserver/app/static/js/src/views/SettingsView.js
yannickperrenet: Can we add a comment to the config section that if a user enabled authentication, they will be automatically redirected to the login page?

services/orchest-ctl/app/cli/main.py
```diff
@@ -203,17 +198,36 @@ def update(
...
-    mode: Mode = typer.Option(
-        Mode.reg, help="Mode in which to start Orchest afterwards."
+    # Support the old way of passing mode so that an old version of
+    # Orchest does not break when updating.
+    mode: str = typer.Option(
+        "reg",
+        help=(
+            "Mode in which to start Orchest afterwards. This option is "
+            "deprecated and will be removed in a future version. Use --dev "
+            "instead of --mode=dev."
```
yannickperrenet, line 208: Why don't we remove it in this PR already?

fruttasecca: Because of the way the updating mechanism works. Updating from the client means pulling the up to date `ctl` image, then sending a `restart` command to it, but the image sending the command will be the "old" one, which still relies on this option. (Assuming I am correct :P)

yannickperrenet: Yes, the update mechanism is quite complex. Good catch.

@ricklamers Do you think it is time to revision the update mechanism? This is definitely not the first time we are running into issues for updates.

ricklamers: @fruttasecca and I spent a bit of time reviewing how we could handle this better (or refactor things to avoid it in the future). As of now, it seems like there aren't strictly better solutions available. The update mechanism isn't wrong by design, it's just a complicated feature.

We can review/refactor/improve in the future. But quick wins seemed out of sight.

yannickperrenet (approved):

PR #199, opened 2021-04-05
Front-end Foundations

joe-bell:
## Architectural Changes

This PR introduces two key architectural changes to the `orchest` monorepo:

### 1. Dependency Management

In it's current form, common JavaScript code is shared between client-side projects via symlinks. It's generally fine as a short-term solution, but it's unstable; with dependencies left unmanaged.

To address this, [**pnpm**](https://pnpm.io/) has been implemented, offering a fast, robust and strict solution for managing dependencies across the project.

In addition to this, each directory in `lib/javascript` is now treated as a local npm package; with their own `dev` and `build` scripts. 

Every local package name is prefixed with `@orchest/`

### 2. Separation of Client and Server 

To make our product more reliable, robust and enable further enhancements to Orchest's Front-end with confidence - a complete separation of client and server was needed.

Each service with a frontend (currently `orchest-webserver` and `authserver`) is now split into two directories:
- `app` - server-side
- `client` - client-side

The `client` directory–similarly to `lib/javascript` directories–is treated as an npm package, with dependencies handled by `pnpm`.

[`vite`](http://vitejs.dev/) is now used in place of esbuild; compiling all client and static files (rather than just the JavaScript bundle).

## Workflow

The goal with this refactor was to make the developer experience as close as possible to how it is currently.

The only real change _needed_, has been documented in [`docs/source/developer_guide/development_workflow`](docs/source/developer_guide/development_workflow.rst)

### Running Commands

Commands for each package can be run in the root via:

```sh
pnpm run -r <script-name>
```

See [pnpm](https://pnpm.io)'s docs for more info

## Going Forwards

With these strong foundations set, we should now be able to make further enhancements to the Front-end with confidence.

Any future changes should be made with the following goals in mind:

> Our Front-end should be:
> 
> 1. **Resilient**
> 2. **Robust**
> 3. **Refined**

### Checklist


ricklamers:
Note: a rather important todo is to configure the `auth-server` to properly build and serve from within the Docker container. Should be fairly simple, but I ran out of time >.<

PR #202, opened 2021-04-13
Add single param get/update for step, pipeline

fruttasecca:
## Description
This PR expands the Orchest SDK to add single parameter `get` and `update` operations for both step and pipeline parameters.
It's still possible to `get` or `update` parameters in a batched manner by using the `get_params` and `update_params` functions.

Signatures:
`def get_step_param(key: str) -> Any:`
`def get_pipeline_param(key: str) -> Any:`
`def update_step_param(key: str, value: Any) -> None:`
`def update_pipeline_param(key: str, value: Any) -> None:`

### Checklist

### DIFF
--- a/orchest-sdk/python/orchest/__init__.py
+++ b/orchest-sdk/python/orchest/__init__.py
@@ -1,9 +1,7 @@
 from orchest.config import Config
 from orchest.parameters import (
-    get_params,
     get_pipeline_param,
     get_step_param,
-    update_params,
     update_pipeline_param,
     update_step_param,
 )
--- a/orchest-sdk/python/orchest/__init__.py
+++ b/orchest-sdk/python/orchest/__init__.py
@@ -1,3 +1,10 @@
 from orchest.config import Config
-from orchest.parameters import get_params, update_params
+from orchest.parameters import (
+    get_params,
+    get_pipeline_param,
+    get_step_param,
+    update_params,
+    update_pipeline_param,
+    update_step_param,
+)
 from orchest.transfer import get_inputs, output
--- a/orchest-sdk/python/orchest/parameters.py
+++ b/orchest-sdk/python/orchest/parameters.py
@@ -13,13 +13,13 @@ from orchest.pipeline import Pipeline, PipelineStep
 from orchest.utils import get_step_uuid
 
 
-def __get_pipeline() -> Pipeline:
+def _get_pipeline() -> Pipeline:
     with open(Config.PIPELINE_DEFINITION_PATH, "r") as f:
         pipeline_definition = json.load(f)
     return Pipeline.from_json(pipeline_definition)
 
 
-def __get_current_step(pipeline: Pipeline) -> PipelineStep:
+def _get_current_step(pipeline: Pipeline) -> PipelineStep:
     try:
         step_uuid = get_step_uuid(pipeline)
     except StepUUIDResolveError:
@@ -34,8 +34,8 @@ def get_params() -> Tuple[dict, dict]:
         A tuple of two elements, where the first is the parameters of
         the current step, the second is the parameters of the pipeline.
     """
-    pipeline = __get_pipeline()
-    step = __get_current_step(pipeline)
+    pipeline = _get_pipeline()
+    step = _get_current_step(pipeline)
     return step.get_params(), pipeline.get_params()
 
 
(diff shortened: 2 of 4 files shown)


orchest-sdk/python/orchest/__init__.py
```diff
@@ -1,3 +1,10 @@
 from orchest.config import Config
-from orchest.parameters import get_params, update_params
+from orchest.parameters import (
+    get_params,
+    get_pipeline_param,
+    get_step_param,
+    update_params,
+    update_pipeline_param,
+    update_step_param,
+)
```
yannickperrenet, line 9: Do we want to expose all of these functions directly? I think we should only expose our "recommended" functions and let the users import the `parameters` module themselves to get another function.

orchest-sdk/python/orchest/parameters.py
```diff
@@ -5,34 +5,38 @@
...
 
 
-def get_params() -> Tuple[dict, dict]:
-    """Gets the parameters of the current step and the pipeline.
-
-    Returns:
-        A tuple of two elements, where the first is the parameters of
-        the current step, the second is the parameters of the pipeline.
-    """
+def __get_pipeline() -> Pipeline:
```
yannickperrenet, line 16: Why the double leading underscores instead of single?

orchest-sdk/python/orchest/__init__.py
fruttasecca: I would leave them exposed, they are part of the SDK so I see no reason in creating a discrepancy.

orchest-sdk/python/orchest/parameters.py
```diff
@@ -54,26 +58,85 @@ def update_params(
...
+
+def get_step_param(key: str) -> Any:
+    """Gets a parameter of the current step by name.
+
+    Args:
+        key: The step parameter to get.
+
+    Returns:
+        The value that was mapped to the step parameter name.
+    """
```
yannickperrenet, line 86: The documentation reads "step parameter name" and "step by name", but the function signature is "key". Unsure how exactly users think about defining the JSON in the UI, do they think of it as "key"s or "parameter"s. My guess is the second and thus I would opt for changing `key: str` to `param_name: str`

orchest-sdk/python/orchest/parameters.py
fruttasecca: No particular reason, can make it into a single.

orchest-sdk/python/orchest/parameters.py
```diff
@@ -54,26 +58,85 @@ def update_params(
...
+
+    Returns:
+        The value that was mapped to the pipeline parameter name.
+    """
+    pipeline = __get_pipeline()
+    params = pipeline.get_params()
+    return params[key]
+
+
+def update_step_param(key: str, value: Any) -> None:
```
yannickperrenet, line 107: I think this function should internally call the `update_params`.

yannickperrenet (changes requested):
A few minor wording changes have to be made.

On another note, shouldn't we go for something like a class instead of defining all these functions? For example:
```python
class Params:
  def get(name: str) -> Any
```
Although this is less inline with the `transfer` module. Just wanted to share my thought of taking a more class oriented approach.

orchest-sdk/python/orchest/parameters.py
yannickperrenet: Yes, I prefer single as it is inline with PEP8.

fruttasecca:
> A few minor wording changes have to be made.
> 
> On another note, shouldn't we go for something like a class instead of defining all these functions? For example:
> 
> ```python
> class Params:
>   def get(name: str) -> Any
> ```
> 
> Although this is less inline with the `transfer` module. Just wanted to share my thought of taking a more class oriented approach.

Given the limited scope of the parameters module I think we can leave it as it is right now.

orchest-sdk/python/orchest/__init__.py
yannickperrenet: We do the same for the `transfer` module. And I would argue that exposing functionality top-level exactly justifies the introduction of such a discrepancy. We can define exactly the function top-level which should be used by the user (in most cases).

fruttasecca: What I dislike is that we would have that a rather "obvious" functionality, i.e. getting all the params, is not exposed. I don't feel strongly about this so I am fine going either way since the user can just import `parameters`, let me know if I should pull the trigger.

yannickperrenet: I think the main power is that we can choose what we want to expose at the top-level and thus pointing users to use certain functions rather than others. So I would rather not make "getting all the params" and "updating all the params" top-level, and let users import `parameters` to get the functionality.

yannickperrenet (approved):

ricklamers:
Note, we should have incremented the SDK version to reflect the API change.

PR #203, opened 2021-04-14
Fix preserve port on restart

fruttasecca:
## Description
Fixes the fact that the port which Orchest was started with was not preserved between restarts and updates.

### Checklist


yannickperrenet (approved):
I like using environment variables to pass this information to the correct containers, sounds like the-way-to-do-these-things :)

And seems to work like a charm :+1:

PR #207, opened 2021-04-15
TypeScript Foundations

joe-bell:
## Description

- Adds initial TypeScript set up. Allowing us to start incrementally converting our Front-end codebase to TypeScript as we add new features
- Align `client-` package names

### Checklist


.pre-commit-config.yaml
```diff
@@ -5,8 +5,8 @@ repos:
         name: "pnpm pre-commit"
         language: "system"
         entry: "pnpm run pre-commit"
-        always_run: true
         pass_filenames: false
+        files: \.(js|jsx|json|ts|tsx|html|css|scss)$
```
joe-bell, line 9: @yannickperrenet I think you'll be chuffed to see this - now this only runs on _these_ specific file types (not all files, like before)

ricklamers:
This looks like a good initial step of adding TypeScript.

Two questions:

- It looks like the pnpm linter is now being skipped in the GitHub Action. Or am I seeing that incorrectly?
- I think the Dockerfile pnpm build needs to receive the updated filter (not sure if it’s referenced anywhere else)

joe-bell:
@ricklamers thanks!

1. It does, but [we run it manually instead](https://github.com/orchest/orchest/pull/207/files#diff-ac5eead3f3ce4863c524fff031a87b7aecccb4a0493df087a4e1c704a1505036R31) (the build script will catch out any type errors too). I spent an hour or two debugging this and trying to get it working in Precommit (you can see the commits haha), but I don't think it's worth spending more time on this
2. Fixed, thank you!

.github/workflows/pre-commit.yml
```diff
@@ -23,8 +26,9 @@ jobs:
         with:
           node-version: "${{ steps.nvm.outputs.NVMRC }}"
 
+      # Run our pnpm hook manually, with additional build checks
       - name: Setup pnpm
-        run: npm run setup && pnpm i
+        run: npm run setup && pnpm install && pnpm run build
```
ricklamers, line 31: The pre-commit should not run a full build IMO. For pre-commit I'd rather stick to formatters & linters at most. Limiting their execution time. If you want to build for verification it should be in the test suite not in the pre-commit hook.

ricklamers (changes requested):

.github/workflows/pre-commit.yml
joe-bell: This is the GitHub action workflow – it only tiggers as part of the review step, **not** locally

If you'd prefer I can create a new GitHub Action to run this @ricklamers

ricklamers: Nevermind, I’m apparently blind 😅

joe-bell:
That last commit made me feel dirty, but it'll work for now…

Let's consider ditching `react-codemirror2` in the future to avoid this `global is not defined` issue

ricklamers (approved):

PR #214, opened 2021-04-19
Introduce OrchestContext (and useOrchest)

joe-bell:
## Description

This PR sets the foundations for how we want to manage global/shared state going forwards – via **`useOrchest`**

`useOrchest` (a hook to access `OrchestContext`) returns 3 keys:

```jsx
const {
  // Globally-accessed state
  state,
  // Actions to modify the globally-accessed state
  dispatch,
 // Helpers to filter/find specific areas of the state (e.g. a specific session)
  get,
} = useOrchest();
```

Note: my actual preference would've been to have separate hooks for separate concerns (e.g. sessions, pipelines, projects), but given that most of our components are `class` based, we can only use one context at a time

## How It Works

Under-the-hood, `useOrchest` uses [`React.useReducer()`](https://reactjs.org/docs/hooks-reference.html#usereducer) to handle state, with a private `SessionProvider` (which fetches/updates sessions via [SWR](https://swr.vercel.app/)).

Whiteboard 30 Apr 2021

### Functional Components

> **⚠️ Important Note**
>
> Don't forget to add `// @ts-check`, see **Closing Thoughts**

Make use of the **`useOrchest()`** hook

```jsx
import React from "react";
import { useOrchest } from "@/hooks/orchest";

const SessionsListFC = () => {
  const { state, dispatch } = useOrchest();

  return (
    <article>
      <h1>Sessions</h1>
      <ul>
        {state.sessions?.map(
          (session, i) =>
            session && (
              <li key={i}>
                <dl>
                  <dt>Project UUID</dt>
                  <dd>{session.project_uuid}</dd>
                  <dt>Pipeline UUID</dt>
                  <dd>{session.pipeline_uuid}</dd>
                </dl>

                <button
                  onClick={() =>
                    dispatch({ type: "sessionToggle", payload: session })
                  }
                >
                  Toggle Session
                </button>
              </li>
            )
        )}
      </ul>
    </article>
  );
};
```

## Class Components

> **⚠️ Important Note**
>
> Going forwards we should **not** be using class components, see **Closing Thoughts**

Set the `contextType` to OrchestContext and consume via **`this.context`** 

```jsx
import React from "react";
import { OrchestContext } from "@/hooks/orchest";

class SessionsListCC extends React.Component {
  static contextType = OrchestContext;

  constructor(props, context) {
    super(props, context);
  }

  render() {
    return (
      <article>
        <h1>Sessions</h1>
        <ul>
          {this.context.state.sessions?.map(
            (session, i) =>
              session && (
                <li key={i}>
                  <dl>
                    <dt>Project UUID</dt>
                    <dd>{session.project_uuid}</dd>
                    <dt>Pipeline UUID</dt>
                    <dd>{session.pipeline_uuid}</dd>
                  </dl>
  
                  <button
                    onClick={() =>
                      this.context.dispatch({ type: "sessionToggle", payload: session })
                    }
                  >
                    Toggle Session
                  </button>
                </li>
              )
          )}
        </ul>
      </article>
    );
  }
}
```

## Closing Thoughts

This PR just sets the foundations, it doesn't refactor the whole codebase (we don't have the time for that right now). So to make our lives easier going forwards, the following points should be considered for new changes

1. **Do** add `// @ts-check` to the top of a file (to include it in our TypeScript checks/linting)
1. **Do not** use class components – they don't support type-checking, and they don't support multiple contexts

## Checklist


services/orchest-webserver/client/src/components/ProjectSelector.jsx
```diff
@@ -56,60 +47,60 @@ class ProjectSelector extends React.Component {
...
-      this.setState({
-        selectItems: this.listProcess(projects),
-        projects: projects,
-      });
+      setSelectItems(listProcess(projectsRes));
+      setProjects(projectsRes);
     });
-  }
-
-  openProjects() {
```
joe-bell, line 93: Don't be alarmed, this wasn't being used/called from anywhere in the codebase!

ricklamers:
This PR is fantastic! It was definitely needed to introduce a more disciplined approach to the global aspects of the Orchest frontend. Even though this is not a full refactor of all global calls, it does lay the foundation to migrate to a context/reducer based approach.

Great job! Let's iron out the kinks and get this merged.

yannickperrenet:
Great work!

I was wondering whether we should write down the use of the `OrchestContext` (your PR description would suffice) for future reference for others. What about the [Implementation details](https://orchest.readthedocs.io/en/latest/developer_guide/implementation_details.html) in the docs?

joe-bell:
@yannickperrenet thank you! Perhaps we could add a new section for Front-end? I feel like `Implementation details` is quite long already

yannickperrenet:
> Perhaps we could add a new section for Front-end? I feel like Implementation details is quite long already

Feel free to structure it any way you like, as long as it is reasonable of course. To me it was more about the idea of writing it down, where exactly I don't have a strong opinion on.

yannickperrenet:
Although I am not afraid to make the implementation details longer still since a toc is automatically generated for users to navigate the implementation details. This has the added benefit of not showing them content they don’t really need.

Seems like I am pro implementation details after all :)

joe-bell:
I think our implementation details would become quite bloated and difficult to digest – it's better off being specific to the product itself imo

Given that I'm going to be adding/maintaining a lot of Front-end-specific documentation that is both product/non-product related, I'll create a separate document

Issue #215, opened 2021-04-20
JupyterLab pip / python usage should show a warning


ricklamers:
First time users of Orchest might expect to be able to continue doing things they did before when it comes to installing Python packages. Therefore something like the following is not unreasonable for a user to do:

- Open JupyterLab
- Try to import a package: `import pandas as pd`
- Open JupyterLab terminal
- Run command: `pip install pandas`
- Try running `import pandas as pd` again in the notebook

In Orchest, this won't work because JupyterLab runs in a separate container from its kernels (through the Jupyter Enterprise Gateway).

A solution could be to warn users when they invoke pip in JupyterLab that packages should be installed through the environments (using the setup_script.sh or custom environments).

```
 __          __              _             
 \ \        / /             (_)            
  \ \  /\  / /_ _ _ __ _ __  _ _ __   __ _ 
   \ \/  \/ / _` | '__| '_ \| | '_ \ / _` |
    \  /\  / (_| | |  | | | | | | | | (_| |
     \/  \/ \__,_|_|  |_| |_|_|_| |_|\__, |
                                      __/ |
                                     |___/ 

# Please use Orchest environments to install pip packages.
# JupyterLab runs in a separate environment so won't affect your Jupyter Notebook kernels.
__________________________________________________________________________________________
```

PR #216, opened 2021-04-21
Catch pip commands to display warning message

yannickperrenet:
## Description

Resolves: #215 

## Issues

It seems like the JupyterLab terminal manipulates the `$PATH` variable, because it always shows:
```bash
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
even though in the Docker container itself it shows the updated path:
```bash
PATH=/usr/local/orchest/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
To make matters even stranger, installing JupyterLab locally and running the terminal the `$PATH` variable is as expected equal to the `$PATH` of the process starting `jupyter lab`.

I have also tried modifying the `start.sh` script in the `jupyter-server` to:
```bash
# This still did not change the $PATH in the terminal
PATH=/usr/local/orchest/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin jupyter lab ...

# This actually prevented JupyterLab from starting as the `jupyter` command
# could no longer be found. This means that the PATH variable is indeed used.
PATH=/usr/local/orchest/bin jupyter lab ...
```

EDIT: It is probably good to note that the `$ORCHEST_VERSION` does get correctly set in the environment (both in the Docker container as well Jupyter terminal).


yannickperrenet:
A quick and dirty fix would be to put our `pip` and `pip3` executables in `/usr/local/sbin`. Then everything just works, it is just less clean.

ricklamers:
I've implemented it in a slightly different way. This includes the warning when pip is invoked differently, like `python -m pip`.

Not the prettiest way to get a warning to the user, but I do think it does its job reliably.

Note, the warning disappears if the user manually upgrades pip in the JupyterLab terminal. Although it does show the warning  _on_ the actual upgrade.

PR #224, opened 2021-05-02
Feature: Services

ricklamers:
## Description

This PR adds the idea of "services" to Orchest. A service can be any container that can run as part of a session, e.g. a streamlit app to be interacted with.

Example service JSON configuration:
```json5
{
  "services": {
    "streamlit": {
      "command": [],
      "env_variables": {
        "BASE_PATH": "$BASE_PATH_PREFIX_80",
        "MY_ENV": "123"
      },
      "env_variables_inherit": [
        "ABC"
      ],
      "image": "httpd:2.4", // or "environment@<uuid>"
      "name": "streamlit",
      "ports": [
        80
      ],
      "preserve_base_path": false,
      "binds": {
        "/project-dir": "/my-project-dir-path-in-image",
        "/data": "/my-data-path-in-image"
      },
      "scope": [
        "interactive",
        "noninteractive"
      ]
    }
  }
}
```
## Todo

- [ ] Cleaning up code
- [ ] Adding section to the docs on services
- [ ] Create a front-end for the services so that the user does not have to edit the raw JSON configuration for services


ricklamers:
BASE_PATH should be able to be referenced in all fields in the services JSON in the pipeline.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,10 +141,28 @@ def launch(
...
+            # host_userdir!
+            userdir_path_component = "/userdir/"
+            relative_project_dir = project_dir[
+                project_dir.find(userdir_path_component) + len(userdir_path_component) :
+            ]
+            with open(
+                os.path.join("/userdir", relative_project_dir, pipeline_path), "r"
+            ) as f:
+                self.pipeline = Pipeline.from_json(json.load(f))
+        except (IOError, Exception) as e:
```
fruttasecca, line 162:
```suggest
        except Exception as e:
```

Not sure if it was meant to be catching all exceptions or `Exception` is a leftover

fruttasecca:
Should the memory and jupyter server(s) be services? It would increase consistency and would make for a nice, self-documenting feature, since "services" in the pipeline settings would already be populated with some entries. I am aware there might be some technical difficulties into making both of them into services, but it's worth reasoning about.

>Handle environment variables in a way consistent with step env vars

Assuming those are to be secrets (non versioned) the naive solution would be to have services inherit the pipeline environment variables, but we run into the risk of collisions, secrets leaking into services that did not need them, etc. The way services are currently declared might be at odds with what we want to do. We could use a slightly higher level input form instead of having the user writing jsons in code mirror, and take care of what should go in the pipeline def json and what should be secret and written to the db behind the scenes.

ricklamers:
Environment variables are now passed, but only when explicitly request:

```
[{
name: "my-service",
image: "httpd:2.4",
environment_inherit: ["ENV_VAR_NAME"]
}]
```

ricklamers:
> Should the memory and jupyter server(s) be services? It would increase consistency and would make for a nice, self-documenting feature, since "services" in the pipeline settings would already be populated with some entries. I am aware there might be some technical difficulties into making both of them into services, but it's worth reasoning about.

They are not optional, and URL routing might be different to services. So I don't think they should be unified into services. As I feel like services should just be user defined services. But they are very similar internally indeed. We should re-use them as much as possible. In way they already share implementation internally (e.g. they're both assigned to `self._containers[resource]`).

ricklamers:
>We could use a slightly higher level input form instead of having the user writing jsons in code mirror

I think this is something we should iterate toward. Not necessarily release v1. But do agree it would be nicer.

Secrets can now be passed using environment variable inheritance.

ricklamers:
> - [x] Handle easy and reliable pointing to network addresses in pipeline code. Differentiate between user address and container address. (Traffic shouldn’t go out of the container network for pipeline code)

This is now handled by the Orchest SDK.

It returns: `{"internal_urls": [...], "external_urls": [...], "ports": [80, 8081]}`. I.e. one endpoint for each port. At the moment we only support TCP port forwarding.

ricklamers:
One pet peeve is the necessity to tweak container services to incorporate a base path. This is due to how the `nginx-proxy` (in this implementation) forwards requests.

The structure is: <origin><base_path><port><application_path>
http://127.0.0.1:8000/service-webserver-uuiduuid-uuiduuid_80_/some-path.txt

Or broken into its components:
origin: http://127.0.0.1:8000
base_path: /service-webserver-uuiduuid-uuiduuid
port: _80_
application_path: /some-path.txt

Would like to do it in a 'cleaner' way. But this works reliably and allows you to flexibily run services on any TCP port in the service container without requiring _any_ changes to the way in which we currently host Orchest on AWS and how it runs locally.

ricklamers:
Another potential issue is that we now pull the containers for services from DockerHub when we start the session. This could take a long time and could make the session start request time out. Should we bite the bullet and make the session start (POST) a polling operation?

fruttasecca:
>Another potential issue is that we now pull the containers for services from DockerHub when we start the session. This could take a long time and could make the session start request time out. Should we bite the bullet and make the session start (POST) a polling operation?

I think so

services/orchest-api/app/tests/input_operations/pipeline.json
```diff
@@ -3,6 +3,7 @@
     "uuid": "pipeline-uuid",
     "settings": {},
     "parameters": {},
+    "services": [],
```
yannickperrenet, line 6: Should not forget to also update the JSON schema in the docs.

ricklamers: Good point!

ricklamers:
Note: we should not access the `userdir/` in the `orchest-api`. This requires a refactor of https://github.com/orchest/orchest/pull/224/files#diff-3c5dc561a3261d8aacd881b4cb16e43c99215b5a7966faf9aadc43734dd6302bR158

orchest-sdk/python/orchest/services.py
```diff
@@ -0,0 +1,99 @@
...
+        {"internal_urls": [], "external_urls": [],
+            ... user specified service fields}
+
+        one for each port specified in the service specification.
+    """
+    pipeline = _get_pipeline()
+
+    for service in pipeline.properties.get("services"):
+        if service["name"] == name:
+            return _generate_urls(service, pipeline)
```
yannickperrenet, line 76: > I am currently going over the services PR in depth so this comment might answer itself later on.

It seems like services are defined as a list and thus allowing multiple services with the same name. This would return the first occurrence only. I think names should be unique and this should be enforced by the data structure in the pipeline definition, i.e. making it a dictionary with nested dictionaries instead of a list of dictionaries.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -125,6 +135,7 @@ def launch(
         project_uuid: str,
         pipeline_path: str,
         project_dir: str,
+        session_type: SessionType,
```
yannickperrenet, line 138: Should be included in docstring as well.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -40,6 +49,7 @@ class Session:
     """
 
     _resources: Optional[list] = None
+    _run_config = None
```
yannickperrenet, line 52: I don't understand why this is a class attribute? Makes more sense to me if it were a initialization argument (i.e. in `__init__`)

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,22 +150,51 @@ def launch(
...
             host_userdir,
+            session_type,
             self.network,
+            self.pipeline,
+            self.pipeline.properties.get("uuid"),
+            self._run_config,
         )
+
+        services = [key for key in container_specs.keys() if key.startswith("service-")]
+        self._resources += services
```
yannickperrenet, line 190: I don't think this has the intended behavior, because `self._resources` is a class attribute and thus modifying it as such would modify it for all sessions.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,22 +150,51 @@ def launch(
...
             project_uuid,
             pipeline_path,
             project_dir,
             host_userdir,
+            session_type,
             self.network,
+            self.pipeline,
+            self.pipeline.properties.get("uuid"),
+            self._run_config,
         )
```
yannickperrenet, line 187: To me all these arguments have become a complete mess. I don't think it makes sense to pass `self.pipeline` and `self.pipeline.properties.get("uuid")` separately if one is included in the other (same goes for all the elements we extract from the `run_config` which we would now be completely passing as well).

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,22 +150,51 @@ def launch(
                 interactive sessions).
             pipeline_path: Path to pipeline file (relative to
                 project_dir).
-            project_dir: Path to project directory.
+            project_dir: Path to project directory. (Absolute host path)
```
yannickperrenet, line 153: Can definitely be formatted better: 
```
project_dir: (Absolute host) Path to project directory.
```

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,22 +150,51 @@ def launch(
...
+
+        # Load pipeline definition
+        try:
+            # Find project_dir from userdir
+            # NOTE: This assumes project_dir is always a child path of
+            # host_userdir!
+            userdir_path_component = "/userdir/"
+            relative_project_dir = project_dir[
+                project_dir.find(userdir_path_component) + len(userdir_path_component) :
+            ]
```
yannickperrenet, line 166:
```suggestion
            relative_project_dir = project_dir.split("/userdir/")[1]
```

services/orchest-api/app/app/core/sessions.py
```diff
@@ -139,22 +150,51 @@ def launch(
...
+            with open(
+                os.path.join("/userdir", relative_project_dir, pipeline_path), "r"
+            ) as f:
+                self.pipeline = Pipeline.from_json(json.load(f))
```
yannickperrenet, line 170: We should not actually be accessing content from the `userdir` here. As a general concept we try not to access the `userdir` directly from the `orchest-api`

services/orchest-api/app/app/core/sessions.py
```diff
@@ -199,7 +239,7 @@ def shutdown(self) -> None:
                 APIError,
                 ContainerError,
             ) as e:
-                current_app.logger.error(
+                logging.error(
```
yannickperrenet, line 242: Why are we no longer using the `current_app.logger`? As far as I remember this is preferred because calling `logging.error` will straight up call the `root` logger.

Same goes for all other `logging` statements in this module.

@ricklamers @fruttasecca What is the current state on the logging? I remember there being issues, but please help me recall those.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -566,13 +608,153 @@ def _get_mounts(
...
+        pipeline_uuid: UUID of pipeline.
+        network: Docker network. This is put directly into the specs, so
+            that the containers are started on the specified network.
+        session_type: Type of session: interactive, or noninteractive,
+
+    Optional:
+        run_uuid: For NonInteractive sessions only.
+            UUID of the pipeline run.
+        run_config: For NonInteractive sessions only. Contains run
+            config information like user_env_variables.
```
yannickperrenet, line 644: The fact that they are optional should be indicated using the `typing.Optional[str]` instead of in the docstring.

services/orchest-api/app/app/core/sessions.py
yannickperrenet: Although in the rare case where someone names their project `userdir` this suggestion does not work. Possibly a similar case can be constructed for the initial implementation where a user clones Orchest in a subdirectory of a user-created `userdir`. Not sure whether the latter actually implies other issues in the codebase.

services/orchest-api/app/app/core/sessions.py
fruttasecca: I think there is still some unknown/undefined behaviour when it comes to logging, as far as I remember it came down to some exceptions or traces not being logged.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -566,13 +608,153 @@ def _get_mounts(
...
+        except Exception as e:
+
+            logging.error("Failed to fetch user_env_variables: %s [%s]" % (e, type(e)))
+
+            traceback.print_exc()
+
+            user_env_variables = {}
+
+        environment = service.get("environment", {})
+
```
fruttasecca, line 705: If we are to allow env variables that are specific to a service we have to store those in the DB, since they are to be treated as secrets.

orchest-sdk/python/orchest/services.py
```diff
@@ -0,0 +1,108 @@
...
+
+from orchest.config import Config
+from orchest.error import ServiceNotFound
+from orchest.pipeline import Pipeline
+
+
+def _get_pipeline() -> Pipeline:
+    with open(Config.PIPELINE_DEFINITION_PATH, "r") as f:
+        pipeline_definition = json.load(f)
+    return Pipeline.from_json(pipeline_definition)
```
yannickperrenet, line 18: This function is actually an exact duplicate of `parameters._get_pipeline()`. Should probably just put this function in the `utils.py` instead.

orchest-sdk/python/orchest/services.py
```diff
@@ -0,0 +1,108 @@
...
+    Raises:
+        ServiceNotFoundError: The service given by name ``name``
+            could not be found.
+
+    """
+    pipeline = _get_pipeline()
+
+    services = pipeline.properties.get("services")
+    if name in services:
+        return _generate_urls(services["name"], pipeline)
```
yannickperrenet, line 87: This should probably be:

```python
return _generate_urls(services[name], pipeline)
```
So without the `""` around `name`

orchest-sdk/python/orchest/services.py
```diff
@@ -0,0 +1,108 @@
...
+
+    raise ServiceNotFound("Could not find service with name %s" % name)
+
+
+def get_services() -> List[Dict[str, List[Any]]]:
```
yannickperrenet, line 92: The return type should be corrected as it is now a dictionary and not a list.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -100,6 +100,11 @@ def post(self):
                 "env_variables": get_proj_pip_env_variables(
                     post_data["project_uuid"], post_data["pipeline_uuid"]
                 ),
+                # NOTE: the definition of a service is currently
+                # persisted to disk and considered to be versioned,
+                # meaning that nothing in there is considered to be
+                # secret. If this changes, this dictionary needs to have
+                # secrets removed.
```
yannickperrenet, line 107: This also holds for the anonymized telemetry pings that we are sending!

services/orchest-api/app/app/core/tasks.py
```diff
@@ -233,14 +234,15 @@ def start_non_interactive_pipeline_run(
...
+    host_base_user_dir = os.path.split(host_userdir)[0]
 
     # To join the paths, the `run_dir` cannot start with `/userdir/...`
     # but should start as `userdir/...`
-    run_config["project_dir"] = os.path.join(host_base_user_dir, run_dir[1:])
-    run_config["run_endpoint"] = f"jobs/{job_uuid}"
     run_config["pipeline_uuid"] = pipeline_uuid
     run_config["project_uuid"] = project_uuid
+    run_config["project_dir"] = os.path.join(host_base_user_dir, run_dir[1:])
+    run_config["run_endpoint"] = f"jobs/{job_uuid}"
```
yannickperrenet, line 245: The comment should have moved with the two files as the comment is applicable to the `"project_dir"` line.

services/orchest-webserver/app/app/analytics.py
```diff
@@ -211,3 +211,29 @@ def check_active(app):
     except OSError as e:
         app.logger.debug("Exception while reading request log recency %s" % e)
         return False
+
+
+def send_session_start(app, session_config):
+    props = {
+        "project_uuid": session_config["project_uuid"],
+        "pipeline_uuid": session_config["pipeline_uuid"],
+        "services_count": len(session_config["services"]),
```
yannickperrenet, line 220: Wouldn't it be interesting to also add the names of the services that are started? To give us an idea of the type of services that are used.

fruttasecca: That could make sense but I wonder if it would be a privacy issue. Both service name or image name could be seen as such.

services/orchest-api/app/app/schema.py
```diff
@@ -69,6 +72,48 @@
...
+        "name": fields.String(required=True, description="Name of the service"),
+        "command": fields.String(required=False, description="Docker command"),
+        "entrypoint": fields.String(required=False, description="Docker entrypoint"),
+        "ports": fields.List(
+            fields.String, required=True, description="List of service exposed ports"
+        ),
+        "scopes": fields.List(
+            fields.String, required=True, description="interactive/noninteractive"
+        ),
+        "env_variables": fields.Raw(
```
yannickperrenet, line 88: This serves mostly as a note, but I found it somewhat confusing that we have `env_variables` and `environment_inherit`. I think they should be closer together, e.g. `env_variables` and `env_variables_inherit`. But this depends on how we want to name it in the UI.

services/orchest-api/app/app/schema.py
```diff
@@ -69,6 +72,48 @@
...
+        "environment_inherit": fields.List(
+            fields.String,
+            required=True,
+            description=(
+                "List of env vars to inherit from project and pipeline env vars"
+            ),
+        ),
+        "project_directory": fields.List(
+            fields.String, required=True, description="List of service exposed ports"
+        ),
```
yannickperrenet, line 104: My guess is that this should be a String and not a List? Currently it is an exact copy of the `"ports"` (except for the name change).

fruttasecca: Indeed, will fix.

services/orchest-api/app/app/schema.py
fruttasecca: Agree with the naming change, will notify @joe-bell @ricklamers once the change is in.

services/session-sidecar/app/main.py
```diff
@@ -0,0 +1,114 @@
...
+    return os.path.join(get_log_dir_path(), f"{service}.log")
+
+
+def create_log_dir():
+    log_dir_path = get_log_dir_path()
+    if not os.path.exists(log_dir_path):
+        try:
+            os.makedirs(log_dir_path)
+        except OSError as exc:
+            raise exc
```
yannickperrenet, line 27: Could make shorter to:
```python
        try:
            os.makedirs(log_dir_path, exist_ok=True)
        except OSError as exc:
            raise exc
```

I think it is also good to be aware of the [note from the Python docs](https://docs.python.org/3/library/os.html#os.makedirs): 
> makedirs() will become confused if the path elements to create include pardir (eg. “..” on UNIX systems).

services/session-sidecar/app/main.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def get_service_name_from_log(log_line):
+    metadata_regex = r"-metadata-end\[\d+\]"
+    name_end = re.search(metadata_regex, log_line).start()
+    prefix = "user-service-"
+    name_start = log_line.find(prefix) + len(prefix)
+    return log_line[name_start:name_end]
```
yannickperrenet, line 35: Not sure how long the `log_line` could be, but you can make this faster by using one regex with groups that requires only one pass. To get the name of the service you can then just access the value of the specific group.

fruttasecca: AFAIK log_line can have arbitrary size (would be different with UDP).  The metadata itself is about 80 characters, so the search itself, both regex and string `search` would stop early regardless of the size of the string. This particular function runs once per service so it's not really a priority when it comes to optimization.

services/session-sidecar/app/main.py
```diff
@@ -0,0 +1,114 @@
...
+            return
+
+        service_name = get_service_name_from_log(bytes.decode(data))
+        logging.info(f"{service_name} sent its first data.")
+
+        with open(get_service_log_file_path(service_name), "w") as log_file:
+            # Used by the log_streamer.py to infer that a new session
+            # has started, i.e. the previous logs can be discarded.
+            log_file.write("%s\n" % str(uuid.uuid4()))
+            log_file.flush()
```
yannickperrenet, line 62: Just so that I understand, you are generating a UUID here to get a random line in the logs?

services/session-sidecar/app/main.py
```diff
@@ -0,0 +1,114 @@
...
+    # Cleanup old service logs on session start, so that if a container
+    # does not emit any output, both by design or because it failed to
+    # start, no old logs will be shown.
+    log_dir_path = get_log_dir_path()
+    for fname in os.listdir(log_dir_path):
+        path = os.path.join(log_dir_path, fname)
+        if (
+            # Pipeline steps logs are of no concern to the sidecar.
+            # NOTE: Should it take care of those as well?
+            not re.search(r"^[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}\.log$", fname)
```
yannickperrenet, line 99: Could you explain the regex? I don't understand what it is trying to match. In particular the `a-f`.

services/session-sidecar/app/main.py
fruttasecca: Yes, I am using the same pattern that we are using for steps logs. The file streamer has this contract to understand that some logs belong to a different session, i.e. different UUID implies different session.

services/session-sidecar/app/main.py
```diff
@@ -0,0 +1,114 @@
...
+                # the only way to know if the ending newline is "real"
+                # or not is to check the length, note that there could
+                # be a false positive if a message with a newline has
+                # exactly this length.
+                if len(data) == 16456 and data[-1] == "\n":
+                    data = data[:-1]
+                # Remove syslog metadata. Note: this is faster than
+                # using a regex but less safe. TODO: decide what to
+                # use.
+                data = data[data.find("]: ") + 3 :]
```
yannickperrenet, line 79: My guess is that speed is fairly important here as all log streaming goes through this code?

services/session-sidecar/app/main.py
fruttasecca: It's trying to match `<uuid4>.log`. Essentially, I wanted to avoid cleaning up step logs on session start (despite it being required, but it already happens in another place, and steps are of no concern to the sidecar).

services/session-sidecar/app/main.py
fruttasecca: Indeed. Some other logging drivers provide a json but parsing the json gets costly as the message gets longer and these more advanced driving logs allow little control on compression (i.e. I could not stop message compression regardless of what settings I was providing past some arbitrary message size), compression is explicitly discouraged in the docker docs to avoid wasting cpu time.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -1,25 +1,43 @@
...
 class IP(NamedTuple):
     jupyter_EG: str
     jupyter_server: str
 
 
+def _inject_message_as_service(ip, port, service, msg):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.connect((ip, port))
+    # TODO: find a way to avoid having to fake being a syslog message.
+    msg = f"user-service-{service}-metadata-end[0000]: {msg}"
```
yannickperrenet, line 36: Could this causes with the regex we are user in case someone would name the `service` to be `user-service` or maybe `user-service-something`? Although the regex matches eagerly, but wanted to quickly check with you guys.

fruttasecca: Given how ` get_service_name_from_log` in the `sidecar` `main.py` module is implemented, a service being named `user-service*` should not be a problem (will test).  It _could_ become a problem if the service was named like `"*-metadata-end\[\d+\]"`, but that is not allowed at the GUI level, because we are enforcing the pattern `/^[0-9a-zA-Z\-]{1,36}$/`, which excludes `]`.

I will enforce said regex at the API level as well.

services/orchest-webserver/app/app/analytics.py
yannickperrenet: I say we include it, because it contains valuable information to us. If users don't like it, then we can always remove it.

@ricklamers Do you want to chip in here?

services/session-sidecar/app/main.py
yannickperrenet: Could you add that to the docstring? I think it makes total sense now that you've said that.

services/session-sidecar/app/main.py
yannickperrenet: I see, didn't recognize it at first.

services/orchest-webserver/app/app/analytics.py
fruttasecca: While we are at it we can discuss how a service in the pipeline definition is anonymized, i.e. what fields we are removing. Currently removing: `command, environment, environment_inherit, entrypoint, binds` (not committed yet).

ricklamers: I think service name is fine to include. The exclusions you mention are sensible to me, although binds is very unlikely to contain sensitive info so I’d rather include it.

fruttasecca: The tradeoff between the risk of ingesting sensible (or user identifying) data and gaining any useful info with raw `binds` does not seem great to me. I will anonymize said binds and you can take a look at it and let me know if you agree.

fruttasecca:
```python
def anonymize_service(service):
    service = copy.deepcopy(service)
    service.pop("command", None)
    service.pop("entrypoint", None)
    service.pop("env_variables", None)
    service.pop("env_variables_inherit", None)

    binds = service.get("binds", {})
    # Maps source to the depth of the target directory.
    for source, target in list(binds.items()):
        binds[source] = len(os.path.normpath(target).split(os.sep))

    return service
```

ricklamers: Let’s drop `binds` then

services/orchest-api/app/app/core/sessions.py
yannickperrenet: I am pretty sure the recommended way is to use `current_app.logger`, but I don't remember the last problem that popped up around logging. In my view we should not be changing the logging behavior in this PR (and so sticking with the `current_app.logger`) and go over it another time. This just makes a mess out of the logging.

I would say we revert all the logging changes made here and stick with using the `current_app.logger`.

orchest-sdk/python/orchest/__init__.py
```diff
@@ -5,4 +5,5 @@
     update_pipeline_param,
     update_step_param,
 )
+from orchest.services import get_service
```
yannickperrenet, line 8: Not sure about you guys, but I think I would often opt for using `get_services` because it allows me to inspect all the services without knowing their name up front. So I think we should add it here.

fruttasecca: Fine with me.

services/orchest-api/app/app/core/sessions.py
yannickperrenet: I reverted it back to `current_app.logger`

yannickperrenet:
A few things that stood out to me when playing with the PR:
* In the pipeline settings configurations tab, shouldn't the "services" be removed in favor of the dedicated services tab?
* The "add service" in the services tab in the pipeline settings doesn't do anything currently.

joe-bell:
> The "add service" in the services tab in the pipeline settings doesn't do anything currently

That's because I'm building it @yannickperrenet 😆

yannickperrenet:
A few comments while reviewing the PR (open to view): 

<details><summary>BUGS</summary>

1. 
image

This warning never goes away. Only when closing and reopening the settings. To reproduce:
1. Add a service
2. Save
3. Get the warning
4. Restart the session
5. Still see the warning
This felt a bit strange, since I did exactly the steps described in the warning yet the warning stayed there.

---
2.

The `code-server` service did not work. To be more specific: the code-server container started, but it did not show any files.

```text
Unable to open 'code-server': Unable to read file 'vscode-remote://service-code-server-e4bae339-0915b350:8080/home/coder/code-server' (Error: Unable to resolve non-existing file 'vscode-remote://service-code-server-e4bae339-0915b350:8080/home/coder/code-server').
```

To reproduce:
1. Through the pipeline settings click "add service"
2. Choose VSCode
3. Restart the session
4. Wait for the code-server to become available
5. Visit the URL
6. Get the error message
</details>

<details><summary>Nitpicks</summary>

* It felt somewhat strange that the "services" button in the pipeline editor only shows me something and I have to go to the settings in order to change the services. Should we add a link "settings" to the floating window that appears?
* Why is the service in the list called "VSCode" but when you look in the table after creation it says "code-server". This is confusing.
image
* Starting the session can take a very long time because it could be that new images are being pulled. Shouldn't we have something that informs the user about this? Currently, the user would just wait for the session to become available and starting to wonder why it is so slow (and possibly expecting it to continue to be slow in the future).
</details>

orchest-sdk/python/orchest/services.py
```diff
@@ -0,0 +1,109 @@
...
+        + "-"
+        + service_uuid.split("-")[0]
+    )
+
+    external_urls = {}
+    base_paths = {}
+    for port in service.get("ports", []):
+        pbp = "pbp-" if service.get("preserve_base_path", False) else ""
+        base_path = f"/{pbp}{container_name}_{port}"
+        external_url = f"http://{{host_name}}:{{port}}{base_path}/"
```
yannickperrenet, line 42: I don't completely understand what the use of this `external_url` is. For example, you get the following:
```text
{'code-server': {'binds': {'/project-dir': '/home/coder/project'},
  'command': 'code-server --auth none',
  'image': 'codercom/code-server:latest',
  'name': 'code-server',
  'ports': [8080],
  'internal_hostname': 'service-code-server-e4bae339-0915b350',
  'external_urls': {8080: 'http://{host_name}:{port}/service-code-server-e4bae339-0915b350_8080/'},
  'base_paths': {8080: '/service-code-server-e4bae339-0915b350_8080'}}}
```
What do I do as a user with this `external_urls`? Do I figure out the `host_name` and `port` myself?

orchest-sdk/python/orchest/config.py
```diff
@@ -16,6 +16,8 @@ class Config:
     PROJECT_UUID = os.getenv("ORCHEST_PROJECT_UUID")
     PIPELINE_UUID = os.getenv("ORCHEST_PIPELINE_UUID", "")
     PIPELINE_PATH = os.getenv("ORCHEST_PIPELINE_PATH")
+    RUN_UUID = os.getenv("ORCHEST_RUN_UUID")
+    RUN_TYPE = os.getenv("ORCHEST_RUN_TYPE")
```
yannickperrenet, line 20: Not sure how big of a deal this is, but these two values would be `None` inside the Jupyter kernels. But my guess is (based on their name) that they don't need to be known in Jupyter kernels?

services/orchest-api/app/app/core/tasks.py
```diff
@@ -165,6 +166,8 @@ def run_pipeline(
     """
     run_config["pipeline_uuid"] = pipeline_definition["uuid"]
     run_config["project_uuid"] = project_uuid
+    if "run_type" not in run_config:
+        run_config["run_type"] = "interactive"
```
yannickperrenet, line 170: This should probably get a comment stating that the `run_type` needs to be injected and is already done so for `"non-interactive"` if applicable.

services/orchest-webserver/client/src/utils/service-templates.jsx
```diff
@@ -0,0 +1,82 @@
...
+  vscode: {
+    label: "VSCode",
+    icon: <IconVSCode />,
+    config: {
+      binds: {
+        "/project-dir": "/home/coder/project",
+      },
+      command: "code-server --auth none",
+      image: "codercom/code-server:latest",
+      name: "code-server",
```
yannickperrenet, line 52: Should this be "VSCode" instead? I felt it to be confusing selecting "VSCode" but getting "code-server"

orchest-sdk/python/orchest/config.py
fruttasecca: Indeed, in Jupyter kernels this is correctly handled.

orchest-sdk/python/orchest/services.py
fruttasecca: >Do I figure out the host_name and port myself?

Yes, if Orchest is deployed we don't know what the hostname is going to be (and possibly port).

yannickperrenet: So how then does the user interact with the service from within the pipeline? For example a redis service.

Do the users need to construct the URL themselves?

fruttasecca:
```python
redis_host = orchest.get_service("redis")["internal_hostname"]
redis_client = redis.Redis(host=redis_host, port=6379, db=0)
```

yannickperrenet: Funny, I made exactly that example as well.

> `internal_hostname` must be used when using services from within a kernel or step.

Makes sense to me. A section in the docs would definitely be helpful to our users though.

fruttasecca: Indeed, the docs must (and will :P) explain such things.

fruttasecca:
About the nitpicks:

>? Currently, the user would just wait for the session to become available and starting to wonder why it is so slow (and possibly expecting it to continue to be slow in the future).

The docs need for sure to mention this, not sure where/if this can be placed in the GUI.

I have a nitpick to add, the red bar stating `Note: changes to services require a session restart to take effect.` is at the top of a service details, but I find myself scrolling down to the point that the bar is not visible anymore

yannickperrenet:
Another nitpick. The term "Mounts" was a little confusing, I think it should instead state "Target mount paths" (since the user does actually not choose the path that is mounted)

image

ricklamers:
> A few comments while reviewing the PR (open to view):
> 
> BUGS
> Nitpicks

The bugs should be fixed. Of the nitpicks I fixed the first 2. The last one I think is acceptable for this v1.

docs/source/user_guide/services.rst
```diff
@@ -0,0 +1,176 @@
...
+
+- **name**: The name of the server.
+- **image**: The image to use for the service.
+
+Non required fields:
+
+- **command**: Service start command, e.g. what process the service will run.
+- **entrypoint**: `command` and `entrypoint` are a 1:1 mapping to Docker, refer to the `Docker docs <https://docs.docker.com/engine/reference/builder/#cmd>`_ for their difference and gotchas.
+- **inherited environment variables**: A list of environment variable names that will be
+  inherited from the project and pipeline environment variables.
```
yannickperrenet, line 112: Does this not include the environment variables set in jobs?

docs/source/user_guide/services.rst
```diff
@@ -0,0 +1,176 @@
...
+- **environment variables**: Key-value pairs of environment variables, take priority over
+  the inherited ones. Note that, while project and pipeline environment variables
+  are considered as `secrets`, services environment variables aren't and will
+  be persisted in the pipeline definition file.
+- **scope**: To specify if the service should be running in interactive mode, jobs, or both.
```
yannickperrenet, line 117: Naming the scope to be "jobs" instead of "non-interactive" makes a lot of sense to me and I am sure it is a lot more obvious for the users as well.

Should we define scope to be "pipeline editor" or "jobs" instead?

fruttasecca: Makes sense to me, @ricklamers @joe-bell ?

docs/source/user_guide/services.rst
fruttasecca: It does, will rephrase.

Issue #226, opened 2021-05-05
Shutting down a session does not abort interactive runs


fruttasecca:
Given an interactive run that is in progress, shutting down a session will not cancel/abort said run, leading to an unexpected state.

Proposed solution: a session shutdown will lead to the abortion of any running interactive run.

PR #227, opened 2021-05-05
Fix abort run on session shutdown

fruttasecca:
## Description
This PR makes it so that an ongoing interactive run will be aborted if the session is shut down or the memory server is restarted through the API, note that, as it stands, the GUI will not allow a user to restart the memory server if an interactive run is ongoing, but the session stop and memory server restart behaviours have been made consistent.

Fixes: #226 

### Checklist


yannickperrenet (approved):
LGTM!

PR #228, opened 2021-05-05
Upgrade Orchest SDK and memory server to pyarrow 4.0.0

fruttasecca:
## Description
This PR upgrades the Orchest SDK and memory server service to `pyarrow` 4.0.0. The Orchest SDK requirements have intentionally been set with `pyarrow>=1.0.0,<=4.0.0` since testing has shown that the SDK would not break if a user was to install `pyarrow` release 1/2/3 when building an environment.
Some unused packages have also been removed from the SDK.

### Checklist

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -52,7 +52,7 @@ exclude_patterns = [""]
 
 # Autodoc configurations.
 # Mock dependencies that are not available at build time.
-autodoc_mock_imports = ["boto3", "pyarrow", "sqlalchemy"]
+autodoc_mock_imports = ["pyarrow"]
 # The first line of the docstring can be considered to be the function's
 # signature (if it looks like one).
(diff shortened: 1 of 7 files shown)


docs/source/conf.py
```diff
@@ -52,22 +53,23 @@
 
 # Autodoc configurations.
 # Mock dependencies that are not available at build time.
-autodoc_mock_imports = ["boto3", "pyarrow", "sqlalchemy"]
+autodoc_mock_imports = ["pyarrow"]
```
yannickperrenet, line 56: Good catch, nice! For your reference this was in there for the old data sources implementation. Without mocking the dependencies the autogenerated docs would not build.

yannickperrenet (approved):
Looks like short and easy PR, but there was quite some hidden work and reasoning behind the scenes ;)

:+1:

Issue #213, opened 2021-04-19
Clean up old Orchest images after update


ricklamers:
Currently the updater does not remove the old images (neither command line, nor web updater). 

We should keep in mind that we would prefer to support letting Orchest run during updating. So ideally it would remove dangling images only on e.g. start such that it can clean images that are not in use anymore.

yannickperrenet:
The `update` is split from `restart` because a `restart` would clean up images and thereby also shutting down the Orchest application before starting it. When invoking an `update` it is up to the user whether they also want to `restart`, to help the user decide we actually advice them (through an echo to the terminal) to `restart` for the changes to take effect.

The web updater actually runs a restart for the user and so old images are cleaned up.

What is happening in the `update` however is that environment images are cleaned up as otherwise an old version of the SDK could be used when running Orchest. So it might be the case that, if a pipeline is running (and thus uses environment images) that then the environment images cannot be cleaned/removed and thus resulting in an error (which then results in the `restart` command in the web updater to never be triggered, although unlikely due to the extensive try-except block structure).

So this issue is somewhat connected to #197.

I will look into it the moment my internet improves as currently running the updater would take too long.

ricklamers:
I don’t think Docker cleans up the old images (those running during update) that are now no longer tagged because the update replaced it during pull. Or am I missing something?

PR #236, opened 2021-05-10
Improv remove orchest dangling images on update and start

fruttasecca:
## Description

Fixes issue #213

### Checklist

### DIFF
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -84,8 +84,6 @@ class OrchestApp:
                 on start.
 
         """
-        self.resource_manager.remove_orchest_dangling_imgs()
-
         # Check whether the minimal set of images is present for Orchest
         # to be started.
         pulled_images = self.resource_manager.get_images()
--- a/services/orchest-ctl/app/utils.py
+++ b/services/orchest-ctl/app/utils.py
@@ -117,4 +117,4 @@ def update_git_repo():
 def is_dangling(image: Mapping) -> bool:
     """Checks whether the given image is to be considered dangling."""
     tags = image["RepoTags"]
-    return not tags or (len(tags) == 1 and tags[0] == "<none>:<none>")
+    return not tags or "<none>:<none>" in tags


services/orchest-ctl/app/utils.py
```diff
@@ -111,3 +111,8 @@ def update_git_repo():
     exit_code = script_process.wait()
 
     return exit_code
+
+
+def is_dangling(image):
+    tags = image["RepoTags"]
+    return not tags or (len(tags) == 1 and tags[0] == "<none>:<none>")
```
yannickperrenet, line 120: Just to make sure. If an image is dangling, will it always have only one tag? Or should it instead be `len(tags) > 0` (as I figured you probably put it in as a check for `tags[0]`)?

services/orchest-ctl/app/docker_wrapper.py
```diff
@@ -474,6 +515,13 @@ def remove_jupyter_build_imgs(self):
         jupyter_build_imgs = self.get_jupyter_build_imgs()
         self.docker_client.remove_images(jupyter_build_imgs, force=True)
 
+    def remove_orchest_dangling_imgs(self):
+        orchest_dangling_imgs = set(self.get_orchest_dangling_imgs())
+        containers, _ = self.get_containers(full_info=True)
+        images_in_use = {c["ImageID"] for c in containers}
+        orchest_dangling_imgs = orchest_dangling_imgs - images_in_use
+        self.docker_client.remove_images(orchest_dangling_imgs, force=True)
```
yannickperrenet, line 523: I feel like it should be added to the docstring that only the dangling images that are not in use are being removed.

services/orchest-ctl/app/utils.py
fruttasecca: It is partially a check for `tags[0]` being there, but I think it should not be `len(tags) > 0` for the fact that I have never seen such an example where tags have more than one element with `<none>:<none>` being one of them, so I went for the safe route to avoid considering something I don't know about as dangling.

services/orchest-ctl/app/docker_wrapper.py
fruttasecca: Will do.

services/orchest-ctl/app/orchest.py
```diff
@@ -84,6 +84,8 @@ def start(self, container_config: dict):
                 on start.
 
         """
+        self.resource_manager.remove_orchest_dangling_imgs()
+
```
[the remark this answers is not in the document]
fruttasecca: The reason for it to be on start is that in an upcoming branch/PR we have made it so that some containers will keep running on update, so we won't be able to fully clean up old images there, the cleanup of those images would then lag by 1 update cycle/call.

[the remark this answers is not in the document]
fruttasecca: About the overhead, I guess you had a lot of dangling images because I am getting pretty much no overhead at all. Let me know if you would still rather not have it on start after my previous comment.

services/orchest-ctl/app/utils.py
yannickperrenet: Can you make sure? Otherwise we can just make it into: 
```python
return not tags or "<none>:<none>" in tags
```

Seems like it is possible to add multiple tags: [stackoverflow](https://stackoverflow.com/questions/21928780/create-multiple-tag-docker-image)

services/orchest-ctl/app/orchest.py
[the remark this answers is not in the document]
yannickperrenet: I don't see a problem with lagging behind 1 update cycle, but I do see a problem with increasing the `start` command time as this is the primary access point of a user.

Plus to me it makes sense running cleanup code on an update, but on a start not so much.

PR #246, opened 2021-05-13
Stop Orchest on CLI update

fruttasecca:
This PR makes it so that `./orchest update` will lead to a full stop of Orchest if Orchest is running when not called with `--mode=web`.

### Checklist

### DIFF
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -360,7 +360,8 @@ class OrchestApp:
         if mode == "web":
             utils.echo("Update completed.")
         else:
-            utils.echo("Update completed. Run ./orchest start to start Orchest.")
+            utils.echo("Update completed. To start Orchest again, run:")
+            utils.echo("\torchest start")
 
     def version(self, ext=False):
         """Returns the version of Orchest.
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -296,6 +296,7 @@ class OrchestApp:
         utils.echo("Updating...")
 
         _, running_containers = self.resource_manager.get_containers(state="running")
+
         if utils.is_orchest_running(running_containers):
             utils.echo(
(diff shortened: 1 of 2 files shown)


services/orchest-ctl/app/orchest.py
```diff
@@ -353,19 +357,7 @@ def update(self, mode=None, dev: bool = False):
...
-            utils.echo("Update completed.")
-        else:
-            # Let the user know they need to restart the application
-            # for the changes to take effect. NOTE: otherwise Orchest
-            # might also be running a mix of containers on different
-            # versions.
-            utils.echo(
-                "Don't forget to restart Orchest for the changes to take effect:"
-            )
-            utils.echo("\torchest restart")
```
yannickperrenet, line 368: I think we should show the user a message (if `mode != "web"`) that Orchest is in a stopped state and that the user can do a `orchest start` to start it again. Otherwise there might be confusion where a running Orchest application will be stopped after update.

services/orchest-ctl/app/orchest.py
```diff
@@ -357,7 +357,10 @@ def update(self, mode=None, dev: bool = False):
         # Delete Orchest dangling images.
         self.resource_manager.remove_orchest_dangling_imgs()
 
-        utils.echo("Update completed.")
+        if mode == "web":
+            utils.echo("Update completed.")
+        else:
+            utils.echo("Update completed. Run ./orchest start to start Orchest.")
```
yannickperrenet, line 363: Can you make it './orchest start' instead? So it is clear that that is the part the user can run?

yannickperrenet: Or some other indication ;)

yannickperrenet: I like the:
```
To start Orchest again, run:
    orchest start
```
Because the additional tab makes it clear that it is a command that can be run.

fruttasecca: alright

PR #245, opened 2021-05-12
Upgrade pnpm and add sessionsKillAll

joe-bell:
## Description

- bumps pnpm to a version that fixes common Vite bugs
- Fixes the Jupyter configure view behaviour

### Checklist

- [ ] The documentation reflects the changes.
- [ ] I have manually tested the application to make sure the changes don’t cause any downstream issues, which includes making sure `./orchest status --ext` is not reporting failures when Orchest is running.

- [ ] In case I changed one of the services’ `models.py` I have performed the appropriate database migrations.

PR #252, opened 2021-05-14
Enhance Sessions Behaviour

joe-bell:
### Stopping (Client, Backend)

Adds client and backend support of stopping `LAUNCHING` sessions as well as `RUNNING` sessions
 
### Fetching (Client)

Previously we fetched multiple sessions in `OrchestContext` using `Promise.all` (due to Flask's limitation of being unable to fetch sessions in parallel)

Little did I know until I worked on #245 that we actually already have an endpoint to handle fetching all sessions… 
`/catch/api-proxy/api/sessions/`

This PR removes all the custom SWR cache key/Promise.all() logic in favour of the single endpoint; which is more efficient, cleaner and perhaps easier to understand for those who need to make amends in the future.

Session-dependent views **must** be wrapped in `SessionConsumer` which will take care of only polling the endpoint where it's needed (we can do this hooks, but this is a caveat of having a codebase that's mostly class-based)

### Checklist


joe-bell:
⚠️ Not to be merged until @fruttasecca has integrated the backend support for this

PR #253, opened 2021-05-17
Refactor: Promise.all() session fetching → api/sessions/

joe-bell:
⚠️ Not to be merged before #252

---

## Description

Previously we fetched multiple sessions in `OrchestContext` using `Promise.all` (due to Flask's limitation of being unable to fetch sessions in parallel)

Little did I know until I worked on #245 that we actually already have an endpoint to handle fetching all sessions… 
`/catch/api-proxy/api/sessions/`

This PR removes all the custom SWR cache key/Promise.all() logic in favour of the single endpoint; which is more efficient, cleaner and perhaps easier to understand for those who need to make amends in the future

### Checklist


joe-bell:
To be included in #252

PR #254, opened 2021-05-17
Fix user configuration migration on JupyterLab upgrade

yannickperrenet:
### Description
On JupyterLab upgrade its old build files will be removed, whilst trying to maintain the user configurations.

### How to test
I tested by swapping between commit `853809f2752c31f58528cc3a860bab9d5ca59003` (a commit before the JupyterLab upgrade) and rebuilding Orchest each time (which is quick due to Docker's caching system).

My flow:
* `sudo rm -rf lab && git checkout lab && ll lab` in `userdir/.orchest/user-configurations/jupyterlab`
* `git checkout 853809f2752c31f58528cc3a860bab9d5ca59003`
* `scripts/build_container.sh`
* Rebuilding the user configured JupyterLab (in which I put `pip show jupyterlab` to inspect that it is running on an older/newer version).
* This now populates the `lab` directory.
* Stop Orchest and rebuild master (after a `git checkout master`)
* Rebuild user configured JupyterLab
* Start a session (to trigger the changes in `start.sh`)
* See that the directory structure of the `lab` directory is updated.

### Checklist

### DIFF
--- a/services/jupyter-server/start.sh
+++ b/services/jupyter-server/start.sh
@@ -26,27 +26,27 @@ build_path=/jupyterlab-orchest-build
 userdir_path=/usr/local/share/jupyter/lab
 
 ###
-# The lockfile is used to make sure
+# The lockdir is used to make sure
 # that this JupyterLab start script
 # is only executed for one instance
 # at a time.
 ###
-lockfile=$userdir_path/.bootlock
+lockdir=$userdir_path/.bootlock
 
-await_lock() {
-    until [ ! -f $lockfile ]
+acquire_lock() {
+    while :
     do
+        if mkdir $lockdir 2>/dev/null; then
+            break
+        fi
+
         echo "Awaiting boot lock..."
         sleep 1
     done
 }
 
-write_lock() {
-    touch $lockfile
-}
-
 release_lock() {
-    rm -f $lockfile
+    rm -rf $lockdir
 }
 
 start_jupyterlab(){
@@ -54,8 +54,7 @@ start_jupyterlab(){
     jupyter lab --LabApp.app_dir="$userdir_path" "$@" --collaborative
 }
 
-await_lock
-write_lock
+acquire_lock
 
 pre_installed_extensions=("orchest-integration" "visual-tags" "nbdime-jupyterlab")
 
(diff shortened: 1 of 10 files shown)


ricklamers:
LGTM, will test tomorrow before merging

yannickperrenet:
When it comes to the lock file, I do see how it can solve issues where multiple jupyter-servers are invoking the `start.sh` but how does it work when multiple JupyterLabs are operating on the same set of files (since the same files are mounted in each container)? Although one could argue that JupyterLab is somewhat solving this case already given that locally you can run multiple JupyerLabs (that are then using the same set of config files -- unless they do some magic there).

services/jupyter-server/start.sh
```diff
@@ -19,6 +19,44 @@
...
+}
+
+start_jupyterlab(){
+    release_lock
+    jupyter lab --LabApp.app_dir="$userdir_path" "$@" --collaborative
+}
+
+await_lock
+write_lock
+
```
fruttasecca, line 59: I am afraid this would still lead to a race condition. Without introducing other dependencies we could use `mkdir`, since that is atomic and we can use the exit code to detect if the process was able to acquire the lock (create the directory). If we want to use more. Otherwise `flock` could be a nice solution as well.

ricklamers:
> Although one could argue that JupyterLab is somewhat solving this case already given that locally you can run multiple JupyerLabs (that are then using the same set of config files -- unless they do some magic there).

I think we don't really have great options other than assuming this.

services/orchest-webserver/app/app/__init__.py
```diff
@@ -123,6 +123,22 @@ def create_app():
                         % (proj.uuid, e, type(e))
                     )
 
+        # To avoid multiple removals in case of a flask --reload, so
+        # that this code runs once per container.
+        try:
+            os.mkdir("/tmp/jupyter_lock_removed")
```
yannickperrenet, line 129: So this is essentially a second lock?

fruttasecca: Yeah this is consistent with what we did in the `orchest-api` to avoid a flask reload leading to undefined behaviours, see the [init ](https://github.com/orchest/orchest/blob/d1d8abea71ddabffb3159ed3fbda51379f3a34eb/services/orchest-api/app/app/__init__.py#L102)file.

yannickperrenet: Just wanted to point it out. You both make valid points ;)

PR #261, opened 2021-05-21
Fix: "Text is hard to read in the 'environments' tab"

joe-bell:
## Description

- Slightly increases spacing between form fields (I tried making it larger but it was just too much)
- Uses Stitches-driven vars to help migrate styles slowly over to Stitches (and at least align our use-case styles in the meantime)

It's important to know that this is really just something to improve short-term, long-term as we migrate to Stitches we can look into improving our overall forms design and structure

|Before|After|
|--|--|
| Screenshot 2021-05-21 at 14 05 35 | Screenshot 2021-05-21 at 14 03 05 | 

| Diff |
| --|
| image |

### Checklist

### DIFF
--- a/services/orchest-webserver/client/src/styles/main.scss
+++ b/services/orchest-webserver/client/src/styles/main.scss
@@ -30,7 +30,6 @@
 @import "ansi";
 @import "themes/jupyter";
 
-$generalPadding: 20px;
 $mdcComponentPadding: 16px;
 $codeSize: 14px;
 $defaultFontSize: 15px;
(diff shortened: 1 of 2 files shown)


services/orchest-webserver/client/src/styles/main.scss
```diff
@@ -30,7 +30,6 @@
 @import "ansi";
 @import "themes/jupyter";
 
-$generalPadding: 20px;
```
joe-bell, line 33: Note: This has been replaced by Stitches-driven CSS variables from our design system

PR #267, opened 2021-05-24
Fix "Double Click" Behaviour

joe-bell:
## Description

When in "read only" mode, double-clicking on a pipeline step will now trigger the "view file" action rather than erroneously trying to open Jupyter.

### Checklist


ricklamers:
LGTM!

Issue #273, opened 2021-05-27
Scheduled jobs are scheduled to run at the wrong time


yannickperrenet:
**Describe the bug**  
Scheduled jobs are scheduled at the wrong time, to be exact (in my case) they are scheduled to run 2 hours later than defined -- I am in UTC+2.

Could be applicable to cron jobs as well. Did not test this.

**To Reproduce**  

1. Create a new "Scheduled" job
2. Note the time at which you schedule the job
3. Run the job
4. Open the job

Note: the "Scheduled to run" time is off by the number of hours the user's timezone is different from UTC. In my case, I scheduled the job to run at `5:35 PM` and inspecting the job shows "Scheduled to run" to be `7:35 PM`. (I am in UTC+2.)

**Expected behavior**  
The job should run at the defined scheduled time.

yannickperrenet:
I think [this code](https://github.com/orchest/orchest/blob/master/services/orchest-webserver/client/src/components/DateTimeInput.jsx#L26) causes the issue (note the `+ "Z"`):
```js
  getISOString() {
    return parseISO(
      this.state.dateValue + " " + this.state.timeValue + "Z"
    ).toISOString();
  }
```

The backend expects to be receiving UTC datetime strings, but the date entered by the user is actually not adjusted to the timezone of the user.

@ricklamers

yannickperrenet:
This happens for cron jobs as well.

To reproduce:
1. Create a cron job with schedule `* 17 * * *`
2. Run the job
3. Open the job again through the job list view

Note that it states: "Scheduled to run" `May 31, 2021 7:00 PM`. Which also has this timezone issue.

yannickperrenet:
For cron jobs, the current implementation might actually be the correct way. Consider the following:

* The cron string does not contain any timezone information. Should the cron string be modified based on the timezone of the user's client, e.g. user X is in UTC and schedules at `* 17 * * *` and user Y is in UTC+2; should user Y see `* 17 * * *` or `* 19 * * *`?
* Team members might be in different timezones, how do they communicate the schedule of a job? Do they all refer to their local timezone or to the UTC standard in the team?

ricklamers:
I think we decided the cron strings would be entered in UTC.

The JS code you linked seems correct to me. The dates entered by the user (scheduled to run at or cron jobs) are in UTC.

So I think the implementation should:
- Show all dates in user local time
- Get all server dates in UTC
- Enter all dates in UTC

fruttasecca:
[...]

PR #282, opened 2021-05-31
Fix scheduled jobs timezone mismatch

yannickperrenet:
## Description

Fixes: #273

### DIFF
--- a/services/orchest-webserver/client/src/views/JobView.jsx
+++ b/services/orchest-webserver/client/src/views/JobView.jsx
@@ -479,7 +479,7 @@ const JobView = (props) => {
                     : state.job.schedule}
                   {state.job.schedule !== null && (
                     <Text as="em" css={{ lineHeight: "normal" }}>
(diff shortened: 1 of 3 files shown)


services/orchest-webserver/client/src/views/JobView.jsx
```diff
@@ -479,7 +479,7 @@ const JobView = (props) => {
                     : state.job.schedule}
                   {state.job.schedule !== null && (
                     <Text as="em" css={{ lineHeight: "normal" }}>
-                      {cronstrue.toString(state.job.schedule)}
+                      {cronstrue.toString(state.job.schedule) + " (UTC)"}
```
yannickperrenet, line 482: I added this bit, because I felt that even though a user sees a help message when entering the cron string, the same user will probably have forgotten about it when inspecting the job later (or another user who didn't schedule it in the first place).

ricklamers (approved):

PR #283, opened 2021-06-01
Add watch scripts to lib packages

joe-bell:
## Description

Adds `esbuild --watch` flags to `lib` packages to enable hot-reloading

(in the same manner that [`design-system` watches](https://github.com/orchest/orchest/blob/master/lib/design-system/package/package.json#L21) for changes)

### Checklist


joe-bell:
cc @ricklamers

Issue #259, opened 2021-05-21
Migrate React Class Components to Functional Components


joe-bell:
A lot of the workarounds we're doing with SWR and global state are a consequence of using class components.

We can't use hooks or consume more than one context at a time. 

To allows us to move forwards without workarounds, we should consider **incrementally** (not one big refactor) migrating all our existing class components (that depend on external state) to functional components

In the meantime, any new components should be written as Functional Components

yannickperrenet:
@ricklamers This can be closed right?

ricklamers:
Yes 👍

PR #286, opened 2021-06-01
Feature: logs view

ricklamers:
## Description

This PR adds a full screen log viewer to the pipeline view. It depends on https://github.com/orchest/orchest/pull/224 and should be merged after that lands on master.

The log view allows you to inspect the logs of any pipeline step and service.

### Checklist


yannickperrenet:
>  It depends on #224 and should be merged after that lands on master.

We could simply merge this branch in the `feature-services` branch first and then merge `feature-services` into `master`.

services/orchest-webserver/client/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,178 @@
+import React from "react";
+import { RefManager, uuidv4 } from "@orchest/lib-utils";
+import { XTerm } from "xterm-for-react";
+import { FitAddon } from "xterm-addon-fit";
+
+class LogViewer extends React.Component {
```
joe-bell, line 6: All new components in `orchest-webserver` should be in the Functional Component syntax please 🙏 
(See #266)

services/orchest-webserver/client/src/views/LogsView.jsx
```diff
@@ -0,0 +1,205 @@
...
+import io from "socket.io-client";
+import {
+  MDCButtonReact,
+  MDCLinearProgressReact,
+  MDCDrawerReact,
+} from "@orchest/lib-mdc";
+import { OrchestContext, OrchestSessionsConsumer } from "@/hooks/orchest";
+import { getPipelineJSONEndpoint } from "../utils/webserver-utils";
+
+class LogsView extends React.Component {
```
joe-bell, line 19: Similarly to above, please convert this to a Functional Component

joe-bell (changes requested):
Generally happy with the new front-end changes!

My only request is that we make sure any new components/views within `orchest-webserver` are written as typed Functional Components (#259) with `// @ts-check`, otherwise we’re just needlessly adding to the pile of components that need to be converted. This'll enable us to make crucial improvements to context/shared state in the future

There are plenty of existing examples in the codebase that should help, but I’m also happy to pair on it if you get stuck

services/orchest-webserver/client/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,178 @@
+import React from "react";
```
joe-bell, line 1: Don't forget type checks!

```suggestion
// @ts-check
import React from "react";
```

services/orchest-webserver/client/src/views/LogsView.jsx
```diff
@@ -0,0 +1,205 @@
+import React from "react";
```
joe-bell, line 1:
```suggestion
// @ts-check
import React from "react";
```

services/orchest-webserver/client/src/components/LogViewer.jsx
ricklamers: This is actually a component rename, so I'll leave this a class component for now. It's relatively complex with the socket.io connection handling, so I'd rather do that refactor when we have more time.

services/orchest-webserver/client/src/views/LogsView.jsx
ricklamers: Good idea! Refactored this to a functional component.

services/orchest-webserver/client/src/components/LogViewer.jsx
ricklamers: Good catch.

services/orchest-webserver/client/src/views/LogsView.jsx
ricklamers: Good catch!

ricklamers:
Moving this PR to a new PR that merges into feature-services.

ricklamers:
> Now that logs have their own button in the pipeline editor, should they be removed from the properties pane?

I think they can stay there, the log viewing component is being re-used. Checking the logs there is less invasive (pane instead full view), so I like it as an addition.

> When opening the logs view I find it hard to navigate to the step I want to see the logs for. What about topological sorting? Would love to hear other ideas ;)

Like the idea! Implemented a topological sort that made sense to me.

PR #289, opened 2021-06-02
Feature: logs view

ricklamers:
## Description

This PR adds a full screen log viewer to the pipeline view. It depends on https://github.com/orchest/orchest/pull/224 and should be merged after that lands on master.

The log view allows you to inspect the logs of any pipeline step and service.

### Checklist


services/orchest-webserver/client/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,139 @@
+// @ts-check
+import React from "react";
+import { RefManager, uuidv4 } from "@orchest/lib-utils";
+import { XTerm } from "xterm-for-react";
+import { FitAddon } from "xterm-addon-fit";
+
+class LogViewer extends React.Component {
```
joe-bell, line 7: Don't forget to make this one a Functional Component too please 🙏

ricklamers: See my response https://github.com/orchest/orchest/pull/286#discussion_r643840816

joe-bell: Fair enough, thanks for clarifying

joe-bell (approved):

joe-bell:
Approved on the front-end side, but I think you should get a review from someone else before merging 👀

yannickperrenet:
I just noticed the following in the pipeline settings view (don't think it was added in this PR though):
image
I don't really like the way it looks at the left side, you can clearly see it is a block overlay.

services/orchest-webserver/client/src/views/LogsView.jsx
```diff
@@ -0,0 +1,262 @@
...
+  const topologicalSort = (pipelineSteps) => {
+    let sortedStepKeys = [];
+
+    pipelineSteps = createOutgoingConnections(pipelineSteps);
+
+    // Add self and children (breadth first)
+    let addSelfAndChildren = (step) => {
+      if (sortedStepKeys.indexOf(step.uuid) == -1) {
+        sortedStepKeys.push(step.uuid);
+      }
```
yannickperrenet, line 68: Not sure you important optimization is here, but I noticed:

This check can be removed, because a step will always be added to `sortedStepKeys` as a childStep. Except for the case of the root steps, so in this edge case these can be added to the `sortedStepKyes` upfront.

services/orchest-webserver/client/src/views/LogsView.jsx
```diff
@@ -0,0 +1,262 @@
...
+    });
+  };
+
+  const disconnectSocketIO = () => {
+    if (sio) {
+      sio.disconnect();
+    }
+  };
+
+  const topologicalSort = (pipelineSteps) => {
```
yannickperrenet, line 59: In the strict sense of a [Topological sort](https://en.wikipedia.org/wiki/Topological_sorting), this function is incorrect.

Take the example: A -> B and A -> C and B -> C then the topological sort would be A, B, C. However with the current implementation A, C, B is also possible. Note: when trying this out and creating this graph, the sorting of the pipeline.json using the UUIDs could actually end up with A,B,C.

yannickperrenet: Could also choose to rename the function to `breathFirstSort`.

Issue #247, opened 2021-05-14
Add warning when creating a Job for a project directory with a large size


ricklamers:
Inform the user about the concept of `/data` when the project directory size exceeds a threshold, like 50 MB.

This avoids users creating Jobs that take forever to get created.

PR #292, opened 2021-06-03
Add warning when snapshot size exceeds 50MB

yannickperrenet:
## Description

Closes: #247

### Implementation details

The warning is added to the modal when creating a job (but only visible if the snapshot size would exceed the 50MB threshold):
image

Note: This is the same style of warning when creating an invalid cron string.

### Todo

- [x] Fixing the styling
  - [x] The icon is misaligned (due to [this code](https://github.com/orchest/orchest/blob/master/services/orchest-webserver/client/src/styles/main.scss#L912) which is done to fix the alignment for the cron string warning.
  - [x] The hyperlink to the docs should probably be prettier.

### DIFF
--- a/services/orchest-webserver/app/app/utils.py
+++ b/services/orchest-webserver/app/app/utils.py
@@ -93,6 +93,29 @@ def get_project_directory(project_uuid, host_path=False):
     return os.path.join(USER_DIR, "projects", project_uuid_to_path(project_uuid))
 
 
+def get_project_snapshot_size(project_uuid, host_path=False):
+    """Returns the snapshot size for a project in MB."""
+
+    def get_size(path, skip_dir):
+        size = 0
+        for root, dirs, files in os.walk(path):
+            size += sum(os.path.getsize(os.path.join(root, name)) for name in files)
+            if skip_dir in dirs:
+                dirs.remove(skip_dir)
+
+        return size
+
+    project_dir = get_project_directory(project_uuid, host_path=host_path)
+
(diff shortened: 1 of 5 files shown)


ricklamers:
Neat! Can you make the link in the warning white for readability?

ricklamers:
I like how it's up front!

yannickperrenet:
@joe-bell Can you make the two style changes, i.e. the hyperlink and alignment of the icon?

services/orchest-webserver/app/app/utils.py
```diff
@@ -93,6 +93,29 @@ def get_project_directory(project_uuid, host_path=False):
...
+            for skip_dir in skip_dirs:
+                if skip_dir in dirs:
+                    dirs.remove(skip_dir)
+
+        return size
+
+    project_dir = get_project_directory(project_uuid, host_path=host_path)
+
+    # This does not count towards size for snapshots.
+    skip_dirs = [".orchest", ".git"]
```
yannickperrenet, line 113: For speed I ignored the `.git` directory to be searched, but I just came across this post on stackoverflow [5GB .git directory](https://stackoverflow.com/questions/5613345/how-to-shrink-the-git-folder/8483112). And so the size approximation would be far off. 

I will change the code to longer ignore `.git` and put a comment to keep evaluating whether it needs to be ignored for optimization purposes.

yannickperrenet:
> @joe-bell Can you make the two style changes, i.e. the hyperlink and alignment of the icon?

No longer needed ;)

Issue #164, opened 2021-03-09
Publish the `orchest-sdk` to PyPI on release


ricklamers:
This will make it easier for people to use older versions of the `orchest-sdk` if necessary.

PR #300, opened 2021-06-08
Publish Python SDK on change

yannickperrenet:
## Description

Closes: #164 

Note: The SDK is published when a PR changes its code instead of on release. After all, having the SDK on PyPi is so that users can install the `orchest-sdk` for their own custom images. This then makes them responsible for having an SDK version that is compatible with the version of Orchest.

### Todo

- [ ] Add a table to the SDK (Python) `README.md` that indicates compatibility between SDK versions and Orchest versions.


yannickperrenet:
@ricklamers @fruttasecca What do you think about the flow of publishing the SDK on commit on master vs on release?

On another note, we need to test whether it indeed only publishes if the SDK files are changed (on commit on master).

yannickperrenet:
> On another note, we need to test whether it indeed only publishes if the SDK files are changed (on commit on master).

Can confirm. Tested in another repo.

fruttasecca:
Agree with the publish on a commit on master approach. 

>This then makes them responsible for having an SDK version that is compatible with the version of Orchest.

This is not super nice but it's somewhat unavoidable given that there will be different versions on`pypi`. Would it make sense to have the SDK communicate to the user that there is a mismatch with the Orchest version? Right now there is no real way to match SDK and Orchest versions unless you dive into the repo, right?

yannickperrenet:
> Would it make sense to have the SDK communicate to the user that there is a mismatch with the Orchest version?

Not a bad idea! We can just add some code in the SDK that runs on import (so in the `__init__.py`) that checks the `ORCHEST_VERSION` environment variable.

fruttasecca:
> > Would it make sense to have the SDK communicate to the user that there is a mismatch with the Orchest version?
> 
> Not a bad idea! We can just add some code in the SDK that runs on import (so in the `__init__.py`) that checks the `ORCHEST_VERSION` environment variable.

It might also make sense to move the table from the readme to the docs.

yannickperrenet:
> It might also make sense to move the table from the readme to the docs.

I'd rather not add (or move) the table to the docs, because the purpose of the docs is to inform users about Orchest and the SDK is pre-installed in Orchest. Therefore, adding a table would only cause confusion to the majority of users, yet for the users who need to know the information should easily be able to find it (on PyPi it shows you the `README` when you visit the [project](https://pypi.org/project/orchest/)).

fruttasecca:
>on PyPi it shows you the `README` 

All good then.

PR #305, opened 2021-06-09
Use a different logger when running as celery

fruttasecca:
## Description
Fixes an issue where code shared by the `orchest-api` and `celery-worker` would fail because it contained a resource that was not available to the `celery-worker`, the app logger.

### Checklist

### DIFF
--- a/services/orchest-api/app/app/core/sessions.py
+++ b/services/orchest-api/app/app/core/sessions.py
@@ -237,6 +237,7 @@ class Session:
 
 
         """
+        logger = utils.get_logger()
 
         # TODO: make convert this "pipeline" uuid into a "session" uuid.
         orchest_services = _get_orchest_services_specs(
@@ -251,9 +252,7 @@ class Session:
                 container = self.client.containers.run(**orchest_services[resource])
                 self._containers[resource] = container
             except Exception as e:
-                current_app.logger.error(
-                    "Failed to start container %s [%s]." % (e, type(e))
-                )
+                logger.error("Failed to start container %s [%s]." % (e, type(e)))
                 raise errors.SessionContainerError(
                     "Could not start required containers."
                 )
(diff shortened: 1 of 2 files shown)


ricklamers (approved):

ricklamers:
LGTM!

yannickperrenet (changes requested):
I would like a short discussion wrt logging before merging this.

fruttasecca:
To bring some more context:
- `logging.error` is used by services that do not run `flask` or `celery`
- `current_app.logger` is used by services running flask
- `utils.get_logger` is used by code that is shared between a service running `flask` and a service running `celery`

What we could do is move `get_logger` in the internal library but that would require the internal library adding `celery` & `flask` to its requirements, or a clever hack that, in a cascading manner, tries to import `flask` and `celery` returning the first logger it is able to get, with fallback to the `logging` module.

yannickperrenet:
> What we could do is move get_logger in the internal library but that would require the internal library adding celery & flask to its requirements, or a clever hack that, in a cascading manner, tries to import flask and celery returning the first logger it is able to get, with fallback to the logging module.

I prefer taking this approach, given that we can make it a "smart" implementation. This way we encapsulate all logging calls so that developers do not have to be aware of your aforementioned rules.

Maybe something along the lines of
```python
try:
    import celery
except ModuleNotFoundError:
    ....
```

fruttasecca:
Yeah that would be the approach to go for the implementation. Happy to take care of this but it should be in a different PR imo.

ricklamers:
Agree it should be a different PR. Not exactly sure if such a logger get utility is preferable as a util. The internal function is then tied to the implementation details of its consumers (i.e. it needs to know about Celery and Flask). Sounds like a code smell to me.

yannickperrenet:
> Happy to take care of this but it should be in a different PR imo.

That works for me!

> Sounds like a code smell to me.

To me it doesn't. It is a utility after all.

yannickperrenet (approved):

fruttasecca:
I think it's not super clean as well but it's a tradeoff. One issue I can see with this approach is that it might introduce some kind of leaky abstraction, given that a service would get its logger from said function while having to configure it, which means we will be operating under the assumption that the logger retrieved through said function and the configuration set by the service are pointing to the same logger.
It doesn't look like the logging configurations are exactly the same, i.e. there are some unique service related entries.

Issue #264, opened 2021-05-23
Handle Jupyter notebook kernel writing more flexibly in pipeline save


ricklamers:
We are currently making to many assumptions about the structure of the JSON encoding of the Jupyter notebook when saving the pipeline file.

When the notebook file deviates from the expected format the save endpoint produces an error:

```
ERROR:app:app.py - [23/May/2021 12:30:59] - Exception on /async/pipelines/json/52d41e28-a23d-42da-90c8-ecea9068ed97/0915b350-b929-4cbd-b0d4-763cac0bb69f [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 272, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 272, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/orchest/services/orchest-webserver/app/app/views/views.py", line 703, in pipelines_json
    pipeline_set_notebook_kernels(
  File "/orchest/services/orchest-webserver/app/app/utils.py", line 678, in pipeline_set_notebook_kernels
    if notebook_json["metadata"]["kernelspec"]["language"] != language:
KeyError: 'language'
```

**To Reproduce**  
Save a notebook that doesn't contain the above metadata->kernelspec->language JSON structure.

yannickperrenet:
We probably need to have a discussion here on how we plan to solve this issue.

Some notes while thinking about this:
[...]

PR #308, opened 2021-06-09
Handle Notebook writing on pipeline save more robustly

yannickperrenet:
## Description
Fixes: #264


services/orchest-webserver/client/src/views/PipelineSettingsView.jsx
```diff
@@ -533,7 +533,7 @@ const PipelineSettingsView = (props) => {
               },
             });
           } else {
-            console.error("Could not save pipeline.json");
+            console.error("Could not save: pipeline definition OR Notebook JSON");
```
yannickperrenet, line 536: I do think the structure in this function is a bit strange, because I don't see this `else` getting hit.

The Python API endpoint returns a `400` status code whenever the `success` is `false`, thereby triggering the `catch` instead. How was this designed to work @ricklamers @joe-bell ?

ricklamers: I think this else branch can be removed iff the server never returns success != true. Which may be the case now (the backend endpoint and frontend calls to it sometimes go a bit out of sync).

yannickperrenet: > I think this else branch can be removed iff the server never returns success != true

The problem is the combination of `success = false` and `status_code = 400`, which cause the `else` to not be triggered. 

I will remove it for now, since the endpoint does not return the combination `success = false` and `status_code < 300`.

PR #318, opened 2021-06-11
Even Further Functional Component Transformation

joe-bell:
## Description

1. Upgrades `pnpm` (we're quite a bit behind and the latest version works fine both locally and in the CI)
2. Further implementation of #259    
Will need particular support on this to test the BuildPendingDialog (which is tricky to recreate locally)

### Checklist

- [x] In case I changed code in the `orchest-sdk`, I updated its version according to [SemVer](https://semver.org/) in its `_version.py` and updated the version compability table in its `README.md`


services/orchest-webserver/client/src/App.tsx
```diff
@@ -26,7 +26,22 @@ import PipelinesView from "./views/PipelinesView";
 import JobsView from "./views/JobsView";
 
 import $ from "jquery";
-import "./utils/overflowing";
+
+$.fn.overflowing = function () {
```
joe-bell, line 30: Moved this here to help with later jQuery removal (so it doesn't get missed off)

ricklamers (approved):
Great job moving us closer to all functional components. Don't see anything that stands out, in manual testing (I see the checkmark for manual testing hasn't been checked in the PR description) I would try to make sure that all areas of the app that consume these components are checked.

LGTM!

PR #320, opened 2021-06-14
Release/v2021.06.3

yannickperrenet:
## Description

Fixes the broken update mechanism.


yannickperrenet:
Note that the failed `pre-commit` check is flaky and due to an error in getting the file differences (we use `on` file change in the workflows).

PR #330, opened 2021-06-20
Run a pipeline through the CLI

yannickperrenet:
## Description
Adds functionality to the CLI to run a pipeline as a one-time job.

To try it out:
```bash
./orchest run --job=test --project=quickstart --pipeline="california-housing"
```

## Questions
* Q: Do we want it as a hidden or public CLI command? I would opt for public since we already indicate we are in alpha.
* A: Ended up with public.

* Q: The CLI now reads the pipeline definition (which it needs to create the strategy json for the job) from the `userdir`. Is this against our design decisions? And if so, do we want to add a dedicated endpoint webserver?
* A: Used already existing webserver endpoint.

* Q: How do we want to handle environment variables? I now opted to let the users use project and pipeline environment variables, because passing variables through the CLI, e.g. `--name=value`, would cause them to be in your shells history (which is undesirable as env variables can be secrets).
* A: Out of scope for now. Can definitely include later.

### DIFF
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -474,7 +474,6 @@ class OrchestApp:
 
         wait_msg_template = "[{endpoint}]: Could not reach Orchest."
         base_url = f"{ORCHEST_WEBSERVER_ADDRESS}{{path}}"
-        abs_projects_path = "/orchest-host/userdir/projects"
 
         # Get project information.
         _, resp = utils.retry_func(
@@ -486,7 +485,6 @@ class OrchestApp:
             # NOTE: We use here that a project name/path is unique.
             if project["path"] == project_name:
                 project_uuid = project["uuid"]
-                project_path = project["path"]
                 break
         else:
             utils.echo("The given project does not exist.")
@@ -501,23 +499,13 @@ class OrchestApp:
         for pipeline in resp["result"]:
             if pipeline["name"] == pipeline_name:
                 pipeline_uuid = pipeline["uuid"]
-                pipeline_path = pipeline["path"]
                 break
         else:
             utils.echo("The given pipeline does not exist in the given project.")
             return
 
-        # Start pipeline run.
+        # Draft job.
         utils.echo("Creating draft job.")
-        # TODO: Probably want to get the pipeline definition through the
-        # webserver instead of having the orchest-ctl read it from the
-        # userdir.
-        full_pipeline_path = os.path.join(
-            abs_projects_path, project_path, pipeline_path
-        )
-        with open(full_pipeline_path, "r") as f:
-            pipeline_definition = json.load(f)
-
         _, resp = utils.retry_func(
             utils.get_response,
             _wait_msg=wait_msg_template.format(endpoint="jobs"),
@@ -542,7 +530,6 @@ class OrchestApp:
         # required environment builds which we need to await before
         # queueing the job.  This behavior is equal to the behavior
         # through the `orchest-webserver`.
-        utils.echo("Queueing job.")
         repeat = True
         while repeat:
             try:
@@ -559,7 +546,7 @@ class OrchestApp:
                 utils.echo(
                     "It seems like Orchest experienced an internal server error."
                 )
-                return
+                raise typer.Exit(code=1)
             else:
                 # Check for status code 201 because it is a POST
                 # request.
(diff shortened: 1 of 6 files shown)


services/orchest-ctl/app/orchest.py
```diff
@@ -465,6 +462,126 @@ def add_user(self, username, password, token, is_admin):
...
+        )
+        job_uuid = resp["uuid"]
+
+        # NOTE: When creating a draft for a job, this triggers the
+        # required environment builds which we need to await before
+        # queueing the job.  This behavior is equal to the behavior
+        # through the `orchest-webserver`.
+        utils.echo("Queueing job.")
+        repeat = True
+        while repeat:
```
yannickperrenet, line 547: I made this an infinite loop as it is impossible to tell how long an environment build could last.

services/orchest-ctl/app/orchest.py
```diff
@@ -539,3 +661,38 @@ def get_required_images(language: Optional[str], gpu: bool = False) -> List[str]
...
+    def parse_parameters_format(value: dict) -> dict:
+        """Parses the parameter values to the correct format.
+
+        The format for the strategy JSON parameters is:
+            {a: "[1]", b: "[2]"}
+        whereas, the given value is:
+            {a: 1, b: 2}
+
+        """
+        return {p: str([v]) for p, v in value.items()}
```
fruttasecca, line 685: Can this be moved outside of `construct_parameters_payload`?

fruttasecca:
>Do we want it as a hidden or public CLI command? I would opt for public since we already indicate we are in alpha.

It being public is fine imo, although the reason for having this feature is somewhat internal and the "cause" of this feature may later go away, moreover, it being public or not plays into what level of correctness/completeness we want this feature to be in, see the next points.

>The CLI now reads the pipeline definition (which it needs to create the strategy json for the job) from the userdir. Is this against our design decisions? And if so, do we want to add a dedicated endpoint webserver?

I think this is definitively getting us into weird territory. The issue with moving this to the webserver is that then you would have to wait for an environment to be ready in the webserver, which is strange as well. This is because we are creating a job draft and confirming it within the same "action", splitting the action into 2 CLI calls is a lot of work and not really user friendly, and we would probably have to support a `--auto-confirm` option anyway.
At the broader level it almost feels like we need an API service that is parallel to what we have been doing with the webserver:
- user interacting with the GUI -> webserver -> orchest-api
- user interacting with the CLI -> api-server -> orchest-api

This can technically be bundled within the webserver although I feel like the webserver is doing a lot already, e.g. the log streamer, being an actual web server, being a proxy for the orchest-api etc. I am just throwing this out there because implementing a new service that, abstractly, can assume the role of a user-facing api might be too costly and not justifiable if we are simply interested in this specific CLI command.  
All in all the most important point  I want to make is that we need to be really aware of what kind of design we are getting into if we allow `orchest-ctl` to do this, all other alternatives look more costly.

>How do we want to handle environment variables? I now opted to let the users use project and pipeline environment variables, because passing variables through the CLI, e.g. --name=value, would cause them to be in your shells history (which is undesirable as env variables can be secrets).

I would say we should allow passing secrets as such, stating in the `help` message about the shell history, and have a similar message be echo'd by the command while running if any env variable is passed this way. We could also make it possible to pass a secrets file, if this feature (the CLI command) is public the importance of said file is higher.

services/orchest-ctl/app/orchest.py
```diff
@@ -465,6 +462,131 @@ def add_user(self, username, password, token, is_admin):
...
+                "name": job_name,
+                "parameters": [],
+                "pipeline_run_spec": {
+                    "run_type": "full",
+                    "uuids": None,  # argument is ignored due to "full"
+                },
+            },
+            method="POST",
+        )
+        job_uuid = resp["uuid"]
```
fruttasecca, line 539: Minor issue but worth noting: both this query and the one confirming the draft assume that everything went well, right? Might need to check the reponse code to make it more robust. This also plays into (very improbable) race conditions.

services/orchest-ctl/app/orchest.py
yannickperrenet: It can, but I don't really see why. Can you elaborate?

I like the fact it is a nested function because it clearly indicates that it is not (to be) used outside of the function it is defined in.

services/orchest-ctl/app/orchest.py
yannickperrenet: Will add!

services/orchest-ctl/app/orchest.py
fruttasecca: It being nested made me think it was dependant on a locally scoped variable while it was not.  I am not feeling strongly about it although it felt somewhat strange.

yannickperrenet:
Probably good to mention, but if anything fails after the draft job has been created the draft job will still be in the DB (and UI). It is not cleaned.

I would say this is the appropriate behavior, because users don't lose the state of the project dir on creation of the job.

fruttasecca:
lgtm :)

fruttasecca (approved):

PR #349, opened 2021-06-23
v2021.06.4

ricklamers:


yannickperrenet:
Can we remove the "Orchest's current status is online." in the Orchest settings? This message does not provide any additional information. (Also when restarting the message does not have to be displayed.)

yannickperrenet:
As a general question, should the help button (top right corner) be a toggle? So that clicking it twice shows you your previous view again.

yannickperrenet:
Now that settings was moved from the top bar to the left pane menu, the ordering of the items feels a bit strange to me. In particular, the "projects" feels like it is out of place.

fruttasecca:
>Now that settings was moved from the top bar to the left pane menu, the ordering of the items feels a bit strange to me. In particular, the "projects" feels like it is out of place.

Agree, feels like "projects" should be after environments

ricklamers:
> As a general question, should the help button (top right corner) be a toggle? So that clicking it twice shows you your previous view again.

I think this is pretty strange UX for a separate page. Would prefer to leave it as is, you can just use the browser back button.

yannickperrenet:
Lastly, I found a bug in the pipeline editor. 

#### To reproduce
1. Open a pipeline
2. Delete a connection
3. Try to delete a selection (can also be single) of steps using `<Del>` or `<Backspace>`. Neither work.

ricklamers:
> Can we remove the "Orchest's current status is online." in the Orchest settings? This message does not provide any additional information. (Also when restarting the message does not have to be displayed.)

I already updated the message (it just shows the approximate duration instead of "Restarting, ..."). I'd rather keep the Orchest status sentence there because it feels more explicit. Implicit status through a loader feels like bad UX.

yannickperrenet (approved):

services/orchest-webserver/client/src/utils/wrap-number.js
```diff
@@ -0,0 +1,13 @@
...
+ * Wraps a number when it leaves the specified range.
+ * @param {number} min
+ * @param {number} max
+ * @param {number} index
+ * @returns number
+ */
+export const wrapNumber = (min, max, index) => {
+  const rangeSize = max - min;
+  return ((((index - min) % rangeSize) + rangeSize) % rangeSize) + min;
+};
```
yannickperrenet, line 13: Not sure what exactly this functions needs to be, but seems a bit funky to me.

yannickperrenet: Pretty sure that can be rewritten to
```js
  return ((index - min) % rangeSize) + min;
```

yannickperrenet: Turns out javascript returns a negative remainder in the case where `index` is negative. Therefore this (mathematically correct) rewrite does not work in javascript.

Issue #326, opened 2021-06-15
Environment images are not available for job runs after update


ricklamers:
**Describe the bug**  
Environment images are not available for job runs after update

**To Reproduce**  

1. Create a cron job
2. Update Orchest
3. Environment image is deleted
4. Build is not triggered for environments referenced in cron jobs
5. When a job triggers a pipeline run, the run fails because no environment image is available

**Expected behavior**  
The job should succeed in running after an update.

**Environment**

- OS: Linux server
- Browser: Chrome
- Version: 2021.06.1
- Environment: prod, cloud.

yannickperrenet:
Depending on how we approach this issue, it is probably good to be aware of #310.

PR #361, opened 2021-06-28
Improv/job environments consistency

fruttasecca:
## Description
This PR solves #326, making it so that environments are not available to a user to be used again for new jobs or interactive sessions after an update, but retaining the underlying resource ("build") backing said environment so that jobs that depend on them are not broken.  

Moreover, this PR makes it so that all runs of a job will use the same environment build(s), regardless of the environment being rebuilt. Before this change, this consistency was only guaranteed across runs belonging to the same batch, i.e. a set of runs running the same pipeline but with different parameters, triggered, for example, by a cron job or a one-off run.

Fixes: #326

### Checklist


fruttasecca:
@ricklamers @joe-bell If we agree that a job should snapshot/lock environments on POST, which equals to the moment the project directory is also snapshotted,  there needs to be some minor GUI work consisting of moving the `checkGate` call, currently done `EditJobView.jsx`, to `JobList.jsx`.  
Wasn't super sure about not breaking any internal state (and about us agreeing about the aforementioned thing) so I kinda left this to the front-end specialists :pray: .

services/orchest-api/app/app/utils.py
```diff
@@ -356,6 +357,67 @@ def lock_environment_images_for_session(
     return env_uuid_docker_id_mappings
 
 
+def lock_environment_images_for_job(
+    job_uuid: str, project_uuid: str, environment_uuids: Set[str]
+) -> Dict[str, str]:
+    """Retrieve the docker ids to use for the runs of a job.
+
+    See lock_environment_images_for_run for more details.
```
fruttasecca, line 365: This "pattern" is definitively getting annoying but given the complexity of what's happening here I opted for not refactoring it in this PR. Currently, I am tinkering with the idea of using postgres advisory locks (application-level locks) to make this logic simpler. Another alternative would be to incorporate this change into the task of abstracting the container/image environment behind a dedicated module.

docs/source/user_guide/jobs.rst
```diff
@@ -76,3 +78,8 @@ To inspect the result of your job, simply click on the job you just created, cho
 pipeline run (the one you want to inspect) and open *View pipeline*. The pipeline is now opened in
 :ref:`read-only mode <read-only mode>` giving you the opportunity to check the logs or to open the
 HTML version of you notebooks.
+
+.. tip::
```
yannickperrenet, line 82: Definitely a nitpick, but I would call it a `note` since a `tip` implies something the user can do to improve things. But this works out of the box :tada:

lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -28,6 +28,15 @@
...
 # Environments
 ENVIRONMENT_IMAGE_NAME = "orchest-env-{project_uuid}-{environment_uuid}"
+# NOTE: the build_uuid is rather important, since it allows us to
+# differentiate between builds of the same environment. Without that,
+# different env builds would contend for the same tag (the same
+# environment image removal name), leading to issues in the case of
+# multiple environment images "versions" surviving through multiple
+# updates because of jobs.
+ENVIRONMENT_IMAGE_REMOVAL_NAME = (
+    "removed-orchest-env-{project_uuid}-" "{environment_uuid}-{build_uuid}"
```
yannickperrenet, line 38: Formatting probably auto-changed by black on pre-commit hook:
```python
"removed-orchest-env-{project_uuid}-{environment_uuid}-{build_uuid}"
```

services/orchest-api/app/app/models.py
```diff
@@ -592,6 +600,42 @@ def __repr__(self):
...
+        nullable=False,
+        index=True,
+        primary_key=True,
+    )
+    orchest_environment_uuid = db.Column(
+        db.String(36), unique=False, nullable=False, primary_key=True, index=True
+    )
+    docker_img_id = db.Column(
+        db.String(), unique=False, nullable=False, primary_key=True, index=True
+    )
```
yannickperrenet, line 629: Since we are communicating through Docker tags, could it be a problem that the same image ID is used when an image only has a different tag?

From the docs: "An image will be listed more than once if it has multiple repository names or tags."
Example:
```
postgres                  9                   746b819f315e        4 days ago          213.4 MB
postgres                  9.3                 746b819f315e        4 days ago          213.4 MB
postgres                  9.3.5               746b819f315e        4 days ago          213.4 MB
postgres                  latest              746b819f315e        4 days ago          213.4 MB
```
Reference: [Docker docs on images](https://docs.docker.com/engine/reference/commandline/images/)

services/orchest-ctl/app/orchest.py
```diff
@@ -137,10 +137,13 @@ def update(self, mode=None, dev: bool = False):
...
-        # having environments with mismatching Orchest SDK versions.
-        logger.info("Deleting user-built environment images.")
-        self.resource_manager.remove_env_build_imgs()
+        # Add a tag to user environment images to mark them for removal.
+        # The orchest-api will deal with the rest of the logic related
+        # to making environment images unavailable to users on update
+        # to avoid the issue of having environments with mismatching
+        # Orchest SDK versions.
+        logger.info("Tagging user-built environment images for removal.")
+        self.resource_manager.tag_environment_images_for_removal()
```
yannickperrenet, line 146: I think it is better practice if the `tag_environment_images_for_removal` function takes the format of the strings (.e.g `ENVIRONMENT_IMAGE_NAME`) as arguments instead of having it read from the `_config`.

yannickperrenet:
Great work! Didn't expect to require changes in so many different places.

This is definitely an improvement ;)

lib/python/orchest-internals/_orchest/internals/config.py
fruttasecca: indeed, will fix

services/orchest-ctl/app/orchest.py
fruttasecca: Given the name of the function, wouldn't passing those strings as arguments imply that removal tags are arbitrary? i.e. it's possible to mark for removal with different tags? The current implementation suggests a stronger contract imo.

services/orchest-api/app/app/models.py
fruttasecca: Given the current logic, not specific to this PR but with the overall logic of environment images and image locking, I don't see any particular issue related to this. Do you have anything in mind?  
I think this is just another point in favour of abstracting this resource, right now we can think (assume) that the logic is correct but said logic risks being scattered around the codebase and increases the risk of mistakes as it becomes more complex.

services/orchest-ctl/app/orchest.py
yannickperrenet: I agree it is indeed a stronger contract, but something that should be controlled by the caller not the function itself. It reduces the extensibility of the function.

I am fine leaving it as is though.

yannickperrenet (approved):

services/orchest-ctl/app/orchest.py
fruttasecca: I think I will opt for leaving it as it is until a pattern emerges that leads us to this discussion/issue again then.

yannickperrenet:
So actually this PR breaks the `orchest run` command, e.g. 
```bash
./orchest run --job=test --project=quickstart --pipeline="california-housing"
```
in the edge case where the environments are building. 

### To reproduce
1. Import the quickstart pipeline
2. Note that this is automatically building the environments
3. While the environments are building, run the above command

This will lead to an error whereas it used to wait for the environments to finish building. 

My guess is that the following commit introduced the bug: https://github.com/orchest/orchest/pull/361/commits/414993dcf6e56f3c1ae97ee07ff84ca6902f9725

yannickperrenet:
The above error can actually be triggered through the UI as well. 

### To reproduce
1. Import the quickstart pipeline
2. Note that this is automatically building the environments
3. While the environments are building, try to create a job

PR #378, opened 2021-07-08
Project, pipeline, job renaming

fruttasecca:
## Description
This PR introduces renaming projects, pipelines and jobs, along with moving pipelines. This is a transparent process that automatically takes care of maintaining the consistency and correctness of the pipeline file, which points at other files through a relative path.

### Checklist

#### Implementation details of interest to the front end implementation
Renaming a project
- constraint: cannot happen while a project has an active interactive session (the backend will reject with a 409, returned json has a `code` entry equal to 0)
- constraint: cannot happen if a project has the same name (the backend will reject with a 409, returned json has a `code` entry equal to 1)
- endpoint: PUT `/async/projects/<project_uuid>` 
- payload  `{"name":"<new name>"}`

Renaming a pipeline:
- POST async/pipelines/json/<project_uuid>/<pipeline_uuid>
-  the payload is the pipeline json, if the pipeline `name` has been changed the backend will take care of maintaining consistency w.r.t other entities, for example, the `job.pipeline_name` entry, that should be used by the front end when displaying the pipeline name in the job details view

Moving a pipeline:
- constraint: cannot happen while a project has an active interactive session (the backend will reject with a 409)
- endpoint: PUT `/async/pipelines/<project_uuid>/<pipeline_uuid>` 
- payload  `{"path":"<new name>"}`

Renaming a job:
- endpoint: PUT `/catch/api-proxy/api/jobs/<job uuid>`
-  payload  `{"name":"<new name>"}`
-  the payload can contain any other data related to the endpoint, i.e. the name can be updated along with other properties


joe-bell:
Note for reviewers/contributors: this PR contains a schema change

yannickperrenet:
I have a question. Why do we not treat the "Renaming a pipeline" the same as "Moving a pipeline"? Same as the `mv` command.

services/orchest-api/app/app/models.py
```diff
@@ -22,6 +25,19 @@ class BaseModel(db.Model):
...
+
+        Can be used to sanitize a status update.
+        """
+        data = copy.deepcopy(data)
+        columns = [c.name for c in cls.__table__.columns]
+        for key in list(data.keys()):
+            if key not in columns:
+                data.pop(key)
+        return data
+
```
yannickperrenet, line 40: The following should be a little faster, since a deepcopy is only created for entries that are a column. (Btw, you can also just do `for key in data` as that is short for `for key in data.keys()` -- no conversion to `list` is needed.)
```suggestion
        ans = {}
        columns = [c.name for c in cls.__table__.columns]
        for key, value in data.items():
            if key in columns:
                ans[key] = copy.deepcopy(value)
        return ans
```

services/orchest-webserver/app/app/core/pipelines.py
```diff
@@ -193,3 +212,144 @@ def _revert(self):
...
+
+    Moves a pipeline, moving it to another path.
+    """
+
+    def _transaction(
+        self, project_uuid: str, pipeline_uuid: str, new_project_relative_path: str
+    ):
+
+        pipeline = (
+            Pipeline.query.with_for_update()
```
yannickperrenet, line 228: As a general question, would `Pipeline.query.with_for_update` not lock the entire `Pipeline` table? Would it work to first do the `filter` and then call `with_for_update` to lock only those rows?

I could definitely be wrong here ;)

services/orchest-webserver/app/app/core/pipelines.py
```diff
@@ -193,3 +212,144 @@ def _revert(self):
...
+        new_path = get_pipeline_path(None, project_uuid, pipeline_path=new_path)
+        project_path = os.path.abspath(get_project_directory(project_uuid))
+        new_path_abs = os.path.abspath(new_path)
+        if not new_path_abs.startswith(project_path):
+            raise ValueError(
+                "New pipeline path points outside of the project directory."
+            )
+
+        # Update the pipeline definition by adjusting the step file
+        # paths, since they should be relative to the pipeline file.
```
yannickperrenet, line 288: Good catch!

services/orchest-webserver/app/app/core/projects.py
```diff
@@ -177,6 +181,84 @@ def _revert(self):
         Project.query.filter_by(uuid=self.collateral_kwargs["project_uuid"]).update(
             {"status": "READY"}
         )
+        db.session.commit()
+
+
+class RenameProject(TwoPhaseFunction):
```
yannickperrenet, line 187: This is most likely a front-end concern, but I got the thought here...

Should we do anything, e.g. a dialog, when renaming a project in case a user has set up a remote git repo? Because renaming the project would yield a different name from the remote.

Probably better to be completely left to the user is my guess.

services/orchest-webserver/app/app/core/pipelines.py
fruttasecca: `with_for_update` simply appends the `FOR UPDATE `  clause of postgres to the query, where "FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update."

fruttasecca: >app.config["SQLALCHEMY_ECHO"] = True

this setting can be used to get what sql queries the backend is producing, to verify said statement

services/orchest-webserver/app/app/core/projects.py
fruttasecca: What issue would this lead to?

fruttasecca:
>I have a question. Why do we not treat the "Renaming a pipeline" the same as "Moving a pipeline"? Same as the mv command.

After an internal discussion, we decided to leave the file name separated from the actual pipeline name. Open to revisit this, not sure to what extent we will break existing pipelines/projects, etc.

services/orchest-webserver/app/app/views/views.py
```diff
@@ -368,6 +372,15 @@ def discoverFSCreatedProjects():
         ]
         new_project_paths = set(project_paths) - set(existing_project_paths)
 
+        # Do not do project discovery if a project is moving, because
```
yannickperrenet, line 375: I was a bit confused when coming across the function `discoverFSCreatedProjects`, why is it defined within the `register_views` function? It is not a view?

services/orchest-webserver/app/app/core/pipelines.py
yannickperrenet: Alright that clarifies things. Thank you!

services/orchest-webserver/app/app/core/projects.py
yannickperrenet: Wouldn't really call it an issue. Just something a user might not be aware of when doing the operation. 

Food for thought, don't think we need to do anything here ;)

yannickperrenet:
Pretty darn great PR, given all the edge cases and possibly race conditions. Great job! :muscle:

services/orchest-webserver/app/app/core/projects.py
fruttasecca: Alright, I thought there was some issue that I was not seeing, ty for clarifying.

services/orchest-webserver/app/app/views/views.py
fruttasecca: Good catch these 2 functions should be moved out of there, will find them a home.

Issue #376, opened 2021-07-08
Environment variable names containing spaces are not handled correctly.


fruttasecca:
**To Reproduce**  
Steps to reproduce the behaviour:

1. Create a project or pipeline environment variable with a name containing spaces, e.g. `  a   `, or  `hello world`
2. Start a session
3. Said variables will not be set, i.e. they won't exist in the environment

This is likely affecting the value of said environment variables as well.
Edit: this is only affect the **name** of environment variables.

**Expected behavior**  
Said environment variables should be set in the environment. Assuming that we want to allow such environment variable names.

yannickperrenet:
> Assuming that we want to allow such environment variable names.

I don't think we do. I wouldn't know how to create such a variable, nor how to get its value.

And from [stackoverflow](https://stackoverflow.com/a/2821183/14839751):
> So while the names may be valid, your shell might not support anything besides letters, numbers, and underscores.

I do think we can validate env names when entered through the UI, e.g. conforming with the regex `[a-zA-Z_]+[a-zA-Z0-9_]*`. @joe-bell Thoughts?

fruttasecca:
I must highlight that such a thing is currently possible in docker, e.g. `docker run --rm -it -e " A "=1 python:3.7 python`. I agree that it is confusing and is likely to break depending on the base image so I would be in favour of not allowing it. The check should be both at the client and backend level.

PR #382, opened 2021-07-12
Environment variables name validation

fruttasecca:
## Description
Adds validation to environment variable names.
Fixes: #376 

@joe-bell, when it comes to the front end, the following things needs to be validated:
- project env variables
- pipeline env variables
- jobs env variables
- services env variables and inherited env variables

Only the name of environment variables needs to be validated. The backend considers `^[0-9a-zA-Z\-_]+$` as valid, the BE and the FE are to be in sync with this.

### Checklist

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -216,6 +216,17 @@ def docker_images_rm_safe(docker_client, *args, attempt_count=10, **kwargs):
         time.sleep(1)
 
 
+def are_environment_variables_valid(env_variables: Dict[str, str]) -> bool:
+    return isinstance(env_variables, dict) and all(
+        [
+            isinstance(var, str)
+            and re.match(r"^[0-9a-zA-Z\-_]+$", var)
+            and isinstance(value, str)
+            for var, value in env_variables.items()
+        ]
+    )
+
+
 def is_service_name_valid(service_name: str) -> bool:
     # NOTE: this is enforced at the GUI level as well, needs to be kept
     # in sync.
@@ -254,13 +265,7 @@ def is_service_definition_valid(service: Dict[str, Any]) -> bool:
                 for var in service.get("env_variables_inherit", [])
             ]
         )
-        and isinstance(service.get("env_variables", {}), dict)
-        and all(
-            [
-                isinstance(var, str) and isinstance(value, str) and var
-                for var, value in service.get("env_variables", {}).items()
-            ]
-        )
+        and are_environment_variables_valid(service.get("env_variables", {}))
     )
 
 
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -219,12 +219,19 @@ def docker_images_rm_safe(docker_client, *args, attempt_count=10, **kwargs):
 def are_environment_variables_valid(env_variables: Dict[str, str]) -> bool:
     return isinstance(env_variables, dict) and all(
         [
-            isinstance(var, str)
-            and re.match(r"^[0-9a-zA-Z\-_]+$", var)
-            and isinstance(value, str)
+            is_env_var_name_valid(var) and isinstance(value, str)
             for var, value in env_variables.items()
         ]
     )
+
+
+def is_env_var_name_valid(name: str) -> bool:
+    # Needs to be kept in sync with the FE.
+    return isinstance(name, str) and re.match(r"^[0-9a-zA-Z\-_]+$", name)
+
+
+def make_env_var_name_valid(name: str) -> str:
+    return re.sub("[^0-9a-zA-Z\-_]", "_", name)
 
 
 def is_service_name_valid(service_name: str) -> bool:
@@ -261,7 +268,7 @@ def is_service_definition_valid(service: Dict[str, Any]) -> bool:
         and isinstance(service.get("env_variables_inherit", []), list)
         and all(
             [
-                isinstance(var, str) and var
+                is_env_var_name_valid(var)
                 for var in service.get("env_variables_inherit", [])
             ]
         )
--- a/services/orchest-api/app/app/apis/namespace_jobs.py
+++ b/services/orchest-api/app/app/apis/namespace_jobs.py
@@ -13,6 +13,7 @@ from sqlalchemy.orm import joinedload, undefer
 
 import app.models as models
 from _orchest.internals import config as _config
+from _orchest.internals import utils as _utils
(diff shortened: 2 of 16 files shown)


lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -216,6 +216,24 @@ def docker_images_rm_safe(docker_client, *args, attempt_count=10, **kwargs):
         time.sleep(1)
 
 
+def are_environment_variables_valid(env_variables: Dict[str, str]) -> bool:
+    return isinstance(env_variables, dict) and all(
+        [
+            is_env_var_name_valid(var) and isinstance(value, str)
+            for var, value in env_variables.items()
+        ]
+    )
```
yannickperrenet, line 225: I think this function should be rewritten to raise a `TypeError` in case the argument does not have a valid type. Otherwise it is unclear why the function fails, e.g. the environment variables are valid but just passed in the wrong data structure.

Same argument for the `is_env_var_name_valid` function.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -688,6 +689,8 @@ def _transaction(
                         "a job which is not a cron job."
                     )
                 )
+            if not _utils.are_environment_variables_valid(env_variables):
+                raise ValueError(("Invalid environment variables definition."))
```
yannickperrenet, line 693:
```suggestion
                raise ValueError("Invalid environment variables definition.")
```

services/orchest-api/app/app/apis/namespace_pipelines.py
```diff
@@ -34,9 +35,12 @@ def get(self):
     @api.marshal_with(schema.pipeline)
     def post(self):
         """Create a new pipeline."""
+        pipeline = request.get_json()
+        pipeline["env_variables"] = pipeline.get("env_variables", {})
+        if not _utils.are_environment_variables_valid(pipeline["env_variables"]):
+            return {"message": ("Invalid environment variables definition.")}, 400
```
yannickperrenet, line 41: Do you think these "pipeline validity checks" should be extended to check more than only the environment variables? If so, then I think it would be good to move this logic into a function `is_pipeline_definition_valid` (or something like that). Making that change now allows us to more easily extend the validity checks.

services/orchest-webserver/app/app/compat.py
```diff
@@ -1,9 +1,41 @@
...
+            tmp = {}
+            for key, value in env_variables.items():
+                valid_key = _utils.make_env_var_name_valid(key)
+
+                # Do not lose variables to collisions.
+                i = 2
+                while valid_key in tmp:
+                    valid_key = f"{valid_key}_{i}"
+                    i += 1
+                tmp[valid_key] = value
```
yannickperrenet, line 25: Wouldn't this be a leaky abstraction?

What if:
1. User enters an invalid project level environment variable directly in the pipeline definition
2. The user views the environment variables of a project
3. The Orchest webserver loads and migrates the pipeline definition for the view and displays the migrated environment variables
4. The user might be amazed or doesn't even notice that the variables are changed without him/her doing anything.

yannickperrenet:
It seems like there is still work to be done on the front-end? For example when I enter an invalid env var name, e.g. `hello world`, then the `SAVE` button doesn't work but at the same time doesn't provide a dialog explaining why.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: Wouldn't that be the case for all validation code in `_orchest/internals/utils.py`? And about the reason of failure, this highlights an issue that we have about bubbling up feedback to the client. The webserver and the client sometimes communicate through the status code, sometimes the client uses the error message of the webserver, etc.

The issue I have with having a function that answers a question raising exceptions to communicate is that the code gets awkward really easily. Not advocating against it, but we need to find a solution that marries into the way we intend to bubble up feedback.

services/orchest-api/app/app/apis/namespace_pipelines.py
fruttasecca: Having an `is_pipeline_definition_valid` would not hurt but would not remove `are_environment_variables_valid` calls from numerous places (only from the `json` endpoint I believe) given that, for example, POSTing to namespace_pipelines does not imply sending a pipeline definition, same for the projects namespace, services, etc.

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: > this highlights an issue that we have about bubbling up feedback to the client.

I 100% agree. 

Let's say with the current implementation we internally pass the wrong data structure to the function, then this is reasonably hard to debug since we don't distinguish this case from an invalid key name.

services/orchest-webserver/app/app/compat.py
fruttasecca: >User enters an invalid project level environment variable directly in the pipeline definition

Note that the only env vars that are stored in the pipeline definition are the ones related to services, proj/pipeline env vars are at the DB level. Nevertheless, the example can be made with services env vars.

How do you distinguish an entry that has been manually modified to be "incorrect" from a pipeline that is genuinely migrating? Does this mean that we should allow users manually modifying their pipeline to be incorrect in the more general case? (again, because then we can't distinguish a genuine migration from intentional/manual incorrectness).
Also, let's say this kind of migration of env vars does not happen, that leads to the client having an incorrect pipeline definition that won't be accepted on save.

fruttasecca:
>It seems like there is still work to be done on the front-end? For example when I enter an invalid env var name, e.g. hello world, then the SAVE button doesn't work but at the same time doesn't provide a dialog explaining why.

Indeed, the PR won't be complete until @joe-bell adds the finishing touches to the client, see the description for what changes are required.

services/orchest-webserver/app/app/compat.py
yannickperrenet: > Also, let's say this kind of migration of env vars does not happen, that leads to the client having an incorrect pipeline definition that won't be accepted on save.

This is indeed the argument to leave it inplace, the migration is there so that the application does not fail on an invalid pipeline definition. Just wanted to highlight that in order to not fail we are actually changing user code (since I consider the pipeline definition to be user code) and it is good to be aware of this.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: I agree about it being difficult to debug. Most if not all validations that are done by the backend are just there in case client level validations miss something, so it has been a long time since we had a definition/validation issue to debug, nevertheless, I agree.

PR #383, opened 2021-07-12
Fix query args passing at `/login`

yannickperrenet:
## Description

The view to load (either `login` or `admin`) in the `auth-server` was determined by 
```javascript
let view = window.location.href.split("/").slice(-1)[0];
```
which incorrectly handles URLs such as `http://localhost:8000/login?redirect_url=/environments` (`view` would be `environments`).

This PR fixes the issue, making the following possible again:
* http://localhost:8000/login?redirect_url=/environments
*  http://localhost:8000/login?redirect_url=/projects&import_url=https://github.com/orchest/quickstart

> FYI the issue was (unknowingly) reintroduced in a6770357512250425ac927aa1cae6227ed760f1f.

### DIFF
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -142,6 +142,9 @@ def register_views(app):
                     db.session.commit()
 
                     redirect_url = request.args.get("redirect_url", "/")
+                    import_url = request.args.get("import_url")
+                    if import_url is not None:
+                        redirect_url += "?" + "import_url=" + import_url
 
                     resp = redirect_response(redirect_url, redirect_type)
                     resp.set_cookie("auth_token", token.token)
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -141,10 +141,15 @@ def register_views(app):
                     db.session.add(token)
                     db.session.commit()
 
-                    redirect_url = request.args.get("redirect_url", "/")
-                    import_url = request.args.get("import_url")
-                    if import_url is not None:
-                        redirect_url += "?" + "import_url=" + import_url
+                    # Returns a shallow mutable copy of the immutable
+                    # multi dict.
+                    request_args = request.args.copy()
+                    redirect_url = request_args.pop("redirect_url", "/")
+                    query_args = "&".join(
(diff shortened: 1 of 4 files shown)


ricklamers:
The `/login` endpoint has a different JS client altogether.

Hence, I don’t think this ever caused an issue. Although arguably the new way to determine the first path component is more robust.

services/auth-server/client/src/App.tsx
```diff
@@ -25,13 +25,18 @@ export default class App extends React.Component {
   }
 
   render() {
+    const view = window.location.pathname.split("/")[1];
```
ricklamers, line 28: This will parse the URL on each app render, this is undesirable. This code should still only run when the root React component is first initialized.

services/auth-server/app/app/views.py
```diff
@@ -141,7 +141,15 @@ def handle_login(redirect_type="client"):
                     db.session.add(token)
                     db.session.commit()
 
-                    redirect_url = request.args.get("redirect_url", "/")
+                    # Returns a shallow mutable copy of the immutable
```
ricklamers, line 144: The `redirect_url` should include any query arguments if necessary. The other query arguments shouldn’t be passed.

ricklamers (changes requested):
See comments

ricklamers:
Please pay attention to the fact that the full redirect_url will be/should be properly URL encoded.

`/login?redirect_url= %2Fprojects%26import_url%3Dhttps%3A%2F%2Fgithub.com%2Forchest%2Fquickstart`

services/auth-server/app/app/views.py
yannickperrenet: Not sure I follow here, because something like http://localhost:8000/login?redirect_url=/projects&import_url=https://github.com/orchest/quickstart should also work which contains two query arguments: `redirect_url` and `import_url`. Good to note here that `import_url` only works on the `/projects` view.

services/auth-server/client/src/App.tsx
yannickperrenet: Will move :+1:

fruttasecca (approved):

joe-bell (approved):

PR #384, opened 2021-07-13
Introduce Cypress

joe-bell:
> **WORK IN PROGRESS**

## Description

A working branch to integrate Cypress e2e testing:
1. Installs Cypress with the default template, along with extra tweaks to work with both PNPM and TypeScript

### Checklist

- [ ] In case I changed code in the `orchest-sdk`, I updated its version according to [SemVer](https://semver.org/) in its `_version.py` and updated the version compability table in its `README.md`


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Pre cleanup
+        run: |
+          ./orchest stop
+          [ ! -z "$(docker ps -a -q)" ] && docker rm $(docker ps -a -q)
+          sudo bash scripts/clean_userdir.sh
+
+      - name: Build and Start Orchest
+        run: |
+          bash scripts/build_container.sh
+          ./orchest start
```
yannickperrenet, line 54: Might be good to keep in mind if we want to test the `--cloud` configuration as well.

fruttasecca: Indeed, we can either use `--cloud` by default or scale up the number of jobs. Let's see what the run times looks like for 1 browser 1 start mode.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+      # In case some failed job has left state.
+      - name: Pre cleanup
+        run: |
+          ./orchest stop
+          [ ! -z "$(docker ps -a -q)" ] && docker rm $(docker ps -a -q)
+          sudo bash scripts/clean_userdir.sh
+
+      - name: Build and Start Orchest
+        run: |
+          bash scripts/build_container.sh
```
yannickperrenet, line 53: I think this is indeed the way to go, but we have to keep in mind the fact that Docker could incorrectly build due to cache issues. Although this has not happened in a while now.

fruttasecca: Yeah, that issue is indeed a pain to deal with. Not sure we can avoid it without having ephemeral instances.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+          # Recommended: pass the GitHub token lets this action correctly
+          # determine the unique run id necessary to re-run the checks
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Post cleanup
+        run: |
+          ./orchest stop
+          [ ! -z "$(docker ps -a -q)" ] && docker rm $(docker ps -a -q)
+          sudo bash scripts/clean_userdir.sh
```
yannickperrenet, line 73: Regarding cleanup, who is responsible for cleaning up testcase state? And what if the tests exit unexpectedly?

Could become a problem where e.g. an environment image sticks around and tests failing. Or because the images stick around, the docker installation blows up leaving us no disk space on the instance.

fruttasecca: Right now the only state that this action is not taking care of (I think) are environment builds and services (because a service might be a pulled image). The former could be taken care of using tags, the latter requires a bit more smartness. Good catch.

joe-bell:
⚠️ **THIS PR IS NOT READY FOR REVIEW**

This is going to be a pretty big PR, so to avoid bloating the commit trail let's leave reviews until the PR leaves its "DRAFT" status

cc @yannickperrenet @fruttasecca

fruttasecca:
Update:

A few issues have been found & fixed during the development of this PR:
- a race condition that led to dangling environment images in the case of a very quick abortion of the build
- a race condition that led to dangling environment images if the abortion of a build arrived too late to the worker
- a race condition occurring when multiple project deletions for the same project were issued
- a category of issues related to rejected asynchronous client requests has been found, which led to client errors in case of view changes. The issue itself is not very disrupting, but it made integration testing more difficult. Instances of this issue have been fixed by `catching` the promise/
- the path validation for moving a pipeline has been improved
- the path validation for creating a step has been improved, and it is now consistent with the action of pipeline moving by the fact that it also takes care of creating directories that do not exist. Moreover, the path of a step does not being with a `/` anymore, which is consistent with pipeline paths.

yannickperrenet:
I will do a more in-depth review later, but before I forget I wanted to note that we should improve the docs on how to run the tests locally. Which in turn can also help us to maintain the runner that we have.

fruttasecca:
I'll update the docs with a section related to the integration tests. (or maybe tests in general)

cypress/fixtures/custom/pipelines/data-passing/B.ipynb
```diff
@@ -0,0 +1,57 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
```
yannickperrenet, line 4: Wouldn't it be easier to use Python scripts here instead of Notebooks? Or are we particularly testing for Notebooks? Keen to hear your thoughts.

cypress/integration/examples/actions.spec.ts
```diff
@@ -0,0 +1,313 @@
+context("Actions", () => {
+  beforeEach(() => {
+    cy.visit("https://example.cypress.io/commands/actions");
+  });
```
yannickperrenet, line 4: When it comes the `examples` directory, can't we delete it?

docs/source/developer_guide/testing.rst
```diff
@@ -0,0 +1,98 @@
+.. _testing:
+
+Testing
+=======
```
yannickperrenet, line 4: As mentioned in https://github.com/orchest/orchest/pull/384#issuecomment-903853423 the docs will be revised. Just mentioning here so that we can have a dedicated thread.

yannickperrenet: I think we should add something to the "development workflow" docs that either links to these "testing" section or notes how they can be run.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,61 @@
...
+			[y]* ) break;;
+			[n]* ) exit;;
+			* ) echo "${msg}";;
+		esac
+	done
+fi
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "${DIR}/../"
+
```
yannickperrenet, line 56: > In my local Docker environment I still had a `orchest-env...` image which made all the `beforeEach` fail

Should we also add something for this in this script? Would definitely make running the script a lot easier without having to first check the docs.

fruttasecca: Yes, the script could actually take care of deleting all environments so that this is taken care of and doesn't need to be referenced in the docs.

cypress/integration/examples/actions.spec.ts
fruttasecca: Yes, we can indeed delete it. We initially went for not deleting it, happy to delete it in hindsight.

scripts/run_integration_tests.sh
fruttasecca: Moreover, the script could actually check if orchest is running and reject if it's not.

yannickperrenet: > Moreover, the script could actually check if orchest is running and reject if it's not.

I like this addition. What if the script starts Orchest in case it is not running yet? Similarly it shouldn't reject if there is still an `orchest-env` instead it should just clean it up and run the tests.

cypress/integration/examples/actions.spec.ts
yannickperrenet: I don't see any reason to leave in all this code that isn't even ours.

cypress/fixtures/custom/pipelines/data-passing/B.ipynb
yannickperrenet: Let's start with Notebooks and if we end up changing them often we can revisit the decision and move to scripts :+1:

cypress/integration/file-system-interactions.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { PROJECTS_DIR, SAMPLE_PROJECT_NAMES } from "../support/common";
+
+describe("file system interactions", () => {
```
yannickperrenet, line 3: How about adding:
* Deleting a pipeline through the FS

@ricklamers This would also be a question for you, but why do we strictly enforce a naming convention on the project names if they are created through the UI but still "detect" FS projects that don't follow the convention:
1. Try to create a project named `hello world` through the UI --> ERROR
2. Try to create a project named `hello world` through the FS --> UI picks up and calls `git init` inside the detected project

cypress/integration/interactive-runs.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+  STEPS,
+  PIPELINES,
+  SAMPLE_PROJECT_NAMES,
+  SAMPLE_PIPELINE_NAMES,
+  SAMPLE_STEP_NAMES,
+  setStepParameters,
+} from "../support/common";
+
+describe("interactive runs", () => {
+  beforeEach(() => {
```
yannickperrenet, line 14: (Not saying we have to add all these tests in this PR, just so we can write them down somewhere.)

I think we are missing some tests around the interaction between interactive runs and environments. For example, what if a used environment is currently building. A bit more of the edge cases.

cypress/integration/pipelines.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    // pipeline.
+    let originalPath = `${PROJECTS_DIR}/${SAMPLE_PROJECT_NAMES.P1}/${expectedFile}`;
+    let copyPath = `${PROJECTS_DIR}/${SAMPLE_PROJECT_NAMES.P1}/copy.orchest`;
+    cy.exec(`cp ${originalPath} ${copyPath}`);
+    // Reload to force the discovery.
+    cy.visit("/pipelines");
+    cy.findAllByTestId(TEST_ID.PIPELINES_TABLE_ROW).should("have.length", 2);
+  });
+
+  it("deletes a pipeline through the fs", () => {
```
yannickperrenet, line 72: I see the pipeline FS interaction is here. Maybe we should change the names of the modules as I don't think it is clear that this test is here instead of in the other. ([reference](https://github.com/orchest/orchest/pull/384#discussion_r695489820))

cypress/integration/file-system-interactions.spec.ts
fruttasecca: >Deleting a pipeline through the FS

This is already being tested but I ended up placing it in the wrong file (`pipelines.spec.ts`), will move it.

>Moving a project through the FS (essentially just renaming it)

Will add

>why do we strictly enforce a naming convention on the project names if they are created through the UI but still "detect" FS projects that don't follow the convention

This is a good point,  we should make an issue out of it, as this PR is very large already.

yannickperrenet: The issue: https://github.com/orchest/orchest/issues/441

cypress/integration/interactive-runs.spec.ts
fruttasecca: I have added this case to our internal documentation.

docs/source/developer_guide/testing.rst
fruttasecca: I have updated the testing docs, given the update to the `run_integration_tests.sh` script I haven't found a lot to add, let me know if anything comes up to you.

yannickperrenet (approved):

yannickperrenet:
LGTM!

PR #390, opened 2021-07-16
Add Testing Guidelines

joe-bell:
## Description

Add testing docs/guidelines

### Checklist

- [ ] In case I changed code in the `orchest-sdk`, I updated its version according to [SemVer](https://semver.org/) in its `_version.py` and updated the version compability table in its `README.md`

### DIFF
--- a/docs/source/developer_guide/testing.rst
+++ b/docs/source/developer_guide/testing.rst
@@ -1,0 +1,75 @@
+.. _testing:
+
+Testing
+=======
+
+We use `Cypress <http://cypress.io/>`__ for all things testing – a key part of our GitHub Action-based continuous integration.
+
+CLI
+---
+
+Run the `Test Runner <https://www.cypress.io/features>`__ against `localhost:8000 <https://localhost:8000>`__ (regardless of production or development mode)
+
+::
+
+    pnpm run cy:open
+
+Commands
+--------
+
+Orchest includes a selection of `custom commands <https://docs.cypress.io/api/cypress-api/custom-commands#Parent-Commands>`__ to ease the writing of tests:
+
+-  ``cy.setOnboardingCompleted(<value>)``
+   override the local storage value for whether or not to display the Onboarding Dialog.
+-  ``cy.getOnboardingCompleted()``
+   get the local storage value for whether or not the Onboarding Dialog has been completed.
+
+As well as: 
+
+- `Cypress LocalStorage Commands <https://github.com/javierbrea/cypress-localstorage-commands>`__ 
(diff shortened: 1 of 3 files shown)


docs/source/developer_guide/testing.rst
```diff
@@ -0,0 +1,75 @@
...
+    cy.findByTestId("some-test-id").should("exist");
+
+Grouping
+~~~~~~~~
+
+Organizing tests can be tricky, but it doesn't have to be - think of it like a tree:
+
+-  Wrap your entire specification with a ``describe()``.
+-  Group similar behaviours with ``context()``.
+
```
fruttasecca, line 59: It might be useful to have a couple more lines in here explaining the way we are organizing tests. The current explanation doesn't give me any gain in terms of concepts/insights.

docs/source/developer_guide/testing.rst
```diff
@@ -0,0 +1,92 @@
...
+   * Use hooks to run `shared checks and cleanup
+     <https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Hooks>`_.
+   * Write each test in an ``it()``.
+
+If you're still stuck, see `onboarding.spec.ts
+<https://github.com/orchest/orchest/blob/master/cypress/integration/onboarding.spec.ts>`_.
+
+Slowing Things Down
+"""""""""""""""""""
```
yannickperrenet, line 82: Would consider this a subsection of the styleguide, instead rather a section at the same level.

docs/source/developer_guide/testing.rst
joe-bell: Potentially – I was just hesitant about duplicating too much content from the Cypress docs (I tried adding links to those code blocks but wasn't able to with RST)

Is there anything in particular you could suggest that would benefit you?

fruttasecca: >I was just hesitant about duplicating too much content from the Cypress docs

Understandable

>Is there anything in particular you could suggest that would benefit you?
What comes to mind is that it says "like a tree" but after reading through the section I have a very vague idea of how this tree looks like. Fine to leave it like this if it's easily understood by looking at the tests or cypress docs.

joe-bell: Let me know if this sounds better!
https://github.com/orchest/orchest/pull/390/commits/1a889aef86837dd9e4eb463234227a1f6d9e6373

fruttasecca: Yep that helps, thank you :).

PR #394, opened 2021-07-16
Improve analytics data structure

yannickperrenet:
## Description

This PR introduces changes to our analytics module so that all data passes through the same logic, ensuring that we can easily work with the data thanks to the common structure. Additionally, the new implementation provides (inside the `_Anonymizer` class) an extensive overview of all the events that we support.

Structure always in the following format:
```
        {
            "event_properties": ...  # anonymized event properties
            "derived_properties": ...  # derived from removed properties
            "app_properties": ...  # Orchest application properties
        }
```

### How to test
1. `./orchest start --log-level=DEBUG`
2. Trigger some events
3. Check the `docker logs orchest-webserver` and note entries such as `Successfully sent analytics event 'session start'`.

In addition I checked the PostHog data to make sure entries are created.

### DIFF
--- a/services/orchest-webserver/app/app/analytics.py
+++ b/services/orchest-webserver/app/app/analytics.py
@@ -1,6 +1,6 @@
 import copy
 import logging
-import os
+import os  # noqa: F401
 import time
 import uuid
 from enum import Enum
--- a/services/orchest-webserver/app/app/analytics.py
+++ b/services/orchest-webserver/app/app/analytics.py
@@ -51,6 +51,7 @@ def send_event(
         app: The Flask application that received the event.
         event_name: The name of the event. Events should be named like:
             "[noun] [verb]", such as "movie played" or "movie updated".
+            Regex for noun: [a-z\-]+, e.g. "pipeline-run".
         event_properties: Any information that describes the event. This
             information will be anonimzed and send to the telemetry
             service.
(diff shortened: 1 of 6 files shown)


services/orchest-webserver/app/app/analytics.py
```diff
@@ -119,142 +340,20 @@ def send_anonymized_pipeline_definition(app, pipeline):
         step.pop("file_path")
         step_parameters_count += len(step.pop("parameters", {}))
 
-        env = step.get("environment", "")
-        if len(env):
+        env = step.get("environment")
+        if env is not None:
```
fruttasecca, line 344: Not sure if intended, but note that a step with no defined environment will have a step equal to `""`.

fruttasecca (approved):
LGTM, a nitpick: using an enum to formalize what events are supported would make it more clear and less error-prone given that you don't end up writing the actual event name when sending events, which would abstract away the need for bubbling up requirements like what an event name should look like. 

Much needed streamlining of the analytics module!

services/orchest-webserver/app/app/analytics.py
yannickperrenet: Great catch. Fixed.

yannickperrenet:
> using an enum to formalize what events are supported

I think this is great feedback. The top-level `send_event` function would become something like:
```python
analytics.send_event(
  app,
  analytics.EventName.JOB_CREATE,
 {},
)
```
Did I understand that correctly?

fruttasecca:
>Did I understand that correctly?

Yes

About the exact pattern (`analytics.EventName.JOB_CREATE`), I have seen different flavours:
- `analytics.JOB_CREATE`
- `analytics.Event.JOB_CREATE`
- `analytics.EventName.JOB_CREATE`

I don't feel strongly about anything in particular, just enumerating.

yannickperrenet:
The pre-commit `flake` hook fails with reason:

> F401 'os' imported but unused

But this is untrue seen that `os` is used [here](https://github.com/orchest/orchest/pull/394/files#diff-2954dbcbe7988e4447096015b9005b83085f14f7d4286ac33ffd0f4c8c92aec8R107).

fruttasecca:
lgtm

PR #404, opened 2021-07-29
Allow the user to control the level of job runs parallelism

fruttasecca:
## Description

Allows the user to control the level of parallelism of job runs through the settings. Once the settings are saved Orchest must be restarted for the change to take effect.

### Checklist

### DIFF
--- a/docs/source/user_guide/other.rst
+++ b/docs/source/user_guide/other.rst
@@ -18,20 +18,28 @@ Example content:
 .. code-block:: json
 
    {
-     "TELEMETRY_UUID": "69b40767-e315-4953-8a2b-355833e344b8",
+     "AUTH_ENABLED": false,
+     "MAX_JOB_RUNS_PARALLELISM": 1,
      "TELEMETRY_DISABLED": false,
-     "AUTH_ENABLED": false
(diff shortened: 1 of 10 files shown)


fruttasecca:
@ricklamers One last point is where should we mention the fact that a restart is required to have the change take effect (aside from the docs). Could be a single "static" sentence or something with more smartness that looks at what field have changed.

ricklamers:
> @ricklamers One last point is where should we mention the fact that a restart is required to have the change take effect (aside from the docs). Could be a single "static" sentence or something with more smartness that looks at what fields have changed.

It would be nice if we could in the frontend settings JSON editor detect the change and show a generic “your changes require a restart of Orchest”. That view also contains the restart button.

We’d add this config variable key to the list of configs that triggers it. I think there’s already a similar mechanism for AUTH_ENABLED. But that doesn’t require a restart, just a page reload. However, it should offer guidance as to how to detect old/new values (alleviating the need to reimplement that bit).

Can you implement it? I’ll double check the FE bits to be sure 👍🏻

ricklamers:
Other than the mention, LGTM! Nice job. Parallelism 🔥🚀🚀🚀

Issue #392, opened 2021-07-16
`orchest-ctl` does not report installation failure if the device is running out of space


fruttasecca:
**Describe the bug**  
Installation failure due to running out of disk space is silent.

**To Reproduce**  
Steps to reproduce the behaviour:

- have a disk that is running out of space
- `./orchest install`
- orchest reports that all services and images have been installed
- `./orchest start` reports that some images are missing, i.e. "are you sure Orchest is installed?"

**Expected behaviour**  
`orchest-ctl` should report that the installation process has failed

**Context**
This might be due to `aiodocker`.

yannickperrenet:
Multiple users have run into this issue so far, and the fact that this issue is hard to debug (especially for new users) indicates that we need a solution.

Possible solutions:
* Have the `orchest` shell script or the `orchest-ctl` (if possible since it is a Docker container) check the available disk size and report if it does not suffice to install Orchest.
* Have the installation process error on invalid disk space.

Additionally we could:
* Mention something like "At least 20GB disk size available" in the main README under requirements (in the installation section)

yannickperrenet:
Since we are now nudging the user towards disk size availability on installation failure I no longer consider this a bug, but instead an improvement we can make over the already improved messaging.

yannickperrenet:
Not directly related, but related to available disk size (after successful installation): https://github.com/orchest/orchest/pull/463.

yannickperrenet:
With the improvements we have made so far to address this issue I think we can consider it solved, even though we have taken a different path to solve the issue.

PR #421, opened 2021-08-17
Improve messages for `orchest install`, `update` and `version`

yannickperrenet:
## Description

The improved messages should help the users debug issues when installing or updating Orchest. For example, when the user is running out of disk space and therefore the installation of Orchest does not complete successfully, we now prompt the user.

Secondly, the old proposed `orchest --help` does very little to verify the installation. Now we propose using `orchest version --ext` to do so, because it checks whether:
* all Orchest owned images are pulled
* all images run the same version of Orchest

Also a partial solution for #392. 
 
### Checklist

### DIFF
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -19,14 +19,7 @@ from typing import List, Optional, Set, Tuple
 
 import typer
 
-from app import spec, utils
-from app.config import (
-    ALLOWED_BUILD_FAILURES,
-    INTERNAL_ERR_MESSAGE,
-    ORCHEST_IMAGES,
-    ORCHEST_WEBSERVER_ADDRESS,
-    _on_start_images,
-)
+from app import config, spec, utils
 from app.debug import debug_dump, health_check
 from app.docker_wrapper import DockerWrapper, OrchestResourceManager
 
@@ -66,7 +59,22 @@ class OrchestApp:
         utils.echo(
             "Checking whether all containers are running the same version of Orchest."
         )
-        self.version(ext=True)
+        try:
+            self.version(ext=True)
+        except typer.Exit:
+            pass
+
+        pulled_images = self.resource_manager.get_images()
+        missing_images = set(req_images) - set(pulled_images)
+        if missing_images:
+            utils.echo("Installation was unsuccessful.")
+            utils.echo(
+                "Could not pull the specified set of images. Please"
+                " make sure you have enough available disk space."
+            )
+            raise typer.Exit(code=1)
+        else:
+            utils.echo("Installation was successful.")
 
     def update(self, mode=None, dev: bool = False):
         """Update Orchest.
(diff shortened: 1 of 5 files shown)


services/orchest-ctl/app/orchest.py
```diff
@@ -66,7 +59,22 @@ def install(self, language: str, gpu: bool = False):
...
-        self.version(ext=True)
+        try:
+            self.version(ext=True)
+        except typer.Exit:
+            pass
+
+        pulled_images = self.resource_manager.get_images()
+        missing_images = set(req_images) - set(pulled_images)
+        if missing_images:
+            utils.echo("Installation was unsuccessful.")
```
ricklamers, line 70: Any reason why we’re not printing the missing images itself. We do do it elsewhere.

services/orchest-ctl/app/orchest.py
```diff
@@ -159,17 +167,35 @@ def update(self, mode=None, dev: bool = False):
...
+        if version_exit_code == 2:
+            # Could be caused because we are currently in a release
+            # process causing different container versions to be up on
+            # DockerHub. Or the user did not have enough disk space
+            # available.
+            utils.echo("Update was unsuccessful.")
+            utils.echo(
+                "Orchest was unable to pull the latest"
+                " version of its currently pulled images. Please make sure you"
+                " have enough disk space available. Otherwise you can try updating"
```
ricklamers, line 188: “again in about 30 minutes” is a bit odd imo. Would prefer something like “try again later”.

ricklamers:
Some minor comments. Otherwise, lgtm! 🙌🏻

services/orchest-ctl/app/orchest.py
yannickperrenet: "Try again later" is a bit vague. I would rather give a better indication of what they can expect. I will make it a mix of both :+1:

services/orchest-ctl/app/orchest.py
yannickperrenet: I don't see the direct value of printing it, since it already prints the pulled images (thus we can deduce the missing once). But I do agree the user should be able to see it, so I will use the logging module (which the user can also read by running `./orchest -vv version --ext`)

services/orchest-ctl/app/orchest.py
```diff
@@ -402,20 +434,38 @@ def version(self, ext=False):
...
+            utils.echo("\torchest install")
+            raise typer.Exit(code=1)
+        elif len(stdout_values) > 1:
+            # If not all versions are the same.
             utils.echo(
                 "Not all containers are running on the same version of Orchest, which"
                 " can lead to the application crashing. You can fix this by running:",
             )
             utils.echo("\torchest update")
             utils.echo("This should get all containers on the same version again.")
```
fruttasecca, line 467: Could we log the values for debugging purposes?

fruttasecca (approved):
lgtm

services/orchest-ctl/app/orchest.py
yannickperrenet: I don't see what this brings us given that the version values will have been echo'd to STDOUT already and that the `orchest-ctl` runs with the `--rm` flag (and so the logs will not be accessible after the `orchest-ctl` has run).

Let me know whether that changes your mind.

fruttasecca: You are right, disregard my comment.

PR #424, opened 2021-08-19
Cron job pausing

fruttasecca:
## Description

This PR introduces cron jobs pausing. A running cronjob can be paused, and a paused cronjob can be resumed. Runs of a cronjob that have already been scheduled (PENDING, STARTED) will continue running when the cron job is paused. 

### Checklist

### UI changes

image

image

image

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_jobs.py
+++ b/services/orchest-api/app/app/apis/namespace_jobs.py
@@ -286,6 +286,48 @@ class JobDeletion(Resource):
             return {"message": "Job does not exist."}, 404
 
 
+@api.route("/cronjobs/pause/<string:job_uuid>")
+@api.param("job_uuid", "UUID of job")
+@api.response(404, "Job not found")
+class CronJobPause(Resource):
+    @api.doc("pause_cronjob")
+    @api.response(200, "Cron job paused")
+    def post(self, job_uuid):
+        """Pauses a cron job."""
+
+        try:
+            with TwoPhaseExecutor(db.session) as tpe:
+                could_pause = PauseCronJob(tpe).transaction(job_uuid)
+        except Exception as e:
+            return {"message": str(e)}, 500
+
+        if could_pause:
+            return {"message": "Cron job pausing was successful."}, 200
+        else:
+            return {"message": "Could not pause cron job."}, 409
+
+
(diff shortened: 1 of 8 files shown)


services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -286,6 +286,48 @@ def delete(self, job_uuid):
             return {"message": "Job does not exist."}, 404
 
 
+@api.route("/cronjobs/pause/<string:job_uuid>")
+@api.param("job_uuid", "UUID of job")
+@api.response(404, "Job not found")
+class CronJobPause(Resource):
```
yannickperrenet, line 292: Why did you go for creating additional endpoints instead of sending a PUT request to the jobs endpoint? Just want to understand your thoughts here :)

yannickperrenet (approved):
LGTM! Nice one :+1:

ricklamers:
This PR is _insanely great_. Awesome job.

I'm also interested in the reasoning behind POST v.s. incorporating it in a PUT. But I must say it's kind of a toss-up here.

ricklamers (approved):

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: The hand-wavy reasoning is that it felt more explicit and declarative to POST an "action" rather than PUT something that isn't actually an attribute. This (PUTting a non-attribute) actually happens for the `confirm_draft` argument of `UpdateJob`, I preferred to not make things worse by having something like `pause_cronjob` and `resume_cronjob` in the PUT endpoint, it would make `UpdateJob` bigger and with more edge cases, which is not something desirable especially for a `2PF`. As (if) the number of "actions" grows, `UpdateJob` would become harder to maintain.

Nevertheless, as @ricklamers  says, this isn't too far from a toss-up. I kept second-guessing myself on this, so definitively open for debate (although I am quite happy about this functionality having its own easy to read & maintain little endpoints).

yannickperrenet: Sounds perfectly reasonable to me. Thanks for writing this up <3

PR #430, opened 2021-08-20
Integrate algolia search into docs

yannickperrenet:
**NOTE:** does not work yet!

## Description

Powers the documentations search by algolia.

## Why it did not work -> SOLVED
We use the `press` theme for our docs ([reference](https://github.com/schettino72/sphinx_press_theme)), however in their `layout.html` they have
```html
      <script src="{{ pathto('_static/theme.js', 1)}}" defer></script>
```
which prohibits the search from working. If you comment this line out, then the search works.

To, locally, comment this line out you can:
1. `python -m site` to find your site packages
2. open `sphinx_press_theme/layout.html`

> which prohibits the search from working

Why and how exactly, I do not know...

### DIFF
--- a/docs/source/_templates/searchbox.html
+++ b/docs/source/_templates/searchbox.html
@@ -1,32 +1,10 @@
-<link
-  rel="stylesheet"
-  href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css"
-/>
-
 <div id="searchbox" class="searchbox" role="search">
   <div class="caption">
     <span class="caption-text">{{ _('Quick search') }}</span>
     <div class="searchformwrapper">
       <form class="search" action="{{ pathto('search') }}" method="get">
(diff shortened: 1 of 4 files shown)


docs/source/_templates/searchbox.html
```diff
@@ -0,0 +1,32 @@
...
+    <div class="searchformwrapper">
+      <form class="search" action="{{ pathto('search') }}" method="get">
+        <input id="doc_search" type="text" name="q" />
+        <input type="submit" value="{{ _('Search') }}" />
+        <input type="hidden" name="check_keywords" value="yes" />
+        <input type="hidden" name="area" value="default" />
+      </form>
+    </div>
+  </div>
+</div>
```
yannickperrenet, line 18: Once the functionality is there, this should be restyled. The custom css that we have added for the docs is probably also no longer need since no "Search" button is required anymore once docsearch is integrated.

Issue #144, opened 2021-02-08
Allow duplication of existing job


ricklamers:
Enable duplicating an existing job entry. This will make it easier to re-run jobs, especially in the one-off running context.

We will have to take into account whether someone wants to run the same job configuration (e.g. the parameters/environment variable configuration) for the snapshotted version of the job that is being duplicated or the current version of the project.

If the current version is being used we need to reconcile the potential change in parameters/environment variables.

Arguably for re-running a job of the current snapshot we should extend the existing job (thus adding to the pipeline runs that previously ran as part of this job).

yannickperrenet:
When implementing this, I think we should reason about the potential abstractions it would introduce for #200.

ricklamers:
https://github.com/orchest/orchest/pull/433

ricklamers:
> Arguably for re-running a job of the current snapshot we should extend the existing job (thus adding to the pipeline runs that previously ran as part of this job).

For completeness, this was dropped from the first implementation spec.

Issue #432, opened 2021-08-20
Allow editing the name of a job in the job view.


fruttasecca:
Changing the name of a job should be possible in the job view, and not only in edit mode.

**Additional context**  
This would improve consistency given it's possible to change the name of a pipeline through its settings, and not only in the list of pipelines in `pipelines`. Moreover, this makes sense in the context of #144.

Issue #391, opened 2021-07-16
API - client response/feedback consistency


fruttasecca:
We haven't a consistent and pre-defined way of communicating feedback between the APIs (webserver, orchest-api) and the client.
Some behaviour that is in the code:
- communication through HTTP error codes
- when there aren't enough HTTP error codes, communication through custom codes
- communication through messages
- bubbling up the API error message to the user

We need a predefined way with pre-defined responsibilities, especially when it comes to error messages.

yannickperrenet:
Additional context can be found in [this PR comment](https://github.com/orchest/orchest/pull/382#discussion_r668510240).

yannickperrenet:
A good example is the verification that is needed around the pipeline definition. Multiple orchest services use the pipeline definition, e.g. the `orchest-webserver` to display the pipeline editor, the `orchest-api` to run pipelines and the `orchest-sdk` to transfer data. The question becomes, where should the verification of the pipeline definition take place?
* Should the `orchest-webserver` always make sure that the pipeline definition it passes is valid?
* Should the `orchest-api` verify the definition when it is passed one through an API call?

No matter where the verification takes place, the user should always get informative messages (if anything is wrong).

yannickperrenet:
Some ideas we have been having.

1. Communicate using error codes
2. If the error codes for certain responses are equal, then use the JSON structure that is returned as well

When it comes to the JSON structure we can use a common format for all responses. This can be something like:
```
{
  # A enum-like reason where the API spec notes the possibilities
  "reason": "name_too_long",
  # A sentence explanation of the reason
  "msg": "The given name was too long to be ...",
  # Some data that we want to pass along 
  "data": { ... }
}
```

PR #433, opened 2021-08-20
Job Duplication

fruttasecca:
## Description

This PR introduces job duplication, which allows duplicating a job, creating a job with the latest code and environments but inheriting the settings of the job that is being duplicated.

The environment variables of the new job are resolved by merging the current project and pipelines environment variables, then adding the inherited job environment variables. Note that this implies that environment variables that were "inherited" by the job that is being duplicated from project/env variables cannot be distinguished from environment variables that were defined in the job. This might be of interest given that if those project/pipeline variables do not exist anymore it might make more sense to not include them in the newly defined job.

Job parameters are resolved by:
 - removing parameters that do not exist anymore.
 - adding new parameters, i.e. parameters that exist in the latest version of the pipeline but do not exist for the job that is being duplicated.
  - parameters that are not to be removed nor are new, i.e. parameters that exist both for the latest pipeline definition and the job use the values that were defined in the job that is being duplicated.
  - this is true for both the strategy json and the job parameters, where each job parameter effectively represents the parameterization of a run.
  - the job parameters are updated in a way that, from the client POV and given the logic in the job view, preserves the runs selection and their ordering, along with the constraint that every run parameterization should be unique.

If the job that is being duplicated is a cronjob, the new job will have its schedule.

Fixes: #144 (partially) 

### Checklist

### Things that need to be discussed further:
- where in the UI should job duplication happen? It's currently done by a button in the `JobView`, which implies that only non `DRAFT` jobs can be duplicated (although it's technically possible to duplicate `DRAFT` jobs)
- env variables inheritance (see the env variables paragraph)

Current UI changes
image

### DIFF
--- a/services/orchest-webserver/app/app/core/jobs.py
+++ b/services/orchest-webserver/app/app/core/jobs.py
@@ -1,0 +1,300 @@
+import copy
+import json
+import uuid
+from collections import defaultdict
+
+import requests
+from flask import current_app
+
+from _orchest.internals import config as _config
+from app import error
+from app.models import Pipeline, Project
+from app.utils import (
+    create_job_directory,
+    get_environments,
+    get_environments_from_pipeline_json,
+    get_pipeline_json,
+    get_project_directory,
+    pipeline_uuid_to_path,
+)
+
+
+def create_job_spec(config) -> dict:
+    """Returns a job spec given the provided configuration.
+
+    Args: Initial configuration with which the job spec should be built.
+        project_uuid, pipeline_uuid, pipeline_run_spec, pipeline_name,
+        name are required. Optional entries such as env_variables can be
+        used to further customize the initial state of the newly created
+        job.
+
+    Returns:
+        A job spec that can be POSTED to the orchest-api to create a job
+        that is a duplicate of the job identified by the provided
+        job_uuid.
+    """
+    job_spec = copy.deepcopy(config)
+    pipeline_path = pipeline_uuid_to_path(
+        job_spec["pipeline_uuid"], job_spec["project_uuid"]
+    )
+    job_spec["pipeline_run_spec"]["run_config"] = {
+        "host_user_dir": current_app.config["HOST_USER_DIR"],
+        "project_dir": get_project_directory(job_spec["project_uuid"], host_path=True),
+        "pipeline_path": pipeline_path,
+    }
+
+    job_spec["pipeline_definition"] = get_pipeline_json(
+        job_spec["pipeline_uuid"], job_spec["project_uuid"]
+    )
+
+    # Validate whether the pipeline contains environments
+    # that do not exist in the project.
+    project_environments = get_environments(job_spec["project_uuid"])
+    project_environment_uuids = set(
+        [environment.uuid for environment in project_environments]
+    )
+    pipeline_environment_uuids = get_environments_from_pipeline_json(
+        job_spec["pipeline_definition"]
+    )
+
+    missing_environment_uuids = pipeline_environment_uuids - project_environment_uuids
+    if len(missing_environment_uuids) > 0:
+        raise error.EnvironmentsDoNotExist(missing_environment_uuids)
+
+    # Jobs should always have eviction enabled.
+    job_spec["pipeline_definition"]["settings"]["auto_eviction"] = True
+
+    job_uuid = str(uuid.uuid4())
+    job_spec["uuid"] = job_uuid
+    create_job_directory(job_uuid, job_spec["pipeline_uuid"], job_spec["project_uuid"])
+    return job_spec
+
+
+def duplicate_job_spec(job_uuid: str) -> dict:
+    """Returns a job spec to duplicate the provided job.
+
+    Args:
+        job_uuid: UUID of the job to duplicate.
+
+    The project, pipeline, name, schedule, env variables and parameters
+    of the "parent" job are inherited. Env variables are resolved
+    according to the following expression:
+        job_spec["env_variables"] = {
+            **current_project_env_vars,
+            **current_pipeline_env_vars,
+            **parent_env_vars}
+
+    Paremeters are resolved by:
+        - removing parameters that do not exist anymore.
+        - adding new parameters, i.e. parameters that exist in the
+          latest version of the pipeline but do not exist for the job
+          that is being duplicated.
+        - parameters that are not to be removed nor are new, i.e.
+          parameters that exist both for the latest pipeline definition
+          and the job use the values that were defined in the job that
+          is being duplicated.
+        - this is true for both the strategy json and the job
+          parameters, where each job parameter effectively represents
+          the parameterization of a run.
+        - the job parameters are updated in a way that, from the client
+          POV and given the logic in the job view, preserves the runs
+          selection and their ordering, along with the constraint that
+          every run parameterization should be unique.
+
+    Returns:
+        A job spec that can be POSTED to the orchest-api to create a job
+        that is a duplicate of the job identified by the provided
+        job_uuid.
+    """
+    resp = requests.get(
+        f'http://{current_app.config["ORCHEST_API_ADDRESS"]}/api/jobs/{job_uuid}'
+    )
+    if resp.status_code == 404:
+        raise error.JobDoesNotExist()
+
+    parent_job = resp.json()
+    job_spec = {}
+    job_spec["draft"] = True
+    job_spec["name"] = parent_job["name"]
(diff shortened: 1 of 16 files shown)


yannickperrenet:
First off, I like the decisions you made (and explained in the PR message) around the parameters/env vars to choose on overlap etc. So well done :)

I think we can make an additional improvement where "duplicate job" duplicates the job (with this new implementation) but also renames the job to something like "Duplicate of <name>" (similar to how Google Drive does it).

services/orchest-webserver/client/src/views/JobView.tsx
```diff
@@ -448,6 +512,13 @@ const JobView: React.FC<TViewProps> = (props) => {
             icon="arrow_back"
             onClick={returnToJobs.bind(this)}
           />
+          <span className="float-right">
+            <MDCButtonReact
+              label="Duplicate job"
```
yannickperrenet, line 517: Could the label "Duplicate job" be confusing to the users? Depending on how someone interprets "duplicating" it might or might not do what the person expects. It is good to keep in mind here that a user doesn't necessarily know that the only the job configuration (whatever that may mean) is duplicated, but new code and new environments are used.

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -175,75 +172,101 @@ def catch_api_proxy_jupyter_builds_most_recent():
...
-        json_obj = request.json
+        try:
+            job_spec = jobs.create_job_spec(request.json)
+        except error.EnvironmentsDoNotExist as e:
+            return (
+                jsonify(
+                    {
+                        "message": environments_missing_msg.format(
+                            missing_environment_uuids=[",".join(e.environment_uuids)]
+                        ),
```
yannickperrenet, line 195: Wouldn't it be more helpful if we add the environment names to the message as well?

This would require a change in the `create_job_spec` method.

services/orchest-webserver/app/app/core/jobs.py
```diff
@@ -0,0 +1,300 @@
...
+    job_spec["pipeline_name"] = parent_job["pipeline_name"]
+    job_spec["pipeline_run_spec"] = {"run_type": "full", "uuids": []}
+
+    if (
+        Project.query.filter_by(
+            uuid=job_spec["project_uuid"],
+        ).one_or_none()
+        is None
+    ):
+        raise error.ProjectDoesNotExist()
```
yannickperrenet, line 131: I think it is indeed good to add this case even though with the current implementation the user can't duplicate the job if the containing project was first deleted. :+1:

services/orchest-webserver/app/app/core/jobs.py
```diff
@@ -0,0 +1,300 @@
...
+    get_environments,
+    get_environments_from_pipeline_json,
+    get_pipeline_json,
+    get_project_directory,
+    pipeline_uuid_to_path,
+)
+
+
+def create_job_spec(config) -> dict:
+    """Returns a job spec given the provided configuration.
```
yannickperrenet, line 23: Nitpick: "Returns a job spec based on the provided configuration"

services/orchest-webserver/app/app/core/jobs.py
```diff
@@ -0,0 +1,300 @@
...
+    # and adding env variables that were not part of the project and
+    # pipeline env vars. NOTE: we currently have no way to discern
+    # old env variables that the job inherited from project/pipeline
+    # env variables. If we could, old project/pipeline env vars that
+    # do not exist anymore could be discarded.
+    job_spec["env_variables"] = {
+        **project_env_vars,
+        **pipeline_env_vars,
+        **parent_env_vars,
+    }
```
yannickperrenet, line 164: We should probably preface this with a `TODO` comment for once we move to Python3.9, since that includes https://www.python.org/dev/peps/pep-0584.

services/orchest-webserver/app/app/views/orchest_api.py
ricklamers: Not sure if the information is present in that context. As those names can change maybe we can resolve the names on the client (to be the actuals) instead of having a layer of "materialized names/cached names".

services/orchest-webserver/app/app/core/jobs.py
ricklamers: Although the new syntax can be considered an improvement. I don't think a TODO is warranted. Thanks for keeping up recent on the Python language @yannickperrenet!

ricklamers:
LGTM! I think it's good if we can exhaustively test the PR for edge cases (it seems quite tricky to test). I think we can add it to Cypress to add basic coverage of the high level UI actions that you can take. As a first suggestion:

Duplicating a job where there are changes between the old and new parameters (cases: parameter newly added, already exists but different default, parameter removed).

So far I couldn't find any issues. Great job :fire:!

The e2e Cypress tests can be added later (can be a separate PR imo).

services/orchest-webserver/app/app/core/jobs.py
ricklamers: P.S. Slightly out of scope I think for this PR.

services/orchest-webserver/app/app/views/orchest_api.py
ricklamers: Slightly out of scope I think for this PR.

services/orchest-webserver/client/src/views/JobView.tsx
ricklamers: Added this in a commit.

ricklamers:
My last commit solves #432 by the way.

services/orchest-webserver/client/src/views/JobView.tsx
```diff
@@ -389,6 +395,64 @@ const JobView: React.FC<TViewProps> = (props) => {
...
+              } catch (error) {
+                console.log(error);
+              }
+            }
+          });
+      })
+      .catch((result) => {
+        if (result.reason === "gate-failed") {
+          setTimeout(() => {
+            orchest.requestBuild(
```
ricklamers, line 443: Not sure why this requires the timeout.

fruttasecca: That's just me misinterpreting the existing FE logic :D.

services/orchest-webserver/app/app/core/jobs.py
fruttasecca: Indeed, just a layer of "safety" for very improbable race conditions.

services/orchest-webserver/app/app/views/orchest_api.py
fruttasecca: >Wouldn't it be more helpful if we add the environment names to the message as well?

I agree. I think this ties (once again) into #391. Let's keep this use case/issue in mind.

fruttasecca:
@ricklamers 

About the tests, what we really want to test imo is `duplicate_job_spec` in `services/orchest-webserver/app/app/core/jobs.py `. It might be less costly to do that through the non-integration tests. Although the integration tests will be "better" since they are testing the real thing.

>The e2e Cypress tests can be added later (can be a separate PR imo).

I agree with that. If we are ok with adding some non-integration tests to this PR those might take a fraction of the time and allow us to sleep better.

yannickperrenet:
> If we are ok with adding some non-integration tests to this PR

I like this idea. The fact that we are about to introduce integration tests doesn't mean we shouldn't ever write unit tests any more. And this test could indeed be a well scoped unit test.

services/orchest-webserver/app/app/core/jobs.py
yannickperrenet: Works for me.

services/orchest-webserver/app/tests/test_jobs.py
```diff
@@ -0,0 +1,433 @@
...
+
+def _get_parent_job_spec(
+    project_uuid="project-uuid",
+    pipeline_uuid="pipeline-uuid",
+    name="job-name",
+    schedule=None,
+    pipeline_name="pipeline-name",
+    env_variables={},
+    strategy_json={},
+    parameters=[{}],
```
yannickperrenet, line 21: In Python you should not create mutable default arguments ([some explanation](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) together with "correct" alternative).

fruttasecca: Yep, my bad, will fix.

yannickperrenet (approved):

Issue #454, opened 2021-09-01
Kernel containers started through Jupyter EG are not reliably cleaned up on session stop


fruttasecca:
**Describe the bug**  
The kernel containers do not get reliably cleaned up by Jupiter EG since on session stop we kill the container running EG.

**To Reproduce**  
1. Make an environment where the `orchest/bootscript.sh` had its content replaced with `sleep 1000`
2. Start a kernel from said environment by opening a step with that environemnt in jupyterlab
3.  Stop the session. 
4. The container will be running.

**Additional context**  
This does not seem to be the case if the notebook contains a long-running cell. Needs to be investigated further.

Issue #452, opened 2021-08-31
Clean up containers in prod mode when EG fails to spawn a kernel


ricklamers:
**Describe the solution you'd like**  
Clean up containers in prod mode when EG fails to spawn a Docker kernel.

**What does your solution aim to solve?**  
Make sure old environment images can be removed by the build process to save disk space.

Issue #352, opened 2021-06-24
Should be clear to users that an environment build requires Jupyter kernel restart


yannickperrenet:
**Steps to reproduce**  
1. Import a non-installed package, e.g. `import typer`, inside your Jupyter kernel.
2. Go to environments and `pip install typer` and build
3. Go back to your pipeline and do an interactive run --> Works
4. Go to your open jupyter file and try to run the cell --> Fails

**Description of solution**  
I don't think the above behavior should be fixed, I am just not sure that this behavior is clear to the user. Writing this down in the docs only reaches a small portion of users, instead the behavior should be explainable on the basis of the concepts in Orchest.

Does a user know / Can a user reason that the above issue can be fixed by a kernel restart?

yannickperrenet:
One way of solving this (but its probably too noisy) is to have a dialog whenever a user builds an environment that is actively in use by kernels.

yannickperrenet:
Now that #839 is about to get merged (fixing #141), the kernel (re)start becomes more "powerful". However, from a user perspective there is nothing that indicates its consequences. This needs to be clearer!

PR #455, opened 2021-09-02
Fix/session environment images handling

fruttasecca:
## Description

This PR makes the following changes: 

- sessions will now try to cleanup dangling images on shutdown, so that any dangling image that could not be deleted because of a session can be deleted asap, thus freeing the disk without waiting for the next time said environment will be built.
- sessions will take care of killing kernel containers on shutdown. This was previously a responsibility of Jupyter EG, but doing it ourselves ensures that the containers are actually killed if they fail to start.
- image locking for a session happens for all images that could potentially be used, i.e. it's not limited to environments used by services as base images anymore but also to step environments. This fixes an issue where a user would not get a prompt before deleting an environment potentially in use by a session (e.g. a kernel), leading to the environment being deleted only at the database level, but actually still existing in the docker environment.
- whenever a new build succeeds for an environment, the newly produced image is locked for existing interactive sessions making use of the same environment. This is because the newly built image might get used by a kernel that gets started after the build.

Fixes: #454, #452

### Checklist

During the implementation of this PR an issue between `APScheduler` and `eventlet` has been found. It looks like the scheduler does not get control as long as there is an ongoing call to the flask webserver. This might not be a problem normally, given that the server is idle most of the time (and thus the scheduler actually gets control), but it's a problem if we do any sort of waiting or sleeping in an API call (which we do) which results relies on an effect of a task started by the scheduler (we rely on this once, in the `DeleteImage` `2PF`). Example of what happen(ed) in one endpoint (used to remove an image):
- delete image is issued
- we stop all sessions using the image, the actual removal of the containers happens in a background task which will be run by the scheduler
- we delete the image. Given that other entities that rely on the image might still be stopping/aborting (like pipeline runs), we try N times to delete it, assuming that eventually said containers will shut down.
- the containers that should supposedly be shut down when shutting down a session will not shut down until we have made all N tries, only then the webserver will be idle and the scheduler will get control. In this particular case, the containers that are causing problems are kernel containers, the ones in use by interactive runs or jobs have already shut down.
- the scheduler gets control, kils the containers
- the env image still exists

See https://github.com/orchest/orchest/pull/455/commits/a9c007bf1b4aa69d31b36b0218ebddbc005a3212 for the current workaround.


services/orchest-api/app/app/apis/namespace_environment_builds.py
```diff
@@ -134,6 +138,34 @@ def put(self, environment_build_uuid):
...
+
+                pipeline_uuids = set()
+                for mapping in existing_env_mappings:
+                    pipeline_uuids.add(mapping.pipeline_uuid)
+                for pipeline_uuid in pipeline_uuids:
+                    lock_environment_images_for_session(
+                        env_build.project_uuid,
+                        pipeline_uuid,
+                        set([env_build.environment_uuid]),
+                    )
```
yannickperrenet, line 167: Just to make sure; the old environment builds are freed (in another function call, not this one) if possible (not in active use by e.g. a job) right?

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -185,8 +139,18 @@ def _collateral(self, project_uuid: str, environment_uuid: str):
...
+            # where the aborted runs are still using the image. Moreover
+            # , session stopping is executed by a background thread in
+            # the scheduler, so kernel containers might still be
+            # running, the time.sleep() in this function does not seem
+            # to pass control back to the scheduler thread, so we are
+            # forced to have this code run in another process.
+            docker_images_rm_safe(docker_client, image_name)
+            # Make the process kill itself so that no follow up code
+            # is called.
+            os.kill(os.getpid(), signal.SIGKILL)
```
yannickperrenet, line 156: Seems like a perfectly reasonable workaround to me :+1:

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -157,24 +157,18 @@ def _transaction(self, session_config: Dict[str, Any]):
...
-                session_config["project_uuid"], env_as_services
+                session_config["project_uuid"], set(session_config["environments"])
             )
         except errors.ImageNotFound as e:
             raise errors.ImageNotFound(
-                "Pipeline services were referencing environments for "
-                f"which an image does not exist, {e}."
+                "Pipeline is referencing environments for which an image does not "
+                f"exist, {e}."
             )
```
yannickperrenet, line 172: In the last PR I made I added an additional error to be (potentially) raised by `get_env_uuids_to_docker_id_mappings`, should we also catch that here (my bad for not adding it in my PR)?

To me it somewhat ties into #391. In some situations we are not even catching the errors raised by this function at all.

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: Not sure if I understand the quetion, but locking images for a session is done through a specific table `InteractiveSessionImageMappings` so this is orthogonal to other image lockings (e.g. for jobs).  What we are doing here is making sure that the locked images for a session actually reflect what images the session might be using, since depending on the timeline of environment builds and kernel starts a session might be using different images (docker ids) related to the same environment.

imo it would make sense to make it so that a session runs always with a specific environment image (docker id), meaning that regardless of new builds any newly started kernels should use the image that has been locked on session start (same for interactive runs). It would be consistent with jobs and make for an easier mental model.

services/orchest-api/app/app/apis/namespace_sessions.py
fruttasecca: Yep I will add that.

services/orchest-api/app/app/apis/namespace_environment_builds.py
yannickperrenet: Excuse my bad wording, but I meant to ask when environment images are removed. Is this on session kill or whenever the environment is build again (yielding a new and old image, thus making it possible to remove the old)? 

Can you make the second point you are making into an issue? Given that a kernel restart is required on environment build (see #352) it might make sense to require a session restart for new environments to be removed. However! Big but here! I think the value our pipeline editor brings is that you can do a lot of "hacky" stuff (just like you can run notebooks cells out of order) so it is super important how we end up choosing to implement which environment is used in a session (and when it can be changed).

yannickperrenet (approved):

services/orchest-api/app/app/apis/namespace_environment_builds.py
fruttasecca: >Excuse my bad wording, but I meant to ask when environment images are removed. Is this on session kill or whenever the environment is build again (yielding a new and old image, thus making it possible to remove the old)?

Session kill will try to delete dangling images. Whenever an environment is built again an attempt to clean up danglings images is made, but it might not go through since the nameless image is actually in use by a session, job, etc. The session attempting to cleanup dangling images is due to the fact that otherwise you would have to trigger a rebuid to cleanup dangling images that where locked by a session.

In a more abstract way, we have entities that lock environment images (docker id). A rebuild of an environment can make an environment image "old", and will attempt to remove the image, but will not do so if the image is in use by a running session/job etc. To make sure that the old image is deleted the moment it is not in use anymore, we make the entity attempt a cleanup the moment said entity stops using images (e.g. session shut down).

fruttasecca: >Can you make the second point you are making into an issue? Given that a kernel restart is required on environment build (see #352) it might make sense to require a session restart for new environments to be removed. However! Big but here! I think the value our pipeline editor brings is that you can do a lot of "hacky" stuff (just like you can run notebooks cells out of order) so it is super important how we end up choosing to implement which environment is used in a session (and when it can be changed).
>EDIT: I think it is easy for us to add more rules around how the system can be used, but we do really need to be careful not to destroy the beautiful life of users inside Orchest.

Absolutely agree, I'll make an issue out of the idea, and we can discuss this later.

ricklamers: Just pitching in my two cents here, I definitely don't think we should require a InteractiveSession restart to start using the new environment images for kernels or InteractivePipelineRuns. Updating your environment is a very common step during pipeline building and a session restart is more than annoying.

services/orchest-webserver/app/app/utils.py
```diff
@@ -316,11 +317,68 @@ def get_pipeline_json(pipeline_uuid, project_uuid):
...
+                step["kernel"].get("name", "python")
+            )
+            language_envs = envs_by_language.get(step_language, set())
+            language_envs_uuids = envs_uuids_by_language.get(step_language, set())
+
+            # The environment needs to be fixed.
+            if step["environment"] not in language_envs_uuids:
+                env_to_set = ""
+                if language_envs:
+                    env_to_set = next(iter(language_envs))
```
yannickperrenet, line 353: I don't understand the logic behind `next(iter(language_envs))` what exactly is this supposed to solve? It just picks the first element from the `language_envs` right, which could be any (but we don't care which one it picks)?

yannickperrenet: I guess this comment can be ignored as we will revert the entire commit that introduced this change.

PR #458, opened 2021-09-03
Fix integration tests hanging on Mac

iannbing:
## Description

When I run Cypress on Mac with scripts like `pnpm run cy:open`, the tests hang and then throw timeout because it couldn't find the test id. According to the [official doc](https://github.com/testing-library/cypress-testing-library#config-testidattribute),  the configuration can be done on level up in `cypress/support/index.ts` instead of inside of the test context. This solves my issue on Mac.

```
import { configure } from "@testing-library/cypress";

configure({ testIdAttribute: "data-test-id" });
```

### Checklist

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -60,10 +60,6 @@ declare global {
   }
 }
 
-before(() => {
-  cy.configureCypressTestingLibrary({ testIdAttribute: "data-test-id" });
-});
-
 Cypress.Commands.add("setOnboardingCompleted", (value: TBooleanString) => {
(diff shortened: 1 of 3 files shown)


ricklamers (approved):
Good catch on the `if (queryArgs.read_only !== "true")` clause. Looks like the old one was incorrect. I assume the wrong logic was recently introduced?

fruttasecca (approved):
lgtm :)

iannbing:
@ricklamers yes, it was caused by my recent PR (removing `==="true"`, you know the history 🙂).

PR #463, opened 2021-09-07
disk information in settings

fruttasecca:
## Description
This PR adds an entry to the settings view that shows the disk usage of the disk that docker is making use of.

image

### Checklist

### DIFF
--- a/services/orchest-webserver/app/app/views/views.py
+++ b/services/orchest-webserver/app/app/views/views.py
@@ -266,20 +266,24 @@ def register_views(app, db):
     @app.route("/async/host-info", methods=["GET"])
     def host_info():
         disk_info = subprocess.getoutput(
-            "df /config --output=used,avail,pcent | sed -n '2{p;q}'"
+            "df -BGB /config --output=size,avail,pcent | sed -n '2{p;q}'"
         )
         disk_info
 
-        used, avail, pcent = disk_info.strip().split()
-        # Incoming data is in KBs.
-        used = float(used) / 10 ** 6
-        avail = float(avail) / 10 ** 6
+        # Incoming data is in GB (-BGB)
+        size, avail, pcent = disk_info.strip().split()
+        # Remove the "GB"
+        avail = int(avail[:-2])
+        # Account for the 5% reserved root space, so that used + avail
+        # add up to the total disk size the user would see in a file
+        # explorer.
+        used = int(size[:-2]) - avail
         pcent = float(pcent.replace("%", ""))
 
         host_info = {
             "disk_info": {
-                "used_gbs": used,
-                "avail_gbs": avail,
+                "used_GB": used,
+                "avail_GB": avail,
                 "used_pcent": pcent,
             }
         }
--- a/services/orchest-webserver/app/app/views/views.py
+++ b/services/orchest-webserver/app/app/views/views.py
@@ -1,5 +1,6 @@
 import json
 import os
+import subprocess
 import uuid
 
 import docker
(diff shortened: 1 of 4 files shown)


services/orchest-webserver/app/app/views/views.py
```diff
@@ -262,6 +263,29 @@ def user_config():
 
         return current_config
 
+    @app.route("/async/host-info", methods=["GET"])
+    def host_info():
+        disk_info = subprocess.getoutput(
+            "df /config --output=used,avail,pcent | sed -n '2{p;q}'"
```
yannickperrenet, line 269: Didn't know this actually works inside a Docker container :+1:

yannickperrenet (approved):

services/orchest-webserver/app/app/views/views.py
fruttasecca: The documentation I could find is somewhat unclear and I think this is the result of how we are running the containers. I have had no chance of trying this on mac nor WSL2 so we should perhaps do that before merging.

yannickperrenet: @ricklamers Mister cross-platform can you verify :100: ?

ricklamers: > @ricklamers Mister cross-platform can you verify :100: ?

If Huang can verify on macOS. I think it should be fine. On both WSL2 and macOS the container just runs in a Linux VM. So I don't expect the df behavior to be different.

I'll test on WSL2.

yannickperrenet:
@iannbing  @ricklamers Can this be merged?

services/orchest-webserver/app/app/views/views.py
iannbing: It works fine on Mac

services/orchest-webserver/client/src/views/SettingsView.tsx
```diff
@@ -401,6 +401,61 @@ const SettingsView: React.FC<TViewProps> = () => {
...
+                      <span className="float-right">
+                        {state.hostInfo.disk_info.avail_gbs.toFixed(2) +
+                          " GB free"}
+                      </span>
+                    </div>
+                  </React.Fragment>
+                );
+              } else {
+                return (
+                  <React.Fragment>
```
iannbing, line 432: React.Fragment is not needed since there's only one child 🙂

services/orchest-webserver/client/src/views/SettingsView.tsx
```diff
@@ -401,6 +401,61 @@ const SettingsView: React.FC<TViewProps> = () => {
           </div>
           <div className="clear"></div>
         </div>
+        <div className="columns">
+          <div className="column">
+            <p>Disk usage.</p>
+          </div>
+          <div className="column">
+            {(() => {
```
iannbing, line 409: Normally in React JSX, ternary is preferred, like this:
```
<div className="column">
            {state.hostInfo !== undefined ? (
              <>
                <MDCLinearProgressReact
                  classNames={["disk-size-info"]}
                  progress={state.hostInfo.disk_info.used_pcent / 100}
                />
                <div className="disk-size-info">
                  <span>
                    {state.hostInfo.disk_info.used_gbs.toFixed(2) + " GB used"}
                  </span>
                  <span className="float-right">
                    {state.hostInfo.disk_info.avail_gbs.toFixed(2) + " GB free"}
                  </span>
                </div>
              </>
            ) : (
              <MDCLinearProgressReact
                classNames={["push-down", "disk-size-info"]}
              />
            )}
</div>
```

But since the instant triggering syntax is used in most of the codebase, it's fine to keep it consistent. We can clean up in later stages 🙂

services/orchest-webserver/app/app/views/views.py
ricklamers: Note: the `df` integer represents kilobits (1000 bits) so we need to convert properly if we want to report in GB = giga byte.

ricklamers:
LGTM. Minor point: there should be some spacing between the text below the bar. We could just add a "push-up" class.

ricklamers:
Great job! :rocket:

services/orchest-webserver/app/app/views/views.py
ricklamers: WSL2 works fine, with the caveat it will be reporting the virtual disk use/availability if Orchest is cloned in the distro's virtual disk. Which I think is fine to the extent it "works as advertised".

fruttasecca:
Updated look
image

PR #468, opened 2021-09-09
Apply react router

iannbing:
## Description

Apply [react-router](https://reactrouter.com/) to replace `window.orchest.loadView`.

With current `loadView` mechanism, we have following limitations:
- if a sub component has a reload-current-page behavior, we cannot extract this sub component (i.e. circular dependency, parent imports child and child imports parent, because child needs to reload parent using `loadView`). This is blocking us from breaking big components to small ones.
- unable to have an overview of all the pages in the app and their corresponding route. This indirectly leads to an inconsistent usage of querystring and props.
- ~~rely fully on queryArgs to initialize the component, while queryArgs is not always consistent amongst all components.~~

Usually, we would like to refactor in small parts, but two routing mechanisms co-exist in one app will more-likely introduce more bugs and unexpected behaviors. So, this PR will inevitably be a big bang.

Major changes:
- Make use of react-router-dom library as much as possible and reduce the amount of custom implementation regarding routing
- Make paths more semantic and configurable (see `routingConfig.tx`), so we don't create path based on component names.
- Make use of the route parameters to replace queryArgs. Most components are self-content by getting important data from the path (e.g. projectId). Thus, we could reduce the need to call BE in order to get data.
- Create one-off `useCustomRoute` hook to collect all route parameters, that is, we no longer need to check the parent, nor the caller of `loadView` to figure out what's been passed in.
- ProjectSelector is responsible for updating projectUuid in Orchest context when projectUuid is provided via URL. Therefore, we could remove all 'projectSet' action dispatcher in other places. 

To-Do:

- [x]  regression testing
- [x] update Cypress tests

### Checklist


lib/javascript/utils/src/untyped.ts
```diff
@@ -34,27 +34,3 @@ export class RefManager {
     });
   }
 }
-
```
iannbing, line 37: I added type definitions, so this class was moved to `typed.ts`

ricklamers:
Great cleanup!

> while queryArgs is not always consistent amongst all components.

Do you have an example of this kind of inconsistency?

iannbing:
I give a second look and realize why I had this impression (for the record, the usage of queryArgs is consistent, my apologies) 🙂
`props.project_uuid` and `props.queryArgs.project_uuid` both appear in the codebase, and it takes some time to understand that `props.queryArgs` is passed from `loadView` and should **only** be used in "view" components. However, this pattern is rather hidden and it's hard to hold when more contributors join forces.

services/orchest-webserver/client/src/components/PipelineList.tsx
```diff
@@ -45,45 +44,47 @@ const PipelineList: React.FC<any> = (props) => {
...
-        </span>,
-        <SessionToggleButton
-          project_uuid={context.state.project_uuid}
-          pipeline_uuid={pipeline.uuid}
-          switch={true}
-          className="consume-click"
-        />,
-      ]);
-    }
+    // @TODO Get the current Project on the Pipelines page
```
ricklamers, line 47: I guess this is old. @iannbing?

ricklamers (approved):

services/orchest-webserver/client/src/components/PipelineList.tsx
iannbing: Removed 👍

PR #469, opened 2021-09-09
Close linked issues automatically on PR merge

yannickperrenet:
## Description

Because we use git-flow and a default branch of `master` we can not make use of issue linking in PRs... which is pretty annoying in my opinion. The added workflow in this PR parses the PR body and closes the mentioned issues on successful merge of the PR.

Syntax:
```
Fixes: #<issue-number>, #<another-issue>, ...
```

### DIFF
--- a/scripts/close_associated_issues.py
+++ b/scripts/close_associated_issues.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+import argparse
+import os
+import re
+import sys
+from http import client
+
+GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
+
+
+def close_issue(conn: client.HTTPSConnection, issue_number: str) -> bool:
+    api_path = "/repos/orchest/orchest/issues/{issue_number}"
+    issue_url = api_path.format(issue_number=issue_number)
+    conn.request(
+        "PATCH",
+        issue_url,
+        '{"state": "closed"}',
+        headers={
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"token {GITHUB_TOKEN}",
+            "User-Agent": "request",
+        },
+    )
+    r = conn.getresponse()
+    return r.status == 200
+
+
+def get_issues_from_pr_body(body: str) -> list[str]:
+    pattern = r"Fixes: (#\d+(?:\, )?)*"
+    match = re.search(pattern, body)
+
+    issue_numbers = match.group(0).lstrip("Fixes: ").split(", ")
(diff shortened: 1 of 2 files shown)


scripts/close_associated_issues.py
```diff
@@ -23,6 +23,7 @@ def close_issue(conn: client.HTTPSConnection, issue_number: str) -> bool:
         },
     )
     r = conn.getresponse()
+    r.read()  # NOTE: make sure to read the response!
```
fruttasecca, line 26: Is this related to this snippet from the `http` client docs?

>Note that you must have read the whole response before you can send a new request to the server.

fruttasecca (approved):
lgtm :)

scripts/close_associated_issues.py
yannickperrenet: Yes. I will change the comment to make that clearer :+1:

AlexandreDecan:
May I ask you what is exactly the issue with git-flow? I do not get why you cannot make use of issue linking in PRs :-) Don't get me wrong, this is not a criticism or an ironic comment, it's a real question ;-) I don't really know this workflow :)

Thank you!

yannickperrenet:
Hi @AlexandreDecan! That is a valid question, let me explain.

We use the [gitflow workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) meaning that we merge PRs into `dev` before they land on `master`. However, linking a PR to an issue works as follows (from the [GitHub docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)):

> Note: The special keywords in a pull request description are interpreted when the pull request targets the repository's default branch. However, if the PR's base is any other branch, then these keywords are ignored, no links are created and merging the PR has no effect on the issues. If you want to link a pull request to an issue using a keyword, the PR must be on the default branch.

So we would have to set `dev` to the default branch, which has some undesired side effects (from the [GitHub docs](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#about-the-default-branch)):

> The default branch is the branch that GitHub displays when anyone visits your repository. The default branch is also the initial branch that Git checks out locally when someone clones the repository.

Therefore the linking sadly does not work and so we chose to implement it ourselves, since manually having to close issues could introduce human errors leaving the backlog in an incorrect state.

I hope that answers your question.

AlexandreDecan:
Yes, thank you! I wasn't aware that issue linking works only for the default branch. Thanks!

PR #472, opened 2021-09-14
Remove useDocumentTitle

iannbing:
## Description

- use window.document.title in Route render props
- delete all usage of useDocumentTitle
- fix LogsView not loaded issue: jobUuid is null when not given, not undefined.

### Checklist


ricklamers (approved):
Nice cleanup 🙌🏻

PR #473, opened 2021-09-15
Feat/example page

iannbing:
## Description

A new "Examples" page next to the Projects View.

### Checklist


ricklamers:
Wow! This is awesome. Great to see it working so quickly.

Minor tweaks:
- `.contribute-card-title { font-size: var(--fontSizes-2xl); }`
- .example-tag { font-size: var(--fontSizes-xs); }`
- Hover of the button shouldn't make the button jump:

image

Great job @iannbing!

ricklamers:
Love the animation on the warning 👍

ricklamers:
One issue I currently have is that there is no "success" confirmation on import. This was less of a problem in the original import dialog because it used to close over the "Projects" page so you'd see the new entry.

@iannbing what do you think?

ricklamers:
@iannbing I think the title of the cards should link to the GitHub repo with a "open_in_new" icon behind it. What do you think?

iannbing:
@ricklamers Good point! How about adding a Confirm dialog after success, with two options: "Go the the example" (primary button) and "Continue browsing".

ricklamers:
Not all Orchest curated examples will by on the user(org in this case) `orchest`. For the collection of Orchest examples we decided to use https://github.com/orchest-examples. So we should check for both `orchest` and `orchest-examples` as owners that go in the first tab.

ricklamers:
> @ricklamers Good point! How about adding a Confirm dialog after success, with two options: "Go the the example" (primary button) and "Continue browsing".

For the text for the first CTA I'd suggest "View pipelines" followed by "Continue browsing".

ricklamers:
LGTM! Awesome job

PR #474, opened 2021-09-16
Updated login view for Orchest Cloud

ricklamers:
## Description

The new login screen takes into account the new design for when the instance is part of Orchest Cloud.

Furthermore, I've added automatic login redirect if a user is already authenticated.

@fruttasecca @yannickperrenet please double check the security logic.

### Checklist


yannickperrenet:
image

The font here seems to be different than inside the product? Or am I just seeing things? :grimacing:

yannickperrenet:
I think it is a bit strange that, having AUTH enabled and going to `/login/admin` shows you the regular admin page.

image

ricklamers:
> I think it is a bit strange that, having AUTH enabled and going to `/login/admin` shows you the regular admin page. But this happens only sometimes (other times the path can not be found -- a `401` is returned)... not sure how to trigger it.
> 
> image

This is a valid endpoint as long as you’re authenticated. That’s either when auth is disabled (then all requests are authenticated) or when auth is enabled and you have a valid session cookie.

ricklamers:
> image
> 
> The font here seems to be different than inside the product? Or am I just seeing things? 😬

Should be the same (Inter weights 400 and 600)

fruttasecca:
>But this happens only sometimes (other times the path can not be found -- a 401 is returned)... not sure how to trigger it.

same

fruttasecca:
The page is super nice by he way :+1:

fruttasecca:
One nitpick:
- login as any user
- go back to `http://localhost:8000/login` (don't logout)
- input random username and password
- login

ricklamers:
> go back to http://localhost:8000/login (don't logout)

This would redirect you (unless you use the browser back button). But you already have a valid auth session so the login will just pick up on this and redirect you.

fruttasecca:
>This would redirect you (unless you use the browser back button)

It's not actually redirecting me when I am going directly to the specified URL through the browser. Don't think this is a problem.

yannickperrenet:
> not sure how to trigger it

I can not reproduce in incognito mode so I guess it has something to do with caching.

Otherwise I can do the following:
1. visit localhost:800/login/admin (fails)
2. visit localhost:800/login
3. visit localhost:800/login/admin (works)
4. visit localhost:800/login/admin (fails)

PR #475, opened 2021-09-17
Issue sync

yannickperrenet:
## Description

Adds a custom action that allows us to sync issue actions to [Linear](https://linear.app/).

## Note

From the GitHub docs:

> This event will only trigger a workflow run if the workflow file is on the default branch.

Therefore this PR is made into `master` to get the workflow file on our default branch.


.github/custom-actions/github-linear-sync/src/index.js
```diff
@@ -0,0 +1,151 @@
...
+          await getLinearIssueId(issueContent.html_url, true),
+          { stateId: statusMap.triage }
+        );
+
+      // NOTE: Let's exclude this through GitHub Actions for now, since
+      // changing labels of old issues would require the archive to be
+      // searched. This can be slow and we don't really care about the
+      // label change.
+      case "unlabeled":
+      case "labeled":
```
fruttasecca, line 121: Might be missing something, but once I read "Let's exclude this through GitHub Actions for now" I expected a break right after the `unlabeled` case.

fruttasecca (approved):
lgtm :)

.github/custom-actions/github-linear-sync/src/index.js
yannickperrenet: I understand the confusion, I just didn't want to remove the code since it works (depending on some caveats). I figured we can easily support it in the future (by triggering the event in the GitHub Action) once we better understand our Linear usage.

fruttasecca: I see, alright.

PR #476, opened 2021-09-17
Disable close for import dialog

iannbing:
## Description

this is to prevent user losing import progress. The downside is that if user was just checking the functionality, user cannot easily close it with ESCAPE key or mouse click outside.

### Checklist


iannbing:
Discussed with @fruttasecca. Merging into `dev`.

PR #486, opened 2021-09-23
Enable authless services

ricklamers:
## Description

Enable services that have no authentication. This can be useful for OAuth flows or hosting Streamlit dashboard.

Fixes: https://github.com/orchest/orchest/issues/487

### Checklist

### DIFF
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -10,12 +10,12 @@ from werkzeug.security import check_password_hash, generate_password_hash
 from _orchest.internals.utils import _proxy
 from app.connections import db
 from app.models import Token, User
-from app.utils import get_user_conf
+from app.utils import get_auth_cache, get_user_conf, set_auth_cache
 
 # This auth_cache is shared between requests
 # within the same Flask process
-auth_cache = {}
-auth_cache_age = 3  # in seconds
+_auth_cache = {}
+_auth_cache_age = 3  # in seconds
 
 
 def register_views(app):
@@ -255,36 +255,9 @@ def register_views(app):
 
         return jsonify(data_json), 200
 
-    # This is for service auth caching
-    def set_auth_cache(
-        project_uuid_prefix, session_uuid_prefix, requires_authentication
-    ):
-        global auth_cache
-
-        auth_cache["%s-%s" % (project_uuid_prefix, session_uuid_prefix)] = {
-            "date": datetime.datetime.now(),
-            "requires_authentication": requires_authentication,
-        }
-
-    def get_auth_cache(project_uuid_prefix, session_uuid_prefix):
-        global auth_cache
-        key = "%s-%s" % (project_uuid_prefix, session_uuid_prefix)
-
-        if key not in auth_cache:
-            return {"status": "missing"}
-
-        if (
-            datetime.datetime.now() - auth_cache[key]["date"]
-        ).total_seconds() > auth_cache_age:
-            return {"status": "expired"}
-
-        return {
-            "status": "available",
-            "requires_authentication": auth_cache[key]["requires_authentication"],
-        }
-
     @app.route("/auth/service", methods=["GET"])
     def auth_service():
+        global _auth_cache, _auth_cache_age
 
         # Bypass definition based authentication if the request
         # is authenticated
@@ -308,7 +281,9 @@ def register_views(app):
             app.logger.error("Failed to parse X-Original-URI: %s" % original_uri)
             return "", 401
 
-        auth_check = get_auth_cache(project_uuid_prefix, session_uuid_prefix)
+        auth_check = get_auth_cache(
+            project_uuid_prefix, session_uuid_prefix, _auth_cache, _auth_cache_age
+        )
         if auth_check["status"] == "available":
             if auth_check["requires_authentication"] is False:
                 return "", 200
@@ -340,14 +315,16 @@ def register_views(app):
                 # Always check first service that is returned,
                 # should be unique
                 if services[0]["service"]["requires_authentication"] is False:
-                    set_auth_cache(project_uuid_prefix, session_uuid_prefix, False)
+                    set_auth_cache(
+                        project_uuid_prefix, session_uuid_prefix, False, _auth_cache
+                    )
                     return "", 200
                 else:
                     raise Exception("'requires_authentication' is not set to False")
 
             except Exception as e:
-                set_auth_cache(project_uuid_prefix, session_uuid_prefix, True)
+                set_auth_cache(
+                    project_uuid_prefix, session_uuid_prefix, True, _auth_cache
+                )
                 app.logger.error(e)
                 return "", 401
-
-    # End of service auth
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -3,6 +3,7 @@ import os
 import secrets
 import uuid
 
+import requests
 from flask import jsonify, redirect, request, send_from_directory
 from werkzeug.security import check_password_hash, generate_password_hash
 
@@ -10,6 +11,11 @@ from _orchest.internals.utils import _proxy
 from app.connections import db
 from app.models import Token, User
 from app.utils import get_user_conf
+
+# This auth_cache is shared between requests
+# within the same Flask process
+auth_cache = {}
+auth_cache_age = 3  # in seconds
 
 
 def register_views(app):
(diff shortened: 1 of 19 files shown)


services/orchest-webserver/client/src/components/ProjectSelector.tsx
```diff
@@ -59,37 +59,47 @@ const ProjectSelector = (_, ref: TProjectSelectorRef) => {
...
-        // navigate ONLY if user is at the project root
-        if (matchProjectRoot) onChangeProject(newProjectUuid);
       })
       .catch((error) => console.log(error));
   };
 
   // sync state.projectUuid and the route param projectUuid
   React.useEffect(() => {
-    dispatch({ type: "projectSet", payload: projectUuidFromRoute });
+    if (projectUuidFromRoute) {
```
ricklamers, line 68: These changes are a fix that I found and discussed with @iannbing

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+        project_uuid_prefix, session_uuid_prefix, requires_authentication
+    ):
+        global auth_cache
+
+        auth_cache["%s-%s" % (project_uuid_prefix, session_uuid_prefix)] = {
+            "date": datetime.datetime.now(),
+            "requires_authentication": requires_authentication,
+        }
+
+    def get_auth_cache(project_uuid_prefix, session_uuid_prefix):
```
yannickperrenet, line 269: These two functions are helper functions and not view, so I think it makes more sense to move them outside of the `register_views` function.

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+        global auth_cache
+        key = "%s-%s" % (project_uuid_prefix, session_uuid_prefix)
+
+        if key not in auth_cache:
+            return {"status": "missing"}
+
+        if (
+            datetime.datetime.now() - auth_cache[key]["date"]
+        ).total_seconds() > auth_cache_age:
```
yannickperrenet, line 278: Would be better to also add `auth_cache_age` to the `global auth_cache` line.

services/auth-server/app/app/views.py
```diff
@@ -11,6 +12,11 @@
 from app.models import Token, User
 from app.utils import get_user_conf
 
+# This auth_cache is shared between requests
+# within the same Flask process
+auth_cache = {}
+auth_cache_age = 3  # in seconds
```
yannickperrenet, line 18: Better to make variables "internal" using a prepended underscore (see [google's styleguide](https://google.github.io/styleguide/pyguide.html#25-global-variables)).

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+        original_uri = request.headers.get("X-Original-URI")
+
+        if original_uri is None:
+            return "", 401
+
+        try:
+            # expected uri:
+            # /pbp-service-[service-name]-
+            # [pipeline_uuid_prefix]-[session_uuid_prefix]_[port]/...
+            components = original_uri.split("/")[1].split("_")[-2].split("-")
```
yannickperrenet, line 304: Just to be sure, the `uuid_prefix` does not contain a dash `-` right?

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+
+                if len(services) > 1:
+                    raise Exception(
+                        "Filtered /api/services endpoint "
+                        "should always return a single service"
+                    )
+
+                # Always check first service that is returned,
+                # should be unique
+                if services[0]["service"]["requires_authentication"] is False:
```
yannickperrenet, line 342: Instead of checking whether something `is False` it is instead more Pythonic to do
```python
if not services[0]["service"]["requires_authentication"]
```
([reference](https://google.github.io/styleguide/pyguide.html#2144-decision))

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+                    set_auth_cache(project_uuid_prefix, session_uuid_prefix, False)
+                    return "", 200
+                else:
+                    raise Exception("'requires_authentication' is not set to False")
+
+            except Exception as e:
+                set_auth_cache(project_uuid_prefix, session_uuid_prefix, True)
```
yannickperrenet, line 349: Strictly speaking this `set_auth_cache` would also be run if `No services found`, I don't think that should happen right? Since then `requires_authentication` is set to `True` , whilst that does not need to be the case.

services/auth-server/app/app/views.py
```diff
@@ -248,3 +254,100 @@ def get_users():
...
+                app.logger.error(e)
+                return "", 401
+
+    # End of service auth
```
yannickperrenet, line 353: Any particular reason this is needed?

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+        # The session URLs contain information about the
+        # project UUID and session UUID.
+        # It's only the first path component of the UUID string.
+        # E.g. [766907b5]-cb1e-4e6f-985b-ebded3a55435
+        # Similarly it's only the first path component of the UUID
+        # of the session UUID.
+
+        project_uuid_prefix_filter = None
+        if "project_uuid_prefix" in request.args:
+            project_uuid_prefix_filter = request.args["project_uuid_prefix"]
```
yannickperrenet, line 34: Isn't this the same as:
```python
project_uuid_prefix_filter = request.args.get("project_uuid_prefix")
```
or does the `request.args` not have a `.get()`?

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+
+        session_uuid_prefix_filter = None
+        if "session_uuid_prefix" in request.args:
+            session_uuid_prefix_filter = request.args["session_uuid_prefix"]
+
+        # Get services of the InteractiveSessions
+        query = models.InteractiveSession.query
+
+        if project_uuid_prefix_filter:
```
yannickperrenet, line 43: This needs to be explicitly compared to `None`:
```python
if project_uuid_prefix_filter is not None:
```

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+                    "service": service,
+                    "type": "INTERACTIVE",
+                    "project_uuid": session.project_uuid,
+                    "pipeline_uuid": session.pipeline_uuid,
+                }
+                for service in session.user_services.values()
+            ]
+            for session in sessions
+        ]
+        session_services = list(chain(*session_services))
```
yannickperrenet, line 70: Isn't this the same as:
```python
session_services = []
for session in sessions:
  for service in for service in session.user_services.values()
    session_services.append({
          "service": service,
          "type": "INTERACTIVE",
          "project_uuid": session.project_uuid,
          "pipeline_uuid": session.pipeline_uuid,
    })
```
which is a lot more readable imo.

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # Get services of the Job runs
+        query_runs = models.NonInteractivePipelineRun.query.with_entities(
+            models.NonInteractivePipelineRun.job_uuid,
+            models.NonInteractivePipelineRun.uuid,
+        )
+
+        if project_uuid_prefix_filter:
```
yannickperrenet, line 78: Again needs to be explicitly compared with `None`

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+            query_runs = query_runs.join(models.Job).filter(
+                models.Job.project_uuid.startswith(project_uuid_prefix_filter)
+            )
+
+        if session_uuid_prefix_filter:
+            query_runs = query_runs.filter(
+                models.NonInteractivePipelineRun.uuid.startswith(
+                    session_uuid_prefix_filter
+                )
```
yannickperrenet, line 87: This is probably my misunderstanding, but why does the uuid of the `NonInteractivePipelineRun` have to start with the `session_uuid_prefix`?

services/orchest-api/app/app/apis/namespace_services.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+        active_pipeline_runs = query_runs.filter_by(status="STARTED").all()
+
+        job_ids = set([run.job_uuid for run in active_pipeline_runs])
```
yannickperrenet, line 92: You can actually also make a `set` comprehension:
```python
job_ids = {run.job_uuid for run in active_pipeline_runs}
```

services/orchest-webserver/client/src/components/ServiceForm.tsx
```diff
@@ -400,7 +400,38 @@ const ServiceForm: React.FC<any> = (props) => {
...
+                  <i
+                    className="material-icons inline-icon push-left"
+                    aria-describedby="tooltip-urls"
+                  >
+                    info
+                  </i>
+                </h3>
+                <MDCTooltipReact
+                  tooltipID="tooltip-urls"
+                  tooltip="The URLs that will be available to communicate with the service. These are all proxied by Orchest."
```
yannickperrenet, line 416: Maybe add the word "directly", i.e.: `(...) to communicate with the service directly.`

yannickperrenet:
Apart from some beauty changes, the overall (high-level) structure looks good to me. Only question I have here is: https://github.com/orchest/orchest/pull/486#discussion_r715598558

Nice work! :muscle:

services/auth-server/app/app/views.py
ricklamers: Correct

services/auth-server/app/app/views.py
ricklamers: That catches more types than False which is not my intention.

services/auth-server/app/app/views.py
ricklamers: It should happen, because the cached value should then be to not authenticate the request.

services/auth-server/app/app/views.py
ricklamers: I thought it could give clarity as the file grows

services/orchest-api/app/app/apis/__init__.py
```diff
@@ -28,4 +29,5 @@
 api.add_namespace(ns_projects)
 api.add_namespace(ns_runs)
 api.add_namespace(ns_sessions)
+api.add_namespace(ns_services)
```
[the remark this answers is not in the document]
ricklamers: I discussed this with @fruttasecca. The reasoning is that even though services are part of sessions and runs, we don't want an external service (in this case the auth-server) to have to be concerned with the internal representation details of services. That's why the new services namespace feels like a wrapper of existing entities. Not sure what the cleaner way would be.

services/orchest-api/app/app/apis/namespace_services.py
ricklamers: It is, I can clean it up to be more concise

services/orchest-api/app/app/apis/namespace_services.py
ricklamers: Fair enough. I think the weirdness was the result of refactoring during implementation. Those algebraic style rewrites then do not lead to the most straightforward code.

services/orchest-api/app/app/apis/namespace_services.py
ricklamers: It's because the session_uuid for a service is the run UUID in case of a NONINTERACTIVE service.

services/auth-server/app/app/views.py
yannickperrenet: Feels like reuse of something that it is not meant for to be honest.

> the cached value should then lead to not authenticate the request.

I agree with this point, but the service is not defined which is another case than `requires_authentication`. So I agree with the result just not with the way to get there. In my opinion it is more an "invalid request"

services/auth-server/app/app/views.py
yannickperrenet: Ohh didn't get that, sorry. I see you already removed it :+1:

services/orchest-api/app/app/apis/__init__.py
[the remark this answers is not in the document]
yannickperrenet: I agree that the services namespace is needed, now that the auth-server needs services information. 

Its just that now the services namespace is introduced it is confusing that the services namespace doesn't manage the services, but the sessions namespace does. Not something we can easily get around, but definitely something to give another thought imo.

services/orchest-api/app/app/apis/namespace_services.py
yannickperrenet: Yea totally get that ;)

services/orchest-api/app/app/apis/namespace_services.py
yannickperrenet: Ahaa nice catch! Could you add that as a comment?

services/auth-server/app/app/views.py
ricklamers: I changed this to not set the cache in case a request was made that had an unexpected result (no service exists, multiple services exist).

services/orchest-api/app/app/apis/namespace_services.py
ricklamers: Added.

services/orchest-api/app/app/apis/__init__.py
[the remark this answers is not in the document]
fruttasecca: As long as `namespace_services` offers a clean  & restful interface to the "users" of the API the actual data model can be refactored later if need be, imo. This is because currently the namespace only offers read capabilities (i.e. `GET`) so its fine to see it as a view over the data, which can be more entangled that the interface of the namespace suggests.

PR #506, opened 2021-09-30
Show loading screen until iframe load completes.

ricklamers:
## Description

Show loading screen until iframe load completes.

### Checklist


fruttasecca:
lgtm :musical_note:

PR #535, opened 2021-11-04
Improv/update logging

### LINKED ISSUE #464: Make update text stream non-buffered for the image pulling phase
**Describe the solution you'd like**  
Make update text stream non-buffered for the image pulling phase

**What does your solution aim to solve?**  
A relatively unresponsive updater will make users navigate away or frustrated.

### LINKED ISSUE #465: Make update text output less verbose
**Describe the solution you'd like**  
Make update text output less verbose.

We could have more verbose outputs based on whether the `dev` flag has been set.

**What does your solution aim to solve?**  
We shouldn't show users confusing blobs of text that represent internal mechanisms.

fruttasecca:
## Description

This PR reduces the verbosity of the updater when in `web` mode, along with making the progress bar unbuffered.

Fixes #464
Fixes #465

### Checklist

For reference, here are the old update logs

reference

while these are the new ones

image

### DIFF
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -180,7 +180,7 @@ class OrchestApp:
             )
         try:
             version_exit_code = 0
-            self.version(ext=True, verbose=False)
+            self.version(ext=True)
         except typer.Exit as e:
             version_exit_code = e.exit_code
 
--- a/services/orchest-ctl/app/orchest.py
+++ b/services/orchest-ctl/app/orchest.py
@@ -150,7 +150,12 @@ class OrchestApp:
         pulled_images = self.resource_manager.get_images()
         to_pull_images = set(config.ORCHEST_IMAGES["minimal"]) | set(pulled_images)
         logger.info("Updating images:\n" + "\n".join(to_pull_images))
-        self.docker_client.pull_images(to_pull_images, prog_bar=True, force=True)
+        self.docker_client.pull_images(
+            to_pull_images,
+            prog_bar=True,
+            mode=mode,
+            force=True,
+        )
 
         # Add a tag to user environment images to mark them for removal.
         # The orchest-api will deal with the rest of the logic related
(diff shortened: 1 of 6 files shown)


services/update-server/app/app/views.py
```diff
@@ -17,6 +17,7 @@
 
 def log_update(message):
     try:
+        message = message.replace("!!\n", "")
```
fruttasecca, line 20: Note that this implies that the `!!\n` sequence will be visible for every user coming from an older version of Orchest, since the latest `update-server` is pulled as part of the update.

ricklamers:
LGTM! Glad you found a way to get the unbuffering resolved.

If somehow we'd genuinely want to print !!\n we couldn't do that now. But I think that's acceptable 👍

fruttasecca:
>If somehow we'd genuinely want to print !!\n we couldn't do that now. But I think that's acceptable

I could add a check in the update server to only remove `!!\n` from lines containing `Pulling images` but I didn't want to create more coupling, altought it might be preferable if we wanted to print `!!\n.` Perhaps we could use a less common substring, although the reason to use `!!\n` is that it didn't feel too out of place to users updating from an old version.

services/orchest-ctl/app/orchest.py
```diff
@@ -168,12 +173,14 @@ def update(self, mode=None, dev: bool = False):
...
-        )
+        if mode != "web":
+            utils.echo(
+                "Checking whether all containers are running the same version of "
+                "Orchest."
+            )
         try:
             version_exit_code = 0
-            self.version(ext=True)
+            self.version(ext=True, verbose=False)
```
yannickperrenet, line 183: During the update process it could be that a user pulls different versions of different images, so I do still think this should be verbose.

services/update-server/app/app/views.py
yannickperrenet: Due to the coupling with the `orchest-ctl` for the `!!\n` value, do we want to put it in the internal config instead?

yannickperrenet:
Minimal fix really. I like it :+1:

services/update-server/app/app/views.py
fruttasecca: Yes, let's do it, it's also a good excuse to use a value that will be more difficult to output by accident.

yannickperrenet (approved):
:metal:

Issue #422, opened 2021-08-17
Failed to delete 'parent' Orchest environment image on update


ricklamers:
**Describe the bug**  
Failure to delete 'parent' Orchest environment image on Orchest update.

**To Reproduce**  
Update Orchest that has a project with a built environment in it.

The parent environment image (e.g. orchest/base-image-py) is not deleted and the updated throws a DockerError.

**Expected behavior**  
The parent environment image should be deleted (on boot, as that cleans up old environments) and no error should be thrown during update.

**Environment**

- OS: Linux
- Browser: Chrome
- Version: 2021.08.2
- Environment: production
- Logs:
```
connector: <aiohttp.connector.UnixConnector object at 0x7ff6045df5e0>
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7ff604621f40>, 177.202228266)]']
ERROR: Unclosed connector
client_session: <aiohttp.client.ClientSession object at 0x7ff6045dfc10>
ERROR: Unclosed client session
aiodocker.exceptions.DockerError: DockerError(409, 'conflict: unable to delete c082e13c3e46 (cannot be forced) - image has dependent child images')
raise DockerError(response.status, json.loads(what.decode("utf8")))
File "/usr/local/lib/python3.8/site-packages/aiodocker/docker.py", line 268, in _do_query
resp = await self._coro
File "/usr/local/lib/python3.8/site-packages/aiodocker/utils.py", line 309, in __aenter__
async with self._query(
File "/usr/local/lib/python3.8/site-packages/aiodocker/docker.py", line 293, in _query_json
return await self.docker._query_json(
File "/usr/local/lib/python3.8/site-packages/aiodocker/images.py", line 227, in delete
await self.aclient.images.delete(image, force=force)
File "/usr/local/lib/python3.8/site-packages/app/docker_wrapper.py", line 198, in _remove_image
await asyncio.gather(
File "/usr/local/lib/python3.8/site-packages/app/docker_wrapper.py", line 201, in _remove_images
return future.result()
File "/usr/local/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
return loop.run_until_complete(main)
File "/usr/local/lib/python3.8/asyncio/runners.py", line 44, in run
asyncio.run(self._remove_images(image_ids, force=force))
File "/usr/local/lib/python3.8/site-packages/app/docker_wrapper.py", line 216, in remove_images
self.docker_client.remove_images(orchest_dangling_imgs, force=True)
[...]

PR #538, opened 2021-11-04
Fix/stale orchest image deletion

fruttasecca:
## Description
This PR:
- moves the responsibility of deleting dangling Orchest images from `orchest-ctl` to the `orchest-api`, so that the deletion of those images is not staggered but happens on every update.
- fixes a bug where unused environment images were not being cleaned up once a job reached an end state if they were not marked for removal.

Fixes: #422 

### Checklist


yannickperrenet (approved):
:rocket:

Issue #501, opened 2021-09-28
Increase default parallelism of Celery workers to 4 (both Job and Interactive queues)


ricklamers:
In addition, we should add a config like `MAX_JOB_RUNS_PARALLELISM` for interactive run parallelism `MAX_INTERACTIVE_RUNS_PARALLELISM`.

The idea is that you can work on multiple pipelines interactively at the same time. Sometimes, even in the same project. This example comes up in the [OAuth included QuickBooks pipeline](https://github.com/ricklamers/orchest-quickbooks-oauth). Here we can have the OAuth token refresh pipeline running jointly with the actual data fetching pipeline.

PR #539, opened 2021-11-04
Increased run parallelism for jobs and interactive runs

yannickperrenet:
## Description

Fixes: #501 

This PR also introduces the `GlobalOrchestConfig` class and refactors all access to the global Orchest config file (`~/.config/orchest/config.json` on the host) to use the class. The class ensures all configuration options have a default value and are validated when changed, and certain preset options in the file cannot be changed when running with `--cloud`.

### Checklist

### Other

@fruttasecca Can you back up your current config file and then build the branch and try to use Orchest? That would be great, thank you!

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -27,8 +27,8 @@ class GlobalOrchestConfig:
 
     # Defines default values for all supported configuration options.
     _default_values = {
-        "MAX_JOB_RUNS_PARALLELISM": 1,
-        "MAX_INTERACTIVE_RUNS_PARALLELISM": 1,
+        "MAX_JOB_RUNS_PARALLELISM": 4,
+        "MAX_INTERACTIVE_RUNS_PARALLELISM": 4,
         "AUTH_ENABLED": False,
         "TELEMETRY_DISABLED": False,
         "TELEMETRY_UUID": str(uuid.uuid4()),
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 import logging
 import os
@@ -93,6 +95,9 @@ class GlobalOrchestConfig:
             >>> # Save the updated (and automatically validated) config
             >>> # to disk.
             >>> requires_orchest_restart = config.save(flask_app=app)
+            >>> # Just an example output.
+            >>> requres_orchest_restart
+            ... ["MAX_INTERACTIVE_RUNS_PARALLELISM"]
 
         """
         unmodifiable_config, current_config = self._get_current_configs()
@@ -104,7 +109,7 @@ class GlobalOrchestConfig:
         # Flatten into regular dictionary.
         return dict(self._values)
 
-    def save(self, flask_app=None) -> Optional[bool]:
+    def save(self, flask_app=None) -> Optional[list[str]]:
         """Saves the state to disk.
 
         Args:
@@ -114,9 +119,9 @@ class GlobalOrchestConfig:
 
         Returns:
             * `None` if no `flask_app` is given.
-            * `True` if Orchest needs to be restarted for the changes to
-              the global config to take effect.
-            * `False` otherwise.
+            * List of changed config options that require an Orchest
+              restart to take effect.
+            * Empty list otherwise.
 
         """
         state = self.as_dict()
@@ -176,18 +181,19 @@ class GlobalOrchestConfig:
     def __getitem__(self, key):
         return self._values[key]
 
-    def _changes_require_restart(self, flask_app, new: dict) -> bool:
+    def _changes_require_restart(self, flask_app, new: dict) -> list[str]:
         """Do config changes require an Orchest restart.
 
         Compares the Orchest global config values in the flask app to
-        the new values and determines whether the changes require a
+        the `new` values and determines whether the changes require a
         restart of the Orchest application.
 
         Returns:
-            `True` if the Orchest application needs to restart if `new`
-            becomes the new config. `False` otherwise.
+            A list of strings representing the changed configuration
+            options that require a restart of Orchest to take effect.
 
         """
+        res = []
         for k, val in self._config_values.items():
             if not val["requires-restart"]:
                 continue
@@ -200,9 +206,9 @@ class GlobalOrchestConfig:
 
             old_val = flask_app.config[k]
             if new.get(k) is not None and new[k] != old_val:
-                return True
+                res.append(k)
 
-        return False
+        return res
 
     def _validate_dict(self, d: abc.Mapping) -> None:
         """Validates the types and values of the values of the dict.
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -209,13 +209,8 @@ class GlobalOrchestConfig:
                 try:
                     unmodifiable_config[k] = deepcopy(current_config[k])
                 except KeyError:
-                    # Otherwise you could get a default value of
-                    # `AUTH_ENABLED` being `False`. This is probably not
-                    # what you want.
-                    raise _errors.CorruptedFileError(
-                        "When running Orchest with '--cloud' all of"
-                        f" {self._cloud_unmodifiable_config_values} have to be defined."
-                    )
+                    # Fall back on default values.
+                    ...
 
         return unmodifiable_config, current_config
 
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -1,8 +1,11 @@
+import json
 import logging
 import os
 import re
 import time
 import uuid
+from collections import ChainMap, abc
+from copy import deepcopy
 from typing import Any, Dict, Optional
 
 import docker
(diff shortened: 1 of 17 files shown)


fruttasecca:
>@fruttasecca Can you back up your current config file and then build the branch and try to use Orchest? That would be great, thank you!

will do

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -1,15 +1,217 @@
...
+            >>> # Automatically fill in default values when accessing
+            >>> # the configuration values.
+            >>> d = config.get()
+
+        """
+        self._values = ChainMap(self._get_current_config(), self._default_values)
+
+    def get(self) -> dict:
+        # Flatten into regular dictionary.
+        return dict(self._values)
```
fruttasecca, line 71: Is there a particular reason for this method name? Or can the name be made more specific to what happens?

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -1,15 +1,217 @@
...
+                if not isinstance(d[k], type(val)):
+                    raise TypeError(
+                        "Given dictionary has uncompatible types for certain values."
+                    )
+            except KeyError:
+                # We let it pass silently because it won't break the
+                # application in any way.
+                logger.debug(
+                    "User tried to add a config value that is not used by Orchest."
+                )
```
fruttasecca, line 169: I might be missing some context, doesn't a `KeyError` here actually imply that the passed dictionary `d` is missing an entry that is instead supposed to be in there?

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -1,15 +1,217 @@
...
+            e.g. parallelism > 0.
+            `False` otherwise.
+
+        """
+        # Check for types.
+        for k, val in self._default_values.items():
+            try:
+                if not isinstance(d[k], type(val)):
+                    raise TypeError(
+                        "Given dictionary has uncompatible types for certain values."
```
fruttasecca, line 162: Note that `isinstance(True, int)` evaluates to True because boolean is (surprisingly) a subclass of int. Perhaps we could use `type(d[k]) == type(val)`, losing subclassing of config values but I don't think we need that.

services/orchest-webserver/app/app/views/views.py
```diff
@@ -228,47 +228,30 @@ def version():
...
-            save_user_conf_raw(json.dumps(config))
-            current_config = config
-
-        return current_config
+                return user_config.get()
+
+            user_config.update(config)
+            user_config.save(flask_app=app)
+
+        return user_config.get()
```
fruttasecca: Since `update` could throw an exception if a given type requiremrent is not met, should we actually catch that and return  the `user_config` in case of error? Makes for a better UX imo; as of now I can input invalid values and save, the FE will keep showing the invalid values because the call returned a 500.

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: Great catch! Will change that to exact type matching :+1:

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: I could rename it to `get_config`, I can understand if it is easily confused with `dict.get` which only gets a particular value. Is that what you meant?

The fact that it flattens into a regular dict is an implementation detail that should be hidden to the user of the class imo.

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -84,70 +90,56 @@ def save(self, flask_app=None) -> None:
...
-            RuntimeError: The private ChainMap was incorrectly modified
-                outside of the class.
             TypeError: The values of the dictionary that correspond to
                 supported config values have incorrect types.
+            ValueError: The values of the dictionary that correspond to
+                supported config values have incorrect values. E.g.
+                maximum parallelism has to be greater or equal to one.
 
         """
-        d = self._remove_undefined_keys(d)
```
fruttasecca: Not particularly in favour of removing undefined keys or not, but what's the rationale behind not doing that?

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: It does not have to be there. I see that the logging message is wrong, will update it.

For example, the default dict contains keys A and B but the new dict only contains A (since only the value for A needs to be updated), then no exception should be raised. Because the end result of the config will fall back on the value that is already defined for B (either in the current config or in the default values). Essentially only checking the types of updated values.

services/auth-server/app/app/views.py
```diff
@@ -45,11 +46,22 @@ def server_config():
...
-        config = _utils.GlobalOrchestConfig()
-        if not config["AUTH_ENABLED"]:
-            return True
+            # Fall back on the cached value.
+            if not app.config["AUTH_ENABLED"]:
+                return True
+        else:
+            global_orchest_config.save(flask_app=app)
+            if not global_orchest_config["AUTH_ENABLED"]:
+                return True
```
fruttasecca, line 64: Saving the config on every `is_authenticated` call looks a bit odd, why is it needed?

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: I no longer remove undefined keys because undefined keys will not be used by Orchest anyways, and thus not breaking the application. 

If we were to instead remove undefined keys then we could in newer versions of Orchest remove an old config value, then if the user downgrades (or maybe we possibly do this in the future) then that config value would fall back on the default for that version instead of using the user's entered value. Additionally, I felt that for JSON you can just add any data you like as long as the application doesn't access the key - why explicitly check (do more work) whether their is no data in there that isn't used (potentially deep traversing the config).

services/orchest-ctl/app/debug.py
```diff
@@ -333,8 +334,16 @@ def containers_version_dump(
...
+    try:
+        config = _utils.GlobalOrchestConfig()
+    except _errors.CorruptedFileError as e:
+        utils.echo(e)
+        config = _utils.GlobalOrchestConfig.read_raw_current_config(
+            _utils.GlobalOrchestConfig
+        )
+    else:
+        config = config.read_raw_current_config()
+
```
fruttasecca, line 346: This is just a matter of taste, but I guess the `read_raw_current_config` call could be made in the try block?

services/orchest-webserver/app/app/views/views.py
yannickperrenet: You are totally right. Will change.

Whilst making the changes I was actually pretty stunned by the number of `try except` blocks that I was adding, I wasn't sure whether this was good or bad. All I could think was the number of `try except` blocks that aren't there on other operations in the codebase.

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -193,10 +193,29 @@ def _get_current_config(self) -> dict:
...
+                    # `AUTH_ENABLED` being `False`. This is probably not
+                    # what you want.
+                    raise _errors.CorruptedFileError(
+                        "When running Orchest with '--cloud' all of"
+                        f" {self._cloud_unmodifiable_config_values} have to be defined."
+                    )
+
+        return unmodifiable_config, current_config
 
     def read_raw_current_config(self) -> dict:
```
fruttasecca, line 217: Could this be made private? Would be more explicit in the role of the class an an abstraction/encapsulator over the configuration, imo.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: >The fact that it flattens into a regular dict is an implementation detail that should be hidden to the user of the class imo.

Agree.

>I could rename it to get_config, I can understand if it is easily confused with dict.get which only gets a particular value. Is that what you meant?

Yes, perhaps `as_dict`? Not feeling particularly strong about it, so whatever you prefer really.

services/orchest-webserver/app/app/views/views.py
fruttasecca: >Whilst making the changes I was actually pretty stunned by the number of try except blocks that I was adding, I wasn't sure whether this was good or bad. All I could think was the number of try except blocks that aren't there on other operations in the codebase.

I share the feeling, although I have no real input to give.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: I see, alright.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: Looks like I have botched the selection. Meant to highlight `read_raw_current_config`.

fruttasecca:
I really like that we are finally abstracting away the access to the config file :)

services/auth-server/app/app/views.py
yannickperrenet: Funny that you mention actually, because the `auth-server` doesn't change the config file anyways. However, there is an edge case where it does (with the new implementation):
1. User changes the config file through the FS, e.g. removing `AUTH_ENABLED`
2. The auth-server reads the config file.
3. When accessing the `AUTH_ENABLED` in the call, it actually falls back to the default value.

The question then is, do we write the "new" config file yes or no?! I went with yes, but strictly speaking I guess we don't have to.

services/orchest-ctl/app/debug.py
yannickperrenet: It can't also be in the `try` because the initialization of the class would error if for example the type of a config value is wrong whilst the `read_raw_current_config` could actually go through. This is also exactly what we want to learn in the debug dump: what is the config value that was wrongly entered to make the application crash?

If the `read_raw_current_config` fails due to a JSON decode error, then the debug dump fails. But I figured this was okay since it immediately is clear where the error is in the application.

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: It was private when I first created the class, but then the `ctl` makes use of this function so I made it public. 

> Would be more explicit in the role of the class as an abstraction/encapsulator over the configuration, imo.

I definitely agree here. I can just make the `ctl` use the private function, I think that would be fine given that it makes a debug dump and thus needs to go deeper into the application logic. :+1:

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: I like `as_dict` :+1:

(Could even cover the case if we were to ever go to yaml)

services/auth-server/app/app/views.py
fruttasecca: Definitively an interesting question, not sure either. What I found strange is this piece of code is catching an edge case by (indirectly) knowing what the internals of the config clas will do when it comes to saving, defaut values, etc.; which indirectly suggests that the responsibility of calling save to sync disk and the in-memory state of the config is of the rest of the application, depending on what edge case a particular piece of code is aware of or busy with. Perhaps such syncing should happen under the hood everything the config class is created?

I don't want to make this take longer than it needs so take this with a pinch of salt, pretty much just thinking aloud.

services/orchest-ctl/app/debug.py
fruttasecca: I understand, makes sense.

fruttasecca (approved):

services/auth-server/app/app/views.py
yannickperrenet: > Perhaps such syncing should happen under the hood everytime the config class is created?

That also crossed my mind, but in the end I went with making it explicit instead of as a side-effect. The class should do what a user expects and I think automatically syncing on initialization is not one of them. Definitely like the idea though :wink:

yannickperrenet:
FYI The failed integration test was due to flakiness:

> Timed out retrying after 15000ms: Unable to find an element by: [data-test-id="projects-table-row"]

Running it locally passed all tests.

ricklamers:
LGTM, really nice cleanup @yannickperrenet!

PR #545, opened 2021-11-10
Feat/idle checking

fruttasecca:
## Description

This PR introduces an endpoint in the `orchest-api` (also proxied by the `orchest-webserver`) that allows knowing if Orchest is idle or not.
The requirement for Orchest being considered idle are:
- no active clients, meaning no heartbeat from a client for 5 minutes
- no ongoing environment builds
- no ongoing JupyterLab builds
- no ongoing interactive runs
- no ongoing job runs
- no busy kernels, according to the JupyterLab API

Example of the returned data.
```json
{
    "idle": false,
    "details": {
        "no_active_clients": false, 
        "no_ongoing_environment_builds": true, 
        "no_ongoing_jupyterlab_builds": true, 
        "no_ongoing_interactive_runs": true, 
        "no_ongoing_job_runs": true,
        "no_busy_kernels": true
        }
}
```
Client heartbeats are kept track of with a database entry.

@iannbing I ended up using `useInterval` since it looked like a simple solution, let me know what you think :). 

NOTE: the branch contains a schema migration

### DIFF
--- a/services/orchest-api/app/app/core/sessions.py
+++ b/services/orchest-api/app/app/core/sessions.py
@@ -518,7 +518,7 @@ class InteractiveSession(Session):
         # 'last_activity': '2021-11-10T09:04:10.508031Z',
         # 'execution_state': 'idle', 'connections': 2}]
         kernels: List[dict] = response.json()
-        return any([kernel.get("execution_state") == "busy" for kernel in kernels])
+        return any(kernel.get("execution_state") == "busy" for kernel in kernels)
 
     def shutdown(self) -> None:
         """Shuts down the launch.
--- a/services/orchest-api/app/app/core/sessions.py
+++ b/services/orchest-api/app/app/core/sessions.py
@@ -491,18 +491,19 @@ class InteractiveSession(Session):
         """
         IP = self.get_containers_IP()
 
-        self._notebook_server_info = {
-            "port": 8888,
-            "base_url": "/"
-            + _config.JUPYTER_SERVER_NAME.format(
-                project_uuid=session_config["project_uuid"][
-                    : _config.TRUNCATED_UUID_LENGTH
-                ],
-                pipeline_uuid=session_config["pipeline_uuid"][
-                    : _config.TRUNCATED_UUID_LENGTH
-                ],
-            ),
-        }
+        if self._notebook_server_info is None:
+            self._notebook_server_info = {
+                "port": 8888,
+                "base_url": "/"
+                + _config.JUPYTER_SERVER_NAME.format(
+                    project_uuid=session_config["project_uuid"][
+                        : _config.TRUNCATED_UUID_LENGTH
+                    ],
+                    pipeline_uuid=session_config["pipeline_uuid"][
+                        : _config.TRUNCATED_UUID_LENGTH
+                    ],
+                ),
+            }
 
         # https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html
         url = (
--- a/services/orchest-api/app/app/core/sessions.py
+++ b/services/orchest-api/app/app/core/sessions.py
@@ -6,7 +6,7 @@ import traceback
 from abc import abstractmethod
 from contextlib import contextmanager
 from enum import Enum
-from typing import Any, Dict, NamedTuple, Optional
+from typing import Any, Dict, List, NamedTuple, Optional
 from uuid import uuid4
 
 import docker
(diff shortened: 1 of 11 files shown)


yannickperrenet:
> meaning no heartbeat from a client for 5 minutes

5 minutes sounds a little short to me, I would rather bump it to something like 15 (possibly even 30). Otherwise someone might start Orchest, prepare some food/open the door, come back and see Orchest being stopped.

Regarding the example JSON
```json
{
    "idle": false,
    "details": {
        "no_active_clients": false, 
        "no_ongoing_environment_builds": true, 
        "no_ongoing_jupyterlab_builds": true, 
        "no_ongoing_interactive_runs": true, 
        "no_ongoing_job_runs": true,
        "no_busy_kernels": true
        }
}
```
what do you think of removing the `no_` everywhere and negating their boolean value?

services/orchest-api/app/app/core/sessions.py
```diff
@@ -481,6 +481,44 @@ def launch(self, session_config: Dict[str, Any]) -> None:
...
+        """Tells if the session has busy kernels.
+
+        Args:
+            session_config: Requires a "project_uuid" and a
+            "pipeline_uuid".
+
+        """
+        IP = self.get_containers_IP()
+
+        self._notebook_server_info = {
```
yannickperrenet, line 494: How about:
```python
if self._notebook_server_info is None:
   # the code
```

fruttasecca: Sounds good, will do.

services/orchest-api/app/app/utils.py
```diff
@@ -711,3 +712,70 @@ def delete_dangling_orchest_images() -> None:
         env_uuid = img.labels.get("_orchest_environment_uuid")
         if env_uuid is None:
             docker_images_rm_safe(docker_client, img.id)
+
+
+def is_orchest_api_idle() -> dict:
```
yannickperrenet, line 717: I think it is actually `is_orchest_idle` since we also check for active clients.

fruttasecca:
>5 minutes sounds a little short to me, I would rather bump it to something like 15 (possibly even 30). Otherwise someone might start Orchest, prepare some food/open the door, come back and see Orchest being stopped.

Will do.

>what do you think of removing the no_ everywhere and negating their boolean value?

I kept second-guessing myself about this, happy to go for it.

services/orchest-api/app/app/utils.py
```diff
@@ -711,3 +712,70 @@ def delete_dangling_orchest_images() -> None:
...
+        See schema.idleness_check_result for details.
+    """
+    data = {}
+
+    # Active clients.
+    threshold = (
+        datetime.now(timezone.utc)
+        - current_app.config["CLIENT_HEARTBEATS_IDLENESS_THRESHOLD"]
+    )
+    current_app.logger.error(threshold)
```
yannickperrenet, line 730: Should this be `logger.debug` instead? Might be good to also add a message so we understand the log message.

services/orchest-api/app/app/utils.py
```diff
@@ -711,3 +712,70 @@ def delete_dangling_orchest_images() -> None:
...
+        )
+        session_has_busy_kernels = session_obj.has_busy_kernels(
+            {
+                "project_uuid": session.project_uuid,
+                "pipeline_uuid": session.pipeline_uuid,
+            }
+        )
+        data["no_busy_kernels"] = (
+            data["no_busy_kernels"] and not session_has_busy_kernels
+        )
```
yannickperrenet, line 759: Could change this slightly to introduce early stopping so we don't (in all cases) have to loop over all sessions.

services/orchest-webserver/client/src/App.tsx
```diff
@@ -43,6 +45,12 @@ const App = () => {
   // load server side config populated by flask template
   const { config, user_config } = context.state;
 
+  // Each client provides an heartbeat, used for telemetry and idle
+  // checking.
+  useInterval(() => {
+    makeRequest("GET", "/heartbeat");
+  }, 5000);
```
yannickperrenet, line 52: Sending the heartbeat can still be every 5s in my opinion.

services/orchest-api/app/app/utils.py
fruttasecca: That's just me forgetting my debugging stuff around :)), will remove.

services/orchest-api/app/app/utils.py
fruttasecca: Will do.

services/orchest-api/app/app/core/sessions.py
```diff
@@ -481,6 +481,44 @@ def launch(self, session_config: Dict[str, Any]) -> None:
...
+        )
+        response = requests.get(url, timeout=2.0)
+
+        # Expected format: a list of dictionaries.
+        # [{'id': '3af6f3b9-4358-43b9-b2dd-03b51c4f7881', 'name':
+        # 'orchest-kernel-c56ab762-539c-4cce-9b1e-c4b00300ec6f',
+        # 'last_activity': '2021-11-10T09:04:10.508031Z',
+        # 'execution_state': 'idle', 'connections': 2}]
+        kernels: List[dict] = response.json()
+        return any([kernel.get("execution_state") == "busy" for kernel in kernels])
```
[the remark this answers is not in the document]
fruttasecca: Will do.

iannbing (approved):
The changes in `App.tsx` looks good to me 👍

services/orchest-api/app/app/utils.py
```diff
@@ -711,3 +712,67 @@ def delete_dangling_orchest_images() -> None:
...
+        session_has_busy_kernels = session_obj.has_busy_kernels(
+            {
+                "project_uuid": session.project_uuid,
+                "pipeline_uuid": session.pipeline_uuid,
+            }
+        )
+        data["busy_kernels"] = data["busy_kernels"] or session_has_busy_kernels
+        if data["busy_kernels"]:
+            break
```
yannickperrenet, line 756: Could even be:
```python
if session_has_busy_kernels:
  data["busy_kernels"] = True
  break
```
which requires a little less mental overhead in my opinion.

yannickperrenet (approved):
Looks good to me :+1:

services/orchest-api/app/app/utils.py
fruttasecca: You are absolutely right, the simplest solution flew over my head :), wil do.

fruttasecca:
Note: I will postpone merging this while working on cloud-related tasks in case I need to do some adjustments.

PR #548, opened 2021-11-11
Improv/next scheduled job

fruttasecca:
## Description
This PR
- adds a `next_scheduled_job` endpoint that returns the next job that will be scheduled by the `orchest-api`
- refactors the `orchest-api` scheduler code so that the update to the `next_scheduled_time` of a job and the creation of job runs entries happens in the same transation

### Checklist


services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -80,6 +80,38 @@ def post(self):
...
+            next_job = next_job.filter_by(project_uuid=request.args["project_uuid"])
+
+        next_job = (
+            next_job.filter(models.Job.status != "DRAFT")
+            .filter(models.Job.next_scheduled_time.isnot(None))
+            # Order by time ascending so that the job that will be
+            # scheduled next is returned, even if the scheduler is
+            # lagging behind and next_scheduled_time is in the past.
+            .order_by(models.Job.next_scheduled_time)
+            .first()
```
yannickperrenet, line 105: Wouldn't it be faster to just get the minimum instead of having to sort it entirely?

I came across: https://stackoverflow.com/a/11846873/14839751 which looks promising.

yannickperrenet (approved):

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: I think query speed greatly depends on the indexes at hand in this case, i.e. I expect the postgres planner to come up with more or less the same plan for this, but I am happy to look into it since with the ORM layer you never know what happens.

yannickperrenet: It is at the core of the "auto start and stop" feature so I think it is important this piece of code is "optimized".

fruttasecca: I have looked into the matter, the planner is indeed optimizing correctly by making use of the index on the timestamp, given that we are looking for the minimum we also don't run into the issue of `NULL`s causing more overhead.

I have tried different alternatives when it comes to the queries (also looking at the link you provided), in all cases we not only need the timestamp but also the `uuid` of the job (and might need more in the future).

For the following plans, I had to turn sequential scanning off because given the low number of jobs in my database the planner would just go through the data sequentially.

The current implementation results in the following query (let's call it query A)
```
SELECT jobs.uuid AS jobs_uuid, jobs.next_scheduled_time AS
jobs_next_scheduled_time FROM jobs WHERE jobs.status != 'DRAFT' AND
jobs.next_scheduled_time IS NOT NULL
ORDER BY jobs.next_scheduled_time LIMIT 1
```

Other implementations that I have tried, making use of min, tend to result in queries that look like this (let's call this query B):
```
 SELECT jobs.uuid AS jobs_uuid, jobs.next_scheduled_time AS
 jobs_next_scheduled_time FROM jobs WHERE jobs.status != 'DRAFT' AND
 jobs.next_scheduled_time IS NOT NULL AND jobs.next_scheduled_time = (SELECT
 min(jobs.next_scheduled_time) AS min_1 FROM jobs WHERE jobs.status != 'DRAFT'
 AND jobs.next_scheduled_time IS NOT NULL) LIMIT 1;
```

Current indexes, query A
image

Current indexes, query B
image

I have added 2 new indexes that can better support the queries at hand, including filtering by project `uuid`, and if needed, filtering the `status` by equality instead of non-equality.
```
        Index("ix_jobs_next_scheduled_time_status", "next_scheduled_time", "status"),
        Index("ix_jobs_project_uuid_next_scheduled_time_status",  "project_uuid", "next_scheduled_time", "status"),
```
Query A plan when filtering without a project `uuid`
image
Query A plan when filtering with project `uuid`
image

Query B plans, without and with project `uuid`.
image
image

The implementation of query B at the ORM layer is a bit less readable and maintainable (imo).

yannickperrenet: Thanks for that elaborate reponse! :rocket: 

> The implementation of query B at the ORM layer is a bit less readable and maintainable (imo).

I agree.

yannickperrenet (approved):

Issue #425, opened 2021-08-19
Installing ipykernel with pip breaks the Orchest notebook runner


ricklamers:
**Describe the bug**  
Modifying the pip/conda environment breaks the Orchest notebook runner

**To Reproduce**  

1. Create a project
2. Create an environment that modifies the pip/conda packages in a way that break the dependencies of the notebook runner (which depends on e.g. nbconvert).
3. Try to run a pipeline step with a notebook
4. See failure (no logs in the UI, but in the Celery worker the notebook runner script shows a dependency related error)

**Expected behavior**  
Users modifying their dependencies shouldn't affect

**Environment**

- OS: linux
- Browser: chrome
- Version: 2021.08.3
- Environment: production
- Logs: (example dependency related failure)
```
Traceback (most recent call last):
  File "run.py", line 73, in <module>
    main()
  File "run.py", line 50, in main
    nr.run(file_path)
  File "/orchest/services/base-images/runnable-shared/runner/runner/runners.py", line 116, in run
    ep.preprocess(nb, {"metadata": {"path": self.working_dir}})
  File "/orchest/services/base-images/runnable-shared/runner/runner/preprocessors.py", line 95, in preprocess
    super().preprocess(nb, resources, km)
  File "/opt/conda/lib/python3.7/site-packages/nbconvert/preprocessors/execute.py", line 80, in preprocess
    with self.setup_kernel():
  File "/opt/conda/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/opt/conda/lib/python3.7/site-packages/nbclient/client.py", line 457, in setup_kernel
    self.start_new_kernel_client()
  File "/opt/conda/lib/python3.7/site-packages/nbclient/util.py", line 78, in wrapped
    return just_run(coro(*args, **kwargs))
  File "/opt/conda/lib/python3.7/site-packages/nbclient/util.py", line 57, in just_run
    return loop.run_until_complete(coro)
  File "/opt/conda/lib/python3.7/asyncio/base_events.py", line 583, in run_until_complete
    return future.result()
  File "/opt/conda/lib/python3.7/site-packages/nbclient/client.py", line 428, in async_start_new_kernel_client
    await ensure_async(self.kc.wait_for_ready(timeout=self.startup_timeout))
  File "/opt/conda/lib/python3.7/site-packages/nbclient/util.py", line 89, in ensure_async
    result = await obj
[...]

PR #571, opened 2021-12-01
Use separate Python environment to do pipeline runs

yannickperrenet:
## Description
Fixes: #425 

You can try it out by importing: https://github.com/yannickperrenet/orchest-notebook-runner-pr

### Todo
- [X] Run tests: inside Jupyter, as interactive runs and using other environments e.g. R and Python.
- [X] ~~Have a look at Docker's multi-stage builds to have a cleaner build process for the runnable images (now they contain a lot of duplicate code)~~ Not possible, see: https://github.com/moby/moby/issues/3378
- [x] Version lock the `requirements.txt` files in the runnable
- [x] (Optional) Make sure the `orchestdependencies` environment uses the same Python version instead of using `3.10`. This might lead to unexpected errors down the road.
- [x] I came across an alternative way to make sure we can skip cells tagged with `skip` (although this would make us dependent on the implementation of Jupyter's `NotebookClient`): `NotebookClient` which we subclass has an attribute `self.skip_cells_with_tags = ["skip"]`.

### DIFF
--- a/services/base-images/runnable-shared/runner/pyrightconfig.json
+++ b/services/base-images/runnable-shared/runner/pyrightconfig.json
@@ -1,0 +1,19 @@
+{
+  "extraPaths": ["../../lib/python/orchest-internals/"],
+
+  "exclude": ["**/node_modules", "**/__pycache__", ".venv"],
+
+  "reportMissingImports": true,
+  "reportMissingTypeStubs": false,
+
+  "pythonVersion": "3.8",
+  "pythonPlatform": "Linux",
+  "venvPath": "../../../../.venvs",
+  "venv": "base-images-runnable",
+
+  "executionEnvironments": [
+    {
+      "root": "app"
+    }
+  ]
+}
--- a/services/base-images/base-kernel-julia/Dockerfile
+++ b/services/base-images/base-kernel-julia/Dockerfile
@@ -25,11 +25,21 @@ RUN chown $NB_USER -R /orchest/orchest-sdk
 RUN chown $NB_USER -R /orchest/lib
 
 USER $NB_USER
-RUN pip install -r requirements.txt
+RUN pip install -r requirements-user.txt
+# Install Orchest dependencies in our own environment
+RUN conda create -y -n orchestdependencies && \
+    # requirements.txt includes local install and thus we need to use pip
+    conda install -y -n orchestdependencies pip && \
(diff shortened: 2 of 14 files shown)


services/base-images/runnable-shared/runner/pyrightconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extraPaths": ["../../lib/python/orchest-internals/"],
+
+  "exclude": ["**/node_modules", "**/__pycache__", ".venv"],
+
+  "reportMissingImports": true,
+  "reportMissingTypeStubs": false,
+
+  "pythonVersion": "3.8",
```
fruttasecca, line 9: Is there a particular place/file where we could put a comment to remind us to keep this in sync?

Sidenote, running
```python
from platform import python_version
print(python_version())
```
In a cell or a step reports `3.7.6`, while `python --version` in a terminal opened through jupyterlab reports `3.8.12`, is this expected? I have no reason to think this has changed because of this PR, just making sure this isn't an issue.

fruttasecca, line 9: Is there a particular place/file where we could put a comment to remind us to keep this in sync?

Sidenote, running
```python
from platform import python_version
print(python_version())
```
In a cell or a step reports `3.7.6`, while `python --version` in a terminal opened through jupyterlab reports `3.8.12`, is this expected? I have no reason to think this has changed because of this PR, just making sure this isn't an issue.

fruttasecca (approved):
lgtm

services/base-images/runnable-shared/runner/pyrightconfig.json
yannickperrenet: > Is there a particular place/file where we could put a comment to remind us to keep this in sync?

I don't really have any good ideas here. The `pyrightconfig.json` is not that important and only need to update once the parent images (in the `FROM`) of our services move to a newer version of Python.

> Sidenote

This is expected. We build the `jupyter-server` ourselves and use the same Python version as we use for all our other services. On the other hand, when running it inside a cell it is executed in the kernel image, which is build by the jupyter-enterprise-gateway project. I think it is fine for now (although not ideal), since all Python code we run as part of Orchest uses the same Python version and all user code is run using another Python version.

PR #582, opened 2021-12-07
Remove ability to update parameters through code using the SDK

yannickperrenet:
## Description
Removes `orchest.parameters.update_step_param` and `orchest.parameters.update_pipeline_param` as both don't have valid use cases:
* Dynamic parameter updating does not work since pipeline runs that are part of jobs work on their own copy of the snapshot, thereby not affecting other runs.
* Static parameter updating through code does not make sense, since that could would have to be deleted again. Its best to just use the UI.
* Since updating through code is not recommended, it should not be exposed in the first place.

### Checklist
- [X] In case I changed code in the `orchest-sdk`, I updated its version according to [SemVer](https://semver.org/) in its `_version.py` and updated the version compatibility table in its `README.md`


fruttasecca (approved):
lgtm :+1:

PR #583, opened 2021-12-08
Orchest core containers cpu priority over user Orchest containers

fruttasecca:
## Description
This will make it so that, during periods of CPU contention, core Orchest services are given priority over containers that the user has launched through Orchest. This is a minor QOL improvement especially aimed at the cloud. The PR is a single commit and the content is intentionally simple given the not too far k8s migration that might condition or be conditioned by such changes.

### Checklist

s.

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/config.py
+++ b/lib/python/orchest-internals/_orchest/internals/config.py
@@ -10,6 +10,15 @@ PIPELINE_FILE = "/pipeline.json"
 PIPELINE_PARAMETERS_RESERVED_KEY = "pipeline_parameters"
 CLOUD = os.environ.get("CLOUD") == "True"
 GPU_ENABLED_INSTANCE = os.environ.get("ORCHEST_GPU_ENABLED_INSTANCE") == "True"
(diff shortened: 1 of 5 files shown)


yannickperrenet:
I guess this change also makes the host machine more responsive as user code (launched in kernels) now get less CPU time in comparison to system commands issued on the host, e.g. `rm file`?

Have you played around with the new value for cpu shares? If not, then that is also fine :+1:

yannickperrenet (approved):
LGTM!

fruttasecca:
>I guess this change also makes the host machine more responsive as user code (launched in kernels) now get less CPU time in comparison to system commands issued on the host, e.g. rm file?

Yes that's an added benefit, although it might also be unexpected by the user. Perhaps this setting could/should be set at the default value (1024) when not in cloud mode, but I did not want to differentiate cloud vs non-cloud further.

>Have you played around with the new value for cpu shares? 

I was actually testing on the cloud with a much lower value (2) and did not notice any strange or unexpected behaviour. Happy to revisit this post  (or during) k8s migration.

yannickperrenet:
Sounds good. Let's merge :)

PR #584, opened 2021-12-08
Fall back on default values for invalidated global config options 

yannickperrenet:
## Description
We need to fall back on default values as otherwise the application won't start for people who had valid values in Orchest release X, but invalid (due to a change on our side) values in release X+i (where i >= 1)


fruttasecca (approved):
lgtm :)

Issue #567, opened 2021-11-26
Auto-align button breaks Orchest


yannickperrenet:
**Describe the bug**  
Using the auto-align button in certain edge cases breaks Orchest. 

(In other edge cases the auto-align button makes the steps unfindable, but sadly I can't reproduce this issue).

**To Reproduce**  
Steps to reproduce to break Orchest:

1. Open the pipeline editor
2. Create a new step (don't enter anything!) and keep the step open
3. Press auto-align button

Get a completely white screen in Orchest.

Steps to reproduce to make steps unfindable:

1. Open the pipeline editor
2. Create a new step (don't enter anything!) and close the step's properties pane
3. Press auto-align button

It actually deletes the step from the pipeline definition.

**Expected behavior**  

**Error message from the console**  
```
TypeError: Cannot read properties of undefined (reading 'incoming_connections')
```

PR #585, opened 2021-12-09
Fix edge cases for auto-align button

yannickperrenet:
## Description
Fixes: #567 

Also adds an offset so that steps are not placed behind the top buttons ("NEW STEP" etc).


ricklamers:
Your change missed an edge case I discovered earlier. I reinstated the original code. Think it should now cover the edge case you tried to fix too.

Issue #555, opened 2021-11-24
Project creation - Substitute unsupported characters with dashes/underscores


yannickperrenet:
**Describe the solution you'd like**  
When creating a project with an invalid name currently gives an error:

![image](https://user-images.githubusercontent.com/26223174/143219310-d031d56b-d3c4-4d8a-9329-d39dd43b7739.png)

The error can be hard to grasp for (non-)technical users and quite frankly can be a bummer if it is the first project you are creating. Instead of throwing an error, we can just substitute the invalid characters with underscores or dashes. For example: `my first project` automatically becomes `my-first-project`.

**What does your solution aim to solve?**  
Solving the issue for the user (with a sane default) instead of having the application surface an error.

yannickperrenet:
This issue is related (and would be a solution) to #521

Issue #521, opened 2021-10-25
Make invalid project name visually clear


ricklamers:
**Describe the solution you'd like**  
When you enter an illegal project name the input fields don't show visually which of the two is incorrect:

![image](https://user-images.githubusercontent.com/1309307/138682115-69714b79-1e59-493a-8036-c0f70d851961.png)
![image](https://user-images.githubusercontent.com/1309307/138682140-377b9c54-aa27-4727-b3d2-3a11135bcb2e.png)

We should visually change the invalid input to be red (or otherwise, as long as we clearly indicate that it's the faulty form field).

**What does your solution aim to solve?**  
Confusion about the form field validity.

PR #587, opened 2021-12-10
Substitute invalid characters on project name creation

yannickperrenet:
## Description
Resolves: #555, #521 

## Question about implementation
I left the `validateProjectNameAndAlert` code in there as I figured it is defensive programming. However, with this change that code should never run so removing it is also not a bad idea.

What do you think? @iannbing


yannickperrenet:
I ran the integration tests locally and all failures seem to be caused due to flakiness.

ricklamers:
LGTM! Nice QoL improvement.

ricklamers (approved):

Issue #406, opened 2021-08-02
Show warning when orchest.get_inputs() is called twice


ricklamers:
**Describe the solution you'd like**  
Show warning when orchest.get_inputs() is called twice

**What does your solution aim to solve?**  
When a pipeline is run as part of a job calling `orchest.get_inputs()` will fail the second time around because inputs are cleared on read. This does not show in interactive runs because the default there is to have the input stick around after read to facilitate the interactive development process.

ricklamers:
A warning message could be:

```
`get_inputs()` should only be called once. Inputs are cleared on read for inputs passed through the data passing memory server. This is done for efficiency when a pipeline is executed in non-interactive mode.
```

yannickperrenet:
Somewhat similar to: #219. Just mentioning as a reference.

PR #603, opened 2021-12-16
Improv/show warning when orchest get inputs is called twice

fruttasecca:
## Description

Makes it so that a warning will be printed to the user when the SDK `get_inputs` function is called twice. I am making a PR "eagerly" to discuss the current approach. 

Fixes: #406.

Note that this wouldn't work if the module is reloaded (https://docs.python.org/3/library/importlib.html#importlib.reload) (not tested, will do so tomorrow), but imo if a user is doing such tricks within Orchest he has already gone through the warning once, so I am not sure how much weight to give to `reload`.

Note: I am still testing this for edge cases when it comes to sys stdout redirection or other logging shenanigans.

### Checklist


yannickperrenet:
Using the current approach I guess that the warning is displayed when a cell containing the `get_inputs` is run twice? I don't think in this case we want to display the message. @ricklamers 

From the issue description I read it as having at least two cells that call `get_inputs` being an issue, since that would not work in jobs and therefore we want to warn the user beforehand that their script is invalid. But in the example I give, the script is not actually invalid.

fruttasecca:
>Using the current approach I guess that the warning is displayed when a cell containing the get_inputs is run twice? I don't think in this case we want to display the message. @ricklamers

Yes, you are right, this went over my head. I would say this is still preferable to not showing it in kernels, but it's a matter of opinion. Tomorrow I'll check if it's possible to detect this edge case but I think it's unlikely. A user could also call `get_inputs` twice in the same cell but through two different calls, which would make a job fail.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -891,7 +891,22 @@ def _resolve(
...
 
-def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> Dict[str, Any]:
+_times_get_inputs_called = 0
+_GET_INPUTS_CALLED_TWICE_WARNING = (
+    "WARNING: `get_inputs()` should only be called once. By default, in interactive "
+    "mode you can call it multiple times per step for ease of development, while in "
+    "non-interactive mode (jobs) you should always only call it once per step. The "
+    "reason is that, for efficiency, in non-interactive mode the inputs of a step are "
+    "deallocated the first time they are retrieved. To silence this warning, call "
+    "`get_inputs` with ``silence_multiple_calls_warning`` set to ``True``."
```
ricklamers, line 901: Not sure why you have `` `get_inputs` `` with single quotes and ``` ``silence_multiple_calls_warning`` ``` with double?

ricklamers:
I think since it's a warning I think it's fine to show it in interactive mode. Even though we don't know whether it's being called from the same cell in interactive mode, showing the warning to the user educates them that in non-interactive (jobs) pipeline runs the get_inputs() method should be called only once.

Two points:

1) I think we should (in this PR) show the warning about calling `orchest.output(...)` twice too (and use a similar mechanism)
2) I think we should print it to `stderr` instead of `stdout`

Other than that I like it @jacobodeharo :)!

fruttasecca:
>I think we should (in this PR) show the warning about calling orchest.output(...) twice too (and use a similar mechanism)

will do

>I think we should print it to stderr instead of stdout

absolutely

I was toying with the idea of actually making a directory in `/tmp` to catch the case of a module reload or a user opening a process which makes use of the orchest SDK from within a step. Not really sure it's worth the (small) increase in complexity.

orchest-sdk/python/orchest/transfer.py
fruttasecca: I treated `silence_multiple_calls_warning` as a value, i.e. the `True` that comes later, will fix.

yannickperrenet:
I generally don't like the idea of showing a warning to a user without them doing anything wrong, i.e. running the same cell containing `get_inputs` twice. Sure it educates them, but not in a particularly nice way imo.

Almost to the extend where we might as well show the warning every time the function is called.

fruttasecca:
>I generally don't like the idea of showing a warning to a user without them doing anything wrong, i.e. running the same cell containing get_inputs twice. Sure it educates them, but not in a particularly nice way imo.

We can fall back to not showing such warning in kernels theb, since the SDK has a way to find out if it is being called by a kernel, will do.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -891,7 +892,22 @@ def _resolve(
...
 
-def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> Dict[str, Any]:
+_times_get_inputs_called = 0
+_GET_INPUTS_CALLED_TWICE_WARNING = (
+    "WARNING: `get_inputs()` should only be called once. By default, in interactive "
+    "mode you can call it multiple times per step for ease of development, while in "
+    "non-interactive mode (jobs) you should always only call it once per step. The "
+    "reason is that, for efficiency, in non-interactive mode the inputs of a step are "
+    "deallocated the first time they are retrieved. To silence this warning, call "
+    "`get_inputs` with `silence_multiple_calls_warning` set to ``True``."
```
yannickperrenet, line 902: Alternative text:

```
"WARNING: `get_inputs()` should only be called once. By default, in interactive "
    "mode you can call it multiple times per step for ease of development, while in "
    "jobs it can only be called once per step due to deallocation after inputs are retrieved"
    "To silence this warning, call "
    "`get_inputs` with `silence_multiple_calls_warning` set to ``True``."
```

orchest-sdk/python/orchest/transfer.py
```diff
@@ -1040,7 +1063,20 @@ def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> Dict[str,
...
+    "calling the function once. To silence this warning, call `output` with "
+    "`silence_multiple_calls_warning` set to ``True``."
+)
+
+
+def output(
+    data: Any,
+    name: Optional[str],
+    silence_multiple_calls_warning: bool = False,
+) -> None:
```
yannickperrenet, line 1079: This also applies to `output_to_disk` and `output_to_memory` right?

yannickperrenet:
> We can fall back to not showing such warning in kernels then, since the SDK has a way to find out if it is being called by a kernel, will do

Don't think this makes sense though. Because using it in different cells will indeed lead to an error in jobs (this being the main painpoint we are trying to solve in this PR). So we need to be able to distinguish such cases.

orchest-sdk/python/orchest/transfer.py
fruttasecca: Wasn't so sure if that was needed, will do.

orchest-sdk/python/orchest/transfer.py
```diff
@@ -946,6 +964,11 @@ def get_inputs(ignore_failure: bool = False, verbose: bool = False) -> Dict[str,
         data or maintain a copy yourself.
 
     """
+    global _times_get_inputs_called
+    _times_get_inputs_called += 1
+    if not silence_multiple_calls_warning and _times_get_inputs_called > 1:
+        print(_GET_INPUTS_CALLED_TWICE_WARNING, file=sys.stderr, flush=True)
```
yannickperrenet, line 970: Does this color the message red? Like it would when using the `warning` module?

fruttasecca: No it will not, will try it out the `warning` module.

yannickperrenet (approved):
LGTM!

As a final idea, what about moving the note in the docstring (e.g. https://github.com/orchest/orchest/pull/603/files#diff-e5556021bc9658e2eb4369c29f18f0861528f4f00df8642b9e7af298b1354225R648) higher up in the docstrings? Probably makes more sense then having them at the bottom where it is less likely someone will read it.

fruttasecca:
>As a final idea, what about moving the note in the docstring (e.g. https://github.com/orchest/orchest/blob/improv/show-warning-when-orchest-get_inputs-is-called-twice/orchest-sdk/python/orchest/transfer.py#L1025) higher up in the docstrings? Probably makes more sense then having them at the bottom where it is less likely someone will read it.

Not sure if readthedocs is going to keep rendering the warning at the bottom of the function or not, nevertheless, I agree even if it does not affect readthedocs. I guess we want to give the same treatment to the "note" section of `output*` functions?

PR #604, opened 2021-12-17
Add `CONDA_ENV` to specify conda environment to use in runner

yannickperrenet:
## Description
Make it easier for users to install another Python version.

I will add an example to the docs on the `improv/docs` branch :+1:

### DIFF
--- a/services/base-images/base-kernel-julia/Dockerfile
+++ b/services/base-images/base-kernel-julia/Dockerfile
@@ -43,6 +43,11 @@ COPY ./runnable-shared/runner /orchest/services/base-images/runnable-shared/runn
 ENV JUPYTER_PATH=/opt/conda/share/jupyter
 ENV HOME=/home/$NB_USER
 
+# Environment variable that can be set to specify the conda environment
+# to use to start Jupyter kernels. Empty string will use the default
+# conda env.
+ENV CONDA_ENV=""
+
 ARG ORCHEST_VERSION
 ENV ORCHEST_VERSION=${ORCHEST_VERSION}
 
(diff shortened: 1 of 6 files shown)


ricklamers:
Not sure if we can _just_ modify the PATH variable instead of actually calling `conda activate`.

As per their [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment) they:

> Activation entails two primary functions: adding entries to PATH for the environment and running any activation scripts that the environment may contain. These activation scripts are how packages can set arbitrary environment variables that may be necessary for their operation. You can also use the config API to set environment variables.

yannickperrenet:
I see. I had a look at the `activate` script and figured I might as well just change the `PATH` myself. Will change it to activation of the environment and test it :+1:

yannickperrenet:
Changing the `else` in the bootscript to:
```bash
    source /opt/conda/etc/profile.d/conda.sh \
        && conda activate "$CONDA_ENV" \
        && /usr/local/bin/bootstrap-kernel.sh
```
ends up in the kernel not starting. Not really sure what is going on there.

(Also changed the `CONDA_ENV` in the Dockerfile to `"base"`)

yannickperrenet:
Well... we might have to find an alternative to `conda run`: https://github.com/conda/conda/issues/2904

yannickperrenet:
To test. Create an environment containing:
```python
conda create -y -n py38 python=3.8 future
conda install -y -n py38 ipykernel jupyter_client ipython_genutils pycryptodomex future
conda run -n py38 pip install orchest 
```

Next set two environment variables (I recommend using project environment variables):
* `CONDA_ENV=py38`
* `JUPYTER_PATH=/opt/conda/envs/py38/share/jupyter`

ricklamers:
Looks like `conda shell.bash hook` is recommended in a number of places and used by the `conda` project itself in various places (e.g. [here](https://github.com/conda/conda/blob/master/conda/core/initialize.py)).

LGTM like this!

Issue #601, opened 2021-12-16
Ability to delete pipeline runs as part of jobs


yannickperrenet:
**Is your feature request related to a problem? Please describe.**  
You might want to delete pipeline runs of jobs for numerous reasons:
* You want to keep a clean slate (maybe you only want to keep 20 at maximum - that could be another feature though)
* Delete the data of jobs (without deleting its logs)

Deleting jobs through the filesystem is pretty much impossible as you need to know the structure and a way to link the uuids with the correct run you want to delete.

**Describe the solution you'd like**  
Ability in the table to delete pipeline runs as part of the job.

In #398 we will introduce a way to abort an individual pipeline run. Would be great if it can also be deleted.

yannickperrenet:
Related to #550

yannickperrenet:
Has been added in #609.

Issue #550, opened 2021-11-16
Clean up old pipeline runs in jobs


ricklamers:
**Is your feature request related to a problem? Please describe.**  
When running a pipeline as part of a job, the job can create many pipeline runs when e.g. on a short cron schedule (run every 5 minutes). Each run has its own "checkout" of the project files. This can fill the disk pretty rapidly in some cases (large project files, e.g. `node_modules/` in some cases).

**Describe the solution you'd like**  
Introduce an option to retain only the last X number of pipeline runs. This automatically frees up space and allows cron jobs to run for longer durations more reliably.

Credits to Serhii Ostapchuk for bringing up the problem & idea.

PR #605, opened 2021-12-17
Feat/delete individual job pipeline run

fruttasecca:
Adds an endpoint to delete (as in, delete from DB and disk) a specific job pipeline run, the implementation is backed by a TwoPhaseFunction that can be later be reused for automatically cleaning old job pipeline runs, so this is related to both #601 and #550.

The endpoint is in the `orchest-api` as `/cleanup/<string:job_uuid>/<string:run_uuid>` and proxied by the webserver as `/catch/api-proxy/api/jobs/cleanup/<job_uuid>/<run_uuid>`.

The PR also fixes a couple of issues related to updating the end state of a Job.

### Checklist


services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -1004,6 +1030,36 @@ def _collateral(self, project_uuid: str):
...
+        AbortJobPipelineRun(self.tpe).transaction(job_uuid, run_uuid)
+
+        # Deletes cascade to: non interactive runs -> non interactive
+        # run image mapping, non interactive runs -> pipeline run step.
+        db.session.delete(run)
+        return True
+
+    def _collateral(self):
+        # The job run directory is removed by the webserver, since it
+        # owns it.
```
yannickperrenet, line 1059: The pipeline run directories are actually created by a celery task, but I agree it is cleaner to delete the directory in the webserver instead.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -1041,25 +1097,29 @@ def _transaction(self, job_uuid: str, run_uuid: str, status_update: Dict[str, An
 
             # Only non recurring jobs terminate to SUCCESS.
```
yannickperrenet, line 1098: I guess this comment can now be removed as it is no longer of value?

services/orchest-api/app/tests/namespaces/test_namespace_jobs.py
```diff
@@ -396,3 +398,104 @@ def test_pipelinerun_delete(
...
+        ).status_code
+        == 200
+    )
+
+    pipeline_runs = client.get(f"/api/jobs/{job_uuid}/pipeline_runs").get_json()[
+        "pipeline_runs"
+    ]
+    assert pipeline_runs[0]["status"] == "ABORTED"
+    assert pipeline_runs[1]["status"] == "PENDING"
+    assert pipeline_runs[2]["status"] == "PENDING"
```
yannickperrenet: Can't it happen that these pipeline runs are STARTED? Would it be better to check for `!= "ABORTED"`?

Or do we just use the fact that `celery` is mocked?

services/orchest-webserver/app/app/utils.py
```diff
@@ -554,6 +554,19 @@ def remove_job_directory(job_uuid, pipeline_uuid, project_uuid):
...
+
+    job_project_path = os.path.join(
+        current_app.config["USER_DIR"], "jobs", project_uuid
+    )
+    job_pipeline_path = os.path.join(job_project_path, pipeline_uuid)
+    job_path = os.path.join(job_pipeline_path, job_uuid)
+    job_pipeline_run_path = os.path.join(job_path, run_uuid)
+
+    if os.path.isdir(job_pipeline_run_path):
+        rmtree(job_pipeline_run_path, ignore_errors=True)
```
yannickperrenet, line 567: (You can ignore this message for this particular PR.)

I've actually had a question from a user asking whether we can make it possible to delete jobs, without deleting artifacts such as logs. This seems like the place to add it. Thoughts?

@ricklamers @fruttasecca

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -648,6 +649,47 @@ def catch_api_proxy_jobs_cleanup(job_uuid):
...
+                # Issue a cleanup of the job pipeline run to the orchest
+                # api, the run will be aborted if necessary.
+                resp = requests.delete(
+                    f"http://{current_app.config['ORCHEST_API_ADDRESS']}/api/"
+                    f"jobs/cleanup/{job_uuid}/{run_uuid}"
+                )
+
+                remove_job_pipeline_run_directory(
+                    run_uuid, job_uuid, pipeline_uuid, project_uuid
+                )
```
yannickperrenet, line 686: Before deleting the directory, shouldn't we first check the `resp` of the delete request?

services/orchest-webserver/app/app/analytics.py
```diff
@@ -38,6 +38,8 @@ class Event(Enum):
     JOB_CREATE = "job create"
     JOB_DELETE = "job delete"
     JOB_DUPLICATE = "job duplicate"
+    JOB_PIPELINE_RUN_CANCEL = "job pipeline-run cancel"
+    JOB_PIPELINE_RUN_DELETE = "job pipeline-run delete"
```
yannickperrenet, line 42: To follow the naming convention, how about renaming them to: `job-pipeline-run cancel` and `job-pipeline-run delete`.

That also brings me to another point. Internally we use "non-interactive run" as terminology, but doesn't really seem like we use it that often. It definitely is not helpful to users. How about doing "~non-interactive~ pipeline runs" and "interactive pipeline runs".

services/orchest-webserver/app/app/utils.py
yannickperrenet: I created a feature request for this: https://github.com/orchest/orchest/issues/610

yannickperrenet (approved):
Really well done! :)

You've caught a lot of edge cases here, which are hard to reason about. It almost scares me how many there are.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: That's correct, will fix the comment.

services/orchest-api/app/tests/namespaces/test_namespace_jobs.py
fruttasecca: >Or do we just use the fact that celery is mocked?

Yes that's what's currently happening. I would rather leave it as is because this asserts that the state has not changed.

services/orchest-webserver/app/app/views/orchest_api.py
fruttasecca: Yeah that makes sense, will do.

services/orchest-webserver/app/app/analytics.py
fruttasecca: >To follow the naming convention, how about renaming them to: job-pipeline-run cancel and job-pipeline-run delete.

Sure thing

>That also brings me to another point. Internally we use "non-interactive run" as terminology, but doesn't really seem like we use it that often. It definitely is not helpful to users. How about doing "non-interactive pipeline runs" and "interactive pipeline runs". @fruttasecca @ricklamers

Agree about "non-interactive run". I like "job pipeline runs" and "interactive pipeline runs". So we agree on one of the two points :), jokes aside "job pipeline runs" is quite long so I don't feel strongly about it, but I want to be sure users distinguish between the two.

services/orchest-api/app/tests/namespaces/test_namespace_jobs.py
yannickperrenet: :+1:

PR #608, opened 2021-12-21
Max job retained pipelined runs

fruttasecca:
## Description
This PR adds to the BE the functionality to automatically clean up the oldest job pipeline runs. This is done through a new field of the `Job` model, `max_retained_pipeline_runs`. This can be specified both during the creation of a job (POST) or update (PUT). This will make it so that 1) when new pipeline runs are scheduled for the job 2) when a pipeline run of the job gets into an end state the logic will check for the oldest job pipeline runs to delete. A requirement for a pipeline run to be deleted is to be in an end state (`SUCCESS`, `FAILED`, `ABORTED`).

Some details:
- the deletion is asynchronous to new runs being scheduled or a run getting to an end state, that is, the cleanup is not run in the same transaction but after. This is because the `orchest-api` does not control the`userdir`, but the `orchest-webserver` does.
- each job pipeline run (as before) has a `pipeline_run_index`, this is used to find out what are the oldest job pipeline runs to delete, i.e. `highest_index -  max_retained_pipeline_runs` would be the threshold to be considered for deletion
- among the pipeline runs considered for deletion, only the ones in an end state are considered
- `max_retained_pipeline_runs = 0` is a valid value, i.e. every run will cleanup itself after reaching an end state
- most likely, this setting makes sense only for cronjobs, but I have not put a constraint in place, keen to know your opinion about this being a FE "constraint" or an actual BE constraint

Related to #550 

@yannickperrenet  The deletion of the directory of a job pipeline run could have been done in different manners, I went for preserving the constraint that the `orchest-api` does not touch the `userdir`, and for reusing the existing `orchest-webserver` endpoint, so now the `orchest-api` is actually getting in touch with the webserver and it's quite awkward since it will result in a call to the `orchest-api`. Keen to know your opinion :).

:warning: The PR contains schema changes that will alter your DB if you build & run.

### Checklist


ricklamers:
Nice job! LGTM

services/orchest-api/app/app/models.py
```diff
@@ -404,6 +404,15 @@ class Job(BaseModel):
...
     )
 
+    # Max number of pipeline runs to retain. So that any newly created
+    # runs (e.g. in a cronjob) will lead to the deletion of the
+    # existing, oldest runs that are in an end state if the total number
+    # of job runs in the DB gets past this value. A value of -1 means
+    # that there is no such limit. The default value is -1.
+    max_retained_pipeline_runs = db.Column(
+        db.Integer, nullable=False, server_default=text("-1")
+    )
```
yannickperrenet, line 427: (Feel free to ignore.)

Funny, I look for `server_default` vs `default` and you get a post like: https://stackoverflow.com/questions/52431208/sqlalchemy-default-vs-server-default-performance. Showing that a (kw)arg can have quite a lot of impact on how something works and how you can interact with it.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -496,6 +497,66 @@ def post(self, job_uuid):
...
+def _delete_non_retained_pipeline_runs(job_uuid: str) -> None:
+
+    job = (
+        db.session.query(
+            models.Job.max_retained_pipeline_runs,
+        )
+        .filter_by(uuid=job_uuid)
+        .one()
+    )
+    max_retained_pipeline_runs = job.max_retained_pipeline_runs
```
yannickperrenet, line 510: Doesn't the query only return the `max_retained_pipeline_runs`?

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -954,6 +961,17 @@ def _transaction(
                 )
             job.strategy_json = strategy_json
 
+        if max_retained_pipeline_runs is not None:
```
yannickperrenet, line 1007: Don't we need to check that it is an integer `>= 0`?

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -496,6 +497,66 @@ def post(self, job_uuid):
...
+        .one()
+    )
+    max_retained_pipeline_runs = job.max_retained_pipeline_runs
+    current_app.logger.info(
+        f"Deleting non retained runs for job {job_uuid}, max retained pipeline "
+        f"runs: {max_retained_pipeline_runs}."
+    )
+    if max_retained_pipeline_runs < 0:
+        current_app.logger.info("Nothing to do.")
+        return
```
yannickperrenet, line 517: Here is a check for `< 0`, but I think it makes more sense to validate and not allow users to set an invalid value, i.e. `< 0`

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -496,6 +497,66 @@ def post(self, job_uuid):
...
+
+    # How many runs have been created for the job up to now.
+    max_job_runs_index = (
+        db.session.query(func.max(models.NonInteractivePipelineRun.pipeline_run_index))
+        .filter_by(job_uuid=job_uuid)
+        .one()
+    )[0]
```
yannickperrenet: I think this is exactly what [`.scalar()`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.scalar) does.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -496,6 +497,66 @@ def post(self, job_uuid):
...
+        base_url = f'{current_app.config["ORCHEST_WEBSERVER_ADDRESS"]}{path}'
+        resp = requests.delete(base_url)
+        # 404 because there could be concurrent calls to this.
+        if resp.status_code not in [200, 404]:
+            current_app.logger.error(
+                f"Unexpected status code ({resp.status_code}) while deleting run "
+                f"{run.uuid}."
+            )
+        else:
+            current_app.logger.info(f"Successfully deleted run {run.uuid}.")
```
yannickperrenet, line 555: So this is where we call the `orchest-webserver` from the `orchest-api`. In another comment on another PR ([reference[(https://github.com/orchest/orchest/pull/605#discussion_r773905421)) I figured it would be fine to do this, however, by making the relationship bidirectional we might make our life hard in the future. 

How do you feel about putting this in in a celery task instead?

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: It does, essentially I am avoiding the fetch of the other fields. But it does return an object with that property, and (AFAIR) accessing the other properties would make sqlalchemy query the DB.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: Indeed, there should be a check  (`>=-1`), will fix.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: I've made it so that the range of valid values is `>=-1`, with `-1` meaning that all job pipeline runs are retained, so in this case this would be valid.

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: This particular snippet is outdated, couldn't `fixup` the commit since it was worth it to keep track of the change, sorreh.

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: I think you are right :+1: 

I got confused by one of the examples in the SQLAlchemy docs :/

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: You are right :+1:

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: No worries. When I hit the later stages of the review I noticed it was gone, all good ^_^

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: I don't dislike the idea of putting it in a celery task but that:
- makes it asynchronous
- could add some decent overhead to the PR (1 more celery worker, 1 more, optional, model/table to keep track of the progress on this, etc. )

Not sure if it's worth it given that the incoming k8s port will force us to evaluate this again. But I agree that this bidirectional relationship should not be there.

This something that just came to my mind, what if we make use of the `file-manager` service that we are running? Would get rid of the bidirectional relationship while maintaining a synchronous nature.

fruttasecca: After further elaboration we decided to go for a celery task without adding another celery worker, celery queue and table in the DB.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -497,6 +497,101 @@ def post(self, job_uuid):
...
+        self.collateral_kwargs["pipeline_uuid"] = job.pipeline_uuid
+        self.collateral_kwargs["job_uuid"] = job_uuid
+        self.collateral_kwargs["pipeline_run_uuids"] = []
+
+        max_retained_pipeline_runs = job.max_retained_pipeline_runs
+        current_app.logger.info(
+            f"Deleting non retained runs for job {job_uuid}, max retained pipeline "
+            f"runs: {max_retained_pipeline_runs}."
+        )
+        if max_retained_pipeline_runs < 0:
```
yannickperrenet, line 524: I think checking for `-1` explicitly is even better, i.e. `if ... == -1`.

services/orchest-api/app/app/core/tasks.py
```diff
@@ -326,3 +327,24 @@ def build_jupyter(
...
+    pipeline_uuid: str,
+    job_uuid: str,
+    pipeline_run_uuids: List[str],
+) -> str:
+    """Builds Jupyter image, producing a new image in the docker env.
+
+    Returns:
+        Status of the environment build.
+
+    """
```
yannickperrenet, line 345: Docstring is incorrect.

yannickperrenet (approved):
This is awesome! :100:

services/orchest-api/app/app/core/tasks.py
fruttasecca: woopsy, fixing it

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: agree

PR #621, opened 2021-12-29
hide buttons if state.pipelineJson is undefined

iannbing:
## Description

`interactive-run.spec.ts` fails after `cy.reload()` because user could already attempt to press "NEW STEP" when `state.pipelineJson` is not yet loaded. Thus, the action button should only be rendered when `state.pipelineJson` is provided.


yannickperrenet (approved):
Integration tests getting less flaky. I like it!

PR #623, opened 2022-01-03
k8s PR

ricklamers:
## Description

Move from Docker to k8s

## Checklist

## Todo
- [x] Update docs. Example pages that need to be updated: Installation, how to migrate to Kubernetes, how to self-host Orchest.


fruttasecca:
The FM looks really really nice :+1:, congrats!

Some considerations, mostly opinions and UX related:
- it seems like it's possible to select a directory or file, once such a thing is selected, I expected the create new file/directory to create those into the directory I selected or the parent directory of the file I selected, but everything seems to be created by using the `project-dir` as the parent directory
- this seems the case for upload file and directory as well
- I tried to completely collapse the FM to hide it like it can be done in vs code, but the FM column ended up taking the entire editor space
- `right click -> view` doesn't allow to view arbitrary files in jupyterlab, while double-clicking opens them correctly

@yannickperrenet On file creation we check to make sure the user is not trying to get out of the project-dir, while we don't do so for delete, I haven't tried it but it looks like one could delete the `data` or `project-dir` directory with enough trying

yannickperrenet:
> On file creation we check to make sure the user is not trying to get out of the project-dir, while we don't do so for delete, I haven't tried it but it looks like one could delete the data or project-dir directory with enough trying.

I am currently preventing this from the webserver BE. Although the option to delete these files was already not presented by the FE, it is better to prevent it from the BE as well :+1:

ricklamers:
> are we sure we want to show dotfiles? I am not against it, perhaps making them greyed out ala vscode?

VS Code doesn't treat them differently by default. You're probably referencing the `.gitignore`'ed files which it makes darker. I think hence we can treat files/folders prefixed with a dot the same and for now we ignore `.gitignore` files.

image

ricklamers:
>it seems like it's possible to select a directory or file, once such a thing is selected, I expected the create new file/directory to create those into the directory I selected or the parent directory of the file I selected, but everything seems to be created by using the project-dir as the parent directory

I agree we can change the default in the future :+1:

docs/source/getting_started/installation.rst
```diff
@@ -93,84 +85,25 @@ how to do that for Ubuntu Linux.
...
    git clone https://github.com/orchest/orchest.git && cd orchest
 
    # Check out the version you would like to build.
-   git checkout v2022.03.0
+   git checkout v2022.03.6
 
-   # Create the Docker network on which all Orchest services will run.
-   docker network create orchest
+   # Activate `minikube`'s docker
+   eval $(minikube -p minikube docker-env)
```
nhaghighat, line 105: `minikube docker-env` only works in single node, it is good to mention it.

nhaghighat (approved):

PR #642, opened 2022-01-12
 Prepare Flask services to support multiple gunicorn workers

yannickperrenet:
## Description

In order to support multiple gunicorn workers the `create_app` function
can no longer handle DB migrations/upgrades and it is factored out
before the application is started.

In addition, because every gunicorn worker needs its own background
scheduler this should not lead to multiple executions of the same job
(since it is duplicated for every worker). To prevent this the jobs read
from the DB to make sure only one will end up running.

In this process the `is_werkzeug_parent` function was removed since it
was very confusing to work with. Its substitute is
`is_running_from_reloader`.

## Checklist


services/orchest-webserver/app/app/core/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+                # will awake at exactly the right time.
+                .filter(
+                    now
+                    >= models.SchedulerJob.timestamp
+                    + datetime.timedelta(minutes=interval)
+                ).first()
+            )
+
+            # The row was previsouly locked and thus another worker
+            # already handled the telemetry heartbeat signal.
```
fruttasecca: I think the comment could be slightly refactored, `job` will be `None` if no `SchedulerJob` passes the filter, the `with_for_update` is there to avoid race conditions leading to multiple executions. Just making sure a random person 2 months from now has a clear picture of the pattern we are using. Perhaps a module level docstring should explain this, since it's basically the pattern this module revolves around.

yannickperrenet:
I found an issue. The orchest-api no longer correctly deletes dangling resources after a restart, e.g.:
1. Session started
2. orchest restart
3. UI notes (due to response from orchest-api) that the session is still running whilst it isn't

services/orchest-webserver/app/app/core/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+        # Check whether there is already an entry in the DB, if not
+        # then we need to create it.
+        if query.first() is None:
+            db.session.add(models.SchedulerJob(type=job_type))
+        else:
+            now = datetime.datetime.now(datetime.timezone.utc)
+            job = (
+                query.with_for_update()
+                # TODO: check whether this works because the scheduler
+                # will awake at exactly the right time.
```
fruttasecca: Perhaps this warrants the scheduler running way more often, or detaching the scheduler time from the time that's used for filtering?

services/orchest-webserver/app/app/core/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+from _orchest.internals.two_phase_executor import TwoPhaseExecutor, TwoPhaseFunction
+from app import analytics, models, utils
+from app.connections import db
+
+logger = logging.getLogger("job-scheduler")
+
+
+class Scheduler:
+    @staticmethod
+    def handle_telemetry_heartbeat_signal(app: Flask, interval: int = 0) -> None:
```
fruttasecca, line 19: A default value of `0` seems a bit risky imo, perhaps no default value at all?

services/orchest-webserver/app/app/models.py
```diff
@@ -108,3 +110,28 @@ class BackgroundTask(BaseModel):
 
     def __repr__(self):
         return f"<BackgroundTask: {self.uuid}>"
+
+
+class SchedulerJobType(enum.Enum):
+    TELEMETRY_HEARTBEAT = "TELEMETRY HEARTBEAT"
+    ORCHEST_EXAMPLES = "ORCHEST EXAMPLES"
+
```
fruttasecca, line 118: You might have to merge `dev` into this brach and also take care of `POLL_ORCHEST_UPDATE_INFO_JSON`, see [PR](https://github.com/orchest/orchest/pull/625).

services/orchest-webserver/app/app/__init__.py
```diff
@@ -92,84 +154,35 @@ def create_app():
...
         # And every hour.
         scheduler.add_job(
-            fetch_orchest_examples_json_to_disk,
+            Scheduler.handle_orchest_examples,
             "interval",
-            minutes=app.config["ORCHEST_EXAMPLES_JSON_POLL_INTERVAL"],
-            args=[app],
+            minutes=interval,
+            args=[app, interval],
         )
```
fruttasecca, line 186: Related to https://github.com/orchest/orchest/pull/642#discussion_r783737504.

Perhaps we can push this complexity downwards to the scheduler module, having `create_app` only create the scheduler and add a single job that runs every minute or so, something along the lines of `scheduler.handle_recurring_tasks`.  So the webserver does not need to be informed about `POLL_ORCHEST_EXAMPLES_JSON` being true or not, in favour of deleting such decision.

services/orchest-webserver/app/app/core/scheduler.py
yannickperrenet: How about making it:
```python
                # Skip locked rows to prevent concurrent runs.
                query.with_for_update(skip_locked=True)
                # The scheduler will wake up at exactly the right time,
                # we still check so that slow concurrent runs don't also
                # run the collateral in case the row is no longer locked.
                .filter(
                    now
                    >= models.SchedulerJob.timestamp
                    + datetime.timedelta(minutes=interval)
                ).first()
            )

            # Another worker has already handled `handle_func`.
            if job is None:
                # Use kwarg instead of raising an error as an error
                # would be logged by the TPE.
                self.collateral_kwargs["run_collateral"] = False
                return
```

I will then also some additional information in the module docstring.

fruttasecca: Looks good, `skip_locked=True` is a nice addition as long as the scheduler runs often "enough".

fruttasecca:
Also, this PR warrants a bit more manual testing/QA than usual given that it's changing both migration and scheduler code. Same for the other PR.

services/orchest-webserver/app/app/core/scheduler.py
yannickperrenet: The idea of giving the interval a default value of `0` (I should probably add this in the docstring) is that on app initialization you want to execute the job immediately. With the default value it reads as `handle_telemetry_heartbeat_signal` which gives the idea it will handle it right away, whereas having to explicitly add the `interval=0` seemed too verbose.

services/orchest-webserver/app/app/models.py
yannickperrenet: Good one! Will do

services/orchest-webserver/app/app/core/scheduler.py
yannickperrenet: I have chosen for an offset to account for potential lag. Because now we would run the scheduled job on the given interval, whereas a scheduled job that runs more often would run on its short interval times the number of tasks it needs to check.

So I felt the offset approach is less overhead.

services/orchest-webserver/app/app/core/scheduler.py
fruttasecca: I see, makes sense.

PR #643, opened 2022-01-12
Feat/clean up finished pipeline runs

iannbing:
## Description

Make use of the new attribute `max_retained_pipeline_runs` of `/catch/api-proxy/api/jobs/{uuid}` to automatically clean up finished pipeline runs.

Minor improvements:
- Extract two Dialog components from JobList component
- fix: fuzzy_filter should be encoded
- fix: JobStatus icon mismatch

Fixes: #550 

## Checklist


services/orchest-webserver/client/src/jobs-view/JobList.tsx
```diff
@@ -0,0 +1,329 @@
+import { useAppContext } from "@/contexts/AppContext";
```
iannbing, line 1: @ricklamers Original `JobList` was way too big. Thus I split it into smaller components (including two dialogs `CreateJobDialog` and `EditJobNameDialog`), and group them in `jobs-view` folder. Due to files moving, diff is no longer visible. Sorry 😅

yannickperrenet:
Awesome work! I think you came up with a great solution of where to put the setting and how to let users edit it.

Two things I think require changes:

> The outcome of the deleted job is gone forever, then max_retained_pipeline_runs = 0 doesn't seem to produce anything useful for users.

(1) Now that I am using the feature I agree. I think can we can make `1` the lowest possible value. A possible value of `0` might make sense once we keep artifacts such as logs around for users to inspect even if the pipeline run was deleted.

(2) When editing an existing cron job, the new feature does not take its currently configured value. For example when creating the job to only keep the `5` most recent runs and then edit the cron job, you see that the setting is actually deselected and the default value is displayed. Instead the state of all settings should be correctly displayed when editing the job.

iannbing:
@yannickperrenet  nice catch! Indeed EditJobView didn't pick up the config. Just pushed the fix.

ricklamers (approved):

PR #644, opened 2022-01-13
Fix/job runs search ignore step uuids

fruttasecca:
## Description

Fixes an issue where the text search used for filtering job pipeline runs would consider the step uuids that are normalliy serialized with a pipeline run `parameters`.

## Checklist


yannickperrenet (approved):
LGTM! Migrations shouldn't be an issue as the migrations on `dev` where done on the `orchest-webserver`.

PR #645, opened 2022-01-14
Enable jupyterlab widgets by default

yannickperrenet:
## Description

Enables JupyterLab widgets by default.

It is a feature many people use and configuring JupyterLab through our UI is something not that many people do, so instead of users having to add:
```bash
pip install jupyterlab-widgets
```
it is baked into the image.


ricklamers:
Let's bump `jupyterlab-git` to latest (0.34.2). It's worth exploring whether that combination works with the pinned JupyterLab version.

yannickperrenet:
> Let's bump jupyterlab-git to latest (0.34.2). It's worth exploring whether that combination works with the pinned JupyterLab version.

Will test and add (but on a commit to `dev` not this branch).

PR #647, opened 2022-01-14
Fix cross-version data passing for Python

yannickperrenet:
## Description

Use `pickle.DEFAULT_PROTOCOL` for serialization so that data is serialized in a manner that is compatible between Python versions.

**Caveat**: If people were using Python3.8 in some of their steps (its an edge case since we ship with Python3.7.6) then they would be using pickle protocol 5 (the default protocol is 4). This means that data on disk (due to fallback or explicit `output_to_disk` calls) will not be accessible for them anymore. A workaround would be doing `pickle.DEFAULT_PROTOCOL=5`, but I think no one will really be impacted by it and can otherwise safely migrate since the pickled data on disk is mostly a "cache" anyways.

## Checklist


ricklamers:
LGTM!

PR #652, opened 2022-01-17
Update npm packages and improvements

iannbing:
## Description

- Update some heavily-used FE packages to more recent/stable versions, e.g. `swr`, `typescript`. Some necessary refactor was done to incorporate the breaking changes.
- break down big components to smaller ones: JobsView and PipelineSettingsView
- extract hooks for fetching data, e.g. `useFetchPipeline`, `useFetchJobs`, and unify their API
- refactor SessionsContext: move the side effects to useSessionsPoller and make SessionsContext a pure state machine. This also removed some Cypress flakiness
- move a side effect from Routes.ts to HeaderBar. This also fixed services.spec.ts failing.
- ~~previously FE polls sessions at all times, this PR makes it only do polling if the status of the session is transitional (i.e. `LAUNCHING` or `STOPPING`.~~

## Checklist


services/orchest-webserver/client/src/components/Layout/OnboardingDialog/OnboardingDialogCarousel/use-onboarding-dialog-carousel.ts
```diff
@@ -1,49 +1,34 @@
...
-  const initialState = { isAnimating: false, slide: [0, 0] };
+export type Slide = { index: number; direction: number };
 
-  const { data: state, mutate: setState } = useSWR(
-    "useOnboardingCarousel",
-    null,
-    {
-      initialData: initialState,
-    }
-  );
```
iannbing, line 14: this is an abuse of useSWR`, as it's not fetching any data, but was used as a shared state. Replace this with normal React.useState

services/orchest-webserver/client/src/hooks/useFetchJobs.ts
```diff
@@ -3,7 +3,7 @@ import { fetcher } from "@orchest/lib-utils";
 import useSWR from "swr";
 
 export function useFetchJobs(projectUuid: string | undefined) {
-  const { data, error, isValidating, revalidate } = useSWR<Job[]>(
+  const { data, error, isValidating, mutate } = useSWR<Job[]>(
```
iannbing, line 6: the new version of `useSWR` merge the old `revalidate` into `mutate`. Basically, calling `mutate` without providing and args is equivalent to `revalidate`.

services/orchest-webserver/client/src/components/HeaderBar.tsx
```diff
@@ -33,12 +32,24 @@ export const HeaderBar = ({
...
   const { navigateTo } = useCustomRoute();
-  const { state } = useProjectsContext();
+  const { state, dispatch } = useProjectsContext();
+  const location = useLocation();
   const appContext = useAppContext();
-  const {
-    state: { sessions },
-  } = useSessionsContext();
-  const currentSession = sessions.find((session) => isSession(session, state));
+
```
iannbing, line 38: Moved from `Routes.ts`

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
```diff
@@ -159,226 +144,157 @@ const initialState: SessionsContextState = {
...
-      ) || [];
-
-    dispatch({
-      type: "_sessionsSet",
-      payload: {
-        sessions: sessions as IOrchestSession[],
-        sessionsIsLoading: isLoading,
-      },
-    });
-  }, [data, isLoading]);
```
iannbing, line 221: this side effect is doing polling at all times, causing lots of re-rendering, and this side effect is triggered by state updates, and its `mutate` function also will trigger other updates. 
These side effects are moved to `useSessionsPoller`, and `SessionsContext` only maintains minimal amount of states that are needed across components.

services/orchest-webserver/client/src/components/Layout/OnboardingDialog/OnboardingDialogCarousel/use-onboarding-dialog-carousel.ts
ricklamers: Thanks! It was already contentious when it was first put in :)

ricklamers:
LGTM! Aside from the need to poll for the session state when there's a view that depends on the session state, as the state of the session can be changed by a user on a different device.

iannbing:
@ricklamers  thanks for clarifying. I remove the change and added a comment to explain.

services/orchest-webserver/client/src/hooks/useSessionsPoller.ts
```diff
@@ -30,22 +30,22 @@ function convertKeyToCamelCase<T>(data: T | undefined, keys?: string[]) {
...
       project_uuid: string;
       pipeline_uuid: string;
     })[];
     status: TSessionStatus;
   }>(ENDPOINT, fetcher, {
-    refreshInterval,
+    // Ideally we would like to only poll if session status is transitional, e.g. LAUNCHING, STOPPING.
+    // however, currently Orchest doesn't have user sessions (yet)
+    // this session (i.e. pipeline session) is shared by potentially multiple users
+    // in order to facilitate multi-user working at the same time, FE needs to check pipeline sessions at all times
```
ricklamers, line 47:
```
    // Ideally we would like to only poll if session status is transitional, e.g. LAUNCHING, STOPPING.
    // however, Orchest sessions are not bound to a single user, therefore
    // this session (i.e. pipeline session) is used by potentially multiple users
    // in order to facilitate multiple users working at the same time, FE needs to check pipeline sessions at all times
    ```

PR #654, opened 2022-01-19
Feat/helm charts

nhaghighat:
## Description

- Base helm charts added to /deploy folder
- Helm charts of thirdparty component that is used is added.
- nginx-proxy is replaced by K8S Ingress

## Checklist


services/nginx-proxy/orchest.conf
```diff
@@ -24,7 +24,7 @@ server {
 
     proxy_set_header Host $http_host;
 
-    set $auth_server "auth-server";
+    set $auth_server "auth-server.default.svc.cluster.local";
```
fruttasecca, line 27: Should the namespace be `orchest`?

fruttasecca:
I see that the `update-sever` is now a deployment, while in pre-k8s Orchest it is a container that is spawned dynamically and then stops (it actually stops itself). Is this intentional? I actually wouldn't mind having it as a deployment since on update we could replace the pod and the deployment will take care of running with the latest image etc. @yannickperrenet @ricklamers

services/nginx-proxy/orchest.conf
nhaghighat: Goof catch, this file needs to be deleted. because we don't use it anymore

nhaghighat:
> 

I thought update-server was a component which was running all the time, but I think having it as a deployment makes a lot of things easier.

fruttasecca:
>I thought update-server was a component which was running all the time, but I think having it as a deployment makes a lot of things easier.

I agree. Let's wait on @yannickperrenet and @ricklamers

yannickperrenet:
Having the `update-server` become a deployment works for me. Can't think of any immediate issues :thinking:

deploy/helm/charts/auth-server/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+description: The helm chart for aurh-server deployment
+name: auth-server
+version: 0.0.1
+icon: https://orchest.io/docs/source/img/logo.png
```
yannickperrenet, line 5: You probably wanted to link to: https://github.com/orchest/orchest/blob/master/docs/source/img/logo.png

:detective: Also the `description` contains a typo: `aurh-server` -> `auth-server`

deploy/Makefile
```diff
@@ -0,0 +1,43 @@
...
+ARGO_WORKFLOW_DEPLOY_CONFIG := argo-workflow \
+	${SHARED_DEPLOY_CONFIG} thirdparty/argo-workflows
+
+.PHONY: orchest
+orchest: argo-workflow
+	helm install ${ORCHEST_DEPLOY_CONFIG}
+
+.PHONY: jupyter-egw
+jupyter-egw:
+	helm install ${JUPYTER_EGW_DEPLOY_CONFIG}
```
yannickperrenet, line 31: Can we replace `jupyter-egw` with `jupyter-eg` since we have used that naming in other placing and also internally (in conversations etc.)

fruttasecca (approved):

deploy/helm/charts/auth-server/templates/deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+deployment
+*/}}
+---
+{{ include "library.deployment" . }}
+    spec:
+      volumes:
+        - name: userdir
+          hostPath:
+            path: /var/lib/orchest/userdir
+            type: DirectoryOrCreate
```
yannickperrenet, line 11: From the docs I read:
> If nothing exists at the given path, an empty directory will be created there as needed with permission set to 0755, having the same group and ownership with Kubelet.

Even though the mounted directory will always exist, what exactly would the group ID be? In the past permissions have been a issue so that's why I am asking :)

yannickperrenet: As a second question. Why is the path `/var/lib/orchest` used? Does this mean the directory will initially be created?

Currently the structure inside the `userdir` is important as no services (e.g. orchest-webserver) will create its structure if it doesn't exist. A good example is the jupyterlab user configurations that are persisted to disk.

@fruttasecca I remember you asking about the jupyterlab configurations. Was it related to this?

deploy/helm/charts/celery-worker/templates/deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+        - name: "MAX_JOB_RUNS_PARALLELISM"
+          value: "1"
+        - name: "MAX_INTERACTIVE_RUNS_PARALLELISM"
+          value: "1"
+        - name: "ORCHEST_LOG_LEVEL"
+          value: "INFO"
+        - name: "ORCHEST_HOST_GID"
+          value: "1"
+        - name: "ORCHEST_GPU_ENABLED_INSTANCE"
+          value: "FALSE"
```
yannickperrenet, line 30: @fruttasecca The celery-worker doesn't actually need this value right? It was also included the `spec.py` (of the `orchest-ctl`) but I don't know why it needs to be there.

deploy/helm/charts/orchest-webserver/templates/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+            path: /var/lib/orchest/userdir
+            type: DirectoryOrCreate
+        - name: config
+          hostPath:
+            path: /var/lib/orchest/config
+            type: DirectoryOrCreate
+        - name: repo
+          hostPath:
+            path: /var/lib/orchest/repo
+            type: DirectoryOrCreate
```
yannickperrenet, line 23: Might be good to add a comment explaining why the `orchest-webserver` needs this path (I want to mention it here also as it might change since we might no longer use the `orchest-ctl` when running on k8s). Basically the `orchest-webserver` gets the Orchest repo mounted at `/orchest-host` so that it can later pass it to the `orchest-ctl` when starting it (e.g. for updates).

@fruttasecca Correct me if I am wrong/forgetting something.

deploy/helm/charts/orchest-webserver/templates/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+      containers:
+      - name: orchest-webserver
+        image: orchest/orchest-webserver:latest
+        securityContext:
+          privileged: true
```
yannickperrenet, line 28: Why do we need the container a `privileged`?

deploy/helm/charts/celery-worker/templates/deployment.yaml
fruttasecca: It does to request the correct devices (i.e. GPU) in the pre k8s implementation. Can be removed but can also be kept to not have helm charts break orchest to avoid for an easier async development.

deploy/helm/charts/orchest-webserver/templates/deployment.yaml
nhaghighat: This is a debug leftover, will be removed

deploy/helm/charts/orchest-webserver/templates/deployment.yaml
fruttasecca: Correct AFAK

deploy/helm/charts/celery-worker/templates/deployment.yaml
nhaghighat: we can keep it for now, as this branch is going to be merged to `feat/k8s,` not `dev` or `master`, once we continued development, we can remove it later.

deploy/helm/charts/orchest-webserver/templates/deployment.yaml
nhaghighat: So should I remove this mounted directory from the charts?

fruttasecca: I think it can be removed indeed.

yannickperrenet: Probably best to keep it for now and make a note to check again once the port is done.

deploy/helm/charts/celery-worker/templates/deployment.yaml
fruttasecca: Agree

deploy/helm/charts/postgres/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+description: The helm chart for postgres deployment
+name: postgres
+version: 0.0.1
+icon: https://orchest.io/docs/source/img/logo.png
```
yannickperrenet, line 5: I think the icon points to the wrong location in all `Chart.yaml` files.

yannickperrenet:
Do we actually need all the files in the `deploy/thirdparty` directory? Or how does that work? Do we need to start maintaining all these files?

nhaghighat:
> Do we actually need all the files in the `deploy/thirdparty` directory? Or how does that work? Do we need to start maintaining all these files?

The thirdparty directory is added to have the helm charts of needed components, will add a note to mention we are not going to maintain it, but will bump the versions of charts occasionally.

deploy/helm/charts/update-server/templates/deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - name: update-server
+        image: orchest/update-server:latest
+        securityContext:
+          privileged: true
+          runAsUser: 0
+        ports:
+        - containerPort: 80        
+        env:
+        - name: "ORCHEST_PORT"
+          value: "8000"
```
yannickperrenet, line 22: @nhaghighat We need to make sure that this environment variable actually get's set when Orchest is launched. 

Because if we always have the `update-server` running then we need it to know what port Orchest is running on so that after the `update-server` restarts Orchest it again makes it run on the same port. 

Previously the `update-server` would not be running and so the `orchest-ctl` would pass it the right port (which it receives from the `orchest-webserver`).

Issue #614, opened 2021-12-23
Moving "deeper" in views hides left pane menu selection and project selector


yannickperrenet:
**Describe the bug**  
No items in the left pane menu stay selected (e.g. Select "Jobs" in menu when viewing a pipeline part of a job #350) after "moving deeper". For example after opening a pipeline the Pipelines menu item is no longer selected.

In addition, the project selector in the top bar is also "gone" when moving deeper. Now it is unclear what project you are working in.

**To Reproduce**  

1. Open a pipeline
2. See that *Pipelines* is not selected and that the project selected in the top nav bar is gone.

**Expected behavior**  
The item in the left pane menu should stay selected and the project selector needs to remain shown (the latter might have exceptions in certain views).

**Screenshots**  
Neither *Pipelines* is selected, nor the project name is visible. This happens across all view, e.g. when building an environment *Environments* is no longer selected and the project from the top nav bar also disappears. 

![image](https://user-images.githubusercontent.com/26223174/147252314-51a9b924-7f92-4e65-9695-162e3501523c.png)

PR #656, opened 2022-01-20
Fix/main drawer unable to track deeper view

iannbing:
## Description

With this PR, the main navigation bar always highlights the root item that user is currently browsing.

Fixes: #614 

## Checklist


yannickperrenet (approved):
Nice one! :100:

PR #662, opened 2022-01-24
Improv/move sessions poller to HeaderBar

iannbing:
## Description

Move useSessionsPoller into HeaderBar to easier session polling management.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/hooks/useSessionsPoller.ts
+++ b/services/orchest-webserver/client/src/hooks/useSessionsPoller.ts
@@ -34,7 +34,7 @@ function convertKeyToCamelCase<T>(data: T | undefined, keys?: string[]) {
 }
 
 /**
- * useSessionsPoller should only be placed in HeaderBar
+ * NOTE: useSessionsPoller should only be placed in HeaderBar
  */
 export const useSessionsPoller = () => {
   const { dispatch } = useSessionsContext();
@@ -44,7 +44,6 @@ export const useSessionsPoller = () => {
   } = useProjectsContext();
 
   const location = useLocation();
-
   const matches = matchPath(location.pathname, [
     siteMap.pipelines.path,
     siteMap.pipelineSettings.path,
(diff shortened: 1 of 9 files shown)


yannickperrenet (changes requested):
It is now no longer possible to do a jupyterlab build:

1. Start a session of an arbitrary pipeline
2. Go to Settings > Configure JupyterLab (button)
3. Press build
4. Get the prompt asking to stop all sessions
5. Note that you can't get rid of this prompt as it keeps reappearing.

yannickperrenet:
I am running into a few issues.

1. Start a session of an arbitrary pipeline
2. Go to Settings > Configure JupyterLab (button)
3. Press build
4. Get the prompt asking to stop all sessions -> confirm
5. (**first issue**) The "build" button now grays out and the build doesn't actually start
6. (**second issue**) Go to the "pipelines" view and notice how the session still appears to be running (even though it was just killed through the jupyterlab build confirm dialog)

Then lastly some other steps for a final issue. The state of the sessions toggle button in the "pipelines" view is not working correctly:
1. Go to pipelines
2. Start a session through the toggle
3. See that the session never appears to start (whilst when opening the pipeline you see it is indeed started)

Similarly
1. Go to pipelines
2. Open the pipeline and wait in the editor until the session is started
3. Stop the session
4. Go back to pipelines and see that the session is still "running" in the view.

services/orchest-webserver/client/src/hooks/useSessionsPoller.ts
```diff
@@ -30,17 +33,46 @@ function convertKeyToCamelCase<T>(data: T | undefined, keys?: string[]) {
...
   const { setAlert } = useAppContext();
+  const {
+    state: { pipelineUuid, pipelineIsReadOnly },
+  } = useProjectsContext();
+
+  const location = useLocation();
+  const matchRooViews = matchPath(location.pathname, [
+    siteMap.configureJupyterLab.path,
+    siteMap.pipelines.path,
+  ]);
```
iannbing, line 50: @yannickperrenet this was where went wrong. In these two root views, `pipline_uuid` is not needed. Previously I put them together in the match array below.

yannickperrenet (approved):
LGTM! 

If I am super precise I do have a small nitpick, but I will leave it up to you to decide whether it is something you feel needs a fix.

1. Open a pipeline
2. See that the button that indicates the status of a session has a small delay to display that the session is starting

Previously this was instant, whereas now it is a bit strange as the button seems to be pressing itself automatically (instead of being pressed automatically on view open).

yannickperrenet (approved):
LGTM! :1st_place_medal:

PR #664, opened 2022-01-24
Remove unused `/api-proxy/` location from nginx conf

yannickperrenet:
## Description

The rule has probably been there from the early days and was never
removed once we had an api-proxy inside the webserver.

@nhaghighat This means that the [orchest-api ingress](https://github.com/orchest/orchest/blob/feat/k8s/deploy/helm/charts/orchest-api/templates/ingress.yaml) is no longer needed, right?

### DIFF
--- a/services/nginx-proxy/orchest.conf
+++ b/services/nginx-proxy/orchest.conf
@@ -19,14 +19,14 @@ server {
   server_name localhost;
 
   error_log  /var/log/nginx.error.log debug;
-  
+
   location /login {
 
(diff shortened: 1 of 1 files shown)


nhaghighat:
> @nhaghighat This means that the [orchest-api ingress](https://github.com/orchest/orchest/blob/feat/k8s/deploy/helm/charts/orchest-api/templates/ingress.yaml) is no longer needed, right?

That indeed means the ingress for orchest-api can be removed.

Issue #414, opened 2021-08-09
Indicate update possibility due to new Orchest release in the UI


yannickperrenet:
**Example solution**  
A dialog indicating a new version.

![image](https://user-images.githubusercontent.com/26223174/128699819-4442527a-b99f-439a-9ee4-2669436a6ac7.png)

**What does your solution aim to solve?**  
It is beneficial for users to be on the latest version of Orchest. Currently, users need to check themselves whether Orchest has released a new version. In addition updating can also be a hidden feature to some.

ricklamers:
Love this idea ❤️

ricklamers:
(Nitckpick: button spacing)

yannickperrenet:
> (Nitpick: button spacing)

Yes that should definitely be improved

yannickperrenet:
Whoops I miss clicked and accidentally closed the issue. My bad.

PR #665, opened 2022-01-24
Indicate possibility to update if new update available

yannickperrenet:
## Description

Prompts only in a selection of views as to not interrupt users during work.

Closes: #414 

The easiest way to test things is to change [this line](https://github.com/orchest/orchest/blob/master/services/orchest-webserver/app/app/views/views.py#L239) to something like:
```python
        return {"version": "v2020.01.1"}
```
This should trigger the update dialog.

## Checklist


services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (latestVersion > currentVersion) {
+      if (skipVersion === latestVersion) {
+        return false;
+      } else {
+        return true;
+      }
+    } else {
+      return false;
+    }
+  };
```
iannbing, line 59: Usually we would move a pure function out of a React component.

```
const pureFunction = () => {...}
const ReactComponent = () => {...}

```

The reason is that the function within a React component are re-created per render. It would be a waste for pure functions.

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const shouldPrompt = shouldPromptOrchestUpdate(
+      currentVersion,
+      latestVersion,
+      skipVersion
+    );
+    if (shouldPrompt) {
+      promptUpdate(currentVersion, latestVersion);
+    } else {
+      if (!noOp) {
```
iannbing, line 102: perhaps just use `else if`?

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+    // calls to prevent updating the values which would trigger the
+    // useEffect and thereby prompting the user twice. In addition,
+    // we want to be able to tell the user that no update is available
+    // if this function is invoked.
+    const [orchestVersion, updateInfo]: [
+      OrchestVersion,
+      UpdateInfo
+    ] = await Promise.all([
+      fetcher("/async/version"),
+      fetcher("/async/orchest-update-info"),
```
iannbing, line 121: Feed the type to `fetcher` would make it more readable.
```
fetcher<OrchestVersion>("/async/version")
```
and Typescript can infer the type for the results, so you don't need to type the array in such a way.

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+    ]);
+    if (orchestVersion && updateInfo) {
+      handlePrompt(orchestVersion, updateInfo, null, false);
+    }
```
iannbing, line 125: `handlePrompt` is always wrapped by `if(orchestVersion && updateInfo)`. Maybe move this check inside of `handlePrompt`?

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+        cancelLabel: "Skip this version",
+      }
+    );
+  };
+
+  const handlePrompt = (
+    orchestVersion: OrchestVersion,
+    updateInfo: UpdateInfo,
+    skipVersion: string | null,
+    noOp: boolean
```
iannbing, line 89: it's a bit hard to reason what `noOp` means as a parameter of `handlePrompt`. Perhaps something more explicit, like `shouldPromptNoUpdate` (a reverse value of `noOp`)?

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -0,0 +1,135 @@
...
+        );
+      }
+    }
+  };
+  const checkUpdateNow = async () => {
+    // Use fetcher directly instead of mutate function from the SWR
+    // calls to prevent updating the values which would trigger the
+    // useEffect and thereby prompting the user twice. In addition,
+    // we want to be able to tell the user that no update is available
+    // if this function is invoked.
```
iannbing, line 115: Nice comment on why it was implemented like this 👍

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
yannickperrenet: I felt like that was less readable, but the fact that you propose doing an `else if` probably means I had it wrong ;)

Will change :+1:

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
yannickperrenet: I think I like it better having the check outside of `handlePrompt` :thinking: 

This way when quickly reading the `useEffect` it becomes obvious that `handlePrompt` won't run in case any of the values is undefined. Otherwise it is buried slightly deeper in the invocation order requiring you to check out the function as well.

iannbing: Makes sense 👍

iannbing (approved):
Nice job done!

astrojuanlu:
For future reference, this was the line to change to test this feature:

https://github.com/orchest/orchest/blob/354481ff355313b922f1f23a1614a95a5d539326/services/orchest-webserver/app/app/views/views.py#L239

Issue #639, opened 2022-01-11
Post project creation should take you to pipelines view


ricklamers:
**Describe the problem this improvement solves**
It can be confusing if the entry of the new project just appears somewhere in the table.

**Describe the solution you'd like**
Post project creation should take you to pipelines view

PR #666, opened 2022-01-24
#639 go to pipelines after creating project

iannbing:
## Description

Fixes: #639 

## Checklist

### DIFF
--- a/services/orchest-webserver/app/app/views/views.py
+++ b/services/orchest-webserver/app/app/views/views.py
@@ -552,10 +552,10 @@ def register_views(app, db):
 
     @app.route("/async/projects", methods=["POST"])
     def projects_post():
-
+        project_uuid = ""
         try:
             with TwoPhaseExecutor(db.session) as tpe:
-                CreateProject(tpe).transaction(request.json["name"])
+                project_uuid = CreateProject(tpe).transaction(request.json["name"])
         except Exception as e:
 
             # The sql integrity error message can be quite ugly.
@@ -568,7 +568,7 @@ def register_views(app, db):
                 500,
             )
 
-        return jsonify({"message": "Project created."})
+        return jsonify({"project_uuid": project_uuid})
 
     @app.route("/async/projects", methods=["DELETE"])
     def projects_delete():
--- a/services/orchest-webserver/app/app/views/views.py
+++ b/services/orchest-webserver/app/app/views/views.py
@@ -552,10 +552,10 @@ def register_views(app, db):
 
     @app.route("/async/projects", methods=["POST"])
     def projects_post():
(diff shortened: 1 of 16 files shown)


yannickperrenet:
Could we add the same behavior to creating a pipeline, i.e. when creating a pipeline it opens up the pipeline for you.

Makes sense to me if we include it in this PR as well, but I will leave that up to you @iannbing

services/orchest-webserver/app/app/views/views.py
```diff
@@ -552,10 +552,10 @@ def projects_get():
 
     @app.route("/async/projects", methods=["POST"])
     def projects_post():
-
+        project_uuid = ""
```
yannickperrenet, line 555: Strictly speaking this line is not necessary. 

Cool stuff that you are touching some Python btw :wink:

yannickperrenet:
After importing a project (when clicking "import project"), don't we also want to automatically go into the project?

services/orchest-webserver/client/src/projects-view/CreateProjectDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+      dispatch((state) => ({
+        type: "projectsSet",
+        payload: [
+          ...state.projects,
+          {
+            path: projectName,
+            uuid: project_uuid,
+            pipeline_count: 0,
+            job_count: 0,
+            environment_count: 1, // by default, a project gets an environment Python 3
```
yannickperrenet, line 63: Great comment :)

yannickperrenet (approved):
Apart from the aforementioned comments I think we can merge. Great job :100:

services/orchest-webserver/app/app/views/views.py
iannbing: @yannickperrenet Thanks 😄. I thought try/except would create a scope, so I have to declare `project_uuid` outside of it? Actually I think it would be even better moving line 571 inside of try (i.e. below line 558).

iannbing: Just pushed the proposed changes. Let me know if you agree :)

yannickperrenet: > I thought try/except would create a scope, so I have to declare project_uuid outside of it?

try/except does not create scope. In general, in Python you don't have variable scoping (except in functions, classes and a few others - but "blocks" don't create scope).

> Actually I think it would be even better moving line 571 inside of try (i.e. below line 558).

This is mostly a subjective manner I would say, but I agree ;)

iannbing:
> After importing a project (when clicking "import project"), don't we also want to automatically go into the project?
Also added this behavior.

PR #669, opened 2022-01-26
Don't try to fix userdir permissions in `--cloud`

yannickperrenet:
## Description

It shouldn't be needed to fix permissions anyways and for large userdirs this `find` command can take very long.

##### TODO
- [x] Fix integration tests
- [x] Depending on the fix for the tests we need to add `chmod g+s` to `./orchest install` to make sure the sticky bit is always set on the userdir.


fruttasecca (commented):
lgtm assuming the integration tests are not failing because of the change 

minor nitpick not related to this PR, can ` logger.info("Fixing permissions on the 'userdir/'.")` be moved one live above, before the fixing takes place?

yannickperrenet:
Hmmm, all integration tests pass when I run them locally (also when running them with `--cloud`).

PR #671, opened 2022-01-26
k8s api integration part 1

fruttasecca:
## Description

This is the first PR about integrating the k8s API. Changes related to this integration have been distributed across different branches, of the form `feat/k8s-api-integration-part-<number>`.

This PR contains the first changes to integrate with k8s, and introduces environment and jupyter images builds with `kaniko` + `argo`.

Some hackiness is involved, especially in the oldest branches, due to the need to get a vertical slice of a feature running. To try this out you will need to
- start the cluster, `minikube start`
- enable the registry, e.g. `minikube addons enable registry` 
- get the `registry` service ip and replace the hardcoded ip in `connections.py`  and `pipelines.py`, which is currently needed until an upcoming registry PR is merged
- build orchest images on the node, e.g. `eval $(minikube -p minikube docker-env)`, followed by `bash scripts/build_container.sh`
- set the pull policy to `IfNotPresent` for the orchest services of interest in the helm charts, so that the cluster will run with the code from the branch, without pulling the latest orchest relase
- deploy the helm chats, `make orchest` in `deploy`
- port forward the `orchest-webserver` service


services/orchest-api/app/app/connections.py
```diff
@@ -1,5 +1,6 @@
 from docker.client import DockerClient
 from flask_sqlalchemy import SQLAlchemy
+from kubernetes import client, config
```
yannickperrenet, line 3: Might make sense to slightly change the naming of the imports so we don't end up with collisions. For example we also already have `docker.client`. 

Maybe something like: `from kubernetes import client as kubernetes_client`.

services/orchest-api/app/app/core/docker_utils.py
```diff
@@ -1,18 +1,61 @@
...
+                }
+            ],
+            "dnsPolicy": "ClusterFirst",
+            "restartPolicy": "Never",
+            "volumes": [
+                {
+                    "name": "userdir",
+                    "hostPath": {
+                        "path": "/var/lib/orchest/userdir",
+                        "type": "DirectoryOrCreate",
```
yannickperrenet, line 60: Why is the entire `userdir` mounted? Aren't the `/data` directory and `/project-dir` enough?

First thought that I had when I saw the full `userdir` mount was that users can now (I guess) address all the content in the `userdir` when they are building using absolute paths. So let's say we introduce multi-user support where user X can't access a certain directory, then now they actually can.

services/orchest-api/app/app/core/docker_utils.py
```diff
@@ -57,76 +101,61 @@ def build_docker_image(
...
+        pod_name = f"env-build-{task_uuid}"
+        for _ in range(100):
+            resp = k8s_api.read_namespaced_pod(name=pod_name, namespace="argo")
+            status = resp.status.phase
+            if status in ["Running", "Succeeded", "Failed", "Unknown"]:
+                break
+            time.sleep(1)
+        else:
+            # Still Pending, consider this an issue.
+            raise Exception()
```
yannickperrenet, line 138: Probably a little early at this point, but my guess is that we will see the pattern of retrying and sleeping more often so might be good to put it inside a function.

services/orchest-api/app/app/core/docker_utils.py
```diff
@@ -57,76 +101,61 @@ def build_docker_image(
...
-                    complete_logs_file_object.flush()
-
+        w = watch.Watch()
+        for event in w.stream(
+            k8s_api.read_namespaced_pod_log,
+            name=pod_name,
+            namespace="argo",
+            follow=True,
+        ):
+            complete_logs_file_object.write(f"{event}\n")
```
yannickperrenet, line 151: I would be curious to see what the impact is on performance when changing this to (and in the other places where we write `event + \n`):
```python
complete_logs_file_object.write(event})
complete_logs_file_object.write("\n")
```
because when doing `complete_logs_file_object.write(f"{event}\n")` the entire `event` string will get copied to construct a new string (since strings are immutable).

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -205,15 +230,18 @@ def prepare_build_context(task_uuid, project_uuid, environment_uuid, project_pat
...
     check_environment_correctness(project_uuid, environment_uuid, userdir_project_path)
 
-    # make a snapshot of the project state
-    snapshot_path = f"/tmp/{dockerfile_name}"
+    if not os.path.exists("/userdir/.orchest/env-builds"):
+        os.mkdir("/userdir/.orchest/env-builds")
+    # Make a snapshot of the project state, used for the context.
+    snapshot_path = f"/userdir/.orchest/env-builds/{dockerfile_name}"
     os.system('rm -rf "%s"' % snapshot_path)
     os.system('cp -R "%s" "%s"' % (userdir_project_path, snapshot_path))
```
yannickperrenet, line 223: Do we want to use the reworked `copytree` function from the webserver (we can just put it in the internal lib) from https://github.com/orchest/orchest/pull/680?

It would also safe copying large files from the project-dir when doing builds (of course under the assumption that those files are ignored in the `.gitignore` file). Might be good to first reason whether users expect these ignored file to not be present when building :thinking:

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -205,15 +230,18 @@ def prepare_build_context(task_uuid, project_uuid, environment_uuid, project_pat
     Raises:
         See the check_environment_correctness_function
     """
+    create_mock_project(task_uuid, project_path, environment_uuid)
```
yannickperrenet: I guess this is for testing purposes? If so, maybe add a `K8S_TODO` comment like you did in other places.

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -319,8 +348,8 @@ def build_environment_task(task_uuid, project_uuid, environment_uuid, project_pa
                 abort_lambda=lambda: AbortableAsyncResult(task_uuid).is_aborted(),
             )
 
-            # cleanup
-            os.system('rm -rf "%s"' % build_context["snapshot_path"])
+            # Cleanup.
+            # os.system('rm -rf "%s"' % build_context["snapshot_path"])
```
yannickperrenet: I guess this was for testing as well?

services/orchest-api/app/app/core/environment_builds.py
yannickperrenet: Indeed. Removed in a later commit.

services/orchest-api/app/app/core/environment_builds.py
yannickperrenet: Indeed. Removed in a later commit.

yannickperrenet:
Great progress! :muscle: 

My comments are mostly about general cleanup that came to light (an not perse introduced) by these code changes.

services/orchest-api/app/app/core/docker_utils.py
fruttasecca: That's a very good point actually, I think it's preferable to do what you suggested for now, since we don't have enough things in place to propery benchmark this atm.

services/orchest-api/app/app/connections.py
fruttasecca: Yes I agree, I was too hasty on the naming here, will do.

services/orchest-api/app/app/core/docker_utils.py
fruttasecca: Most likely we only need such a thing in this place, but a bit more cleanliness doesn't hurt, will do.

services/orchest-api/app/app/core/docker_utils.py
fruttasecca: The reason is that pre-k8s builds were run by the `celery-worker`, whereas now  the `celery-worker` role in env builds is essentially a sidecar with some added logic, this means that the snapshot can't be kept at the `celery-worker` container level, but must be put in a place that the kaniko container will have to read.

I think we can restrict the mounting to `/userdir/.orchest/env-builds`.

image

yannickperrenet: > I think we can actually restrict the mounting to the exact build context, e.g. `/userdir/.orchest/env-builds/<build context uuid>`

That would be better I think. Although adding a mount for the `/data` might also be a good idea (actually not 100% sure whether we do that currently) so users can still access something from there for the build process, e.g. a binary.

services/orchest-api/app/app/core/docker_utils.py
yannickperrenet: > since we don't have enough things in place to propery benchmark this atm.

Yea I don't think we need to benchmark. If you think what I say makes sense, then I am happy to go for it :)

services/orchest-api/app/app/core/docker_utils.py
fruttasecca: Don't think we were explicitly providing `/data` access previously, but the entire `userdir` was being mounted to the `celery-worker`. I was going for a 1:1 mapping when it comes to some things for the sake of gettings things to run over introducing new stuff, but I am not against this, seems like a low hanging fruit, will do!

services/orchest-api/app/app/core/environment_builds.py
fruttasecca: I am not sure if any breaking changes have been introduced on `dev` since the k8s has been created, not against it. Perhaps we can make another PR merging the "rsync " branch with the k8s one or this branch?

yannickperrenet: I don't think any breaking changes were introduced.

> Perhaps we can make another PR merging the "rsync " branch with the k8s one or this branch?

Whatever you prefer really.

fruttasecca: I am moving `rmtree` and `cptree` to the internal lib since they might be useful across the board as you suggested. In this particular case I think the snapshot would already not contain any file that should be ignored since the snapshot was created using `cptree`.

yannickperrenet: > In this particular case I think the snapshot would already not contain any file that should be ignored since the snapshot was created using copytree

I agree. We could add a flag `ignore=False` which uses the probably faster `cp -R` command and `ignore=True` uses `rsync` to ignore from the `.gitignore`

yannickperrenet (approved):
LGTM! :sunglasses:

nhaghighat:
Great work indeed, LGTM

nhaghighat (approved):

PR #672, opened 2022-01-26
k8s api integration part 2

fruttasecca:
## Description

Second part of the k8s api integration. Makes it so that pipeline runs are run with argo instead of being run by the `celery-worker`.

Refer to https://github.com/orchest/orchest/pull/671 for instructions about trying this out.


yannickperrenet:
Regarding the `Use /project-dir/.orchest for plasma.sock` commit. This is what pre-k8s did not work on Mac, because the project directory lives on the host and is added to the virtualized environment of Docker Desktop. By placing the socket in that directory, the socket actually lives on the host instead of only in the virtualized environment. This then broke the communication over the socket.

@nhaghighat I guess when running k8s using docker desktop this would still lead to breakage?

fruttasecca:
>This is what pre-k8s did not work on Mac

I wasn't sure where such a volume would belong, i.e. helm charts etc., but I would be fine with having a dedicated volume to avoid the issue, so this pretty much depends on what @nhaghighat advises.

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -416,6 +416,33 @@ def get_environment_capabilities(environment_uuid, project_uuid):
...
+    project_dir,
+    pipeline_file,
+) -> List[dict]:
+    """Prepare all volume mounts that are needed to run Orchest."""
+
+    return [
+        {"name": "project-dir", "mountPath": project_dir},
+        {"name": "pipeline-file", "mountPath": pipeline_file},
+        {"name": "data", "mountPath": "/data"},
+    ]
```
yannickperrenet, line 442: These two functions are tightly coupled in since the `name` needs to be the same in both return values. However, from the implementation they are not tightly coupled so that should change in my opinion.

The fact that both function are called independently definitely makes this harder to achieve ^_^

services/orchest-api/app/app/core/tasks.py
```diff
@@ -86,53 +86,19 @@ async def get_run_status(
...
-        await asyncio.gather(
-            *[
-                asyncio.create_task(pipeline.run(task_id, run_config=run_config)),
-                asyncio.create_task(
-                    check_pipeline_run_task_status(run_config, pipeline, task_id)
-                ),
-            ]
-        )
-    # Make sure to cleanup containers in any case.
+        await run_pipeline_workflow(task_id, pipeline, run_config=run_config)
```
yannickperrenet, line 91: So far it has worked fine, but I think that we should use `eventlet` exclusively (which we use because of how we run the Flask services in production - using `gunicorn` and `eventlet`). The fact that we now mix `eventlet` and `asyncio` could lead to unexpected behavior from what I understand.

Don't think this should be fixed in this PR, but if you agree I can open an issue to have a look at this.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: Agree, will fix

services/orchest-api/app/app/core/tasks.py
fruttasecca: That's indeed something we can look into, I am not sure if the `celery-worker` process is actually affected by `eventlet` given that we are running in `fork` mode. tldr: I agree let's keep track of this in an issue.

yannickperrenet: https://github.com/orchest/orchest/issues/681

yannickperrenet:
In `pipelines.py` there are still a number of functions, which I don't think are used anymore: [L397](https://github.com/orchest/orchest/blob/feat/k8s-api-integration-part-2/services/orchest-api/app/app/core/pipelines.py#L397) until [L471](https://github.com/orchest/orchest/blob/feat/k8s-api-integration-part-2/services/orchest-api/app/app/core/pipelines.py#L471)

fruttasecca:
>In pipelines.py there are still a number of functions, which I don't think are used anymore: L397 until L471

AFAIR those have been removed in `feat/k8s-api-integration-part-3`, essentially I wasn't sure where we were going with the registry integration so those changes are a bit scrambled.

> I propose removing those and adding a new TypedDict for the run_config. I always wonder and always need to check the schema, API calls etc to be sure what is contained in the dictionary. Now that the entire file has gotten a big makeover I think it is time to address that pain as well :P

Would you mind if this happens in `feat/k8s-api-integration-part-3`? Less likely to get into conflicts etc. I will post a comment in that PR if you agree.

yannickperrenet:
We can move it to `feat/k8s-api-integration-part-3`. Totally works for me :+1:

yannickperrenet:
> But in k8s, the plasma might be started in another node, shouldn't we change it to listen over Ethernet networks?

Indeed probably best to expose the socket that Plasma is listening on so data can be send to it from steps directly :+1:

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -806,63 +470,332 @@ def remove_containerization_resources_on_docker(self, task_id, run_config):
...
+                            for step in pipeline.steps
+                        ],
+                    },
+                },
+                {
+                    "name": "step",
+                    "securityContext": {
+                        "runAsUser": 0,
+                        "runAsGroup": int(os.environ.get("ORCHEST_HOST_GID")),
+                        "fsGroup": int(os.environ.get("ORCHEST_HOST_GID")),
```
yannickperrenet, line 617: Not sure whether we want to let it fail or provide a default value to make sure the conversion to `int` always works.

yannickperrenet (approved):
LGTM! Just some minor changes following the aforementioned comments.

Don't see anything wrong with logic for running the pipeline workflows. Really well done :smile:. The logic has also become a lot easier to follow imo (expectable since we are essentially offloading the entire scheduling bit lol)

services/orchest-api/app/app/core/pipelines.py
fruttasecca: I have no strong feeling about this, a slight preference over letting it fail because fixing permissions once this is shipped can be more annoying to users

nhaghighat (approved):

PR #673, opened 2022-01-26
k8s api integration part 3

fruttasecca:
## Description

Third part of the k8s api integration.  Introduced changes:
- interactive runs are now running on k8s as argo workflows
- noninteractive and interactive sessions are now a bundle of k8s deployments+services
- the `sessions.py` module has been reworked to have a more functional approach. The main reason is to avoid dealing with 3 different states, DB, orchestration layer, class instance.
- each session runs in its own k8s namespace
- the session sidecar has been reworked since there is no `syslog-address` k8s equivalent. This is a temporary solution, the plan is to eventually use daemonsets to take care of collecting all logs

Things that are "running" but not working yet:
- jupyter-eg and jupyter-server, for which ingress needs to be setup
- user services that make use of `ports`, i.e. that are supposed to be reachable from outside the cluster, ingress needs to be setup
- the session-sidecar is not fully operative and tested, waiting on some permission changes in the helm charts

Refer to https://github.com/orchest/orchest/pull/671 for instructions about trying this out.


fruttasecca:
Note: add module-level docstrings to the `_core.py` sessions module and the `pipelines.py` module.

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -638,3 +638,37 @@ def docker_has_gpu_capabilities(
             pass
         return False
     return True
+
+
+def get_k8s_namespace_name(project_uuid: str, pipeline_or_run_uuid: str) -> str:
```
yannickperrenet: I am not entirely sure what I should pass as the `pipeline_or_run_uuid` argument. 

Questions I would have:
* Would they both yield the same result? Would be unlikely.
* Does the `run_uuid` take precedence if I have it? Otherwise there might be collisions, I am not sure.

I think it needs to be more explicit of what needs to be passed. Currently it feel like you can pass `pipeline_uuid` or `run_uuid` and get returned the same `k8s_namespace_name`.

yannickperrenet: Seems like this was changed in a later commit.

I think for this PR I will just go over all the changes instead of commit by commit due to all the changes :thinking:

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -638,3 +639,38 @@ def docker_has_gpu_capabilities(
...
+    """Get k8s namespace name given a uuid4 string.
+
+    Args:
+        uuid:
+
+    Returns:
+        Name of the namespace to use.
+    """
+    if len(uuid) != 36:
+        raise ValueError("Needs to be a uuid4 string")
```
yannickperrenet, line 654: Should probably say `"The given 'uuid' needs to be a uuid4 string"` so that if we later add another arg to the func, it is still clear what caused the ValueError

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -638,3 +639,38 @@ def docker_has_gpu_capabilities(
...
+
+def get_k8s_namespace_manifest(uuid: str, labels: Dict[str, str]) -> dict:
+    """Returns a k8s namespace manifest given the arguments.
+
+    The namespace name will be set according to get_k8s_namespace_name.
+
+    Args:
+        uuid of the namespace. Will be used for the name.
+        labels: dictionary mapping string to string. Will be used for
+        the namespace labels.
```
yannickperrenet, line 666: Styling needs to be fixed so that we could autogenerate docs (if we ever wanted to).

orchest-sdk/python/orchest/pipeline.py
```diff
@@ -8,23 +8,43 @@ class TypedDict(dict):
...
+    title: str
+    uuid: str
+
+
+class PipelineSettings(TypedDict):
+    auto_eviction: bool
+    data_passing_memory_size: str  # 1GB and similar.
+
+
+class ServiceDefinition(TypedDict):
```
yannickperrenet, line 27: How about moving these (and the other `TypedDict` definition) to a file like (open to suggestions here :P) `types.py` in the internal library. This way we only have to maintain them in one place and incentivize them to be used whenever applicable (otherwise you would have to define the type again in the resp. module).

yannickperrenet: Actually for the SDK we have to redefine them anyways, sorry.

We could still put the definitions from `pipelines.py` in a `types.py` though, right?

fruttasecca: Not against it, essentially we have to decide if every module defines its own types or if we have a "big" `types.py` module. I would be in favour of the latter given that the moment you have cyclic dependencies you end up with the same issue.

services/orchest-api/app/app/apis/namespace_runs.py
```diff
@@ -304,12 +304,14 @@ def _collateral(
         run_config["user_env_variables"] = env_variables
         # For interactive runs the session uuid is equal to the pipeline
         # uuid.
-        run_config["session_uuid"] = pipeline.properties["uuid"]
+        run_config["session_uuid"] = (
+            project_uuid[:18] + pipeline.properties["uuid"][:18]
+        )
```
yannickperrenet, line 309: Comment can get removed imo. Or altered in case you feel like it needs some explanation.

orchest-sdk/python/orchest/pipeline.py
fruttasecca: >We could still put the definitions from pipelines.py in a types.py though, right?

I guess that for non SDK code we can use a `types.py` module in the internal lib while the SDK stays the same until further notice

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+    INTERACTIVE = "interactive"
+    NONINTERACTIVE = "noninteractive"
+
+
+def _create_session_k8s_namespace(
+    session_uuid: str,
+    session_type: SessionType,
+    session_config: Dict[str, Any],
+    wait_ready=True,
+) -> dict:
```
yannickperrenet, line 25: Return type should be `None`

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+            sessions also include jupyter-eg and jupyter-server. These
+            services, along with any user defined service, can be
+            interacted with using the functions in this module through
+            their name.
+        session_config: A dictionary containing the session
+            configuration. Required entries: project_uuid,
+            pipeline_uuid , project_dir, host_userdir,
+            env_uuid_docker_id_mappings, session_type.
+            user_env_variables is a required entry for noninteractive
+            session type, while it's unused for interactive session
```
yannickperrenet, line 85: How about creating a `TypedDict` for `session_config` as well?

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+                    }}
+            }
+        should_abort: A callable that can be used to abort the
+            launch logic. When the callable returns True the launch
+            is interrupted. Note that no resource cleanup takes
+            place, and the caller of launch should make sure to call
+            the cleanup_resources method if desired.
+    """
+    if should_abort is None:
+        should_abort = bool  # Returns False.
```
yannickperrenet, line 137: Can we make this more explicit like:
```python
should_abort = lambda: False
```
because if we later call `should_abort(some_arg)` then it could actually be evaluated to `True`

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+    logger = utils.get_logger()
```
yannickperrenet, line 138: I think it is common practice to instantiate the logger at the top of the module.

services/orchest-api/app/app/core/sessions/_core.py
fruttasecca: That was my idea initially but flake8 didn't like that, will find a workaround.

services/orchest-api/app/app/core/sessions/_core.py
fruttasecca: I think the reason we were doing that already  in another function was circular dependencies, will verify.

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+                session_uuid, session_config, session_type.value
+            )
+        )
+        orchest_session_service_k8s_deployment_manifests.append(
+            _manifests._get_session_sidecar_deployment_manifest(
+                session_uuid, session_config, session_type.value
+            )
+        )
+    else:
+        raise ValueError(f"Invalid session type: {session_type}.")
```
yannickperrenet, line 157: We could also run a static type checker over the code, which should also catch these errors. Although on a different level.

I really like you checking for it explicitly though and our typing is probably not good enough at this point to use a type checker, but felt like mentioning it :)

fruttasecca: Yes I think we should at least transition to pydantic (or similar) eventually

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,362 @@
...
+    logger.info("Waiting for Orchest session service deployments to be ready.")
+    for manifest in orchest_session_service_k8s_deployment_manifests:
+        name = manifest["metadata"]["name"]
+        deployment = k8s_apps_api.read_namespaced_deployment_status(name, ns)
+        while deployment.status.updated_replicas != deployment.spec.replicas:
+            if should_abort():
+                return
+            logger.info(f"Waiting for {name}.")
+            time.sleep(1)
+            deployment = k8s_apps_api.read_namespaced_deployment_status(name, ns)
```
yannickperrenet, line 220: Why are we waiting and not proceeding with creating the user session services?

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -10,29 +24,65 @@
...
     parameters: Dict[str, Any]
-    settings: Dict[str, Any]
+    services: Dict[str, ServiceDefinition]
+    settings: PipelineSettings
+    steps: Dict[str, PipelineStepProperties]
+    uuid: str
+    version: str
+
+
+class RunConfig(TypedDict):
```
yannickperrenet, line 75: :partying_face:

services/session-sidecar/app/main.py
```diff
@@ -17,108 +23,84 @@ def get_log_dir_path():
...
-        default=Config.LOGS_PATH,
-    )
-    parser.add_argument(
-        "-p", "--port", type=int, required=False, default=Config.LISTEN_PORT
-    )
-
-    args = parser.parse_args()
-    return args
+                return
+            else:  # Pending
```
yannickperrenet, line 79: How about adding `logging.debug(f"{service} is pending")`

services/session-sidecar/app/main.py
```diff
@@ -141,11 +123,11 @@ def get_command_line_args():
...
-        server.request_queue_size = 1000
-        server.serve_forever()
+    logging.info(f"Storing logs in {logs_path}.")
+
+    services_to_follow = get_services_to_follow()
+    logging.info(
+        f"Following services: {services_to_follow} for {Config.SESSION_TYPE} session."
+    )
+    pool = ThreadPool(len(services_to_follow))
+    pool.map(follow_service_logs, services_to_follow)
```
yannickperrenet, line 133: Good solution for now :+1:

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -193,10 +192,28 @@ def _transaction(self, session_config: Dict[str, Any]):
...
-    def _background_session_start(cls, app, session_config: Dict[str, Any]):
+    def _should_abort_session_start(cls, project_uuid, pipeline_uuid) -> bool:
+        session_entry = (
+            models.InteractiveSession.query.options(
+                lazyload(models.InteractiveSession.image_mappings)
+            )
+            .filter_by(project_uuid=project_uuid, pipeline_uuid=pipeline_uuid)
+            .one_or_none()
+        )
+        # Has been stopped or is in the process of being stopped.
```
yannickperrenet, line 210: Nice catch! :fishing_pole_and_fish:

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -0,0 +1,709 @@
...
+
+        volumes_mounts["jupyterlab-user-settings"] = {
+            "name": "jupyterlab-user-settings",
+            "mountPath": "/root/.jupyter/lab/user-settings",
+        }
+
+        volumes_mounts["jupyterlab-data"] = {
+            "name": "jupyterlab-data",
+            "mountPath": "/data",
+        }
```
yannickperrenet, line 122: When I first read `jupyterlab-data` I did not guess it was about the `/data` directory. Maybe because I don't couple the `/data` directory to `jupyterlab` (whereas jupyterlab user-settings I do). 

How about `orchest-data-dir`?

fruttasecca: Didn't really think about it when porting the old volume bindings, will take a look at it & adjust.

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -0,0 +1,709 @@
...
+            session_config: See `Args` section in class :class:`Session`
+                    __init__ method.
+            service_config: See `Args` section in class :class:`Session`
+                    __init__ method.
+            session_type: Type of session: interactive, or
+                    noninteractive.
+
+    Returns:
+            Tuple of k8s deployment and service manifests to deploy this
+            user service in the session.
```
yannickperrenet, line 574: Indentation seems off here.

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca: I think the reason here is that in the docker implementation volume & volume mounts were represented by the same object, so the name made more sense. Now we have `volumes` and `volume_mounts` and their names need to match as well, and while for `volume_mounts` `jupyterlab-data` makes sense given that it specifies where it's going to be mounted in jupyterlab, it makes less sense when seen as `volume` name.

yannickperrenet:
> each session runs in its own k8s namespace

I guess this means that all of the following will run within the same namespace (created for the session):
* Orchest services (e.g. memory-server)
* User defined services
* Kernels as started by the jupyter-enterprise-gateway

Given that interactive pipeline runs run within a different namespace, do we address the user defined services with a fully qualified service name?

fruttasecca:
>I guess this means that all of the following will run within the same namespace (created for the session):

Correct

>Given that interactive pipeline runs run within a different namespace

Interactive pipeline runs actually run in the same namespace of the interactive session, if there is something suggesting otherwise it's either a bug or something whose name/docs can be improved.

orchest-sdk/python/orchest/pipeline.py
yannickperrenet: That all sounds great to me :)

services/orchest-api/app/app/core/sessions/_core.py
yannickperrenet: > Because user session services could depend on orchest session services

Forgot about that. This was only the case for the `session-sidecar` right? If that is the case, then I think it makes sense to remove the waiting behavior but I will leave it up to you to decide :)

services/orchest-api/app/app/core/sessions/_manifests.py
yannickperrenet: :+1:

yannickperrenet:
> if there is something suggesting otherwise it's either a bug or something whose name/docs can be improved.

I think it is just me due to all the namespace related changes scattered through the different integration branches. But good to know :+1:

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -8,120 +8,106 @@
...
-) -> Dict[str, Dict]:
-    # Make sure the same set of keys is available for all cases.
-    volumes_mounts = {
-        "kernelspec": None,
-        "jupyterlab-lab": None,
-        "jupyterlab-user-settings": None,
-        "jupyterlab-data": None,
+    volume_mounts["pipeline-file"] = {
+        "name": "pipeline-file",
+        "mountPath": container_pipeline_path,
```
yannickperrenet, line 39: Why do we need this mount? Isn't the pipeline-file included through the project-dir?

fruttasecca: There are a few implementation details/choices from the pre k8s version that made it this way:
- to, potentially, allow containers access to the pipeline definition without access to the project directory
- having the pipeline file accessible at `/pipeline.json` makes it consistent across different projects etc. for ease of debugging, etc.
- surely some other details that I am forgetting

services/orchest-api/app/app/core/sessions/_core.py
fruttasecca: Yeah that's fine with me, we can always get back to this in case something changes.

orchest-sdk/python/orchest/pipeline.py
fruttasecca: I have opted to keep the `types.py` module in the `orchest-api` rather than making it part of the internal library. A pattern that we have in the `orchest-webserver` is the augmentation of data when the webserver acts as a proxy between the client and the `orchest-api`, which would require some more thinking into how (if) we handle types globally, i.e. across all services in the internal lib. I also didn't want to make the PR grow even larger, and the introduction of types in the internal lib felt like a bit of a leap, I think that, if it needs to happen, it should happen in a separate PR where we also introduce `pydantic` or similar libraries.

Let me know if you agree.

yannickperrenet: > I have opted to keep the types.py module in the orchest-api rather than making it part of the internal library.

I like this a lot. Makes total sense :+1:

services/orchest-api/app/app/core/sessions/_manifests.py
yannickperrenet: > I think this also allowed the mount to "persist" across the pipeline file being moved, since bindings point to the inode, AFAIR, don't quote me on this

Pretty sure that is indeed the case (of it pointing to the inode). Could you add it as a comment (that it it works with inode and should allow the pipeline file being moved)?

@nhaghighat Or is this actually not the case?

nhaghighat: I guess if the project-dir is mounted, we don't need to mount this one.

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,296 @@
...
+        )
+
+    logger.info("Waiting for user and orchest session service deployments to be ready.")
+    for manifest in (
+        user_session_service_k8s_deployment_manifests
+        + orchest_session_service_k8s_deployment_manifests
+    ):
+        name = manifest["metadata"]["name"]
+        deployment = k8s_apps_api.read_namespaced_deployment_status(name, ns)
+        while deployment.status.available_replicas != deployment.spec.replicas:
```
nhaghighat, line 182: I guess there is a chance for null pointer exception here. the available replicas might be null in the beginning of deployment creation.

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca: I have actually tested this and this is the case indeed. So I don't see the reason to introduce more changes in the PR tbh, also considering point 1 and 2.

services/orchest-api/app/app/core/sessions/_core.py
fruttasecca: `available_replicas` is equal to `None` when there are no replicas, so I think this is safe

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -0,0 +1,296 @@
...
+    do by restarting it, since clearing would also lose all state.
+    Note that restarting the `memory-server` resets its eviction
+    state, which is exactly what we want.
+
+    """
+    ns = get_k8s_namespace_name(session_uuid)
+    old_replicas = k8s_apps_api.read_namespaced_deployment_status(
+        service_name, ns
+    ).spec.replicas
+    k8s_apps_api.patch_namespaced_deployment_scale(
```
nhaghighat, line 259: I guess deletion of the corresponding pod is better, then deployment creates a new one and then you can do it in one action

fruttasecca: You are right, I looked into this a bit more and there is a `delete_collection_namespaced_pod` function :+1:

nhaghighat (approved):

yannickperrenet (approved):

PR #677, opened 2022-01-27
Feat/docker registry

nhaghighat:
## Description

This PR add docker registry deployment to orchest with SSL.
The Kaniko should be created with additional argument :
`--registry-certificate=/etc/ssl/docker/ca.crt`
and with added mounted secret.
in volumes section:
`
        - name: registry-tls
          secret:
            secretName: registry-tls-secret
`

in volumeMounts section:
`            - mountPath: /etc/ssl/docker
              name: registry-tls-secret
              readOnly: true
`

## Checklist

- [ ] The PR branch is set up to merge into `k8s` instead of `master`.

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -21,10 +21,12 @@ SHARED_DEPLOY_CONFIG := --kubeconfig ${KUBECONFIG} \
 
 REGISTRY_SECRET := ${if ${REGISTRY_SECRET}, ${REGISTRY_SECRET},registry-tls-secret}
 REGISTRY_NAME := ${if ${REGISTRY_NAME}, ${REGISTRY_NAME},docker-registry}
+REGISTRY_PORT := ${if ${REGISTRY_PORT}, ${REGISTRY_PORT},443}
 
 # docker-registry deployment configs
 DOCKER_REGISTRY_SECRET_CONFIG := --set tlsSecretName=${REGISTRY_SECRET}
 DOCKER_REGISTRY_NAME_CONFIG := --set fullnameOverride=${REGISTRY_NAME}
+DOCKER_REGISTRY_PORT_CONFIG := --set service.port=${REGISTRY_PORT}
 
 
 # Orchest cluster resource deployment configs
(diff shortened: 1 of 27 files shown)


fruttasecca (approved):
Can't really comment much on the yaml files, they look correct. I have been developing against this PR/deployment and I haven't met issues.

About the third-party repositories which we store in our repo, is there a reason to not have those as submodules pointing to a specific commit? Aside from the extra step of pulling submodules

fruttasecca:
Forgot: can we make the registry service map 443:5000? That's the port that most clients work with by default

nhaghighat:
> Can't really comment much on the yaml files, they look correct. I have been developing against this PR/deployment and I haven't met issues.
> 
> About the third-party repositories which we store in our repo, is there a reason to not have those as sub modules pointing to a specific commit? Aside from the extra step of pulling sub modules

The main reason that I did not use sub modules, was that sometimes the third-parties are living as part of other git repositories, and it is not possible to have a submodule that points to subdirectories of a git repo.

deploy/Makefile
```diff
@@ -7,11 +7,74 @@ DEBUG := ${if ${DEBUG},--debug --dry-run,}
...
+
+# Orchest deployment configs
+ORCHEST_ENABLE_CONFIG := ${ORCHEST_API_ENABLE_CONFIG} \
+						 ${AUTH_SERVER_ENABLE_CONFIG} \
+						 ${AUTH_SERVER_ENABLE_CONFIG} \
+						 ${FILE_MANAGER_ENABLE_CONFIG} \
+						 ${ORCHEST_WEBSERVER_ENABLE_CONFIG} \
+						 ${POSTGRES_ENABLE_CONFIG} \
+						 ${RABBITMQ_ENABLE_CONFIG} \
+						 ${UPDATE_SERVER_ENABLE_CONFIG}
```
yannickperrenet, line 74: I guess this means that the `update-server` is started when Orchest is deployed? (Fine for now, just checking.)

nhaghighat: Indeed, and we have the option to disable it later

deploy/thirdparty/cert-manager/cert-manager.yaml
```diff
@@ -0,0 +1,5266 @@
+# Copyright 2022 The cert-manager Authors.
```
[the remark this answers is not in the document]
nhaghighat: Indeed, the additional configurations are in cluster-resources in helm directory.

deploy/helm/templates/_cluster-registry.tpl
```diff
@@ -0,0 +1,21 @@
...
+*/}}
+{{- define "library.cluster.registry.certificate" -}}
+{{- .Values.registryCertificate | default "registry-certificate" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Get the dns of docker-registry.
+*/}}
+{{- define "library.cluster.registry.dns" -}}
+{{ printf "%s.%s.svc.cluster.local" .Values.registry.name .Release.Namespace }}
```
yannickperrenet, line 12: We don't set this `.Values.registry.name` do we? Couldn't find it.

nhaghighat: It is seted here:
https://github.com/orchest/orchest/pull/677/files#diff-ca41056a0340e0139d50c70288e712c894b470842211c94554a1a493ed0164f7R35

yannickperrenet: Thanks. Missed that one.

yannickperrenet (approved):
Awesome stuff! :muscle:

PR #686, opened 2022-02-03
Fix read-only mode step status polling and pipeline settings

iannbing:
## Description

Fix unexpected behaviors in PipelineView read-only mode:
- run status polling was not started
- pipeline settings view was broken because env variables was still null in first several renders.

## Checklist


yannickperrenet (approved):
LGTM! 

(Didn't test it though.)

Issue #681, opened 2022-02-01
Use `eventlet` instead of `asyncio`


yannickperrenet:
**Describe the problem this improvement solves**
We currently run the Flask services using `gunicorn` with `eventlet`, therefore it is a good idea to only use `eventlet` and not use other libraries such as `asyncio`. 

**Describe the solution you'd like**
Use `eventlet` instead of `asyncio` everywhere throughout the code base. 

We could also conclude that we only adhere to this rule for the Flask services and not for others (e.g. the `celery-worker` can still use `asyncio`).

PR #689, opened 2022-02-04
Feat/k8s kaniko caching

fruttasecca:
## Description

This PR aims to reduce the build time of environment images. Three main issues so far:
- base image caching (solved by making use of kaniko warmer)
- layer caching and/or "kaniko unpacking rootfs as cmd copy requires it". Difficult since we don't know how aggressively we can cache.
- pushing to the registry is slow

Conclusions:
- pulling cached layers and uncompressing them takes a very long time
- pushing to the registry is still very slow and takes the lion share of the environment build time
- let's see if block storage helps in improving these pain points
- would be nice to be able to trim down our base images

Related kaniko issues:
- https://github.com/GoogleContainerTools/kaniko/issues/875
- https://github.com/GoogleContainerTools/kaniko/issues/1013
- https://github.com/GoogleContainerTools/kaniko/issues/1392
- https://github.com/GoogleContainerTools/kaniko/issues/970

## Remaining TODOs

- [x] make use of a labeled "builder" node?
- [x] endpoint & GUI settings button to clean up cached base images
- [x] try making the build completely reproducible (e.g. by removing references to the `task_uuid` in the setup script etc.) to see if it helps in avoiding a registry push if the image exists already and if the sha is the same
- [x] more fiddling around with kaniko `--flags` to improve performance

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_environment_images.py
+++ b/services/orchest-api/app/app/apis/namespace_environment_images.py
@@ -199,7 +199,7 @@ class DeleteBaseImagesCache(TwoPhaseFunction):
             models.JupyterBuild.status.in_(["PENDING", "STARTED"])
         )
         for jb in jupyter_builds:
-            AbortJupyterBuild(self.tpe)._transaction(jb.uuid)
+            AbortJupyterBuild(self.tpe).transaction(jb.uuid)
 
     def _collateral(self):
         image_utils.delete_base_images_cache()
--- a/services/orchest-api/app/app/apis/namespace_environment_images.py
+++ b/services/orchest-api/app/app/apis/namespace_environment_images.py
@@ -4,11 +4,14 @@ from flask_restx import Namespace, Resource
 from _orchest.internals import config as _config
 from _orchest.internals.two_phase_executor import TwoPhaseExecutor, TwoPhaseFunction
 from _orchest.internals.utils import docker_images_list_safe, docker_images_rm_safe
+from app import models
 from app.apis.namespace_environment_builds import (
+    AbortEnvironmentBuild,
     DeleteProjectBuilds,
     DeleteProjectEnvironmentBuilds,
 )
 from app.apis.namespace_jobs import AbortJob
+from app.apis.namespace_jupyter_builds import AbortJupyterBuild
 from app.apis.namespace_runs import AbortPipelineRun
 from app.apis.namespace_sessions import StopInteractiveSession
 from app.connections import db, docker_client
@@ -39,6 +42,25 @@ class EnvironmentImage(Resource):
             return {"message": str(e)}, 500
 
         return {"message": "Environment image was successfully deleted."}, 200
+
+
+@api.route(
+    "/base-images-cache",
+)
+class BaseImagesCache(Resource):
+    @api.doc("delete-base-images-cache")
+    def delete(self):
+        """Deletes the base images cache.
+
+        All ongoing environment and/or jupyter builds are cancelled.
+        """
+        try:
+            with TwoPhaseExecutor(db.session) as tpe:
+                DeleteBaseImagesCache(tpe).transaction()
+        except Exception as e:
+            return {"message": str(e)}, 500
+
+        return {"message": "Base images cache deletion successful."}, 200
 
 
 @api.route(
@@ -164,3 +186,20 @@ class DeleteImage(TwoPhaseFunction):
             DeleteImage._background_collateral,
             args=[current_app._get_current_object(), project_uuid, environment_uuid],
         )
+
+
+class DeleteBaseImagesCache(TwoPhaseFunction):
+    def _transaction(self):
+        env_builds = models.EnvironmentBuild.query.filter(
+            models.EnvironmentBuild.status.in_(["PENDING", "STARTED"])
+        )
+        for eb in env_builds:
+            AbortEnvironmentBuild(self.tpe).transaction(eb.uuid)
+        jupyter_builds = models.JupyterBuild.query.filter(
+            models.JupyterBuild.status.in_(["PENDING", "STARTED"])
+        )
+        for jb in jupyter_builds:
+            AbortJupyterBuild(self.tpe)._transaction(jb.uuid)
+
+    def _collateral(self):
+        image_utils.delete_base_images_cache()
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -113,6 +113,12 @@ def _get_image_build_workflow_manifest(
                             "--cache=true",
                             "--cache-dir=/cache",
                             "--use-new-run",
+                            # Note: make sure this runs at least with
+                            # kaniko 1.7, see
+                            # https://github.com/GoogleContainerTools/kaniko/pull/1735.
+                            # Essentially pre 1.7 there are some false
+                            # positive cache hits.
+                            "--cache-copy-layers",
                             # This allows us to simplify the logging
                             # logic by knowing that kaniko will not
                             # produce logs. If you need to restore the
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
 from typing import Optional
 
@@ -85,6 +86,10 @@ def _get_image_build_workflow_manifest(
     Returns:
         Valid k8s workflow manifest.
     """
+    verbosity = "panic"
+    # K8S_TODO: pass this env variable to the celery_worker.
+    if os.getenv("FLASK_ENV") == "development":
+        verbosity = "info"
     manifest = {
         "apiVersion": "argoproj.io/v1alpha1",
         "kind": "Workflow",
@@ -113,7 +118,7 @@ def _get_image_build_workflow_manifest(
                             # produce logs. If you need to restore the
                             # previous logic look for commit
                             # "Simplify image build logs logic".
-                            "--verbosity=panic",
+                            f"--verbosity={verbosity}",
                             "--snapshotMode=redo",
                             # From the docs: "This flag takes a single
                             # snapshot of the filesystem at the end of
@@ -274,6 +279,12 @@ def _build_image(
         build_context["snapshot_host_path"],
         build_context["dockerfile_path"],
     )
+
+    if os.getenv("FLASK_ENV") == "development":
+        msg = "Running in DEV mode, kaniko logs won't be filtered."
+        user_logs_file_object.write(msg)
+        complete_logs_file_object.write(msg)
+        complete_logs_file_object.flush()
 
     msg = "Building image...\n"
     user_logs_file_object.write(msg)
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -263,14 +263,16 @@ def _build_image(
     task_uuid,
     image_name,
     build_context,
-    dockerfile_path,
     user_logs_file_object,
     complete_logs_file_object,
 ):
     """Triggers an argo workflow to build an image and follows it."""
     pod_name = f"image-build-task-{task_uuid}"
     manifest = _get_image_build_workflow_manifest(
-        pod_name, image_name, build_context["snapshot_host_path"], dockerfile_path
+        pod_name,
+        image_name,
+        build_context["snapshot_host_path"],
+        build_context["dockerfile_path"],
     )
 
     msg = "Building image...\n"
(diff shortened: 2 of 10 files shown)


services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -41,6 +44,25 @@ def delete(self, project_uuid, environment_uuid):
...
 
+@api.route(
+    "/base-images-cache",
+)
+class BaseImagesCache(Resource):
+    @api.doc("delete-base-images-cache")
+    def delete(self):
+        """Deletes the base images cache.
+
+        All ongoing environment and/or jupyter builds are cancelled.
```
yannickperrenet, line 55: Makes total sense for now, but this will probably lead to a not-so-great experience in a multi-user context. Since this flow is similar for other things as well, e.g. JupyterLab builds, I think it is totally fine to leave it like this (just wanted to mention).

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -164,3 +186,20 @@ def _collateral(self, project_uuid: str, environment_uuid: str):
...
+        env_builds = models.EnvironmentBuild.query.filter(
+            models.EnvironmentBuild.status.in_(["PENDING", "STARTED"])
+        )
+        for eb in env_builds:
+            AbortEnvironmentBuild(self.tpe).transaction(eb.uuid)
+        jupyter_builds = models.JupyterBuild.query.filter(
+            models.JupyterBuild.status.in_(["PENDING", "STARTED"])
+        )
+        for jb in jupyter_builds:
+            AbortJupyterBuild(self.tpe)._transaction(jb.uuid)
```
yannickperrenet, line 202: This should probably call `.transaction` instead of `._transaction`.

services/orchest-api/app/app/core/tasks.py
```diff
@@ -288,3 +288,28 @@ def delete_job_pipeline_run_directories(
...
+    isn't on the same node of the celery worker given the type of
+    operation and it avoids 1 layer of indirection.
+    """
+    try:
+        with os.scandir(CONFIG_CLASS.BASE_IMAGES_CACHE) as entries:
+            for entry in entries:
+                if entry.is_dir() and not entry.is_symlink():
+                    shutil.rmtree(entry.path)
+                else:
+                    os.remove(entry.path)
```
yannickperrenet, line 312: Would it be better to use the `rmtree` from our internal library?

I would also be fine leaving it like this, but on the other hand using the internal library we treat the entire codebase the same. The fact that the celery-worker can run this command is that `eventlet` does not need to patch it, however that can also be seen as an advantage that we can use.

Not really a strong opinion. Can also be something to revision once we pick up https://github.com/orchest/orchest/issues/681. Probably best.

services/orchest-api/app/config.py
```diff
@@ -27,6 +27,11 @@ class Config:
     # activity.
     CLIENT_HEARTBEATS_IDLENESS_THRESHOLD = datetime.timedelta(minutes=30)
 
+    # Image building.
+    BUILD_IMAGE_LOG_TERMINATION_FLAG = "_ORCHEST_RESERVED_LOG_TERMINATION_FLAG_"
+    BUILD_IMAGE_ERROR_FLAG = "_ORCHEST_RESERVED_ERROR_FLAG_"
+    BASE_IMAGES_CACHE = "/tmp/kaniko/cache"
```
yannickperrenet, line 33: I remember reading about issues with using the `/tmp` directory for `gunicorn` but I guess that issue (making it slower) can also apply to Kaniko:

> in AWS an EBS root instance volume may sometimes hang for half a minute (...)

-- [source](https://pythonspeed.com/articles/gunicorn-in-docker/)

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: yep, will fix

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: Agree

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -97,22 +196,193 @@ def _generate_image_build_workflow_manifest(
...
+        complete_logs_file_object.write(msg)
+        complete_logs_file_object.flush()
+
+    # The w.stream loop exits once the pod has finished running.
+    resp = k8s_core_api.read_namespaced_pod(name=pod_name, namespace="orchest")
+    if resp.status.phase == "Failed":
+        msg = "There was a problem pulling the base image."
+        user_logs_file_object.write(msg)
+        complete_logs_file_object.write(msg)
+        complete_logs_file_object.flush()
```
yannickperrenet, line 269: Is there a particular reason we are never flushing to the `user_logs_file_object`?

services/orchest-webserver/client/src/views/SettingsView.tsx
```diff
@@ -484,6 +535,37 @@ const SettingsView: React.FC = () => {
           </div>
           <div className="clear"></div>
         </div>
+
+        <h3>Images Cache</h3>
+        <div className="columns">
+          <div className="column">
+            <p>Clear the base images cache to free up disk space.</p>
```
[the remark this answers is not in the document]
fruttasecca: I didn't go with "image cache" because it might imply more than we intend to users that know about the concept of a registry, cached layers etc. I don't feel strongly about it and would be fine going with "image cache".

yannickperrenet:
Please correct me if I am wrong in describing why the image building is slow (from what I understand).

Essentially, we have now added (in comparison to what we have pre-k8s) a registry that can be on another node and thereby introducing the overhead of pulling the base-image (from the registry) and pushing the resulting image (to the registry). When it comes to the actual layer caching it is pretty much similar as before due to how we create a new image (doing a `RUN` that always gets invalidated because of the `COPY . .`).

So if we want to make the environment building quicker we have to make working with the registry faster, which boils down to reducing the size of the images?

services/orchest-api/app/app/core/image_utils.py
fruttasecca: Performace reason AFAIR, since user logs are also sent through socketio etc, but don't quote me on that :).

services/orchest-webserver/client/src/views/SettingsView.tsx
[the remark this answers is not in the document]
yannickperrenet: The button is there to solve the edge case where dangling base images are present due to an update, right? If this is the case then we probably want to make sure users refrain from pressing this button, since pressing it often will always clear your cache making the builds even slower.

On the other hand, if I misunderstood the addition of the button and it is the case that we want users to press the button as often as possible then we might want to change the wording all together. It becomes more of a goal of what we are after (minimizing disk space) instead of providing the actual functionality to users.

services/orchest-api/app/app/core/image_utils.py
yannickperrenet: Ghehe okay. Let's leave it as is then ;)

yannickperrenet:
When it comes to flags, do we want to use [`cache-ttl-duration`](https://github.com/GoogleContainerTools/kaniko#--cache-ttl-duration)? Although would you say it does not apply to our situation?

services/orchest-webserver/client/src/views/SettingsView.tsx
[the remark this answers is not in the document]
fruttasecca: >The button is there to solve the edge case where dangling base images are present due to an update, right?

More or less. Essentially I didn't like the idea of `feat/k8s` (post merge) to write things to disk that cannot be deleted by the user, and other forms of handling would have required to implement the logic around updates and/or environment images lifecycle. So this GUI change is a sort of a band-aid that I hope we can get rid of later.

services/orchest-api/app/app/core/jupyter_builds.py
```diff
@@ -209,10 +207,18 @@ def build_jupyter_task(task_uuid):
...
             # We get here either because the task was successful or was
-            # aborted, in any case, delete the workflow.
+            # aborted, in any case, delete the workflows.
+            k8s_custom_obj_api.delete_namespaced_custom_object(
+                "argoproj.io",
+                "v1alpha1",
+                "orchest",
+                "workflows",
+                f"image-cache-task-{task_uuid}",
+            )
```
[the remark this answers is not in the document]
fruttasecca: It is indeed a leaky abstraction. I didn't want to rock the boat too much during the k8s migration. One reason why it is/was like this is that `SioStreamedTask` takes a given callable and runs it in a child process, in case the task is aborted `SioStreamedTask` will kill the child process, so the child process can't be trusted to cleanup after itself. We can't do away with this parent/child process and have a single process listen for abortion due to blocking code and/or the fact that it would probably make the code very ugly.

What we could do is to have some kind of `finally` argument for the `SioStreamedTask` although that wouldn't bring much too the table imo. I think this requires a bit more thinking and that the code does indeed need to be restructured but doing it in this PR would increase the scope a bit too much.

services/orchest-webserver/client/src/views/SettingsView.tsx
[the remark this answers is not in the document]
fruttasecca: >Makes sense. The current implementation removes all entries in the kaniko cache right?

The current implementation is pretty much a placeholder, I am waiting for https://github.com/orchest/orchest/pull/697 to see what's the cleanest implementation that can be used. When it comes to the interface and intention of this, currently it is indeed kaniko **base images** cache, so only base images in the base images cache directory. No layers/registry involved.

fruttasecca:
>When it comes to flags, do we want to use [cache-ttl-duration](https://github.com/GoogleContainerTools/kaniko#--cache-ttl-duration)? Although would you say it does not apply to our situation?

I played around with it, didn't seem to bring much to the table (e.g. it won't cleanup expired images or anything like that AFAIR). There were also some issues like https://github.com/GoogleContainerTools/kaniko/issues/896. I also don't think it applies to our situation atm.

fruttasecca:
> registry that can be on another node and thereby introducing the overhead of pulling the base-image (from the registry) and pushing the resulting image (to the registry)

The slowdown happens regardless of the registry being on another node, our current deployment  (or at least my local one) is single node, still, there is a huge overhead. So the base image caching seems to be a must-have in our case.

>When it comes to the actual layer caching it is pretty much similar as before due to how we create a new image (doing a RUN that always gets invalidated because of the COPY . .)

When it comes to layer caching that's indeed similar as before, but the RUN layer is not always invalidated, i.e if nothing changes in the project directory. Although there is quite an overhead in pulling and decompressing cached layers (I guess, again, because of image size) it's still faster than doing a run from scratch, and I guess this holds true if the RUN command actually does something that takes time.

>So if we want to make the environment building quicker we have to make working with the registry faster, which boils down to reducing the size of the images?

Yes, and hopefully block storage helps as well.

When it comes to the performance that we had pre k8s (pre kaniko, to be fair) I don't see us ever getting close to it at the moment, at least until the issues I linked at the top are solved, and some of them seem to require a paradigm change in kaniko, hopefully I am mistaken. 

These are some env build times I recorded, the setup script was something very simple, like an `echo` or `pip install requests` (can't really remember which is which).

Jupyter:
- Cached base image, no cached RUN, 1min 32s
- Cached base image, cached RUN, 1 min 18s

Environment build
- Cached base image, no cached RUN, 4m 30s (most time was spent pushing the image)
- Cached base image, cached RUN, at 4.20  it was done building, it skipped pushing because the resulting image was the same (most time was spent uncompressing the cached layer)

services/orchest-api/app/app/core/jupyter_builds.py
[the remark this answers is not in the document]
yannickperrenet: > so the child process can't be trusted to cleanup after itself.

Forgot about this. Makes total sense. Indeed best to leave it as is :+1:

services/orchest-webserver/client/src/views/SettingsView.tsx
```diff
@@ -266,6 +268,55 @@ const SettingsView: React.FC = () => {
...
+  const deleteBaseImagesCache = () => {
+    return setConfirm(
+      "Warning",
+      "Are you sure you want to clear the base images cache? This will abort all ongoing environment and jupyter builds.",
+      async (resolve) => {
+        setState((prevState) => ({
+          ...prevState,
+          deletingBaseImagesCache: true,
+        }));
+        resolve(true);
```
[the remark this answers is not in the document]
iannbing: Sorry for the late reply. You are correct. the wrapped the resolve function should be called in the end of the promise chain (i.e. line 293 or 296)

ohgenlong:
hope will fix asap. thank you so much

PR #692, opened 2022-02-07
Fix failing end-to-end tests for Orchest update checks

iannbing:
## Description

Cypress tests were failing because cy.intercept was not functioning.

## Checklist

### DIFF
--- a/cypress/cypress/integration/auth.spec.ts
+++ b/cypress/cypress/integration/auth.spec.ts
@@ -2,9 +2,9 @@ import { reset, TEST_ID } from "../support/common";
 
 describe("auth system", () => {
   beforeEach(() => {
+    cy.disableCheckUpdate();
     reset();
     cy.setOnboardingCompleted("true");
-    cy.disableCheckUpdate();
   });
 
   it("can create and delete a user", () => {
--- a/services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
+++ b/services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
@@ -15,19 +15,9 @@ const shouldPromptOrchestUpdate = (
   skipVersion: string | null = null
 ) => {
   // The latest version information has not yet been fetched by Orchest.
-  if (latestVersion === null) {
-    return false;
-  }
-
-  if (latestVersion > currentVersion) {
-    if (skipVersion === latestVersion) {
-      return false;
-    } else {
-      return true;
-    }
-  } else {
-    return false;
-  }
+  if (latestVersion === null) return false;
+  if (latestVersion <= currentVersion) return false;
+  return skipVersion !== latestVersion;
 };
 
 // To limit the number of api calls and make sure only one prompt is shown,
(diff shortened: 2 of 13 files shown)


services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -54,86 +44,90 @@ export const useCheckUpdate = () => {
...
-          setSkipVersion(updateInfo.latest_version);
-          resolve(false);
-          return false;
-        },
-        confirmLabel: "Update",
-        cancelLabel: "Skip this version",
-      }
-    );
-  };
+  const promptUpdate = React.useCallback(
```
iannbing, line 47: @yannickperrenet  the only difference is the wrapping `React.useCallback`. It memoize the function, so it won't trigger `React.useEffect` in every render

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -54,86 +44,90 @@ export const useCheckUpdate = () => {
...
+      } else if (shouldPromptNoUpdate) {
+        setConfirm(
+          "No update available",
+          "There doesn't seem to be a new update available."
+        );
+      }
+    },
+    [setConfirm, promptUpdate]
+  );
+
```
iannbing, line 109: @yannickperrenet same as above.

services/orchest-webserver/client/src/hooks/useLocalStorage.ts
```diff
@@ -26,17 +26,20 @@ export const useLocalStorage = <T>(key: string, defaultValue: T) => {
...
-      const valueToStore =
-        value instanceof Function ? value(storedValue) : value;
-      setStoredValue(valueToStore);
-      cachedItemString.current = JSON.stringify(valueToStore);
-      window.localStorage.setItem(privateKey, cachedItemString.current);
-    } catch (error) {
-      console.log(error);
-    }
-  };
+  const setValue = React.useCallback(
```
iannbing, line 29: @yannickperrenet same, wrapped it with `React.useCallback`

cypress/cypress/integration/auth.spec.ts
```diff
@@ -2,9 +2,9 @@ import { reset, TEST_ID } from "../support/common";
 
 describe("auth system", () => {
   beforeEach(() => {
+    cy.disableCheckUpdate();
```
iannbing, line 5: `cy.intercept` should be put as higher as possible to ensure that API stubbing kicks in. I noticed that many tests failed because of this (but many passed magically while the dialog is open)

yannickperrenet: Makes sense :+1:

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
iannbing: because other `React.useEffect` depends on `promptUpdate`, so if `promptUpdate` is re-created, it triggers `useEffect`. See this [codesandbox](https://codesandbox.io/s/bold-cannon-blx5c?file=/src/App.tsx) for the explanation. Note that in this codesandbox, it only cause one unnecessary render. In our case, it could cause indefinite rendering because the useEffect might cause more state updates.

yannickperrenet: Thanks for the extensive explanation and dedicated codesandbox :muscle:

yannickperrenet (approved):
LGTM! :+1:

PR #694, opened 2022-02-08
Do not run fix userdir permissions after restarting with `--cloud`

yannickperrenet:
## Description

In the edge case where Orchest is updated or restarted directly, the flag to skip the `find` command to fix userdir permissions was not passed. Now it is.

This resulted in the `find` command still running in certain cases, making the restart of Orchest (possibly) very slow.

Should've been caught in #669.


ricklamers:
LGTM!

PR #695, opened 2022-02-08
Fix page loading when visiting file preview, pipeline settings, logs with an interactive run

iannbing:
## Description

When trying to navigate to subsequent views of pipeline editor (i.e. pipeline settings, logs, file preview), the view would break because run uuid was passed along (thus it was considered as a job run, instead of an interactive run).

## Checklist


ricklamers:
LGTM!

ricklamers (approved):

PR #697, opened 2022-02-09
Feat/rook ceph

nhaghighat:
## Description

Please don't use the branch, until upcoming branch, for setting up the cluster itself.

## Checklist


deploy/helm/charts/cluster-resources/charts/ceph/values.yaml
```diff
@@ -23,6 +23,33 @@ cephBlockPools:
...
+  name: ceph-filesystem
+  # see https://github.com/rook/rook/blob/master/Documentation/ceph-filesystem-crd.md#filesystem-settings for available configuration
+  spec:
+    metadataPool:
+      replicated:
+        size: 3
+    dataPools:
+      - failureDomain: osd
+        replicated:
+          size: 3
```
fruttasecca, line 36: I think we want to set a `name` here
>name: (optional, and highly recommended) Override the default generated name of the pool. The final pool name will consist of the filesystem name and pool name, e.g., <fsName>-<poolName>. We highly recommend to specify name to prevent issues that can arise from modifying the spec in a way that causes Rook to lose the original pool ordering

https://github.com/rook/rook/blob/master/Documentation/ceph-filesystem-crd.md#pools.

fruttasecca:
Currently `orchest-ctl` updates orchest, argo, and the docker registry during an `orchest update`. Do we want to do the same for `ceph` and `rook`? I see that as a possible issue since `orchest-ctl` mounts the `userdir`.

nhaghighat:
Closed the PR as the changes merged by https://github.com/orchest/orchest/pull/742

PR #704, opened 2022-02-10
Always return identical error message on invalid login credentials

yannickperrenet:
## Description

This improves security as the attacker won't get additional information, e.g. whether a username actually exists or not.

### DIFF
--- a/services/auth-server/client/src/styles/main.scss
+++ b/services/auth-server/client/src/styles/main.scss
@@ -65,6 +65,7 @@ a.button {
     }
 
     .box {
+      max-width: 380px;
       padding: $generalPadding * 3 $generalPadding * 2;
       box-shadow: 0 15px 30px 8px rgba(0, 0, 0, 0.08);
       display: inline-block;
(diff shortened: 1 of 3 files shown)


services/auth-server/client/src/styles/main.scss
```diff
@@ -65,6 +65,7 @@ a.button {
     }
 
     .box {
+      max-width: 380px;
```
iannbing, line 68: @yannickperrenet not sure how this `max-width` would take effect with the given form ? The width of all the content in the form is quite small (around 300px).

iannbing: I see. It was the error message below. 👍

iannbing (approved):

services/auth-server/client/src/styles/main.scss
yannickperrenet: I've rounded the value of the box up to a "nice looking number" but the visual difference shouldn't be more than a few pixels.

PR #706, opened 2022-02-10
Fix command palette broken when entities with same name

iannbing:
## Description

This PR fixes the following bugs in Command Palette:
- broken if there are entities with same name in the same project
- unable to click to execute command

## Checklist


yannickperrenet (approved):
:100:

PR #709, opened 2022-02-11
Improve error handling for deleting selected rows

iannbing:
## Description

Currently DataTable doesn't handle errors internally when deleting selected rows. This might cause "frozen" state if failed to delete selected rows within DataTable.

## Checklist


yannickperrenet (approved):
LGTM!

I was wondering whether we can now consider https://github.com/orchest/orchest/issues/693 resolved?

> the only possibility to freeze DataTable is that the request is never resolved

Seems like an unlikely edge case right?

Issue #375, opened 2021-07-08
Clashing Service Names


joe-bell:
**Describe the bug**  

As found by @fruttasecca – when a service is edited to have the same name as another, it will completely disappear from the view

**To Reproduce**  

1. Go to a pipeline
2. Click on settings
3. Create a service called `redis`
4. Create another service called `redis1`
5. Rename `redis1` to `redis` to see the bug

**Expected behavior**  
Clashing names shouldn't cause services to remove, we should either:
1. Alert the user to choose a different name
2. Use a UUID/key for each service to prevent removal

**Screenshots**  
N/A

**Environment**

N/A

**Additional context**

PR #710, opened 2022-02-11
Fix service name collision or breakage

iannbing:
## Description

This PR fix the issue for service name collision and breakage.
If user attempts to rename a service with an existing name, the form will show a proper error message and disable all other fields until user gives a unique name.

Fixes: #375 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/components/ServiceForm.tsx
+++ b/services/orchest-webserver/client/src/components/ServiceForm.tsx
@@ -24,6 +24,7 @@ import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import { visuallyHidden } from "@mui/utils";
 import {
+  hasValue,
   makeCancelable,
   makeRequest,
   PromiseManager,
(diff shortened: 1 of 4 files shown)


iannbing:
Note that the uuid of service is created by FE, which is not ideal. I didn't look deeper to check if how BE process the services. @fruttasecca Feel free to chip in.

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -578,10 +578,7 @@ def is_service_definition_valid(service: Dict[str, Any]) -> bool:
 
 def is_services_definition_valid(services: Dict[str, Dict[str, Any]]) -> bool:
     return isinstance(services, dict) and all(
-        [
-            is_service_definition_valid(service) and sname == service["name"]
-            for sname, service in services.items()
-        ]
+        [is_service_definition_valid(service) for _, service in services.items()]
```
fruttasecca, line 581: I think it might disrupt some existing logic where it is/was assumed that the dictionary key and service name were interchangeable. Regardless of this, it might not be feasible because then users won't be able to resolve (i.e. dns) to a given service from within a step/kernel, or the logic around that needs to get more complicated. 

Essentially we want users to be able to refer to service "redis" as `redis`, or at least that's what happens currently in the k8s branch, while prek8s it's something like `service-<project uuid>-<pipeline uuid>` or something like that.

iannbing: Hmm, this means we actually don't allow users to have multiple instances of the same service in a pipeline?

Issue #617, opened 2021-12-24
Session/JupyterLab session start should be gated by environment builds


ricklamers:
**Describe the problem this improvement solves**
When you open the pipeline editor/run a job the environment gate is used to validate if all environments are ready. This doesn't happen when you start a session through the JupyterLab view or pipeline list view with the session toggle.

**Describe the solution you'd like**
Session start/JupyterLab session start should be gated by environment builds like the pipeline editor.

PR #712, opened 2022-02-11
Gate the session toggle

yannickperrenet:
## Description
When user wants to start a session using the session toggle in the pipelines view, then starting the session should also be gated (checked whether all environments have been build).

Fixes: #617 

## How to test
The easiest way to test is changing [L106](https://github.com/orchest/orchest/blob/master/services/orchest-api/app/app/apis/namespace_validations.py#L106) to be:

```python
# validation, action = validate_environment(project_uuid, env_uuid)
validation, action = "fail", "WAIT"
```
and swap the comments after seeing the build dialog (make sure to run Orchest with `--dev` for this to work).

@ricklamers Tagging you along here to make sure I understood your issue correctly.

## Checklist


yannickperrenet:
Going over the changes again I think the PR is incomplete in its current stage. Reason being is that by moving the `checkGate` into the `toggleSession` the button to toggle a session is now in charge to check whether it is allowed to start, whereas we currently have them decoupled and every view that wants to start the session separately manages the `checkGate`. 

For example in the `JupyterLabView` one `useEffect` does a gate check and the other toggles the session. This means that if the gate fails we redirect to the pipeline list (which is bad UX imo), but actually start the session in the background (you can check using `docker ps`). This can prevented by letting the `toggleSession` manage the `checkGate`.

In my opinion it would be cleaner to have the handle to start a session (`toggleSession`) be in charge of ensuring it is allowed to start (makes the code more DRY as well).

@iannbing

yannickperrenet:
The new changes are great (definitely much cleaner, but I think the comments made in my old comment (https://github.com/orchest/orchest/pull/712#issuecomment-1038962202) still apply. We should probably quickly discuss this before deciding whether we want to merge this PR.

@iannbing

iannbing:
@yannickperrenet Now that `toggleSession` includes `requestBuild` mechanism in it, but this doesn't apply to all scenarios. Some views (e.g. /pipeline) needs to check gate when view is loaded. But it does make code cleaner for JupyterLabView.

yannickperrenet:
Thanks @iannbing! Looks good to me :)

PR #715, opened 2022-02-14
Fix hanging data table when a delete dialog is cancelled

iannbing:
## Description

This PR fixes the bug: DataTable is hanging if a delete dialog is cancelled. The default onConfirm and onCancel should call the resolve function in the args.

This PR also cleans up the analytics message. Currently the content includes irrelevant data such as props.

## Checklist


yannickperrenet (approved):
Great stuff! How such a lightweight change can have such an impact on the user experience ^_^

PR #721, opened 2022-02-15
Orchest-api rework cleanup as a command

fruttasecca:
## Description

~This PR makes it so that the `orchest-api` will perform the required cleanups on exit. This helps to maintain state consistent when it comes to shutting down, restarting or stopping Orchest.~

See https://github.com/orchest/orchest/pull/721#issuecomment-1053501770

### DIFF
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -24,8 +24,10 @@ from app.apis.namespace_environment_builds import AbortEnvironmentBuild
 from app.apis.namespace_jobs import AbortJob
 from app.apis.namespace_jupyter_builds import AbortJupyterBuild, CreateJupyterBuild
 from app.apis.namespace_runs import AbortPipelineRun
+from app.apis.namespace_sessions import StopInteractiveSession
+from app.config import CONFIG_CLASS
 from app.connections import db
-from app.core import environments, image_utils
+from app.core import environments
 from app.core.scheduler import Scheduler
 from app.models import (
     EnvironmentBuild,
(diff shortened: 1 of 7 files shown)


services/orchest-api/app/app/__init__.py
```diff
@@ -233,6 +123,17 @@ def create_app(config_class=None, use_db=True, be_scheduler=False, to_migrate_db
...
+        # Needs to happen after the scheduler has been setup given that
+        # some of these tasks will make use of the scheduler.
+        if use_db:
+            initialization_cleanup(app)
+
+            # Trigger a build of JupyterLab if no JupyterLab image is
+            # found for this version and JupyterLab setup_script is
+            # non-empty.
+            if _utils.is_running_from_reloader():
+                trigger_conditional_jupyter_build(app)
```
yannickperrenet, line 135: Should probably be `if not  _utils.is_running_from_reloader():` otherwise the code will only execute in the reloader when running in Flask development mode.

services/orchest-api/app/app/__init__.py
```diff
@@ -359,3 +260,110 @@ def teardown(response):
...
+    # make sure the cleanup is quick!  (Especially when running only a
+    # single gunicorn worker as the entire orchest-api will become
+    # unresponsive.)
+    _cleanup(app, "/tmp/initialization-cleanup-lock")
+
+
+def shutdown_cleanup(app):
+    if os.getenv("FLASK_ENV") == "development":
+        return
+    _cleanup(app, "/tmp/shutdown-cleanup-lock")
```
yannickperrenet, line 277: Shouldn't it be 
```python
if _utils.is_running_from_reloader():
    return
_cleanup(app, "/tmp/shutdown-cleanup-lock")
```
This way we do still run cleanup in development mode (so the code is as similar to PROD as possible) but only in the main thread that starts the Flask development server.

services/orchest-api/app/app/__init__.py
fruttasecca: Indeed

services/orchest-api/app/app/__init__.py
fruttasecca: Yeah that is an alternative as well, wasn't really sure what would be best to have in DEV mode, will make it like you suggest.

services/orchest-api/app/gunicorn_conf.py
```diff
@@ -2,3 +2,26 @@
 workers = 1
 bind = "0.0.0.0:80"
 keepalive = 30
+
+
+def worker_exit(server, worker):
```
yannickperrenet, line 7: Given that this is a `gunicorn` setting ([docs](https://docs.gunicorn.org/en/stable/settings.html?highlight=worker_exit#worker-exit)), it won't trigger `shutdown_cleanup` when running Flask development. So we do have an alternative code path here.

Could there be any case conditions when running multiple gunicorn workers? In addition, what if a gunicorn workers exits due to an error then it will shutdown resources that are actually still in use. Right?

fruttasecca: >Given that this is a gunicorn setting ([docs](https://docs.gunicorn.org/en/stable/settings.html?highlight=worker_exit#worker-exit)), it won't trigger shutdown_cleanup when running Flask development. So we do have an alternative code path here.

You are right, I guess that for `--dev` mode this needs to be handled differently (or same path, see below).

>Could there be any case conditions when running multiple gunicorn workers?
I would have to check if we are locking for update the objects that are being cleaned up etc. I guess currently the code path would not be race condtion free.

>In addition, what if a gunicorn workers exits due to an error then it will shutdown resources that are actually still in use. Right?

Yeah. We can make this happen on the gunicorn master shutdown for additional safety but there is still the fact a SIGTERM could be sent by k8s for unforeseen reasons. I guess the safest way to handle this would actually be an API endpoint which is called by `orchest-ctl` on stop and the `update-server` on update.

yannickperrenet: > but there is still the fact a SIGTERM could be sent by k8s for unforeseen reasons

Indeed. Would be quite a bummer if that has as a side-effect that everything managed by the orchest-api gets stopped as well, e.g. a long running kernel or job.

> I guess the safest way to handle this would actually be an API endpoint which is called by orchest-ctl on stop and the update-server on update.

So the `orchest-ctl` would first query the API to tell it to shut down and after that tear down the namespace (causing a SIGTERM)?

fruttasecca: Yes, I think we also need to do something similar on start, otherwise the cleanup that happens during the initialization of the `orchest-api` will bite us the same way because of an unforeseen SIGTERM sent by k8s. This can get tricky given that the scheduler starts during the `orchest-api` initialization as well. We know that such cleanup needs only to happen on orchest `start` and `stop` so we can make use of that fact. Today (maybe tomorrow as well) I will be quite deep into the rework of `orchest-ctl`, I'll try to come up with something.

yannickperrenet: > I think we also need to do something similar on start, otherwise the cleanup that happens during the initialization of the orchest-api will bite us the same way because of an unforeseen SIGTERM sent by k8s. This can get tricky given that the scheduler starts during the orchest-api initialization as well.

What case do you mean? Do you mean running the cleanup and getting a SIGTERM could leave incorrect state in the DB?

yannickperrenet: Alternatively we could run cleanup the same way we do DB migrations, i.e. moving it to [`start.sh`](https://github.com/orchest/orchest/blob/master/services/orchest-api/start.sh).

fruttasecca: >What case do you mean? Do you mean running the cleanup and getting a SIGTERM could leave incorrect state in the DB?

- `orchest-api` initializes
- does some work
- for whatever reason, the `orchest-api` pod dies
- another pod gets created to replace it
- initializes -> cleanup

>Alternatively we could run cleanup the same way we do DB migrations, i.e. moving it to [start.sh](https://github.com/orchest/orchest/blob/master/services/orchest-api/start.sh).

Covered by the same case

I think we can get away with it by having the files we use for locking be mounted and shared by all `orchest-api` containers, so that the initialization cleanup happens once (`orchest-ctl` will remove the file on start) and happens once on stop (again, `orchest-ctl` removes the lock on stop).

yannickperrenet: > on start/stop, orchest-ctl launches a pod with the orchest-api image with the appropriate command and args to perform a cleanup

Would this be a prerequisite for the `orchest-api` to start or will the "clean-up orchest-api" be started in parallel? If it is a prerequisite then we might as well run it inside the `orchest-api` itself similar to how we run the `migration_manager.py`.

I don't really see the benefit of a dedicated pod over adding it to the `start.sh` script, but maybe I am missing something :thinking:

fruttasecca: >I don't really see the benefit of a dedicated pod over adding it to the start.sh script, but maybe I am missing something thinking

If it lives in the `start.sh` script a pod failure (for whatever reason) will lead to the creation of another `orchest-api` pod, which means that the cleanup code will go through again.

yannickperrenet (approved):
LGTM! 

Just need to address the aforementioned comments :100:

fruttasecca:
Given the highlighted issues the scope of this PR is changing:
- cleanup won't be performed, normally, by the `orchest-api`
- `orchest-ctl` will run a pod on `orchest start` and `orchest stop` that will invoke the cleanup code

Given these, the previous comments are considered as resolved.

PR #722, opened 2022-02-16
New environment edit page

iannbing:
## Description

This PR improves the UX of environment edit page. 
- all options in one view instead of multiple tabs
- support hotkey (Ctrl/Cmd + Enter to build the image)
- accessibility and responsiveness improvements.

## Checklist

## Screenshots
image


astrojuanlu:
May I ask for a screenshot / short screencast? 😇

services/orchest-webserver/client/src/App.tsx
```diff
@@ -119,11 +118,13 @@ const App = () => {
...
           data-test-id="app"
         >
-          <Toolbar />
-          <Box sx={{ overflowY: "auto", flex: 1 }}>
+          {/* <Toolbar /> */}
+          {/* <Box sx={{ overflowY: "auto", flex: 1 }}>
             <Routes />
             <div ref={jupyterRef} className="persistent-view jupyter hidden" />
-          </Box>
+          </Box> */}
```
iannbing, line 125: Will clean this up after finishing a round of Cypress tests

services/orchest-webserver/client/src/components/CommandPalette.tsx
```diff
@@ -185,110 +181,72 @@ const CommandPalette: React.FC = () => {
...
+      // handle failure silently because this is done in the background
+      console.error(`Failed to fetch for command palette: ${error}`);
+    }
+  }, [fetchJobs, fetchPipelines, fetchProjects]);
 
   const onQueryChange = (
     e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>
   ) => {
     setQuery(e.target.value);
   };
```
iannbing, line 249: It's a waste to keep a pure function in a React component. Moved to a more suitable place.

services/orchest-webserver/client/src/components/CommandPalette.tsx
```diff
@@ -164,17 +163,14 @@ const commandsFromJob = (projectPaths: Record<string, string>, job: Job) => {
 
 const CommandPalette: React.FC = () => {
   // global states
-  const {
-    state: { isCommandPaletteOpen },
-    dispatch,
-  } = useAppContext();
   const { navigateTo } = useCustomRoute();
 
+  const [isOpen, setIsOpen] = React.useState(false);
```
iannbing, line 168: it's not necessary to put this state in the Store. CommandPalette is already at the root level. And it doesn't share state with other.

services/orchest-webserver/client/src/components/CronScheduleInput.tsx
```diff
@@ -76,7 +76,7 @@ export const CronScheduleInput: React.FC<{
         value={value}
         onChange={onChangeCronString}
         data-test-id={dataTestId}
-        helperText={!cronDescription ? "Invalid cron string." : ""}
+        helperText={!cronDescription ? "Invalid cron string." : " "}
```
iannbing, line 79: This is to prevent jumping. It's recommended by MUI.

yannickperrenet:
First things first, really great job! :muscle: 

### Design notes
When it comes to the design I get a bit confused when first entering the view by the large black box that will show the build log output. There is no indication what it is for (maybe it needs a header?), but takes up a large portion of the screen real estate. Other than that I really like the design, especially they eye to detail (such as the coloring of the environment setup script). 

Can we make the loading bar red on failure so it is visually clearer that the build failed?
image

Edge case, but strange error when you completely remove the environment name
image

Now that the setup script for environments has a dark colorscheme, can we do the same for the script to configure JupyterLab ([localhost link](http://localhost:8000/configure-jupyter-lab))?

(Super sexy that you show "enter" inside the build button when `ctrl` is pressed.)

#### Text changes
> This will execute when you build the environment. Use it to include your dependencies.

Can we change this to: "Add dependencies to your environment."

> Choose a container image
> The container image will be the starting point from which the environment will be built.

I feel like the tag line is a little complicated to understand. Maybe we can simplify it to something like (@astrojuanlu curious about your opinion as well):
_Choose a base image
The base image can be extended using the environment set-up script_
OR
_Choose a language
Install dependencies using the environment set-up script_
I like the first one better so we also explain some of the concepts. 

> Create custom container image

Better text would be: "Add custom base image" which is also the text of the dialog that shows afterwards. Regardless of the text on the button, I do think it should equal the title of the dialog.

### Functionality notes
None. Works perfect from what I can tell :100:

astrojuanlu:
> Choose a base image
> The base image can be extended using the environment set-up script ([learn more](https://docs.orchest.io/en/latest/fundamentals/environments.html))

I like this 👍🏽 (proposed adding a link to [the docs](https://docs.orchest.io/en/latest/fundamentals/environments.html))

iannbing:
@yannickperrenet @astrojuanlu  Thanks for the feedback. All comments are considered and updated accordingly.
I ended up adding a dark-gray header "Logs" and didn't want to mess up with Xterm UI, until we have more insights in this new design. See the following screenshot.

Screenshot 2022-02-17 at 16 52 38

yannickperrenet (approved):
I absolutely love it! Great great job :heart: 

I haven't actually reviewed the code itself, but the final result is great!

PR #724, opened 2022-02-16
Feat/k8s orchest deployments probes

fruttasecca:
## Description

This PR adds startupProbes, `livenessProbes` and `readinessProbes` to orchest deployments. When no `startupProbe` is necessary, a `readinessProbe` is used both as a health check and a readiness check. `StartupProbes` are used to formalize dependencies among deployments, this is, generally speaking, not strictly necessary but, in the case of the `orchest-webserver`, it provides a better UX given that it won't be ready until the `orchest-api` is ready, which data it depends upon to have a functioning client.

Note: timings, failure thresholds etc. are quite arbitrary, generally speaking I have tried to avoid false positives.

Related dos:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

### DIFF
--- a/deploy/helm/charts/auth-server/templates/deployment.yaml
+++ b/deploy/helm/charts/auth-server/templates/deployment.yaml
@@ -32,3 +32,21 @@ deployment
            mountPath: /userdir
          - name: config
            mountPath: /config
+        startupProbe:
+          httpGet:
+            path: /login/server-config
+            port: 80
+            scheme: HTTP
+          periodSeconds: 1
+          successThreshold: 1
+          timeoutSeconds: 5
+          failureThreshold: 20
+        livenessProbe:
+          httpGet:
+            path: /login/server-config
+            port: 80
+            scheme: HTTP
+          periodSeconds: 10
+          timeoutSeconds: 2
+          successThreshold: 1
+          failureThreshold: 2
--- a/deploy/helm/charts/auth-server/templates/deployment.yaml
+++ b/deploy/helm/charts/auth-server/templates/deployment.yaml
@@ -33,10 +33,12 @@ deployment
          - name: config
            mountPath: /config
         startupProbe:
-          httpGet:
-            path: /login/server-config
-            port: 80
-            scheme: HTTP
+          exec:
+            command:
+              - sh
+              - -c
+              - "nc -zvw1 orchest-database 5432 &&
+                wget -T 1 -t 1 localhost/login/server-config --spider"
           periodSeconds: 1
           successThreshold: 1
           timeoutSeconds: 5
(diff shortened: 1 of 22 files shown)


deploy/helm/charts/auth-server/templates/deployment.yaml
```diff
@@ -31,4 +31,24 @@ deployment
...
          - name: config
-           mountPath: /config
\ No newline at end of file
+           mountPath: /config
+        startupProbe:
+          exec:
+            command:
+              - sh
+              - -c
+              - "nc -zvw1 orchest-database 5432 &&
```
yannickperrenet, line 40: @nhaghighat Or is it better practice not to hardcode the ports here as well and instead use templating?

deploy/helm/charts/rabbitmq/templates/deployment.yaml
```diff
@@ -18,4 +18,16 @@ deployment
...
-           mountPath: /var/lib/rabbitmq/mnesia
\ No newline at end of file
+           mountPath: /var/lib/rabbitmq/mnesia
+        readinessProbe:
+          exec:
+            command:
+              - su
+              - rabbitmq
+              - -c
+              - "/opt/rabbitmq/sbin/rabbitmq-diagnostics ping"
```
yannickperrenet, line 28: Why did you opt for `ping` instead of the `check_port_connectivity` that we currently use? Is that because this is the `readinessProbe`?

yannickperrenet (approved):
Don't think you missed anything :+1:

deploy/helm/charts/rabbitmq/templates/deployment.yaml
fruttasecca: See the comment from https://github.com/orchest/orchest/pull/724/commits/6094f19d70e8bfaf51450ccc73bad007593dedd5. Essentially, it was the safest thing to do.

The comment from https://github.com/orchest/orchest/pull/724/commits/fb0de75ca78d603dbfffe01c63dc8b3b05f93ab2 is unrelated but also interesting.

deploy/helm/charts/auth-server/templates/deployment.yaml
nhaghighat: lot\s of things currently are hard-coded but I also think it is better to template them and move the probes to values files of each chart.

yannickperrenet: I would be fine not doing it in this PR given that other things are also hard-coded :+1:

nhaghighat (approved):

nhaghighat:
Looks good

PR #730, opened 2022-02-18
Orchest-ctl refactoring PR 1, `orchest install`

fruttasecca:
## Description

This first PR is the beginning of the rework of `orchest-ctl`. Main changes:
- `orchest` script
- `orchest-ctl` dockerfile
- initial refactor of `orchest-ctl`
- `orchest install` can now be used to install orchest in the k8s cluster. The only assumed dependency is `minikube`
- the deployment makefile won't assume the usage of ~/.kube/config anymore. If you want to deploy through it you will have to set the `KUBECONFIG` env variable to point to your kube config file, usually found in `~/.kube/config`.

To test this:
- pull the branch
- set `imagePullPolicy: Never` or `IfNotPresent` in the helm charts if you actually want to use the deployment, otherwise it will from the prek8s images from dockerhub
- `eval $(minikube -p minikube docker-env)`
- `bash scripts/build_container.sh -i orchest-ctl`
- `bash orchest install`

Known issues:
- https://github.com/kubernetes/kubernetes/issues/27264
- helm is taking a while to actually start deploying

Loom link https://www.loom.com/share/5121caa2b12f44c290b872db7731f83c

### DIFF
--- a/orchest
+++ b/orchest
@@ -70,7 +70,7 @@ function on_exit {
 trap on_exit EXIT
 
 # Necessary because
-# https://github.com/kubernetes/kubernetes/issues/83242 Not: not
+# https://github.com/kubernetes/kubernetes/issues/83242 Note: not
 # sleeping here because not all sleep implementations support floating
 # point values, and sleeping for 1s introuces a bad delay UX wise. It
 # doesn't go through more than 1 iteration usually.
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -15,6 +15,9 @@ from app.orchest import _k8s_wrapper as k8sw
 
 logger = logging.getLogger(__name__)
 
+
+# This is just another failsafe to make sure we don't leave dangling
+# pods around.
 signal.signal(signal.SIGTERM, lambda *args, **kwargs: k8sw.delete_orchest_ctl_pod())
 atexit.register(k8sw.delete_orchest_ctl_pod)
 


deploy/Makefile
```diff
@@ -92,7 +92,7 @@ DOCKER_REGISTRY_DEPLOY_CONFIG := registry \
 
 .PHONY: orchest
 orchest: initialize
-	helm install ${ORCHEST_DEPLOY_CONFIG}
+	helm install --atomic ${ORCHEST_DEPLOY_CONFIG}
```
nhaghighat, line 95: You can add `--atomic` to `SHARED_DEPLOY_CONFIG`

orchest
```diff
@@ -1,73 +1,31 @@
...
-# When Orchest is not yet installed then the Orchest does not yet exist,
-# however for certain features such as 'orchest run' we need the
-# ctl container to be launched within the network.
-OPTIONAL_NETWORK="--network=orchest"
-[ -z "$($SUDO_PREFIX docker network ls --filter="name=orchest" -q)" ] && OPTIONAL_NETWORK=""
-
-$SUDO_PREFIX docker run \
+# Make sure the orchest namespace exists. Necessary only if Orchest is
+# being installed.
+minikube kubectl -- create namespace orchest  > /dev/null 2>&1
```
nhaghighat, line 5: I think this can be omitted, helm takes care of creating namespace

orchest
```diff
@@ -1,73 +1,31 @@
...
-    -v /var/run/docker.sock:/var/run/docker.sock \
-    -v "${DIR}":/orchest-host \
-    -v "${HOST_CONFIG_DIR}":/config \
-    -e HOST_CONFIG_DIR="${HOST_CONFIG_DIR}" \
-    -e HOST_REPO_DIR="${DIR}" \
-    -e HOST_USER_DIR="${HOST_USER_DIR}" \
-    -e HOST_OS="${HOST_OS}" \
-    ${OPTIONAL_NETWORK} \
-    orchest/orchest-ctl:latest "$@"
+    --wait \
```
nhaghighat, line 24: I think there is no need to have orchest-ctl inside the cluster. orchest-ctl can still live inside docker of the executing command.

deploy/Makefile
fruttasecca: I went for not adding it to avoid unforeseen consequences, i.e. what would happen to a botched upgrade with `--atomic`?

for reference
```
.PHONY: orchest_upgrade
orchest_upgrade:
	helm upgrade ${ORCHEST_DEPLOY_CONFIG}
```

orchest
fruttasecca: Not sure I understand, could you clarify?

Btw, `--wait` is from `kubectl`
>--wait=false: If true, wait for resources to be gone before returning. This waits for finalizers.

deploy/Makefile
nhaghighat: it is mentioned if, during the upgrade with atomic flag, it fails, it rolls back to the previous version.
https://helm.sh/docs/helm/helm_upgrade/

orchest
nhaghighat: you are installing orchest-ctl via kubectl, which means, orchest-ctl will be a pod in k8s, but we still can keep orchest-ctl in the client machine, and use the provided kubeconfig on the machine.
so orchest-ctl can still be spawned via docker (or even a bash script itself)

nhaghighat: > Currently the only requirement is minikube, and it can be later easily made to only be kubectl (with an optional flag to point to the desired kubeconfig), so one downside would be making the requirement list longer.
If we keep orchest-ctl in the client, the requirement will be docker (or helm) and kubeconfig, so the requirement list will not become longer, but the deployment process becomes simpler. as it is executed from the machine itself and it can be debugged easier.
Concurrent upgrade/install is already blocked by helm. for example:
https://github.com/helm/helm/blob/main/pkg/action/upgrade.go#L182

fruttasecca: Note, I am not against the approach you suggested, but I am not sure we should prioritize or commit to it before the k8s MVP is done unless there are evident gains

nhaghighat: >Wouldn't this mean that, at least currently, our requirements would be minikube + docker? Note, I am talking about the requirements to run Orchest altogether, not only orchest-ctl, the aim is to have the installation script in the README of the repo as simple as possible.

For installing orchest we need Kubernetes, and orchest-ctl is just a wrapper around make-file, so if we insist on running orchest in a docker, then our requirement will be docker + Kubernetes, and if we want to run it with kubectl, then requirements will be kubectl + Kubernetes.

>I don't exactly agree with this, to the caller of the orchest script it will look exactly the same and when it comes to local development I would rather have the orchest-ctl live in the cluster (easy inspection with k9s) rather than having this specific component live in another containerization engine. It feels like a regression having to introduce calls to docker. Another point (imo) is that this minikube kubectl -- calls can be later made into kubectl calls, which means that, potentially, one would be able to control orchest through the CLI without having an orchestration engine locally.

orchest-ctl is a control tool, something like kubectl or argo ctl or even helm itself (in the previous version of helm, it was thriller and helm, helm was the client-side component and thriller was server-side component), it is pretty normal to have ctl running on a client itself.

>Another reason (which I hope we won't have to make use of) of having orchest-ctl in-cluster is that we will be able to mount whatever secret/volume etc. can normally be accessed by orchest. If the container lives locally we are out of luck for this use case.

maybe I miss something, but I think for this use case, we should have access to k8s itself, in that case, we can check the volumes or secrets from k8s itself.

fruttasecca: >For installing orchest we need Kubernetes, and orchest-ctl is just a wrapper around make-file

This is not exactly true, `orchest-ctl` has other commands. `orchest install` is, partially, a wrapper around the make file. See `orchest --help`.

>so if we insist on running orchest in a docker, then our requirement will be docker + Kubernetes, and if we want to run it with kubectl, then requirements will be kubectl + Kubernetes.

From the POV of a user trying out Orchest OSS post-k8s it would be:
- install minikube
- run `orchest install`
vs
- install minikube (for the k8s cluster)
- install docker (for running orchest-ctl)
- run `orchest install`

In the future, once we move from `minikube kubectl --` to just using `kubectl`, a user which needs to interact with a remote orchest cluster will only need `kubectl` and point to the right config. If we run `orchest-ctl` locally docker will be a requirement for this use case as well. Technically speaking we can move to using `kubectl` instead of `minikube kubectl --` right now but I don't see the benefit of adding to the requirements.

>orchest-ctl is a control tool, something like kubectl or argo ctl or even helm itself (in the previous version of helm, it was thriller and helm, helm was the client-side component and thriller was server-side component), it is pretty normal to have ctl running on a client itself.

I agree, however:
- what these tools have (among other things) in common is that those are tools that we don't want our users to operate with because it's usually lower level w.r.t. to what they are doing
- we want `orchest-ctl` to do "everything" for the user, from installation to deletion. If any of those operations require a system package we can 1) ask the user to install it (adding to the requirements) 2) provide `orchest-ctl` as a containerized image. I think that this is only the case for installation (helm) most likely, so that's nice.
- given my last sentence it means that, technically, we can have `install` be a "special" operation (maybe a container?) and provide all other CLI operations through a package, which wouldn't require a running container. It takes more time to develop and I don't see the added benefits right now, but I can see how a package with less overhead than an in-cluster container would benefit us as the number of CLI operations increases

>maybe I miss something, but I think for this use case, we should have access to k8s itself, in that case, we can check the volumes or secrets from k8s itself.

We could indeed query and play around with secrets through the API (making secrets travel from the cluster to the user client), but how would we cover having the `orchest-ctl`, let's say, create/delete a file etc.? I am sure there is a way around it but it seems the ergonomics of it would get worse. Same for `orchest-ctl` having to query some existing `orchest` services like the registry etc., a lot of things get easier if you know you are inside the cluster and the `orchest` namespace. Not saying it has to be like this forever, it could look completely different in the future.

Are you against `orchest-ctl` being a container or the container running in-cluster? 

When it comes to the former I am not a fan either, but the reason is that `orchest-ctl` has a different scope than, let's say, argo-ctl, given that it allows you to install/delete the deployment altogheter. The reason is also a bit "historic", i.e. we are migrating from an existing setup.

When it comes to the latter I fail to see the downside at the moment.

nhaghighat (approved):

orchest
fruttasecca: After an internal meeting we have decided to run `orchest-ctl` in-cluster for the time being, with the aim of making it become a package in the future.

deploy/orchest-ctl/rbac/clusterrole.yml
```diff
@@ -0,0 +1,16 @@
...
+    app.kubernetes.io/part-of: orchest
+    app.kubernetes.io/release: orchest
+  name: orchest-ctl
+rules:
+  - apiGroups:
+      - "*"
+    resources:
+      - "*"
+    verbs:
+      - "*"
```
yannickperrenet, line 16: Would this later be limited to exact `apiGroups`, `resources` and `verbs`?

services/orchest-ctl/setup.py
```diff
@@ -5,11 +5,10 @@
     version="0.1.0",
     packages=setuptools.find_packages(),
     install_requires=[
-        "typer",
-        "docker",
+        "kubernetes==21.7.0",
         "six>=1.13.0",
-        "aiodocker==0.21.0",
         "tqdm==4.53.0",
+        "typer",
```
yannickperrenet, line 11: Let's hardcode a version here as well. Suddenly being hit by breakage due to packages updating is pretty annoying (just like the `itsdangerous` case.

deploy/orchest-ctl/rbac/clusterrole.yml
fruttasecca: Ideally yes, this is there currently to avoid playing whac-a-mole with permissions required to install Orchest through helm, although I could see this being handy to avoid the following:
- be strict with those permissions
- discover that `orchest-ctl` actually needs more permissions (i.e. a new feature or CLI command)
- an existing deployment now has to pull the latest repo version to make sure the latest permissions are added

I can see point 3 already happening for `updates` since it would make sure that the `orchest` script is the latest tho, so being strict might not add more to this.

tldr yes but not much of a priority imo

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -0,0 +1,99 @@
...
+        return True
+    elif not any(deployments):
+        return False
+    else:
+        utils.echo(
+            "Unexpected Orchest installation state. Expected to find no deployments at "
+            "all or a complete installation. Please verify your installation.",
+            err=True,
+        )
+        raise typer.Exit(code=1)
```
yannickperrenet, line 34: I think it is easier to read like:
```python
    if all(deployments):
        return True
    elif any(deployments):
        utils.echo(
            "Unexpected Orchest installation state. Expected to find no deployments at "
            "all or a complete installation. Please verify your installation.",
            err=True,
        )
        raise typer.Exit(code=1)
    else:
        return False
```

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -0,0 +1,99 @@
...
+    # attach to the terminal (compare comment #16670 (comment))."
+    # https://github.com/kubernetes/kubernetes/issues/27264
+    # K8S_TODO: find a workaround.
+    utils.echo("Installing...")
+
+    # K8S_TODO: remove DISABLE_ROOK?
+    env = os.environ.copy()
+    env["DISABLE_ROOK"] = "TRUE"
+    process = subprocess.Popen(
+        ["make", "orchest"],
```
yannickperrenet, line 54: I guess this means that after installation you don't have any of the base-images for environments and no session related images? We probably want to be pulling all images during the installation process as otherwise operations through the UI appear very slow.

services/orchest-ctl/app/config.py
```diff
@@ -1,5 +1,21 @@
...
+ORCHEST_DEPLOYMENTS = [
+    "celery-worker",
+    "docker-registry",
+    "file-manager",
+    "orchest-api",
+    "orchest-database",
+    "orchest-webserver",
+    "rabbitmq-server",
+    "update-server",
+    # Bit risky in chase of a name change? K8S_TODO: discuss.
```
yannickperrenet, line 14: Good question. 

@nhaghighat What do you think?

services/orchest-ctl/app/cli/main.py
```diff
@@ -195,8 +195,9 @@ def install(
     # Make sure the permissions of the userdir are correctly set in case
     # Orchest will always be started using `--cloud` in the future (as
     # in other modes the permissions are fixed on start).
-    fix_userdir_permissions()
-    app.install(language, gpu=gpu)
+    # K8S_TODO: is this still needed?
+    # fix_userdir_permissions()
```
yannickperrenet, line 199: The `fix_userdir_permissions()` call is in `install` to ensure that the `userdir` structure (so the folders added to the repo, and now the dynamic `DirectoryOrCreate`) also has the right permissions set. So I guess it is still needed.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -0,0 +1,99 @@
...
+        utils.echo("Installation is already complete. Did you mean to run:")
+        utils.echo("\torchest update")
+        return
+    # "When running a command via kubectl run -it that immediately prints
+    # something, we might lose some lines of the log due to a race of the
+    # execution of the container and the kubectl attach used by kubectl run to
+    # attach to the terminal (compare comment #16670 (comment))."
+    # https://github.com/kubernetes/kubernetes/issues/27264
+    # K8S_TODO: find a workaround.
+    utils.echo("Installing...")
```
yannickperrenet, line 48: Yes a workaround would indeed be good to have as otherwise the user might be waiting a "long" time before getting confirmation that the installation process has actually started (i.e. waiting for the progress bar).

services/orchest-ctl/app/orchest/_core.py
fruttasecca: One issue I can see is that in a multi-node setting it might not always be true that we know where the image will be needed. As of now we might have a way around it by using our assumption that builds always happen on a given designated node and trigger a pull there, if possible.

>no session related images
Same as above, but worse because we don't know on which node a given pod may land. Feels like this might be a job fo daemonsets? @nhaghighat

services/orchest-ctl/app/cli/main.py
fruttasecca: You are right. This might complicate the matter further since this will require mounting to `orchest-ctl`, which is something we wanted to avoid.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: The fact that we need to mount into the pod (thus the pod needs to be in-cluster) + https://github.com/kubernetes/kubernetes/issues/27264 puts us is quite an annoying situation. Let's sleep on it.

fruttasecca: >EDIT: Not saying we need it right now ;)

I am not sure this is something we can skip (very) safely, I can see some migration issues in dramatically changing how we launch `orchest-ctl` post release

services/orchest-ctl/app/cli/main.py
yannickperrenet: Oeff good point. Can't really come up with a solution for that :thinking:

services/orchest-ctl/app/orchest/_core.py
fruttasecca: This has been "fixed' by not using `kubectl run`.

services/orchest-ctl/app/cli/main.py
fruttasecca: We are now mounting the `userdir` directory.

services/orchest-ctl/app/config.py
nhaghighat: we can feed the name as environment variable or config files.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: Not sure what's the time to add this as a daemonset but I guess we might not have the time for that, so I guess this would mean doing a pull for every node, for every image that we want to be there on first use.

services/orchest-ctl/app/config.py
fruttasecca: If we don't want to hardcode the name of orchest deployments I can perhaps parse `deploy/helm/Chart.yaml` for those, not sure how this would look like for third parties since we aren't in control of those files so updating third parties can mean a breakage regardless. Not super happy having `orchest-ctl` parsing yamls tho.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: Yes that's indeed true. Our issue is, imo, that Orchest doesn't communicate in a nice way that something isn't starting because of an ongoing pull, which to users would look like something is broken.

yannickperrenet: > Our issue is, imo, that Orchest doesn't communicate in a nice way that something isn't starting because of an ongoing pull, which to users would look like something is broken.

Indeed. Even if it were clear from the UI it is a poor user experience, so in my opinion a solution is required.

services/orchest-ctl/app/config.py
yannickperrenet: It sucks that we need to make a decision on this now, since after the code has been deployed the upgrade path from different version of Orchest might break (in this case due to a name change). 

Given that the `orchest-ctl` will be version specific (i.e. we also add the helm charts to the image) wouldn't name changes be solved for? (Maybe it is too late and I am not seeing things straight anymore.)

fruttasecca: >Given that the orchest-ctl will be version specific (i.e. we also add the helm charts to the image) wouldn't name changes be solved for? (Maybe it is too late zzz and I am not seeing things straight anymore.)
Yes that should be the case AFAIK, not only `orchest-ctl` but also the `update-server` will likely be version specific, still something to make sure of.

My comment "# Bit risky in chase of a name change? K8S_TODO: discuss." was mostly about the fact that a hardcoded name that you don't control is a bit of a code smell. Parsing yamls instead of doing that would not solve the upgrade issue you mentioned since the yamls will be version specific and live in the `orchest-ctl` and/or `update-server` as well.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: After an internal meeting, we decided to go for a daemonset and not have `orchest-ctl` be in charge of making sure images are pulled on install.

nhaghighat (approved):

services/orchest-ctl/app/config.py
yannickperrenet: So I think we agree that we can leave it as is? 

> Yes that should be the case AFAIK, not only orchest-ctl but also the update-server will likely be version specific, still something to make sure of.

:+1:

yannickperrenet (approved):

PR #731, opened 2022-02-21
Use `-exec ... +` notation on `find` that fixes permissions

yannickperrenet:
## Description
As per comment:
> Use the `-exec ... +` notation to pass all found files to `chmod` at once and reduce the number of invocations.

Thank you @tavianator for making us aware of this option.


fruttasecca:
Thank you @tavianator indeed!

PR #732, opened 2022-02-21
Improv/remove namespace scoped sessions

fruttasecca:
## Description

Makes it so that Orchest sessions do not live in their own k8s namespace anymore. This is (sadly) needed because of the scope of some k8s resources that we will make use of when it comes to cross namespace behaviour.


nhaghighat (approved):

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -230,29 +187,107 @@ def shutdown(session_uuid: str, wait_for_completion: bool = False):
...
+        # Don't let an error here hinder the deletion of other
+        # resources.
         try:
-            k8s_core_api.read_namespace_status(ns)
-        except client.ApiException as e:
-            if e.status == 404:
-                break
-            raise
+            logger.info(f"Deleting {resource_name}.")
+            entities = thread.get()
```
yannickperrenet, line 245: I guess this call is blocking? 

Although it shouldn't be that big of a deal since the "compute" in this for-loop is minimal.

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -265,10 +300,10 @@ def has_busy_kernels(session_uuid: str) -> bool:
 
     """
     # https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html
-    ns = get_k8s_namespace_name(session_uuid)
-    service_dns_name = f"jupyter-server.{ns}.svc.cluster.local"
+    ns = _config.ORCHEST_NAMESPACE
+    service_dns_name = f"jupyter-server-{session_uuid}.{ns}.svc.cluster.local"
```
yannickperrenet, line 304: The full dns name is needed still because the jupyter-server is exposed through an iframe?

fruttasecca: No, this is actually something that can be fixed since the `orchest-api` and `celery-worker` share the namespace with the session, nice catch :).

services/orchest-api/app/app/core/sessions/_core.py
```diff
@@ -230,29 +187,107 @@ def shutdown(session_uuid: str, wait_for_completion: bool = False):
...
+
+    # TODO: move to exception groups
+    # https://www.python.org/dev/peps/pep-0654/.
+    if exceptions:
+        raise errors.SessionCleanupFailedError(str(exceptions))
+
+    if not wait_for_completion:
+        return
+    # Else keep listing resources until no resources are there.
+    tries = 1000
```
yannickperrenet, line 269: I guess the browser request will already timeout before all `tries` are done. Or is this function called in a background task?

fruttasecca: Background task through the scheduler for interactive sessions and celery-worker for non-interactive ones AFAIR, so yes.

yannickperrenet (approved):
LGTM!

services/orchest-api/app/app/core/sessions/_core.py
yannickperrenet: Does any of the `resources_to_delete` take significantly longer to delete than others? Because if that is the case, then we might want to fiddle a bit with the order of the list.

fruttasecca: I guess we can put deployments and pods first, most likely an improvement.

yannickperrenet: Sounds good :+1:

PR #734, opened 2022-02-22
Don't rely on `orchest stop` to kill resources for integration tests

yannickperrenet:
## Description
Why kill resources without using `orchest stop`?
- The command might be broken
- The command might not actually successfully kill all resources due to
  incompatibility issues with the previous deployment

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,18 +4,24 @@ on:
   push:
     branches:
       - dev
-    paths-ignore:
-      - "docs/**"
-      - ".github/**"
-      - "README.md"
+    # Run on all file changes, except for ... but always run when the
+    # workflow for integration tests is changed.
+    paths:
+      - "**"
+      - "!docs/**"
+      - "!.github/**"
+      - ".github/workflows/integration-tests.yml"
+      - "!README.md"
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -56,10 +56,19 @@ jobs:
...
-          ./orchest stop
-          [ ! -z "$(docker ps -a -q)" ] && docker rm -fv $(docker ps -a -q)
+          # Stop Orchest by killing all its resources.
+          # NOTE: We can not make use of the `orchest stop` to do this,
+          # because:
+          # - the command might have been broken by the version of the
+          #   previous deployment.
+          # - first building and then using `orchest stop` wouldn't
+          #   reliably work as the previous deployment does not need to
+          #   be compatible.
```
fruttasecca, line 66: Makes very much sense.

fruttasecca (approved):
Technically speaking we are assuming that `./orchest stop` only stops containers without having other collateral effects, not ideal but it's what actually happens.

yannickperrenet:
> Technically speaking we are assuming that ./orchest stop only stops containers without having other collateral effects, not ideal but it's what actually happens.

Might indeed be good to explicitly mention this in the comments :+1:

PR #739, opened 2022-02-23
Orchest-ctl refactoring PR 2, `orchest version` 

fruttasecca:
## Description

Ports `orchest version` to k8s, and introduces a `--json` flag, e.g. `orchest version --json --ext`:

```json
{
  "cluster_version": "v2022.02.8",
  "deployment_versions": {
    "argo-workflow-argo-workflows-server": "v3.2.6",
    "argo-workflow-argo-workflows-workflow-controller": "v3.2.6",
    "celery-worker": "latest",
    "docker-registry": "2.7.1",
    "file-manager": "latest",
    "orchest-api": "latest",
    "orchest-database": "13.1",
    "orchest-webserver": "latest",
    "rabbitmq-server": "3",
    "update-server": "latest"
  }
}
```

### DIFF
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -152,7 +152,7 @@ def _echo_version(
         }
         if deployment_versions is not None:
             data["deployment_versions"] = deployment_versions
-        print(json.dumps(data, sort_keys=True, indent=2))
+        utils.echo(json.dumps(data, sort_keys=True, indent=2), wrap=False)
 
 
 def version(ext=False, output_json: bool = False) -> None:
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -1,10 +1,12 @@
 """Core functionality of orchest-ctl."""
 import atexit
+import json
 import logging
 import os
 import signal
 import subprocess
 import time
+from typing import Dict, Optional
 
 import typer
 from kubernetes import client as k8s_client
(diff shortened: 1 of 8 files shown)


services/orchest-ctl/app/orchest/_core.py
```diff
@@ -117,3 +119,66 @@ def install():
...
+        ext: If True return the extensive version of Orchest, i.e.
+        including deployment versions.
+        output_json: If true echo json instead of text.
+    """
+    cluster_version = k8sw.get_orchest_cluster_version()
+    if not ext:
+        _echo_version(cluster_version, None, output_json)
+        return
+
+    deployments = k8sw.get_orchest_deployments(config.ORCHEST_DEPLOYMENTS)
```
yannickperrenet, line 171: Might be good to add to the docstring of `get_orchest_deployments` that the those will be returned in the same order as the given `deployments` arg.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -117,3 +119,66 @@ def install():
...
+                    "the 'deploy' directory might fix the issue.",
+                    err=True,
+                )
+    else:
+        data = {
+            "cluster_version": cluster_version,
+        }
+        if deployment_versions is not None:
+            data["deployment_versions"] = deployment_versions
+        print(json.dumps(data, sort_keys=True, indent=2))
```
yannickperrenet, line 155: Can't we use `utils.echo(json.dump(...), wrap=False)` here?

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -117,3 +119,66 @@ def install():
...
+                utils.echo(f"{depl:<{buffer_space}}: {d_version}")
+                if depl in config.DEPLOYMENT_VERSION_SYNCED_WITH_CLUSTER_VERSION:
+                    differences = differences or d_version != cluster_version
+            if differences:
+                utils.echo(
+                    "Some deployment version differs from the cluster version. This is "
+                    "an unexpected state. Upgrading or using the helm charts in "
+                    "the 'deploy' directory might fix the issue.",
+                    err=True,
+                )
```
yannickperrenet, line 148: I think it makes sense to run this `differences` check as well when `output_json=True`. Maybe adding an additional key to the JSON output? What do you think?

Now there is a difference when running with or without `output_json`, it is not just a different output type but you are actually getting a different message.

yannickperrenet (approved):
Looks good to me! :man_dancing:

services/orchest-ctl/app/orchest/_core.py
yannickperrenet: I just looked at the other [PR](https://github.com/orchest/orchest/pull/740) for `orchest status` and I do like it. Can this be refactored to using a similar flow using `config.JSON_MODE`.

> Now there is a difference when running with or without output_json, it is not just a different output type but you are actually getting a different message.

Still not sure about this though. Or is it common among other CLIs that if you pass something like `--json` (or some other machine interpretable output) that you are missing out on other output?

services/orchest-ctl/app/orchest/_core.py
fruttasecca: That's something I forgot to try, will do.

services/orchest-ctl/app/orchest/_core.py
yannickperrenet: Let's leave it as is, I agree with your consensus. When returning JSON output we could always return a non-zero exit code if something is wrong.

nhaghighat (approved):

PR #740, opened 2022-02-23
Orchest-ctl refactoring PR 3, `orchest status`

fruttasecca:
## Description

Ports `orchest status` to k8s, and introduces a `--json` flag, e.g. `orchest status --json`:

```json
{"status": "unhealthy", "reason": ["orchest-api"]}
```

### DIFF
--- a/services/orchest-ctl/app/config.py
+++ b/services/orchest-ctl/app/config.py
@@ -87,6 +87,8 @@ ORCHEST_IMAGES = {
 }
 
 WRAP_LINES = 72
+# Used to avoid outputting anything that isn't the desired json.
+JSON_MODE = False
 
 ORCHEST_WEBSERVER_ADDRESS = "http://orchest-webserver:80"
 
--- a/services/orchest-ctl/app/config.py
+++ b/services/orchest-ctl/app/config.py
@@ -1,6 +1,30 @@
+from enum import Enum
 from typing import List, Set
 
 ORCHEST_NAMESPACE = "orchest"
+
+STATUS_CHANGING_OPERATIONS = ["install", "start", "stop", "restart", "update"]
+
+
+class OrchestStatus(str, Enum):
+    INSTALLING = "installing"
+    RESTARTING = "restarting"
+    RUNNING = "running"
+    STARTING = "starting"
+    STOPPED = "stopped"
+    STOPPING = "stopping"
+    UNHEALTHY = "unhealthy"
+    UPDATING = "updating"
+
+
+ORCHEST_STATUS_CHANGING_OPERATION_TO_STATUS = {
+    "install": OrchestStatus.INSTALLING,
+    "start": OrchestStatus.STARTING,
+    "stop": OrchestStatus.STOPPING,
+    "restart": OrchestStatus.RESTARTING,
+    "update": OrchestStatus.UPDATING,
+}
+
 
 DEPLOYMENT_VERSION_SYNCED_WITH_CLUSTER_VERSION = set(
     [
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -1,4 +1,3 @@
-"""Core functionality of orchest-ctl."""
 import atexit
 import json
 import logging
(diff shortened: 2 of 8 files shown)


services/orchest-ctl/app/config.py
```diff
@@ -1,7 +1,31 @@
...
+    RESTARTING = "restarting"
+    RUNNING = "running"
+    STARTING = "starting"
+    STOPPED = "stopped"
+    STOPPING = "stopping"
+    UNHEALTHY = "unhealthy"
+    UPDATING = "updating"
+
+
+ORCHEST_STATUS_CHANGING_OPERATION_TO_STATUS = {
```
yannickperrenet, line 20: Confusing name imo.

yannickperrenet: Maybe `STATUS_CHANGING_OPERATIONS` can be renamed to something like `COMMANDS` (depending on how exactly you envision it should be extended in the future of course, e.g. when adding a new command).

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -182,3 +182,62 @@ def version(ext=False, output_json: bool = False):
...
+        unhealthy_reason = set()
+        if unhealthy:
+            status = config.OrchestStatus.UNHEALTHY
+            unhealthy_reason.update(unhealthy_deployments)
+            # Assume that, if at least 1 deployment is running, the ones
+            # which are not are unhealthy.
+            if stopped_deployments and running_deployments:
+                unhealthy_reason.update(stopped_deployments)
+            unhealthy_reason = sorted(unhealthy_reason)
+            utils.echo(f"Unhealthy deployments: {unhealthy_reason}.")
```
yannickperrenet, line 231: Don't think there is a difference between unhealthy "reason" and "deployments" here right? Asking because here we print `"Unhealthy deployments: {unhealthy_reason}"` whereas for the `output_json` it is called `"reason"`. Would make more sense if both follow the same terminology.

yannickperrenet (approved):
:100:

services/orchest-ctl/app/orchest/_core.py
fruttasecca: Agree, will go with unhealthy reasons to keep it high level since in the future this might cover more than deployments.

fruttasecca: New output
```json
{
 "reason": {
  "deployments": [
   "orchest-api"
  ]
 },
 "status": "unhealthy"
}
```

services/orchest-ctl/app/config.py
fruttasecca: After a bit of thought, I opted for moving `ORCHEST_STATUS_CHANGING_OPERATION_TO_STATUS` to the only place where it is used so that somebody "exploring" the `orchest-ctl` code won't get confused by looking at the config module. Essentially I am kicking the can down the road because I can't find a more suitable name. Of course this is up for debate.

yannickperrenet: `ORCHEST_OPERATION_TO_STATUS_MAPPING`? The things that makes it confusing currently is that it is unclear that it is read like "status-changing-operation" (an operation that changes the status of Orchest) "to-status".

services/orchest-ctl/app/utils.py
```diff
@@ -16,10 +16,18 @@
 logger = logging.getLogger(__name__)
 
 
+def echo_json(data: dict) -> None:
+    """Wraps typer.echo to output json in a consistent manner."""
+    typer.echo(json.dumps(data, sort_keys=True, indent=True))
```
yannickperrenet, line 21: Maybe add:
```python
if not config.JSON_MODE:
    return
```

services/orchest-ctl/app/config.py
fruttasecca: I see, sure, `ORCHEST_OPERATION_TO_STATUS_MAPPING` works for me.

nhaghighat:
Looks good

nhaghighat (approved):

PR #742, opened 2022-02-24
Feat/integrate with distributed fs

nhaghighat:
## Description

This PR changes the mounting structure to use PVC (persistance volume claims)
Helm charts creating the PVCs and then the orchest uses them. the names of the PVCs are hardcoded, so one of the next jobs should be to make it configurable.

If you want to test this PR with rook-ceph, the kvm have to be installed, after that minikube can be installed with following command:
`./scripts/install_minikube.sh -n 5`

If you want to test it with rook-nfs, there is no need for kvm. (the cluster with kvm also works)

Then you should install docker-registry with the following command in `deploy` directory:
`make docker-registry`
then the docker-registry port should be forwarded.
`export POD_NAME=$(kubectl get pods --namespace orchest -l "app=docker-registry,release=registry" -o jsonpath="{.items[0].metadata.name}")`
`kubectl -n orchest port-forward $POD_NAME 8080:5000`

then the new images can be built on the node(laptop), and tagged and pushed to our docker-registry:
`./scripts/build_container.sh -i celery-worker`
`docker tag orchest/celery-worker 127.0.0.1:8080/orchest/celery-worker`
`docker push 127.0.0.1:8080/orchest/celery-worker`

Then the value of `images.registry` of each chart you just pushed needs to be updated to the service ip of docker registry. 
finally you can start deploying the cluster with:

With rook-ceph:
` ENABLE_ROOK_CEPH=true make orchest`

With rook-nfs:
` ENABLE_ROOK_NFS=true make orchest`

## Checklist


lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -442,19 +443,49 @@ def get_step_and_kernel_volumes_and_volume_mounts(
...
-    volume_mounts.append(
-        {"name": "pipeline-file", "mountPath": container_pipeline_file}
-    )
+    relative_project_dir = get_userdir_relpath(project_dir)
+    relative_pipeline_path = os.path.join(relative_project_dir, pipeline_file)
 
     volumes.append(
-        {"name": "data", "hostPath": {"path": os.path.join(host_user_dir, "data")}}
+        {
+            "name": "userdir-pvc",
```
fruttasecca, line 451: >From the POV of the service using this code, is it necessary for it to know that's it's interacting with a pvc

Same here and for all entries that have now become `*pvc`

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -70,16 +70,16 @@ def _get_base_image_cache_workflow_manifest(workflow_name, base_image: str) -> d
...
     """Returns a workflow manifest given the arguments.
 
     Args:
         workflow_name: Name with which the workflow will be run.
         image_name: Name of the resulting image, can include repository
             and tags.
-        build_context_host_path: Path on the host where the build
-            context is to be found.
+        build_context_path: Path on the container where the build
+            context is to be found, ther userdir should be removed
```
fruttasecca, line 82: Not sure I understand the last part of the docstring.

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -136,11 +136,15 @@ def _get_image_build_workflow_manifest(
                             "--single-snapshot",
                         ],
                         "volumeMounts": [
-                            {"name": "build-context", "mountPath": "/build-context"},
                             {
-                                "name": "kaniko-cache",
+                                "name": "userdir-pvc",
+                                "mountPath": "/build-context",
+                                "subPath": get_userdir_relpath(build_context_path)
```
fruttasecca, line 142: Does the `subPath` mean we are still mounting just the thing that kaniko strictly needs right?

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -136,11 +136,15 @@ def _get_image_build_workflow_manifest(
...
+
+                            },
+                            {
+                                "name": "kaniko-cache-pvc",
                                 "mountPath": "/cache",
-                                "readOnly": True,
                             },
```
fruttasecca, line 148: Can't it be kept `readOnly`? If not perhaps we want a comment in there explaining why

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -419,18 +420,18 @@ def get_environment_capabilities(environment_uuid, project_uuid):
 
 
 def get_step_and_kernel_volumes_and_volume_mounts(
-    host_user_dir: str,
-    host_project_dir: str,
-    host_pipeline_file: str,
+    userdir_pvc: str,
+    project_dir: str,
+    pipeline_file: str,
```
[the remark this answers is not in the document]
nhaghighat: I think it is ok to let the serivce knows it is dealing with PVC, because at the end it uses the pvc to build step and jupyter deployments.

all the path, all over the code is changed to be the path, it is seen by the container.

services/orchest-api/app/app/core/image_utils.py
nhaghighat: indeed, we mount part of the pvc to the container.

services/orchest-api/app/app/core/image_utils.py
nhaghighat: will change the comment to explain more

services/orchest-api/app/app/core/image_utils.py
nhaghighat: I think it is removed accidentally.

deploy/helm/charts/cluster-resources/charts/orchest/values.yaml
```diff
@@ -0,0 +1,3 @@
+userdir:
+  name: userdir
+  storage: 4Gi
```
yannickperrenet: Better to make it something like `20Gi` to be on the safer side. Depending on how "cleanly" you use Orchest, you can easily get past `4Gi`.

deploy/helm/templates/_cluster_orchest.tpl
```diff
@@ -0,0 +1,21 @@
+{{/*
+Get the name of the userdir pvc.
+*/}}
+{{- define "library.cluster.orchest.userdir.name" -}}
+  {{- if .Values.userdir.name -}}
+    {{ .Values.userdir.name | trunc 63 | trimSuffix "-" }}
+  {{- else -}}
+    {{ "user-dir" }}
```
yannickperrenet: Can we make it `"userdir"` instead? Since that is what it is currently called.

services/orchest-api/app/app/core/environment_builds.py
```diff
@@ -206,9 +206,11 @@ def prepare_build_context(task_uuid, project_uuid, environment_uuid, project_pat
...
 
-    env_builds_dir = "/userdir/.orchest/env-builds"
+    env_builds_dir = ".orchest/env-builds"
     # K8S_TODO: remove this?
-    Path(env_builds_dir).mkdir(parents=True, exist_ok=True)
+    Path(f"/userdir/{env_builds_dir}").mkdir(parents=True, exist_ok=True)
+    # We need snapshot subpath to be used in userdir pvc
+    snapshot_subpath = os.path.join(env_builds_dir, dockerfile_name)
     # Make a snapshot of the project state, used for the context.
     snapshot_path = f"{env_builds_dir}/{task_uuid}"
```
yannickperrenet, line 213: I don't think this works as expected right? 

With this change `snapshot_path` now no longer has the `/userdir/` prefix and so (unless `userdir` is the workdir) it won't do set up the build context in the `userdir` and instead will use the relative path of the `env_builds_dir = ".orchest/env-builds"`

Minimal change would have been something like:
```python
snapshot_subpath = os.path.join(env_builds_dir.removeprefix("/userdir/", dockerfile_name))
```
but maybe I am misunderstanding what is happening here (or I am looking at an outdated commit).

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -359,6 +359,7 @@ def run_orchest_ctl(client, command):
         environment={
             "HOST_CONFIG_DIR": os.environ.get("HOST_CONFIG_DIR"),
             "HOST_REPO_DIR": os.environ.get("HOST_REPO_DIR"),
+            #TODO: k8s, probably remove
```
yannickperrenet, line 362: Lets use the same format for the k8s TODOs, so changing this to `# TODO_K8S: probably remove` (otherwise searching across the codebase will be hard.

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -28,15 +28,58 @@ def _get_common_volumes_and_volume_mounts(
     volumes = {}
     volume_mounts = {}
 
+    '''
```
yannickperrenet: This entire string can be removed if we don't use it.

scripts/install_minikube.sh
```diff
@@ -0,0 +1,116 @@
...
+  log "Failed to find binary ${kubectl}}"
+  exit 1
+fi
+
+
+#minikube-status
+#minikube-start ${num_nodes}
+create-and-attach-volumes ${size}
+minikube-stop
+minikube start
```
yannickperrenet, line 114: Should probably be `minikube-start`

yannickperrenet:
Absolutely stellar work @nhaghighat! I am impressed by all the changes you have made (especially given that they are all over the codebase). Really well done :100: :fire:

deploy/helm/templates/_cluster_orchest.tpl
nhaghighat: I guess you are seeing an old version, it is already changed in 
https://github.com/orchest/orchest/commit/dbff5e589e6edf11b78b67dd01a557e05f374a1b#diff-dd107175ef4d1c02b230017a4ea7587a1d207b1fe70e9d530bd3e4498a13a40cR8

services/orchest-api/app/app/core/sessions/_manifests.py
nhaghighat: This is also outdated, it is fixed in later commits

nhaghighat:
> Absolutely stellar work @nhaghighat! I am impressed by all the changes you have made

Thanks so much

yannickperrenet:
Some quick notes, on the provided PR description, that I took whilst testing:

* Had to run `make docker-registry` a second time as the first time I got:

```
helm install registry-resources --namespace orchest   --create-namespace --atomic --timeout 600s --set cluster-resources.enabled=true --set cluster-resources.registry.enabled=true --set cluster-resources.registry.secret=registry-tls-secret --set cluster-resources.registry.name=docker-registry --set cluster-resources.registry.issuer.name=certificate-issuer helm
Error: INSTALLATION FAILED: release registry-resources failed, and has been uninstalled due to atomic being set: Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": dial tcp 10.111.115.214:443: connect: connection refused
make: *** [Makefile:233: bin/registry-resources] Error 1
```

* I assume the `scripts/install_minikube.sh` also needs a line to enable the ingress addon? i.e. `minikube addons enable ingress`)

* To build all the images and prevent the images being pulled from DockerHub I took the approach of ssh'ing to all the minikube nodes and build on the nodes directly:

```
# Do this for every node (`minikube list nodes`)
minikube ssh -n minikube

# Get code so we can build
git clone https://github.com/orchest/orchest \
    && cd orchest \
    && git checkout feat/integrate-with-distribute-fs

# Build minimal set of containers
scripts/build_container.sh -i jupyter-server \
    -i jupyter-enterprise-gateway \
    -i celery-worker \
    -i orchest-api \
    -i orchest-webserver \
    -i session-sidecar \
    -i auth-server \
```
This way I did not have to change the value of `image.registry` in all the `values.yml` files.


<details><summary>In case someone wants to take the `tag` `push` approach after building locally instead:</summary>

```
docker tag orchest/auth-server 127.0.0.1:8080/orchest/auth-server \
    && docker tag orchest/celery-worker 127.0.0.1:8080/orchest/celery-worker \
    && docker tag orchest/jupyter-enterprise-gateway 127.0.0.1:8080/orchest/jupyter-enterprise-gateway \
    && docker tag orchest/jupyter-server 127.0.0.1:8080/orchest/jupyter-server \
    && docker tag orchest/orchest-api 127.0.0.1:8080/orchest/orchest-api \
    && docker tag orchest/orchest-webserver 127.0.0.1:8080/orchest/orchest-webserver \
    && docker tag orchest/session-sidecar 127.0.0.1:8080/orchest/session-sidecar \
    && docker push 127.0.0.1:8080/orchest/auth-server \
    && docker push 127.0.0.1:8080/orchest/celery-worker \
    && docker push 127.0.0.1:8080/orchest/jupyter-enterprise-gateway \
    && docker push 127.0.0.1:8080/orchest/jupyter-server \
    && docker push 127.0.0.1:8080/orchest/orchest-api \
    && docker push 127.0.0.1:8080/orchest/orchest-webserver \
    && docker push 127.0.0.1:8080/orchest/session-sidecar
```

</details>

Issue #201, opened 2021-04-09
Event-based job trigger


yannickperrenet:
Let jobs be triggered by some external event.
* Webhooks?
* HTTP request to trigger a job? (This is already possible, but not considered usable for users)

Currently, a possible workaround looks as follows:
* Create a cron scheduled job that runs as often as possible.
* Let the pipeline poll to check whether it should be triggered.
* If "triggered" continue with the next steps in the pipeline, if not "triggered" then raise an error in the step to make the pipeline fail/not execute.

yannickperrenet:
Once #330 gets merged it is possible to run a pipeline through the CLI. This means that users can set up their own "listener" and trigger a pipeline run using the CLI.

For example, listen for a POST request on their own endpoint and trigger a pipeline run using the CLI.

astrojuanlu:
>  HTTP request to trigger a job? (This is already possible, but not considered usable for users)

@yannickperrenet Could you elaborate?

yannickperrenet:
> HTTP request to trigger a job? (This is already possible, but not considered usable for users)

I've updated the PR description regarding this point. It was possible to directly call the `orchest-api` to run a job, and we used to have (but removed in when porting to Kubernetes) `./orchest run ...` (as part of the CLI) to run a job.

Now we are looking to create a public API so that users can interact programmatically with Orchest as well.

ricklamers:
One of the users mentioned in Slack they'd like to see pipeline step execution information as part of the API response:

>Hi, quick question. Does someone know if  it is posible to capture the exit of a pipeline? For example, a pipeline step that fails or exit with some response

This is something we could embed in the pipeline job API endpoint.

yannickperrenet:
+1 from @franceme.

minitriga:
Hiya, 

About to move into a new job that would require doing integrations between software vendors one way of doing this is having orchest in the middle, to act as the glue between platforms. Being able to trigger via webhook and API would be great. 

For my use case a webhooks would be most preferred as the the initiator can send a webhook to a given address when it has been updated and then the payload could be passed to the first task.

everlectric-technology:
We have some developers who prefer working in python, and some node.js, for this reason, and to make code accessible and well documented for everyone, we are using node-red for node.js and something like orchestra as a python counterpart. Primarily process orchestration would still take place from node-red, due to its number of connectors etc, but would like to have orchest be able to be integrated into this process layer, so that developer teams can collaborate programatically.

PR #744, opened 2022-02-27
Orchest-ctl refactoring PR 4, `orchest start/stop/restart`

fruttasecca:
## Description

Ports `orchest start`, `orchest stop`, `orchest restart` to k8s. Pre-k8s the `orchest-api` was in charge of doing a cleanup on start, this is now orchestrated by `orchest-ctl` on start and stop.


services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -50,6 +50,56 @@ def get_orchest_deployments(
     return responses
 
 
+def match_labels_to_label_selector(match_labels: Dict[str, str]) -> str:
+    return ",".join([f"{k}={v}" for k, v in match_labels.items()])
+
+
+def get_orchest_deployments_pods(
+    deployments: Union[Optional[List[str]], k8s_client.V1Pod] = None,
```
yannickperrenet: Not entirely sure that type is correct. Shouldn't it instead be?
```python
deployments: Optional[Union[List[str], List[k8s_client.V1Deployment]]] = None,
```

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -117,6 +117,7 @@ def install():
     # port = 8001
     # utils.echo(f"Orchest is running at: http://localhost:{port}")
     utils.echo("Installation was successful.")
+    utils.echo("Orchest is running, portforward to the webserver to access it.")
```
yannickperrenet, line 121: Don't think we want to include this message? Should be able to just go to the `minikube ip`.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -283,3 +284,65 @@ def stop():
...
+                for d in k8sw.get_orchest_deployments(deployments_to_start_names)
+                if d is not None and d.status.ready_replicas != d.spec.replicas
+            ]
+            progress.update(len(deployments_to_start) - len(tmp_deployments_to_start))
+            deployments_to_start = tmp_deployments_to_start
+
+    # K8S_TODO: coordinate with ingress for this.
+    # port = 8001
+    # utils.echo(f"Orchest is running at: http://localhost:{port}")
+    utils.echo("Orchest is running, portforward to the webserver to access it.")
```
yannickperrenet, line 404: Same as https://github.com/orchest/orchest/pull/744/commits/9bd640019ac112d4f213e68e31c11bc7d6ca958f#r816040130.

I do agree that we need some form of message, but don't think this is the one.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -236,3 +236,50 @@ def status(output_json: bool = False):
         if status == "unhealthy":
             data["reason"] = {"deployments": unhealthy_deployments}
         utils.echo_json(data)
+
+
+def stop():
```
yannickperrenet, line 254: Mostly a note for myself for later, but this `stop` only takes care of `config.ORCHEST_DEPLOYMENTS` which is the "orchest control-pane". However, other resources (like session resources, e.g. `jupyter-server`) are left running. 

I guess this is taken care of using some form of post-stop-cleanup.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -147,17 +158,38 @@ def _get_ongoing_status_changing_pods() -> List[k8s_client.V1Pod]:
...
+    running of other CLI commands that alter the state of the Orchest
+    deployment.
+    """
+    pod = _get_ongoing_status_changing_pod()
+    if pod.metadata.name != os.environ["POD_NAME"]:
+        cmd = pod.metadata.labels.get("command", "update")
+        if pod.metadata.labels["app"] == "update-server":
+            cmd = "update"
+        else:
+            cmd = pod.metadata.labels["command"]
```
yannickperrenet, line 178: The first `cmd = ...` is obsolete with the current `if`-statement that follows it.

services/orchest-api/app/app/__init__.py
```diff
@@ -233,6 +124,9 @@ def create_app(config_class=None, use_db=True, be_scheduler=False, to_migrate_db
             args=[app],
         )
 
+        if _utils.is_running_from_reloader():
+            trigger_conditional_jupyter_build(app)
```
yannickperrenet, line 128: Same as https://github.com/orchest/orchest/pull/721#discussion_r808810571. 

> Should probably be `if not _utils.is_running_from_reloader():` otherwise the code will only execute in the reloader when running in Flask development mode.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -201,3 +202,87 @@ def get_ongoing_status_change() -> Optional[config.OrchestStatus]:
...
+                ],
+                "volumeMounts": [{"name": "dockersock", "mountPath": "/var/run"}],
+            }
+        ],
+        "volumes": [
+            {
+                "name": "dockersock",
+                "hostPath": {"path": "/var/run", "type": "Directory"},
+            }
+        ],
```
yannickperrenet, line 247: Why are we mounting the `dockersock`? Not sure I follow.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -201,3 +202,87 @@ def get_ongoing_status_change() -> Optional[config.OrchestStatus]:
...
+    },
+    "spec": {
+        "restartPolicy": "Never",
+        "containers": [
+            {
+                "name": "orchest-api-cleanup",
+                "image": f'orchest/orchest-api:{os.environ["ORCHEST_VERSION"]}',
+                "command": ["/bin/sh", "-c"],
+                # Make sure the database is compatible with the code.
+                "args": ["python migration_manager.py db migrate && python cleanup.py"],
```
yannickperrenet, line 226: Regarding the `args`, it feels a bit strange that we are running database migrations on `orchest stop` due to `stop` also running cleanup.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: Indeed

services/orchest-ctl/app/orchest/_core.py
fruttasecca: Agree, it's just a placeholder until we have setup ingress.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: See https://github.com/orchest/orchest/pull/744#discussion_r816078449

services/orchest-api/app/app/__init__.py
fruttasecca: good catch

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -201,3 +202,87 @@ def get_ongoing_status_change() -> Optional[config.OrchestStatus]:
...
+    """
+    manifest = _cleanup_pod_manifest
+    # K8S_TODO: fix this once we move to versioned images.
+    manifest["spec"]["containers"][0]["image"] = "orchest/orchest-api:latest"
+    resp = k8s_core_api.create_namespaced_pod(config.ORCHEST_NAMESPACE, manifest)
+    pod_name = resp.metadata.name
+
+    max_retries = 1000
+    status = None
+    while max_retries > 0:
```
yannickperrenet, line 267: Don't we want to return earlier than after `1000` retries? As a user waiting ~17 minutes for Orchest to start (due to some issue), since cleanup is run before Orchest is started, is too long imo. How about reducing the `max_retries`?

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: My bad, forgot to leave a comment. Essentially it's there so that `orchest-api` doesn't break since we still have to go through a complete cleanup of the codebase that removes references to docker.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: Indeed might sound strange, although I don't feel strongly about it. It might save us some troubles in the case of some inconsistent state of Orchest, but I don't feel attached to it :).

services/orchest-ctl/app/orchest/_core.py
fruttasecca: >I guess this is taken care of using some form of post-stop-cleanup?
yes

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -283,3 +284,65 @@ def stop():
...
+
+def start():
+    depls = k8sw.get_orchest_deployments(config.ORCHEST_DEPLOYMENTS)
+    missing_deployments = []
+    deployments_to_start = []
+    for depl_name, depl in zip(config.ORCHEST_DEPLOYMENTS, depls):
+        if depl is None:
+            missing_deployments.append(depl_name)
+        elif depl.spec.replicas == 0:
+            deployments_to_start.append(depl)
```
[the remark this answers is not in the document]
fruttasecca: >Let's say that for some reason some deployments have replicas > 0 (e.g. due to concurrent start invocations or because of inconsistent state), then (with the current implementation) we would proceed to start the other

We already prevent concurrent invocations of conflicting commands, but let's assume there is some inconsistent state. 

>Does this work with the probes we have set up?

I don't see why not, both the ones that are already running and the newly started ones will keep trying to get to a healthy state by polling, do you have a particular case in mind?

>Also I am thinking whether we have some form of handshake that happens the moment two (or more) services start up at the same time.

Not sure I completely get what you mean, but it doesn't ring a bell when it comes to edge cases atm.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -317,33 +334,47 @@ def start():
...
+    pre_cleanup_deployments_to_start = []
+    post_cleanup_deployments_to_start = []
+    for depl in deployments_to_start:
+        # Don't start those until after cleanup, this way the webserver
+        # won't be available to the user and the orchest-api scheduler
+        # won't run any job while the cleanup is in progress.
+        if depl.metadata.name in ["orchest-api", "orchest-webserver"]:
+            post_cleanup_deployments_to_start.append(depl)
+        else:
+            pre_cleanup_deployments_to_start.append(depl)
```
yannickperrenet, line 368: This might be a bit far fetched, but I was trying to come up with an edge case where this construction of pre-cleanup & post-cleanup & `replicas == 0` (see https://github.com/orchest/orchest/pull/744#discussion_r816048190) would cause an issue.

Let's say the `orchest-api` is still running when invoking `orchest start` (this could be the case if our public API is running (e.g. for #201) and we choose to shut-down the `orchest-webserver`) then the cleanup will actually remove resources (e.g. ongoing runs) that we don't want to have cleaned.

Moreover, it feels like dangerous territory in some sense to have `start` perform cleanup (but I understand why we are doing it). The moment we allow concurrent runs of `start` (maybe once the CLI becomes a binary the `k8sw.abort_if_unsafe()` check is different) or someone invokes `start` whilst Orchest is in an invalid state (e.g. only one Orchest deployment is down) then we cleanup resources.

Not sure whether I am overthinking this, but now that cleanup no longer lives in the `orchest-api` (which needs the cleanup) and it being decoupled we need to be absolutely sure we gate it correctly.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -57,7 +57,7 @@ def match_labels_to_label_selector(match_labels: Dict[str, str]) -> str:
 
 
 def get_orchest_deployments_pods(
-    deployments: Union[Optional[List[str]], k8s_client.V1Pod] = None,
+    deployments: Optional[Union[List[str], List[k8s_client.V1Pod]]] = None,
```
yannickperrenet, line 60: Shouldn't this be `k8s_client.V1Deployment` instead of `V1Pod`?

fruttasecca: Don't think so,we are getting pods here right?

services/orchest-ctl/app/orchest/_core.py
fruttasecca: About the edge case, if we assume an inconsistent state then indeed CLI commands might lead to undesired consequences, like in this case a cleanup stopping runs/sessions/builds etc. I can see that something like this can happen and I would be ok with removing the cleanup on `start`, essentially it's there in case cleanup fails on `stop`.

We can also make the start logic slightly more complex, i.e. cleanup only in some cases, although I am not super stoked at the idea of doing such a thing in this phase.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
yannickperrenet: How about leaving setting it at `999` retries and adding a an `if`-statement like:
```python
if max_retries % 50 == 0:
    utils.echo("Still cleaning up resources...")
```
to indicate that the process is still alive. Otherwise as a user you get all anxious "can I kill it, is it hanging? or should I wait?"

fruttasecca: Not sure how it will play with the progress bar, I'll try this out.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
yannickperrenet: Not sure whether the following could be an issue:
1. Orchest is executing a job
2. A step of the job update its status in the DB
3. `orchest stop` invokes cleanup which invokes DB migrations
4. Another step of the job update its status in the DB

What happens if the DB is used whilst a migration is going on?

services/orchest-ctl/app/orchest/_k8s_wrapper.py
yannickperrenet: If it destroys the progress bar, then lets leave it as is :+1:

services/orchest-ctl/app/orchest/_core.py
yannickperrenet: I would vote for only doing cleanup on `stop`. We currently run it on `start`, but running it on `stop` feels better and solves for more edge cases.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
yannickperrenet: The fact that the argument is called `deployments` and we run:
```python
elif not all(isinstance(depl, k8s_client.V1Deployment) for depl in deployments)
```
it made me believe that we expect a list of `V1Deployment` not `V1Pod`. Right?

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: >What happens if the DB is used whilst a migration is going on?

`orchest-api` and `orchest-webserver` get shut down before running the cleanup, so the step status won't be updated (the cleanup will later mark it as aborted). Assuming the `celery-worker` could update the db directly (which I think should something we should be considering), that just means we have to include the `celery-worker` in the pre cleanup shutdown

services/orchest-ctl/app/orchest/_core.py
fruttasecca: Happy to only run in on `stop`, will do.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: Right, sorry, I was looking at the return type. Will fix!

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: I think it makes sense to have it, it only "disrupts it" by writing the message, then another progress bar is drawn. Given that it won't happen often I think it's good to have it in case, for some reason, things are slow. Btw, I have done this in another branch and PR to avoid getting slowed down by having to switch back and forth and conflicts, hopefully by today we can merge most `orchest-ctl` related PRs.

yannickperrenet (approved):
Great stuff! :man_singer:

nhaghighat (approved):

PR #745, opened 2022-02-27
Orchest-ctl refactoring PR 5, `orchest adduser`

fruttasecca:
## Description

Ports `orchest adduser` to k8s. The command is not completely tested given that it requires the `auth-server` to be running, which is currently not included in our deployment.

### DIFF
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional
 
 import typer
 from kubernetes import client as k8s_client
+from kubernetes.stream import stream
 
 from app import config, utils
 from app.connections import k8s_core_api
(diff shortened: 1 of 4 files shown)


services/orchest-ctl/app/orchest/_core.py
```diff
@@ -407,3 +408,57 @@ def start():
...
+    if is_admin:
+        cmd += " --is_admin"
+    cmd = cmd.split()
+
+    resp = stream(
+        k8s_core_api.connect_get_namespaced_pod_exec,
+        pod.metadata.name,
+        config.ORCHEST_NAMESPACE,
+        command=cmd[:1],
+        args=cmd[1:],
```
yannickperrenet, line 458: Given that you do `cmd.split()` I don't think `cmd[:1]` and `cmd[1:]` get the correct output.

The list would become something like: `cmd = ["python", "add_user.py", ..., "--token", "{token}", ...]`

fruttasecca: Not really sure I understand what you mean, but I think I'll go ahead and just avoid this whole string and split thing in favour of just creating a list directly

yannickperrenet: :+1:

yannickperrenet (approved):
LGTM!

PR #747, opened 2022-03-01
Orchest-ctl refactoring PR 6, `orchest update/hidden-update`

fruttasecca:
Adds functionality to perform a CLI update. To ensure correctness, when updating from `v1` to `v2`, `orchest-ctl:v1` will be in charge of stopping Orchest, while `orchest:ctl:v2` will be in charge of updating it (which currently implies starting it).

Note: this is not about the GUI update, but a dependency of.


yannickperrenet:
Probably best to also quickly update the docs (https://docs.orchest.io/en/latest/getting_started/how_to.html#self-host-orchest) to no longer mention the SSL stuff. Not directly related but removing all the unused code made me think of it.

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -43,28 +44,26 @@ def is_orchest_already_installed() -> bool:
...
-        raise typer.Exit(code=1)
 
-    utils.echo(f"Installing Orchest {orchest_version}.")
+def _run_helm_with_progress_bar(mode: HelmMode) -> None:
+    if mode == HelmMode.INSTALL:
+        cmd = ["make", "orchest"]
+        label = "Installation"
+    elif mode == HelmMode.UPGRADE:
+        cmd = ["make", "orchest-upgrade"]
+        label = "Update"
```
yannickperrenet, line 58: Why not use `label = mode.name` or `label = mode.value` here instead? The fact that `INSTALL = "Install"` in the `Enum` but that the `label` is called `"Installation"` was a small discrepancy to me.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -287,3 +300,61 @@ def orchest_cleanup() -> None:
         )
 
     k8s_core_api.delete_namespaced_pod(pod_name, config.ORCHEST_NAMESPACE)
+
+
+def _get_orchest_ctl_update_post_manifest(update_to_version: str) -> dict:
```
yannickperrenet, line 294: What alternatives have you considered vs the current approach (where we now have to keep the template in sync)?

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -485,10 +486,78 @@ def add_user(username: str, password: str, token: str, is_admin: str) -> None:
...
+            )
+        except k8s_client.ApiException as e:
+            if e.status != 404:
+                raise
+            time.sleep(1)
+        else:
+            status = resp.status.phase
+            if status in ["Succeeded", "Failed", "Unknown", "Running"]:
+                break
+        time.sleep(1)
```
yannickperrenet: Probably needs an extra indentation level so it is inside the `else`. Otherwise we will sleep for 2s in case of the `except` being hit.

services/orchest-ctl/app/orchest/_core.py
fruttasecca: sounds good

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -485,10 +486,78 @@ def add_user(username: str, password: str, token: str, is_admin: str) -> None:
...
+        command=["orchest", "hidden-update"],
+        _preload_content=False,
+    )
+    while resp.is_open():
+        resp.update()
+        # nl=False, wrap=False to not disrupt the progress bar.
+        if resp.peek_stdout():
+            utils.echo(resp.read_stdout(), nl=False, wrap=False)
+        if resp.peek_stderr():
+            utils.echo(resp.read_stdout(), nl=False, wrap=False, err=True)
```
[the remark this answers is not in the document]
fruttasecca: yep

services/orchest-ctl/app/orchest/_k8s_wrapper.py
yannickperrenet: Loading the yaml into a dictionary and change the appropriate values seems safer to me, because:
* Adding options to the yaml will be automatically reflected in the converted object
* Same for removal
* Changes of options are still overwritten on the object through code

Given the fact that Kubernetes has strict yaml configuration specs we know that the option we are overwriting is always valid. 

Converting yaml to a dict should be as simple as `yaml.load(...)` afaik.

yannickperrenet (approved):
:100:

yannickperrenet (approved):
Still looking good to me :wink:

PR #748, opened 2022-03-01
Remove jquery from pipeline editor and add minor UX improvements

iannbing:
## Description

This PR aims at removing jQuery and RefManager from PipelineView component in order to make the implementation closer to the React way. 

The refactoring also comes with some improvements on the drag-n-drop behavior, including:

- when select/drag a step, the step and its connections (lines) will be moved to top, making them visually recognizable
- when a connection line is selected, the start/end dots will also be highlighted.
- several minor bugs were fixed, e.g. connection lines sometimes are visually gone.

## Checklist


yannickperrenet:
Great work! The editor feels smooth to me, although the connections are lagging a bit when selecting many steps and moving them around.

### Notes
> when select/drag a step, the step and its connections (lines) will be moved to top, making them visually recognizable

Now I can do the following, which I don't really like:
image
1. Select connection --> it will be over the unselected step
2. Unselect connection (just click anywhere else in the editor)

Another screenshot:
image
Here, when dragging the step in the lower right corner the connection goes over some steps but not over others. I found this a bit confusing, I would rather default to always have the connections below steps.

**Connections don't always connect to the step**
image

Exact reproduction steps are a bit hard, but I can reproduce every now and then by:
1. Select all steps
2. Move them
3. Make a subselection
4. Move them
All these steps are done in quick succession.

After some more playing with the editor I came up with better steps:
1. Quickly "throw" a step. Meaning that you hold your mouse to select and then quickly move your mouse and let go of the step (don't drag and hold).

**What is the UUID that is shown as part of the steps?**

As you can see in the above screenshots, there is an additional string included below the filename. I guess this was used for testing, but should be removed.

> when a connection line is selected, the start/end dots will also be highlighted.

I really love this! :heart:

yannickperrenet:
### Some more notes

**When wanting to do a run, the button needs to be pressed twice**

1. Drag select a step
2. Press "run selected steps"
3. Note that it didn't work
4. Press "run selected steps" again
5. Now it executes

**Styling of step status seems off**

Not sure whether this is related to the aforementioned UUID that is included, but:
image
As you can see the background of "Failure" is a lot bigger (my guess is that this is caused by the icon as it not only happens for "Failure"

iannbing:
> Connections don't always connect to the step
Thanks for spotting this issue. To precisely reproduce it,
- select any step(s)
- drag a step that is NOT selected
Just fixed it in my latest commit.

I also changed the behavior to "steps are always on top of connections". What I don't like about it is when you move a step over another, you often see the following. If I had to describe it, it feels like the "arm" of the step is twisted.

Screenshot 2022-03-02 at 16 44 55

Earlier I actually implemented exactly as you proposed: move steps to top in the sequence of selection. But then when you create a multi-selection rectangle, the steps are like pop-corns jumping to your eye sight, which is quite annoying.

yannickperrenet:
> **Connections don't always connect to the step**
> After some more playing with the editor I came up with better steps:
> 1. Quickly "throw" a step. Meaning that you hold your mouse to select and then quickly move your mouse and let go of the step (don't drag and hold).

I can still easily reproduce this:
image

> **Styling of step status seems off**

This is still the case. 
1. Do a pipeline run
2. See that the state change (a state that includes an icon, e.g. ":heavy_check_mark: Completed") is not rendered correctly.

> Just fixed it in my latest commit. Could you try again?

Works for me. The pipeline runs issue is fixed and connections are now always displayed below other steps. I agree that we can improve on the connection (z-index) behavior (but priority wise I think it is best to postpone it -- feel free to open an issue ;)

services/orchest-webserver/client/src/utils/webserver-utils.ts
```diff
@@ -1,10 +1,15 @@
...
-import { extensionFromFilename, makeRequest } from "@orchest/lib-utils";
+import {
+  extensionFromFilename,
+  hasValue,
+  makeRequest,
+} from "@orchest/lib-utils";
 import Ajv from "ajv";
 import dashify from "dashify";
 import { format, parseISO } from "date-fns";
+import $ from "jquery";
```
ricklamers, line 12: Still importing jQuery? Might be overlooking something.

ricklamers:
Found a minor issue: when you click on a selected step the properties panel doesn't open. I think it should open, especially as we now no longer open the properties pane on drag single select.

ricklamers:
Minor issue: when zoomed out the panning is no longer "1:1". Your cursor should stick to the same position on the grid (like it is on 100% zoom level).

ricklamers:
Minor nitpick: since you can't drag from a start-connecting-dot on the pipeline step (left circle/dot), when you're **not** dragging a connection (the arrow under the cursor) the start-connecting-dots shouldn't have the "drop zone" appear on hover.

Really like the addition of the visual drop zone though!

ricklamers:
A minor issue when dragging a selection of steps: https://www.tella.tv/video/cl0asjt58000309l82ldkej7s/view

Basically, the step clicked-and-dragged that causes the whole selection to move seems to move more and therefore disconnect with the connection-arrow during the drag.

ricklamers:
Minor nitpick: I think we should make the drag less sensitive. I.e. when you try to click a step it should register as a drag when you drag it _more_. Right now sometimes you try to click a step, move the mouse a bit in the process (unintentionally) and the click is not registered as a click on the step but as a drag.

ricklamers:
Minor nitpick: like Yannick mentioned there's a bit of a delay in the connection position update when moving the steps (the connections seem to update their position ever so slightly delayed). If this is too difficult to fix without a major refactor we can leave it like this. I'd say it's similar in "severity" to what we had before, although it was one of the things that I hoped this refactor would be able to remedy/improve.

ricklamers:
Overall remark about the refactor: the code structure & abstractions seemed to have improved a ton. Great job!

services/orchest-webserver/client/src/utils/webserver-utils.ts
iannbing: This PR doesn't remove all jquery usages in the codebase, but only PipelineView. In fact, jquery is used in `webserver-utils.ts` but it was not imported explicitly (because currently we expose jquery to `window`). This PR also imports jquery explicitly per file, so we can easily find them for future refactoring.

iannbing:
@yannickperrenet @ricklamers the delay issue should be addressed by the new commits. The approach was explained [above](https://github.com/orchest/orchest/pull/748#issuecomment-1058302815), except that it's already quite smooth without canvas API.

yannickperrenet:
> Minor nitpick: I think we should make the drag less sensitive. I.e. when you try to click a step it should register as a drag when you drag it more. Right now sometimes you try to click a step, move the mouse a bit in the process (unintentionally) and the click is not registered as a click on the step but as a drag.

Not sure whether this caused it, but I didn't have the following issue before (I can not let go of the step anymore):

https://user-images.githubusercontent.com/26223174/156831313-fa9939de-f65b-456e-9319-0f3ddf113569.mp4

To reproduce: do the "throwing" of the step again but do it with your cursor close to the dot where connections would connect to or close to the outside of the step (so that the quick drag motion is outside of the step).

https://user-images.githubusercontent.com/26223174/156832115-89b69f5e-1cb6-47ca-9b25-6245e1568209.mp4


> the delay issue should be addressed by the new commits

The delay is indeed a lot less severe. Combining that with the additional re-render after `savePositions` makes the editing experience smooth again. Really well done! :100:

iannbing:
@yannickperrenet I think it's okay from UX perspective since user intended to "drag" (with lower sensitivity, user really needs to drag to trigger the dragging behavior). And user should be aware that they started dragging from the edge. So, I think "dragging without touching the step" in this specific case shouldn't be confusing / annoying.

The only problem is that user cannot "finish" dragging. I added an event listener to handle this case. So user will not be locked up by the dragging behavior :-) You can checkout the last commit I made.

yannickperrenet:
> You can checkout the last commit and give it a try.

Awesome! Works like a charm :100:

yannickperrenet (approved):
Great stuff! :muscle:

PR #749, opened 2022-03-01
Feat/node agent

nhaghighat:
## Description

This branch adds Node-Agent Service, currently it only downloads provided list of images

## Checklist


yannickperrenet:
As a side-note, I will review the commits starting at https://github.com/orchest/orchest/pull/749/commits/9b0eb9e0da26f4fe347f43b2a564e22cc76325b8 as to only review node-agent related changes.

On another note, could you install pre-commit (`pre-commit install`) so that your code is automatically formatted on commit ([docs](https://docs.orchest.io/en/latest/development/development_workflow.html))?

services/node-agent/app/main.py
```diff
@@ -0,0 +1,43 @@
...
+
+    image_puller = ImagePuller(image_puller_interval,
+                               image_puller_policy,
+                               image_puller_retries,
+                               image_puller_images,
+                               image_puller_log_level)
+
+    loop = asyncio.get_event_loop()
+    loop.create_task(image_puller.run())
+    loop.run_forever()
```
yannickperrenet, line 41: From the [Python docs](https://docs.python.org/3/library/asyncio-eventloop.html):
> Application developers should typically use the high-level asyncio functions, such as [asyncio.run](https://docs.python.org/3/library/asyncio-task.html#asyncio.run), and should rarely need to reference the loop object or call its methods.

So you could write this like: 
```python
asyncio.run(image_puller.run())
```

services/node-agent/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.8-slim
+LABEL maintainer="Orchest B.V. https://www.orchest.io"
+
+WORKDIR /orchest/services/node-agent/
+
+# Get all Python requirements in place and install them.
+COPY . ./
+RUN pip3 install -r requirements.txt
```
yannickperrenet, line 8: For faster re-builds, you can change this to:
```dockerfile
COPY ./requirements.txt ./
RUN pip3 install -r requirements.txt

COPY . ./
```
We have done this in other services (e.g. [`orchest-api`](https://github.com/orchest/orchest/blob/master/services/orchest-api/Dockerfile#L4)) as well so that changes of the code don't require the dependencies to be installed again.

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,84 @@
+import time
+import logging
+
+from docker.client import DockerClient
+from docker.errors import NotFound
+
+POLICY_IF_NOT_PRESENT = "IfNotPresent"
+POLICY_ALYWAYS = "Always"
+policies = (POLICY_IF_NOT_PRESENT, POLICY_ALYWAYS)
```
yannickperrenet, line 9: Probably better to change this to an [`Enum`](https://docs.python.org/3/library/enum.html):
```python
from enum import Enum

class Policy(Enum):
    IF_NOT_PRESENT = "IfNotPresent"
    ALWAYS = "Always"
```

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,84 @@
...
+        self.logger.info(f"Pulling image '{image_name}'...")
+        if not self.download_image(image_name):
+            self.logger.warning(f"Image '{image_name}' was not downloaded!")
+
+    def image_exists(self, image_name: str) -> bool:
+        """Checks for the existence of the named image using the configured container runtime."""
+        result = True
+        t0 = time.time()
+        try:
+            DockerClient.from_env().images.get(image_name)
```
yannickperrenet, line 45: Just so that I understand correctly, we use the Docker of the host/node to pull the images to the specific node that the daemonset is running on?

nhaghighat:
> As a side-note, I will review the commits starting at https://github.com/orchest/orchest/commit/9b0eb9e0da26f4fe347f43b2a564e22cc76325b8 as to only review node-agent related changes.
On another note, could you install pre-commit (pre-commit install) so that your code is automatically formatted on commit

Sounds good, will add those.

services/node-agent/app/image_puller.py
nhaghighat: right. we can later add more run-times.

services/node-agent/Dockerfile
nhaghighat: Good idea, will add that

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self.logger.info("Starting image puller.")
+        
+        while True:
+            for image_name in self.images:
+                i = 0
+                while i < self.num_retries:
+                    try:
+                        self.pull_image(image_name)
+                        break
```
yannickperrenet, line 77: We are not making any use of `asyncio` right? 

I do like using an async approach through to pull in all the images. How about making the `self.pull_image` async (pre-k8s code to do so in `orchest-ctl`: [L62](https://github.com/orchest/orchest/blob/73e83431066acec964b6ecaf11dd59af352d219a/services/orchest-ctl/app/docker_wrapper.py#L62))? So it would becomes something along the lines of:

```python
while True:
    i = 0
    while i < self.num_retries:
        await self.pull_images(self.images)
    ...
```
Although if we need to retry we need to make sure to not pull the existing images again (i.e. images X, Y, Z are being pulled and X, Y succeeds but Z does not, then in the retry we only want to try to pull Z again)

services/node-agent/app/main.py
nhaghighat: you are right, currently it is better to use asyncio.run.

services/node-agent/app/main.py
```diff
@@ -0,0 +1,43 @@
+import logging
+import asyncio
+import argparse
+
+from image_puller import *
```
yannickperrenet, line 5: I would change this to:
```python
from image_puller import ImagePuller
```

services/node-agent/app/main.py
```diff
@@ -0,0 +1,43 @@
...
+    image_puller_interval = arguments['image_puller_interval']
+    image_puller_policy = arguments['image_puller_policy']
+    image_puller_retries = arguments['image_puller_retries']
+    image_puller_images = arguments['image_puller_images']
+
+    image_puller = ImagePuller(image_puller_interval,
+                               image_puller_policy,
+                               image_puller_retries,
+                               image_puller_images,
+                               image_puller_log_level)
```
yannickperrenet, line 37: My guess is that you can rewrite this to:
```python
arguments = vars(parser.parse_args())
image_puller = ImagePuller(**arguments)
```

services/node-agent/app/image_puller.py
yannickperrenet: Don't forget you would have to change usage of the constants (e.g. [this line](https://github.com/orchest/orchest/pull/749/commits/9b0eb9e0da26f4fe347f43b2a564e22cc76325b8#diff-25e8025cc5c0bb3a3dfef40ebc002b0679da51b08895fa46fc88c03821b71e80R31)).

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,84 @@
+import time
```
yannickperrenet, line 1: We are trying to improve the documentation of the codebase on a continual basis, could you add a short module level docstring ([example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)) explaining what this module does? That would be super useful.

nhaghighat: Sure

yannickperrenet:
Might be best to change the base of the PR to merge it into https://github.com/orchest/orchest/pull/742 to prevent us accidentally merging this PR first, which would result in the changes from #742 to be included as well.

deploy/helm/charts/node-agent/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+description: The helm chart for node-agent deployment
+name: node-agent
+version: 0.0.1
+icon: https://github.com/orchest/orchest/blob/master/docs/source/img/logo.png
+sources:
+  - https://github.com/orchest/orchest/tree/master/services/node-agent
```
fruttasecca, line 7: I guess we want `orchest-ctl` to be aware of the `node-agent` when starting and stopping Orchest. What would be the best practise way to do that?

services/node-agent/app/main.py
```diff
@@ -0,0 +1,34 @@
...
+                        help='Specifies image puller interval in sec, default 60 sec.', 
+                        default=60, type=int)
+    parser.add_argument('--image-puller-policy', dest='image_puller_policy', nargs='?',
+                        help='Specifies image puller policy. default "IfNotPresent".', 
+                        default=Policy.IfNotPresent, type=Policy)
+    parser.add_argument('--image-puller-retries', dest='image_puller_retries', nargs='?',
+                        help='Specifies image puller number of retries. default 3.', 
+                        default=3, type=int)
+    parser.add_argument('--image-puller-images', dest='image_puller_images', nargs='+',
+                        help='Specifies image puller number of retries.')
```
fruttasecca, line 27: Guess the `help` here needs to be fixed.

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+
+from docker.client import DockerClient
+from docker.errors import NotFound
```
fruttasecca, line 5: Bit of a naive question, with this we are assuming that we have docker on the node, is this always the case with minikube?

services/node-agent/app/image_puller.py
```diff
@@ -0,0 +1,86 @@
...
+        """Pulls the image.
+
+        If the policy is `IfNotPresent` the set of pulled image names is
+        checked and, if present, the method returns.  Otherwise, the pull attempt is made
+        and the set of pulled images is updated, when successful.
+        """
+        if self.policy == Policy.IfNotPresent:
+            if self.image_exists(image_name):
+                return
+            self.logger.warning(f"Image '{image_name}' is not found - attempting pull...")
```
fruttasecca, line 36: Guess the `warning` should actually be a `info`

services/node-agent/app/image_puller.py
fruttasecca: I guess the retry can be at the `pull_image` (note the lack of a "s") level

deploy/helm/charts/node-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
...
+        imagePullPolicy: {{ include "library.spec.images.pullPolicy" . }}
+        command:
+        - python
+        - ./app/main.py
+        args:
+        - --image-puller-log-level=INFO
+        - --image-puller-interval=60
+        - --image-puller-policy=IfNotPresent
+        - --image-puller-retries=3
+        - --image-puller-images=ubuntu:latest
```
fruttasecca, line 26: What's the plan for keeping the images that are being pulled in sync with the cluster version? Currently the orchest `namespace` does have a `version` label that can be used, or we can set the cluster version using an environment variable set in this `daemonset` char

fruttasecca:
I really like this approach :)).

services/node-agent/app/image_puller.py
nhaghighat: I tried minikube with docker and kvm driver, and they both had docker installed.

fruttasecca: I see. I guess we will likely have to revise this once we move away from minikube.

deploy/helm/charts/node-agent/templates/daemonset.yaml
nhaghighat: I think the images to be pulled can be setted with helm, during deployment. and we can use orchest-ctl or makefile to configure helm.

deploy/helm/charts/node-agent/Chart.yaml
fruttasecca: >but for start and stop, the same approach you followed for other components

When it comes to starting/stopping Orchest we are currently scaling the deployments to 0 replicas, what do you propose we do with daemonsets?

nhaghighat: Discussed offline, since node-agent is pretty simple component, we can updated the daemonset manifest for now for updating.

deploy/helm/charts/node-agent/templates/daemonset.yaml
fruttasecca: Is this something that we plan to do in this PR or another one? If the latter feel free to resolve the conversation

nhaghighat: part of it can be done with this PR, let me know the image names and I will add them to the helm

services/orchest-ctl/app/utils.py
```diff
@@ -188,3 +189,16 @@ def retry_func(func, _retries=10, _sleep_duration=1, _wait_msg=None, **kwargs) -
...
+
+def create_required_directories() -> None:
+    for path in [
+        "/userdir/data",
+        "/userdir/jobs",
+        "/userdir/projects",
+        "/userdir/.orchest/env-builds",
+        "/userdir/.orchest/jupyter-builds-dir",
+        "/userdir/.orchest/user-configurations/jupyterlab",
+        "/userdir/.orchest/base-images-cache",
```
yannickperrenet, line 202: @fruttasecca How about moving these paths to the internal library instead? They are already used in a few places so would be best to store them there. (For some reason I didn't think of this when reviewing earlier, sorry.)

deploy/helm/charts/node-agent/templates/daemonset.yaml
yannickperrenet: > let me know the image names and I will add them to the helm

@fruttasecca :smiley:

services/orchest-ctl/app/utils.py
fruttasecca: I think it makes sense. I was actually on the verge of making them an internal config variable like `ENV_BUILDS_DIRECTORY=...` etc. but I didn't want the PR to get larger

deploy/helm/charts/node-agent/templates/daemonset.yaml
fruttasecca: What we want this `daemonset` to pull depends very much on us. Technically speaking all images that we need are pulled in one way or another (e.g. when creating a session), the reason we are pulling is to improve the first experience of a user.

Things that we could pull pre-emptively (https://hub.docker.com/u/orchest for reference):
- `orchest/jupyter-enterprise-gateway`
- `orchest/jupyter-server`
- `orchest/session-sidecar`

Things that we can't pull pre-emptively using this `daemonset`:
- base images, since those are pulled & cached by kaniko

To solve this last point we could make it so that those base images are pulled pre-emptively but that's for another PR imo (if it happens at all). I am a bit afraid of Orchest downloading a large amount of data and doing considerable compute while a user is trying it for the first time. Pre-emptively caching base images could work for our cloud offering tho.

services/node-agent/app/main.py
```diff
@@ -43,6 +43,13 @@
         default=3,
         type=int,
     )
+    parser.add_argument(
+        "--image-puller-thrediness",
+        dest="image_puller_thrediness",
+        nargs="?",
+        help="Specifies number of async pullers to be yse, default 3.",
+        default=3,
+    )
```
yannickperrenet, line 52: Should probably be `--image-puller-threadiness`. Also not sure what `yse` is.

services/node-agent/app/image_puller.py
```diff
@@ -56,17 +65,44 @@ def pull_image(self, image_name: str):
...
+        pull_task = await queue.get()
+
         if self.policy == Policy.IfNotPresent:
-            if self.image_exists(image_name):
-                return
-            self.logger.info(f"Image '{image_name}' is not found - attempting pull...")
+            if self.image_exists(pull_task.image_name):
+                self.logger.info(
+                    f"Image '{pull_task.image_name}' is not found - attempting pull..."
+                )
```
yannickperrenet, line 78: Don't think this is correct. I guess it should be:
```python
if self.image_exists(...):
    return
self.logger.info(...)
```

services/node-agent/app/image_puller.py
```diff
@@ -56,17 +65,44 @@ def pull_image(self, image_name: str):
...
 
-        self.logger.info(f"Pulling image '{image_name}'...")
-        if not self.download_image(image_name):
-            self.logger.warning(f"Image '{image_name}' was not downloaded!")
+        try:
+            self.logger.info(f"Pulling image '{pull_task.image_name}'...")
+            if not self.download_image(pull_task.image_name):
```
yannickperrenet, line 82: Given that `self.download_image(...)` is using the synchronous Docker client, the entire pulling will still be synchronous right?

services/node-agent/app/image_puller.py
```diff
@@ -118,23 +154,16 @@ async def run(self):
...
-                            self.logger.error(
-                                f"Attempt {i} to pull image '{image_name}' "
-                                f"failed with exception: {ex}"
-                            )
+                # Create a pull task, with image_name which will
+                # be consumed by image puller tasks.
+                queue.put_nowait(PullTaskMessage(1, image_name))
+
+            await queue.join()
+            await asyncio.sleep(self.interval)
```
yannickperrenet, line 169: Does this work? 

From what I can tell, once the initially created tasks are resolved, then no tasks are picking up subsequent additions to the queue.

fruttasecca:
Coming back to this, if we want this `daemonset` to pull environment images (which seems a requirement to speedup kernel starts), it will actually need to either query the `orchest-api` to know what images should be pulled or it should query the registry to see what's in there and then pull (I prefer the former, but first the versioned env images PR needs to be merged)

services/node-agent/app/image_puller.py
nhaghighat: It has been changed, now the tasks are non-ending tasks.

services/node-agent/app/image_puller.py
```diff
@@ -52,10 +46,27 @@ def __init__(
...
+            be consumed by puller tasks.
+
+        """
+        while True:
+            for image_name in self.images:
+                # Create a pull task, with image_name which will
+                # be consumed by image puller tasks.
+                await queue.put(image_name)
+
+                await asyncio.sleep(self.interval)
```
yannickperrenet, line 68: I guess it should instead be like:
```python
        while True:
            for image_name in self.images:
                # Create a pull task, with image_name which will
                # be consumed by image puller tasks.
                await queue.put(image_name)

            await asyncio.sleep(self.interval)
```

services/node-agent/app/image_puller.py
```diff
@@ -142,28 +140,29 @@ def download_image(self, image_name: str) -> bool:
...
-                # be consumed by image puller tasks.
-                queue.put_nowait(PullTaskMessage(1, image_name))
+        get_images_future = asyncio.ensure_future(self.get_image_names(queue))
+        pullers = [
+            asyncio.create_task(self.pull_image(queue)) for _ in range(self.threadiness)
+        ]
 
-            await queue.join()
-            await asyncio.sleep(self.interval)
+        await asyncio.gather(*pullers, get_images_future)
```
yannickperrenet, line 168: Might want to wrap this like this:
```python
try:
    ...  # existing code
finally:
    if self._aclient is not None:
        await self._aclient.close()
        self._aclient = None
```
to prevent from additional errors being raises that the client wasn't closed correctly (we had this issue in the old implementation until we explicitly closed the client).

services/node-agent/app/main.py
```diff
@@ -0,0 +1,66 @@
...
+        default=3,
+        type=int,
+    )
+    parser.add_argument(
+        "--image-puller-threadiness",
+        dest="image_puller_threadiness",
+        nargs="?",
+        help="Specifies number of async pullers, default 4.",
+        default=4,
+    )
```
yannickperrenet, line 53: Since everything runs in one thread anyways, we probably want the threadiness to be exactly equal to the number of `--image-puller-images`. So I think I prefer removing this argument and instead just pass `image_puller_threadiness=len(image_puller_images)` to `ImagePuller`. 

Otherwise we might (for no good reason) just be pulling 4 images at the same time (which could prevent all other images from being pulled due to retries).

yannickperrenet: Good point. I am also not sure what would happen if we start pulling a lot of images concurrently. Instead controlling that variable through `threadiness` seems like a good approach. Alright then, let's leave it as is!

deploy/helm/charts/node-agent/values.yaml
```diff
@@ -1,3 +1,9 @@
 images:
   pullPolicy: IfNotPresent
   name: orchest/node-agent:latest 
+
+imagePuller:
+  images:
+  - orchest/jupyter-enterprise-gateway:latest
+  - orchest/jupyter-server:latest
+  - orchest/session-sidecar:latest
```
yannickperrenet, line 9: `jupyter-server` can also be configured by the user. And for consistency best to get all the images from endpoints from the `orchest-api`.

Todo
* `orchest-api` needs to get endpoints for this (for environments we should already have this)

fruttasecca (approved):
lgtm

Issue #210, opened 2021-04-15
Endpoint slash ending consistency


ricklamers:
Currently endpoints are a mix of endpoints ending in slashes `/login/` and not ending in slashes `/pipelines`. Similarly the `orchest-api` includes slash for some paths, but not for others. As does the `orchest-webserver` Flask endpoints.

We should consistently `not` include a trailing slash to avoid future ambiguity.

ricklamers:
We should attempt a refactor of the project to see if we can easily migrate all endpoints.

yannickperrenet:
> We should consistently not include a trailing slash to avoid future ambiguity.

We need to write this down somewhere as it is easily forgotten when adding new endpoints.

yannickperrenet:
I went over all endpoints that we expose inside the application by searching for `route` and going over all files that contain a match. I used `route` because that is what Flask uses, i.e. `@app.route(...)`, and we use Flask for all services that expose endpoints. The files that contain endpoints with trailing slashes are:
* `services/orchest-webserver/app/app/views/orchest_api.py`
* `services/orchest-api/app/app/apis/namespace_jupyter_builds.py`

The refactor *should* not be that time consuming.

ricklamers:
You should take into account all the locations in the code base from which we're connecting to those endpoints (i.e. in the JS client).

yannickperrenet:
> You should take into account all the locations in the code base from which we're connecting to those endpoints (i.e. in the JS client).

Of course!

yannickperrenet:
Might be something to look at during #201.

PR #751, opened 2022-03-02
Feat/k8s gui update

fruttasecca:
## Description

Implements the GUI update functionality for the k8s version. Makes use of the existing `orchest-ctl` pod update functionality
by starting an update through that and then following it with a sidecar.

Requires ingress to be setup to be easily tested. I have tested this end to end with some temporary hacky code that I would rather not commit given the ingress PR is quite near.

Known issue: the Shutdown progress bar is not showing up, so the update logs in the browser will look like this:

```
<rest of the logs>
Update [############################--------] 80%
Update [#########################-----------] 70%
Update [#####################---------------] 60%
Update [##################------------------] 50%
Update [##############----------------------] 40%
Update [##########--------------------------] 30%
Update [#######-----------------------------] 20%
Update [###---------------------------------] 10%
Update [------------------------------------] 0%
Updating...
Shutdown successful.
Shutdown <missing progress bar here>
Shutting down...
```


yannickperrenet:
Every time I see the log output of the update I am totally confused that it is upside down. Why exactly are we doing it that way again? @ricklamers

services/nginx-proxy/orchest.conf
```diff
@@ -73,14 +73,14 @@ server {
...
+  location /update-sidecar {
 
     auth_request /auth;
 
     proxy_request_buffering off;
     proxy_read_timeout 1d;
 
-    set $update_server "update-server.default.svc.cluster.local";
+    set $update_server "update-sidecar.default.svc.cluster.local";
     proxy_pass http://$update_server;
```
yannickperrenet, line 82: Don't think it matters since we are no longer using this file, but `$update_server` should also be renamed to `$update_sidecar`.

services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -150,17 +150,16 @@ def _get_ongoing_status_changing_pod() -> Optional[k8s_client.V1Pod]:
...
-    priority when it comes to conflicts.
+    status changing command, which works as a priority when it comes to
+    conflicts.
 
     Returns:
-        None if no instance of the update-server or orchest-ctl running
-        with state changing commands is not running. That instance pod
-        otherwise.
+        None if no instance of orchest-ctl running with state changing
+        commands is not running. That instance pod otherwise.
```
yannickperrenet, line 160: Should probably be `None if no ... is running` instead of `None if no ... is not running`.

fruttasecca: Indeed :)

fruttasecca:
>Every time I see the log output of the update I am totally confused that it is upside down. Why exactly are we doing it that way again? @ricklamers

Agree

services/update-sidecar/app/app/views.py
```diff
@@ -0,0 +1,32 @@
...
+
+from flask import jsonify
+
+from config import CONFIG_CLASS
+
+
+def register_views(app):
+    @app.route("/update-sidecar/heartbeat", methods=["GET"])
+    def heartbeat():
+        return "", 200
```
yannickperrenet, line 12: Regarding these return types, we want to slowly migrate to more consistency i.e. always returning an object/dictionary (https://github.com/orchest/orchest/issues/391). As far as I remember the general `fetcher` in the frontend only expects dictionaries and not strings.

services/update-sidecar/app/app/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+            namespace=_config.ORCHEST_NAMESPACE,
+            follow=True,
+        ):
+            log_file.write(event)
+            log_file.flush()
+            logging.info(event)
+        logging.info("Update pod exited.")
+
+    with open(CONFIG_CLASS.UPDATE_COMPLETE_FILE, "w") as f:
+        f.write("true")
```
yannickperrenet, line 160: The content that is written to the `UPDATE_COMPLETE_FILE` can be arbitrary right? Might be good to add this as a comment or change the string itself.

services/update-sidecar/app/app/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def create_app():
+    app = Flask(__name__)
+    app.config.from_object(CONFIG_CLASS)
+
+    register_views(app)
+
+    executor.submit(follow_update)
+    return app
```
yannickperrenet, line 173: Submitting to the `executor` here means that we are following the update the moment the app is created (which is the moment the pod is started) right? I think it would be good explaining (maybe in the `main.py`) that the update-sidecar depends on the `UPDATE_POD_NAME`. 

Alternatively the following of the update (`follow_update`) can be done based on an API trigger. What made you choose this route? Simplicity?

services/nginx-proxy/orchest.conf
```diff
@@ -75,8 +75,6 @@ server {
   # Update server
   location /update-sidecar {
 
-    auth_request /auth;
-
```
yannickperrenet, line 79: I am guessing this needs to be updated in the appropriate `ingress.yml` files instead, but those were removed in a previous commit (`helm/charts/update-server/templates/ingress.yaml`). So probably needs to be checked that the token authentication works as expected.

deploy/orchest-ctl/pod.yml
```diff
@@ -1,5 +1,7 @@
 # Note: this yaml file is used is used in the "orchest" script to create
-# the orchest-ctl pod, and in "orchest-ctl" to create the update pod.
+# the orchest-ctl pod, in "orchest-ctl" to create the hidden-update pod
+# and in "orchest-webserver" to create the orchest-ctl update pod for
```
yannickperrenet, line 3: You mention adding the manifest to the `"orchest-webserver"` but you are adding them to the `"orchest-api"` (https://github.com/orchest/orchest/pull/751/commits/674090e9bc216fc7e44806f0b9da9ed5eb8b056e)

services/update-sidecar/app/app/__init__.py
fruttasecca: >Submitting to the executor here means that we are following the update the moment the app is created (which is the moment the pod is started) right?

Yes

>Alternatively the following of the update (follow_update) can be done based on an API trigger. What made you choose this route? Simplicity?

Yes, a few reasons:
- it's simpler than having an API call trigger this task as a side effect (imo)
- in a couple of edge cases the sidecar will cleanup/delete the update-pod
- starting it after the first request for logs has started will add a (albeit small) latency to the first logs arriving to the client

services/orchest-api/app/app/schema.py
```diff
@@ -770,3 +770,13 @@
...
+
+update_info = Model(
+    "UpdateInfo",
+    {
+        "token": fields.String(
+            required=True,
+            description="Token to access the update sidecar.",
+        ),
+    },
+)
```
yannickperrenet, line 782: Can we change the name to something like `UpdateSidecarInfo` as it shouldn't be confused with the current "update-info" that refers to our external endpoint containing the update information, e.g. "latest version is X".

services/nginx-proxy/orchest.conf
fruttasecca: Yeah we will need to do a bit of coordination/integration with the ingress PR.

deploy/orchest-ctl/pod.yml
fruttasecca: Indeed, will fix.

services/orchest-api/app/app/schema.py
fruttasecca: Sure :+1:

services/update-sidecar/app/app/views.py
```diff
@@ -1,18 +1,21 @@
...
-from flask import jsonify
+from flask import jsonify, request
 
-from config import CONFIG_CLASS
+from app.config import CONFIG_CLASS
 
 
 def register_views(app):
-    @app.route("/update-sidecar/heartbeat", methods=["GET"])
+    @app.route("/update-sidecar/heartbeat/", methods=["GET"])
```
yannickperrenet: Why did you add the trailing `/`? We are trying to have the consistency of no trailing slash in endpoints (https://github.com/orchest/orchest/issues/210 and mentioned in the [docs](https://docs.orchest.io/en/latest/development/implementation_best_practices.html#api-design-guidelines)).

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -0,0 +1,71 @@
...
+from _orchest.internals import config as _config
+from app import schema, utils
+from app.connections import k8s_core_api
+
+api = Namespace("ctl", description="Orchest-api internal control.")
+api = utils.register_schema(api)
+
+
+@api.route("/start-update")
+class IdleCheck(Resource):
```
yannickperrenet, line 17: Might want to rename this class.

services/update-sidecar/app/app/views.py
fruttasecca: Was a mistake, it has been removed in a later commit AFAIR.

services/orchest-api/app/app/apis/namespace_ctl.py
fruttasecca: woopsy :))

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -0,0 +1,71 @@
...
+    for env_var in orchest_ctl_container["env"]:
+        if env_var["name"] == "ORCHEST_VERSION":
+            env_var["value"] = update_to_version
+            break
+    orchest_ctl_container["command"] = ["/bin/bash", "-c"]
+    # Make sure the sidecar is online before updating.
+    orchest_ctl_container["args"] = [
+        "while true; do nc -zvw1 localhost 80 > /dev/null 2>&1 && orchest update "
+        "&& break; sleep 1; done"
+    ]
```
yannickperrenet, line 116: @nhaghighat is there a k8s-native approach to specify dependencies for containers in the same pod, something like startupProbes?

deploy/Makefile
```diff
@@ -68,7 +71,7 @@ ORCHEST_ENABLE_CONFIG := ${ORCHEST_API_ENABLE_CONFIG} \
 						 ${ORCHEST_WEBSERVER_ENABLE_CONFIG} \
 						 ${POSTGRES_ENABLE_CONFIG} \
 						 ${RABBITMQ_ENABLE_CONFIG} \
-						 ${UPDATE_SERVER_ENABLE_CONFIG}
+						 ${UPDATE_SIDECAR_ENABLE_CONFIG}
```
yannickperrenet, line 74: Wait, does the update-sidecar always run now?

And the `AUTH_SERVER_ENABLE_CONFIG` above needs to be `CELERY_WORKER_ENABLE_CONFIG` ([L49](https://github.com/orchest/orchest/pull/751/commits/6434e25d77520dc8d046c0fe4704db6d49483857#diff-ca41056a0340e0139d50c70288e712c894b470842211c94554a1a493ed0164f7R49) is miss-typed as well).

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -24,18 +25,66 @@ def post(self):
         token = secrets.token_hex(20)
         # K8S_TODO: query update-info endpoint once we use versioned
         # images.
-        manifest = _get_update_pod_manifest("latest", token)
+        manifest = _get_update_pod_manifest("latest")
         k8s_core_api.create_namespaced_pod(_config.ORCHEST_NAMESPACE, manifest)
+        manifest = _get_update_sidecar_manifest(
+            "latest", manifest["metadata"]["name"], token
+        )
+        k8s_core_api.create_namespaced_pod(_config.ORCHEST_NAMESPACE, manifest)
```
yannickperrenet: From these calls the update-pod and update-sidecar seem loosely coupled, whilst they are actually bidirectionally waiting on each other before executing (update-pod waits with the update and update-sidecar waits with following the logs). Why did you choose to spawn them in different pods instead of the same (the latter somewhat indicating the coupling)?

services/update-sidecar/app/app/views.py
yannickperrenet: Ghehe in a later commit this trailing slash was removed (https://github.com/orchest/orchest/pull/751/commits/cef5346bbcb2801c60558873d94b18566c78a42b#diff-09142d7e04b95ce0c5e2bb1e77381606989644d8fb8cde7ba4f6504e184a40c2R28-R33) again. :+1:

services/update-sidecar/app/app/__init__.py
```diff
@@ -8,27 +8,41 @@
...
     try:
         _follow_update()
     except Exception as e:
         logging.error(e)
         raise
+    finally:
+        # Give time to the webserver to communicate that the update is
+        # done.
+        time.sleep(30)
+        _delete_pod()
```
yannickperrenet, line 45: Not sure whether I have seen this, but does the `orchest-ctl` that is updating also delete itself (its pod) after it is done? Or how is that cleaned up? Time-based?

services/orchest-api/app/app/apis/namespace_ctl.py
fruttasecca: >update-pod waits with the update

The reason is that otherwise the update process might start before the sidecar is ready thus the user would not see any logs until it is. This could happen if, for some reason, the sidecar image is still being pulled.

>update-sidecar waits with following the logs

Not sure it's something we can avoid given we are talking about a sidecar, but I might have misunderstood what you mean.

>Why did you choose to spawn them in different pods instead of the same (the latter somewhat indicating the coupling)?

Because there is no strict necessity for it, other reasons:
- the sidecar is reachable without being authenticated, given that this works by pod (and not by container) it seemed better hygiene not to make `orchest-ctl` live in that pod
- the `update-pod` will delete itself once `orchest-ctl` exits, which would mean a possible race condition, the `update-sidecar` might still be fetching logs or returning them to the client when the termination is issued

services/update-sidecar/app/app/__init__.py
fruttasecca: The `update-sidecar` runs in its own pod, the waiting time is to make it so that the client has the time to fetch the logs.

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -25,12 +25,16 @@ def post(self):
...
-            "latest", manifest["metadata"]["name"], token
+        update_pod_manifest = _get_update_pod_manifest("latest")
+        sidecar_manifest = _get_update_sidecar_manifest(
+            "latest", update_pod_manifest["metadata"]["name"], token
+        )
+        # Create the sidecar first to avoid the risk of the update_pod
+        # shutting down the orchest-api before that happens.
+        k8s_core_api.create_namespaced_pod(_config.ORCHEST_NAMESPACE, sidecar_manifest)
+        k8s_core_api.create_namespaced_pod(
+            _config.ORCHEST_NAMESPACE, update_pod_manifest
```
yannickperrenet, line 36: Don't we still need a way to ensure that the endpoint returns? Although that is probably a very low probability edge case.

fruttasecca: Yes, the probability is quite low, we can make an asynchronous task for it.

services/update-sidecar/app/app/views.py
yannickperrenet: Indeed removed in a later commit.

services/orchest-api/app/app/apis/namespace_ctl.py
yannickperrenet: > Note that the container running the update and the update sidecar are not in the same pod anymore.

I noticed in a later commit :wink: 

All good then :+1:

deploy/Makefile
yannickperrenet: > No, this is just the deployment of a k8s service, to make the sidecar reachable.

I see, my bad. :+1:

services/orchest-api/app/app/apis/namespace_ctl.py
yannickperrenet: > Not against removing it for the purpose of lowering the chance of worse issues.

I am fine leaving it in, just wanted to know the thoughts behind the decision.

> Not sure it's something we can avoid given we are talking about a sidecar, but I might have misunderstood what you mean.

All good. It was just part of the story explaining the bidirectionality. 

> the sidecar is reachable without being authenticated, given that this works by pod (and not by container) it seemed better hygiene not to make orchest-ctl live in that pod

Makes sense!

> the update-sidecar might still be fetching logs or returning them to the client when the termination is issued

Makes sense as well!

> the update-pod will delete itself once orchest-ctl exits

I thought that the `ctl:<current-version>` kills the `ctl:<new-version>` (the one running the `hidden-update` and I think the one you call `update-pod`). Where does it say the `update-pod` kills itself?

fruttasecca: All `orchest-ctl` pods delete themselves, to have stronger guarantees `orchest-ctl` will attempt to delete the pod it creates (`hidden-update`, and the `orchest-api` performing cleanups) when exiting.

services/orchest-api/app/app/apis/namespace_ctl.py
yannickperrenet: Let's leave it as is :+1:

services/orchest-api/app/app/apis/namespace_ctl.py
fruttasecca: >the one running the hidden-update and I think the one you call update-pod

From the `orchest-api`POV the `update-pod` is the `orchest-ctl` pod running `update`. I am trying to make the update process as identical as possible between the CLI and the GUI update. The logs we are fetching are the logs of the `orchest-ctl` pod running `update`.

yannickperrenet: > All orchest-ctl pods delete themselves

Where exactly is this being done? I could not find this.

> From the orchest-apiPOV the update-pod is the orchest-ctl pod running update

I see. I got a little confused by the terminology.

services/update-sidecar/app/app/__init__.py
fruttasecca: >I understood that part, but I was wondering where the ctl:<new-version> is deleting itself  

It is, all `orchest-ctl` pods run the following:

```python
signal.signal(signal.SIGTERM, lambda *args, **kwargs: k8sw.delete_orchest_ctl_pods())
atexit.register(k8sw.delete_orchest_ctl_pods)
```

services/orchest-api/app/app/apis/namespace_ctl.py
fruttasecca: >I see. I got a little confused by the terminology.

That's partially my fault I think :p

```python
signal.signal(signal.SIGTERM, lambda *args, **kwargs: k8sw.delete_orchest_ctl_pods())
atexit.register(k8sw.delete_orchest_ctl_pods)
```
In `services/orchest-ctl/app/orchest/_core.py`

services/update-sidecar/app/app/__init__.py
yannickperrenet: I see it now, thank you! :+1:

services/orchest-api/app/app/apis/namespace_ctl.py
yannickperrenet: Answered in https://github.com/orchest/orchest/pull/751#discussion_r818698233 :+1:

services/nginx-proxy/orchest.conf
fruttasecca: @nhaghighat

ricklamers:
> Every time I see the log output of the update I am totally confused that it is upside down. Why exactly are we doing it that way again? @ricklamers

The main reason is that users would not have to scroll to “keep up with the output”. We can disable reverse output stream if we fix the height of the container to some fraction (e.g.) 50% of the viewport.

yannickperrenet (approved):
Great job! The update process was a biggy :100:

yannickperrenet:
> The main reason is that users would not have to scroll to “keep up with the output”. We can disable reverse output stream if we fix the height of the log container to some fraction (e.g.) 50% of the viewport.

I would like that. Ideally we mimic the same thing that happens when installing something through the terminal. I will create an issue out of this (no reason to include it in this PR).

nhaghighat (approved):

nhaghighat:
Looks good, thanks

PR #756, opened 2022-03-03
Feat/user service ingress rules

nhaghighat:
## Description

This PR creates nginx ingress for session services

It is still a little bit WIP.

## Checklist


fruttasecca:
As you have mentioned this is still WIP, just mentioning that ingress needs to be set up for user session services that need to be reachable from outside the cluster. Given that `ports` is now a required property of Orchest user services we will need to add a boolean property that expresses this reachability requirement, because pre-k8s we were using `ports` to infer that. Some FE changes are require in pipeline -> settings -> services.

Note that the URL through which services will be reachable is returned by the orchest python SDK, so we need to make sure that it is still in sync.

yannickperrenet (commented):
Like you've said, indeed still WIP. But so far so good!

Todo
* Add ingress for publicly exposed user services.

yannickperrenet (approved):
I haven't tested the actual PR but the changes look good to me :+1:

fruttasecca (approved):
same here :)

PR #766, opened 2022-03-07
Environment image versioning

fruttasecca:
## Description

Introduces environment images versioning into Orchest.

The following models have been introduced in the `orchest-api`:
- `Envirornment`
- `EnvironmentImage`

`EnvironmentBuild` has been renamed to `EnvironmentImageBuild`
`JupyterBuild` has been renamed to `JupyterImageBuild`

Environment images are now built and pushed with a given tag, which is an integer that increments on every build.

!:warning: This PR introduces breaking changes to the schema, i.e. you will lose your existing environment builds and locked images data (shouldn't matter given the k8s port, just mentioning)

## Checklist

p.s., sorry for the beefy PR, environment builds & images are cross-cutting and most new lines of code are due to schema migrations or the renaming of files.


lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -77,7 +77,7 @@
 # Networking
 ORCHEST_API_ADDRESS = "orchest-api"
 ORCHEST_SOCKETIO_SERVER_ADDRESS = "http://orchest-webserver"
-ORCHEST_SOCKETIO_ENV_BUILDING_NAMESPACE = "/environment_builds"
+ORCHEST_SOCKETIO_ENV_BUILDING_NAMESPACE = "/environment_image_builds"
```
yannickperrenet, line 80: Since you so consistently renamed, maybe rename the variable as well to `ORCHEST_SOCKETIO_ENV_IMG_BUILDING_NAMESPACE`?

fruttasecca: Makes sense

services/orchest-api/app/app/apis/__init__.py
```diff
@@ -26,6 +27,7 @@
 api.add_namespace(ns_ctl)
 api.add_namespace(ns_env_builds)
```
yannickperrenet, line 28: Rename this to `ns_env_img_builds`?

services/orchest-api/app/app/apis/namespace_environments.py
```diff
@@ -0,0 +1,152 @@
...
+@api.route("/<string:project_uuid>")
+@api.param("project_uuid", "uuid of the project")
+class ProjectEnvironmentList(Resource):
+    @api.doc("get_project_environments")
+    @api.marshal_with(schema.environments)
+    def get(self, project_uuid):
+        """Get all environments of a project."""
+
+        projects = models.Environment.query.filter_by(project_uuid=project_uuid).all()
+        return {"environments": [proj.__dict__ for proj in projects]}, 200
```
yannickperrenet, line 42: `projects` --> `environments`

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -102,7 +102,8 @@ def _get_image_build_workflow_manifest(
                         "args": [
                             f"--dockerfile={dockerfile_path}",
                             "--context=dir:///build-context",
-                            f"--destination={_config.REGISTRY_FQDN}/{image_name}:latest",  # noqa
+                            f"--destination={_config.REGISTRY_FQDN}/{image_name}:"
+                            f"{image_tag}",
```
yannickperrenet, line 106: Can you wrap it like so:
```python
(
  f"--destination..."
  f"{image_tag}"
),
```
I would be tempted (when looking at this code) to add a comma `,` at the end of the `--destination` arg.

services/orchest-api/app/app/apis/__init__.py
fruttasecca: Yep

services/orchest-api/app/app/apis/namespace_environments.py
```diff
@@ -0,0 +1,152 @@
...
+
+@api.route("/")
+class EnvironmentList(Resource):
+    @api.doc("get_environments")
+    @api.marshal_with(schema.environments)
+    def get(self):
+        """Get all environments."""
+
+        projects = models.Environment.query.all()
+        return {"environments": [proj.__dict__ for proj in projects]}, 200
```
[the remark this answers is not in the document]
fruttasecca: whoopsy, fixing

services/orchest-api/app/app/apis/namespace_environments.py
fruttasecca: whoopsy, fixing

services/orchest-api/app/app/core/environments.py
```diff
@@ -21,65 +23,154 @@ def get_env_uuids_missing_image(project_uuid: str, env_uuids: str) -> List[str]:
...
-    env_uuid_image_id_mappings = {}
+    env_uuid_to_image = {}
     for env_uuid in env_uuids:
         if env_uuid == "":
             raise self_errors.PipelineDefinitionNotValid("Undefined environment.")
 
-        # K8S_TODO: fix.
-        env_uuid_image_id_mappings[env_uuid] = _config.ENVIRONMENT_IMAGE_NAME.format(
-            project_uuid=project_uuid, environment_uuid=env_uuid
+        # Note: here we are assuming that the database olds the truth,
```
yannickperrenet, line 50: `olds` --> `holds`

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
```diff
@@ -264,11 +274,39 @@ def _transaction(self, build_request):
...
         # https://stackoverflow.com/questions/9034091/how-to-check-task-status-in-celery
         # task.forget()
+        (
+            models.Environment.query.with_for_update()
+            .filter_by(
+                project_uuid=build_request["project_uuid"],
+                uuid=build_request["environment_uuid"],
+            )
+            .one()
+        )
```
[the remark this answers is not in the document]
fruttasecca: So that there is no risk of race conditions when it comes to the latest `tag`, and no risk of race conditions between an environment being deleted and an ongoing env build request. There are different approaches to it (e.g., keep trying until the  DB doesn't return an error in the case of the incremental tag), this seemed the cleanest codewise and in terms of readability. We can move to more performance and fine grained locking down the road if we need it.

services/orchest-api/app/app/core/environments.py
```diff
@@ -21,65 +23,154 @@ def get_env_uuids_missing_image(project_uuid: str, env_uuids: str) -> List[str]:
...
+        # and that if the record is in the database then the image is in
+        # the registry.
+        env_image = (
+            models.EnvironmentImage.query.filter_by(
+                project_uuid=project_uuid,
+                environment_uuid=env_uuid,
+            )
+            .filter(models.EnvironmentImage.build.has(status="SUCCESS"))
```
yannickperrenet, line 58: Correct me if I am wrong in my understanding, but the new flow would be:
1. Users start an environment build
2. The `orchest-api` created the appropriate `EnvironmentImage` and `EnvironmentImageBuild` entries.

This means that while the `EnvironmentImageBuild` is ongoing, the `EnvironmentImage` can't actually be used as it does not exist yet (or won't even exist in case the build fails). Now to the question, wouldn't it be better to "lock" the `EnvironmentImage` on build instead of having to do the `filter` like above?

The `filter` seems like something that is easily forgotten in future changes leading to `EnvironmentImage`s being used even though they aren't actually build yet.

services/orchest-api/app/app/core/environments.py
```diff
@@ -21,65 +23,154 @@ def get_env_uuids_missing_image(project_uuid: str, env_uuids: str) -> List[str]:
...
+
+
+def _lock_environments(project_uuid: str, environment_uuids: Set[str]):
+    for env_uuid in environment_uuids:
+        models.EnvironmentImage()
+        (
+            models.Environment.query.with_for_update()
+            .filter_by(project_uuid=project_uuid, uuid=env_uuid)
+            .one()
+        )
```
yannickperrenet, line 81: What exactly does the `models.EnvironmentImage()` do?

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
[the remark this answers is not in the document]
yannickperrenet: And the `Environment` stays locked until the next `db.commit` right? (Which can also be automatically invoked when the request returns).

services/orchest-api/app/app/core/environments.py
```diff
@@ -21,65 +23,154 @@ def get_env_uuids_missing_image(project_uuid: str, env_uuids: str) -> List[str]:
...
+    run_image_mappings = []
+    for env_uuid, image in env_uuid_to_image.items():
+        run_image_mappings.append(
+            models.PipelineRunInUseImage(
+                run_uuid=run_id,
+                project_uuid=project_uuid,
+                environment_uuid=env_uuid,
+                environment_image_tag=image.tag,
+            )
+        )
```
yannickperrenet, line 103: Not sure whether I had seen it already, but do we also "unlock" the images after the pipeline run has finished?

yannickperrenet: Nvm, in `interactive_runs_using_environment` (from the next commit) we check whether the run has reached an end state.

services/orchest-api/app/app/core/environments.py
fruttasecca: That's a mistake, will fix.

services/orchest-api/app/app/core/environments.py
fruttasecca: No need to unlock in an active manner given that we query the DB and we filter for a run that must be `PENDING|STARTED`.

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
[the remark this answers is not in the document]
fruttasecca: >And the Environment stays locked until the next db.commit right? 

Yes

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -416,31 +416,7 @@ def build_image(
...
-        models.InteractiveSession.status.in_(["LAUNCHING", "RUNNING"]),
-    ).all()
-
-    jobs = models.Job.query.filter(
-        models.Job.image_mappings.any(docker_img_id=img_id),
-        models.Job.status.in_(["DRAFT", "PENDING", "STARTED", "PAUSED"]),
-    ).all()
-
-    return bool(int_runs) or bool(int_sessions) or bool(jobs)
+    return True
```
yannickperrenet, line 425: I believe this function is no longer used in the codebase, we can be removed.

yannickperrenet:
Apart from the above comments I am left with the following general questions:

1.   Usage and need of `Environment` model.
      
      Using the `Environment` model we can now, when working with `EnvironmentImage`s and `EnvironmentImageBuild`s, lock the `Environment` itself so that users can't delete the respective environment from the UI. Is that right?

2.   The `DELETE` endpoint in `namespace_environments` seems a little scary to me, since it actually aborts everything that uses the environment. This sounds rather devastating as a user would abort (beyond saving as the environment will be removed and I don't think the job would start again) a job when deleting an environment.

     Are we sure we want to support this? In other applications (although I often dislike this), e.g. Linear, you first need to delete all resources (in our case e.g. jobs) that use the resource (here environment) you want to delete.

services/orchest-api/app/app/core/environments.py
fruttasecca: >Correct me if I am wrong in my understanding

You are correct

>Now to the question, wouldn't it be better to "lock" the EnvironmentImage on build instead of having to do the filter like above?

If with "lock" you mean create an `EnvironmentImage` entry on image build success I am on board. It's something I avoided since this approach was a bit ismpler, but I can feel this is actually more desirable.

fruttasecca:
>Apart from the above comments I am left with the following general questions:

- 1 Yes, more or less. I would summarize it in the following way: Orchest has a concept of `Environment` and not mapping said concept to a db entry opens up for more issues or implicit assumptions that we can do away. Coarse locking on the environment has currently a good ROI when it comes to avoiding edge cases.

- 2 This endpoint was actually already there, but under a different name and namespace (because we didn't have an `Environment` resource and namespace at the `orchest-api` level). Before said deletion the client actually checks the `in-use` endpoint to prompt/warn the user.

services/orchest-api/app/app/core/image_utils.py
fruttasecca: Pretty sure I had removed it :thinking: , will do.

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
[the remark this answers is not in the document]
yannickperrenet: On that matter of locking the `Environment` to prevent race conditions. Isn't the edge case super small given that the `TwoPhaseExecutor` calls a `self.session.commit()` right after the `transaction`, meaning that the `Environment` wouldn't actually be locked during the `collateral` (which is the actual build). 

And so it would indeed be (like you said) during the "ongoing env build request" (literally the request, not the building)? That does sound right to me, very smart of you :nerd_face:

services/orchest-api/app/app/core/environments.py
fruttasecca: To add to this, keeping this data instead of deleting *InUseImage rows means we can add showing what env version a job/run was using as a feature down the road, and it will work with existing jobs

services/orchest-api/app/app/core/environments.py
yannickperrenet: > If with "lock" you mean create an EnvironmentImage entry on image build success I am on board.

Yes indeed. Feels like we otherwise might (unknowingly) introduce a bug when forgetting about the implicit relationship.

yannickperrenet:
Your answers to (1) and (2) makes sense to me :+1: 

@ricklamers @iannbing How do you feel about (2). Is this something we might want to revisit in the future?

yannickperrenet (approved):
After making the discussed changes, it is good to go! Great work once again :100:

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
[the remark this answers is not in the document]
fruttasecca: Indeed the locking only happens during the transaction. We are definitively at a very low risk of edge cases when it comes to this, the biggest benefit of this coarse locking is peace of mind/ease of reasoning imo :P.

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -16,6 +18,27 @@
 api = register_schema(api)
 
 
+@api.route("/<string:project_uuid>/<string:environment_uuid>/latest")
```
yannickperrenet, line 21: Up until this point I believe we have always chosen the format:

`@api.route("/latest/<string:project_uuid>/<string:environment_uuid>")`

instead of 

`@api.route("/<string:project_uuid>/<string:environment_uuid>/latest")`

yannickperrenet (approved):
LGTM!

PR #773, opened 2022-03-08
Remove memory-server related docs

yannickperrenet:
## Description

Removes all memory-server related docs. The Kubernetes MVP won't have a memory-server and so we want to remove all user-facing content that mention it.

Related PR: #769 

## Checklist
- [X] The PR branch is set up to merge into `feat/k8s` instead of `master`.


astrojuanlu:
There are still mentions to "the plasma store" in the SDK docs https://docs.orchest.io/en/latest/fundamentals/sdk/index.html?highlight=plasma#api-transfer should they be removed too?

Issue #743, opened 2022-02-25
Open project (pipelines view) after import


ricklamers:
**Describe the problem this improvement solves**
It's unclear whether import was successful if you have many projects and you don't see it listed immediately.

**Describe the solution you'd like**
Open project (pipelines view) after import

yannickperrenet:
> Open project (pipelines view) after import

We are already doing this. If it is not working then it is a bug.

yannickperrenet:
I just tested again on `dev` and after importing a project it is automatically opened (i.e. going to its pipelines view). This worked for both Orchest and non-Orchest owned projects.

ricklamers:
```projects?import_url=https://github.com/orchest-examples/orchest-pipeline-r-python-mix/```

I should have probably mentioned this happens when you're importing a project that's triggered through the query argument parameter `import_url` (e.g. via the Open in Orchest button)

iannbing:
Still couldn't reproduce this issue with the query argument in the Cloud version (i.e. `master` branch) or in `dev` branch locally.

iannbing:
I think I might have found the issue. When importing the project without giving a name, this behavior is indeed broken. Will fix.

PR #778, opened 2022-03-09
Fix not redirecting to Pipeline List after importing if the import URL contains trailing slash

iannbing:
## Description

The regular expression for the import_url should take into account the trailing slash.

Fixes: #743


yannickperrenet (approved):
The fix makes sense to me :+1:

Issue #729, opened 2022-02-17
Service configuration collapses when browser tab regains focus


astrojuanlu:
**Describe the bug**
I have the Pipeline settings -> Services of a Pipeline in one browser tab, and there is one service configured. If I expand the service configuration, leave the tab, and come again, the service configuration collapses and I have to expand it again.

**Expected behavior**
I expected the service configuration to not collapse unless I decide to do it myself.

**To Reproduce**

1. Go to Pipeline settings -> Services
2. Create a service, and save it
3. Keep the service configuration open, and change tab or window
4. Come back to tab, click on any blank area of the page, and see the service configuration collapse.

**Screenshots**
Here I demonstrate the issue with 2 tabs:

https://user-images.githubusercontent.com/316517/154517640-d8517fa3-03bd-41da-86df-e7696201b889.mp4

**Environment**

- OS: Linux Mint
- Browser: Firefox
- Orchest's version (in the settings page): `v2022.02.8`

iannbing:
I think we need more clarifications for this specific case.

Since Orchest allows multiple users managing the same instance at the same time, one of the principle for Orchest FE is to re-validate the UI when tab re-gain focus, in order to ensure (as much as possible) that user is working with the latest server-side state.

And because of this, the form will be refreshed and the service form in the table is reset to default. To solve this issue, we could persist the UI state of the form in the localstorage or a URL query args. @ricklamers I think query args would be a slightly better option. What do you think?

iannbing:
Or we could also introduce auto-save just like environment edit page.

yannickperrenet:
> introduce auto-save, so user doesn't lose input data when switching tabs

I really like this approach (relates to #426). No one expects to lose data by switching tabs, at least I don't :smile: 

> use query args to persist the form state to keep the form open when re-gain focus.

When it comes to this, I don't think it has that much priority but would be a nice quality of life improvement.

Issue #770, opened 2022-03-08
Auto-saving in pipeline settings to not happen after every typed character


yannickperrenet:
**Describe the problem this improvement solves**
When auto-saving after every character, then auto-formatting would lead to "strange" behavior. For example:

1. The pipeline parameters are `{}` (so go to the pipeline's settings)
2. Go into the dictionary like so: `{|}` (where `|` represents the cursor)
3. Press `<Enter>`
4. Expect to see:

    ```plain
    {
    }
    ```
    
    but actually see: `{}|`

(Thanks @fruttasecca.)
 
**Describe the solution you'd like**
Editing the pipeline parameters should behave as expected (for example editing the step parameters works fine).

iannbing:
To clarify, PipelineSettings doesn't have auto-saving yet. PipelineSettings was not properly persisting user's input in general. This issue is addressed in #781.

PR #781, opened 2022-03-09
Fix losing progress on Pipeline Settings when browser tab regaining focus

iannbing:
## Description

When editing Pipeline Settings, user loses their changes if they switch to other browser tabs and come back. This PR persists the changes as subsets of the whole Pipeline settings object. This PR also makes it possible to edit Pipeline Path in Pipeline Settings view. 

Fixes: #729, #770 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
+++ b/services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
@@ -7,7 +7,6 @@ export const ProjectsContext = React.createContext<IProjectsContext>(null);
 export const useProjectsContext = () => React.useContext(ProjectsContext);
 
 type Action =
-  | { type: "pipelineClear" }
   | {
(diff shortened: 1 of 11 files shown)


yannickperrenet:
> This PR also makes it possible to edit Pipeline Path in Pipeline Settings view.

I vaguely remember there being a reason why we did not support this, however, I can't remember what that reason is. Could you verify that changing the path (moving it into another folder for example) does not break the sdk in active kernels?

iannbing:
@yannickperrenet 

As far as I see from the UI logic, to change the pipeline path, user needs to stop session first. 

Screenshot 2022-03-10 at 11 41 06
 
Same restriction is applied. The Pipeline Path field is disabled if session is alive.

Screenshot 2022-03-10 at 11 42 59

yannickperrenet:
Awesome! All good then :smile:

PR #782, opened 2022-03-09
Fix highlighting navigation item "Jobs" when viewing the pipeline of a job run

iannbing:
## Description

Currently if user is viewing the pipeline of a job run, the navigation item "Jobs", instead of "Pipelines", shouldn't be highlighted. 

## Checklist


yannickperrenet (approved):
Great quick fix! Let's merge :)

Would be great to get fixes like these in before the k8s merge.

PR #783, opened 2022-03-11
Show Project Selector in deeper levels of the UI

iannbing:
## Description

Currently Project Selector only appears in `/pipelines`, `/environments`, and `/jobs`, but not in the views at a deeper level, such as `/job`. It's very confusing especially when user have jobs/pipelines with the same name across projects. 

To make it clearer which project user is working on, this PR make ProjectSelector visible as long as the current view is within one project.

Current UI logic: if user switches projects while viewing "Jobs", the view stays at Jobs, e.g. from Project A Jobs -> Project B Jobs. This PR follows the same rationale, see the following example:

Viewing a job in project A, switch to project B, viewing the jobs of project B.

## Checklist

### DIFF
--- a/services/base-images/runnable-shared/runner/requirements-dev.txt
+++ b/services/base-images/runnable-shared/runner/requirements-dev.txt
@@ -1,6 +1,6 @@
-nbformat
-nbconvert
-IPython
-ipykernel
+ipykernel==6.6.0
+ipython==7.31.1
+nbconvert==6.3.0
+nbformat==5.1.3
 pytest
 -e ../../../../lib/python/orchest-internals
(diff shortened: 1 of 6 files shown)


yannickperrenet:
Not sure why the Python tests are failing for you. 

> from ipython_genutils.text import indent
> E   ModuleNotFoundError: No module named 'ipython_genutils'

raised as an error (due to `nbconvert`) does not happen for me locally. Unsure whether this is related due to me running Python3.9 locally and the Actions running Python3.8. 

From https://github.com/ipython/ipython_genutils/tree/0.2.0:
> This package shouldn't exist. It contains some common utilities shared by Jupyter and IPython projects during The Big Split™. As soon as possible, those packages will remove their dependency on this, and this repo will go away.

I created a commit to pin dependencies, let's hope that fixed it :)

yannickperrenet (approved):
As discussed, including the project selector solves a real UX issue of not knowing what project you are currently working on.

The behavior of going to the closest root level view when switching projects makes sense. We agreed on later fixing the issue where swapping projects when looking at a job run (in read-only mode) will move to the others projects pipelines (instead of jobs). The UX issue this solves is simple bigger than the issue of this edge case.

PR #784, opened 2022-03-11
Add route /job-run

iannbing:
## Description

Currently `/pipeline` is used both for editing a pipeline and viewing a read-only pipeline of a job run. This causes quite a few bugs. For example, the root view while viewing a job run becomes `/pipelines`, where it should have been `/jobs`. And read-only state might be reset when user refreshes the page.

This PR creates a new route: `/job-run` and re-use the same component PipelineEditor. `/job-run` shows the pipeline always in the read-only mode, and its root view is `/jobs`. 

Edit: because this PR remove the unnecessary re-rendering. PipelineEditor in read-only mode has missing connection lines until first session polling response. Thus, this PR also fixes this issue.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/job-view/PipelineRunTable.tsx
+++ b/services/orchest-webserver/client/src/job-view/PipelineRunTable.tsx
@@ -160,7 +160,7 @@ export const PipelineRunTable: React.FC<{
     }
 
     navigateTo(
-      siteMap.pipeline.path,
+      siteMap.jobRun.path,
       {
         query: {
           projectUuid: pipelineRun.project_uuid,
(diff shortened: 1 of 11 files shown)


yannickperrenet:
I think it might be best to postpone this PR. All other views as part of jobs still select _Pipelines_.

#### Steps to reproduce
1. Create a job
2. Open a pipeline run
3. See that _Jobs_ is correctly selected
4. Open the settings of the pipeline
5. See that _Pipelines_ is incorrectly selected
6. Repeat this with _Logs_ and when viewing a step (_View file_ or double clicking it)

yannickperrenet (changes requested):
For this PR to get merged it needs to, in my opinion, fully address all edge cases of jobs incorrectly selecting _Pipelines_ in the left pane menu.

iannbing:
@yannickperrenet Said cases are covered in the new commit. the proposed solution is following the same pattern.
Pipeline settings of a job run => `/job-run/pipeline-settings`
Logs => `/job-run/logs`

iannbing:
> The edge case of "View file or double clicking it" is not fixed yet.

Should be fixed in the new commits. Thanks for catching it!

> Also, after refreshed the page (when looking at the pipeline in read-only mode of a job), the connections only appear after a few seconds. I thought that this issue should also have been fixed in this PR?

I made a mistake when doing cherry-picking and rebasing. Just restore the commit.

yannickperrenet (approved):
Great stuff! All works for me now :)

PR #785, opened 2022-03-11
Build images with buildkit

fruttasecca:
This PR replaces kaniko with buildkit for environment and jupyter-server image building. This is for a matter of performance, given that buildkit works with layers and can thus speed up builds through caching layers, pulling base image layers concurrently etc.

Known issues:
- logging is not as clean as kaniko, getting only the build logs is a bit more tedius and leads to overhead
- base image layers seem to be unaffected, leading to an overhead on build due to decompression


services/orchest-api/app/config.py
```diff
@@ -28,9 +28,13 @@ class Config:
...
     # Image building.
-    BUILD_IMAGE_LOG_TERMINATION_FLAG = "_ORCHEST_RESERVED_LOG_TERMINATION_FLAG_"
+    IMAGE_BUILDER_IMAGE = "moby/buildkit:v0.10.0"
+    BUILD_IMAGE_LOG_FLAG = "_ORCHEST_RESERVED_LOG_FLAG_"
     BUILD_IMAGE_ERROR_FLAG = "_ORCHEST_RESERVED_ERROR_FLAG_"
-    HOST_BASE_IMAGES_CACHE = f"/var/lib/orchest{_config.USERDIR_BASE_IMAGES_CACHE}"
+    HOST_KANIKO_BASE_IMAGES_CACHE = (
+        f"/var/lib/orchest{_config.USERDIR_KANIKO_BASE_IMAGES_CACHE}"
+    )
+    HOST_BUILDKIT_CACHE = f"/var/lib/orchest/{_config.USERDIR_BUILDKIT_CACHE}"
```
yannickperrenet, line 37: Didn't these paths change with the userdir-pvc change? Is the `userdir` still located at `/var/lib/orchest/userdir`?

Nonetheless, it needs to be (removing the `/` as the config value already contains a leading `/`):
```python
HOST_BUILDKIT_CACHE = f"/var/lib/orchest{_config.USERDIR_BUILDKIT_CACHE}"
```

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -57,9 +58,12 @@ class OrchestImagesToPrePull(Resource):
     @api.doc("orchest_images_to_pre_pull")
     def get(self):
         """Orchest images to pre pull on all nodes for a better UX."""
-        # K8S_TODO: remove latest once we have versioned images on
-        # dockerhub.
         pre_pull_orchest_images = [
+            # This image is only used in the builder node, pull it
+            # anyway through the daemonset? (it's around 50 MB).
+            CONFIG_CLASS.IMAGE_BUILDER_IMAGE,
```
yannickperrenet, line 64: Sure why not :)

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -9,8 +12,31 @@
 from app.connections import k8s_core_api, k8s_custom_obj_api
 from config import CONFIG_CLASS
 
+# So that pushing to the registry is faster and it always uses the same
+# cache on disk.
+_registry_pod_affinity = {
```
yannickperrenet, line 17: I guess we are talking about scheduling the builder pod on the same pod as the registry? If that is the case then I would like to propose to update the comment so that is a bit clearer.

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -312,37 +391,83 @@ def _build_image(
...
+    # Tells us if we can expect base image layers to be in the cache.
+    needs_to_pull_base_image = not os.path.exists(
+        f'{_config.USERDIR_BUILDKIT_CACHE}/{build_context["base_image"]}/.success.txt'
+    )
+    if needs_to_pull_base_image:
+        msg = "Pulling base image..."
+        _log_(user_logs_file_object, complete_logs_file_object, msg, False)
+
+    # Buildkit will add runtimes to commands, can't be deactivated atm.
+    runtime_regex = re.compile(r"#\d*\s\d*\.\d*\s")
```
yannickperrenet, line 403: I guess we can speed this up by prepending the pattern with `^` to indicate it needs to start with the pattern as otherwise the entire string will be searched (which doesn't seem to be necessary based on how `runtime_regex` is used, but I might very well be mistaken).

yannickperrenet (approved):
LGTM! I haven't tested it as I figured this will happen anyways once it is merged.

The supplied flags to the buildkit container also looked good to me, couldn't find any other flags we might want to add :+1:

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -195,81 +343,14 @@ def _get_image_build_workflow_manifest(
...
-                    ns,
-                    expected_statuses=["Succeeded", "Failed", "Unknown"],
-                    max_retries=1,
-                )
-            except errors.PodNeverReachedExpectedStatusError:
-                user_logs_file_object.write(".")
-            else:
-                user_logs_file_object.write("\n")
-                done = True
+def _log_(user_logs, all_logs, msg, newline=False):
```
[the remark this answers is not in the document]
fruttasecca: Didn't even realize that there was a trailing underscore :detective:  nice catch

services/orchest-api/app/app/core/image_utils.py
fruttasecca: Indeed, will fix

fruttasecca:
Coming back to this, I think we want to make it so that the directory that is being used as a cache is not part of the NFS, i.e. something only local to the node on which the builder is running.

yannickperrenet:
> something only local to the node on which the builder is running.

We start the builder on the same node as the registry right? But I guess we aren't ensured that the registry always starts on the same node, or are we? Leading to loss of cache.

> postpone such things once we work on the environment images lifecycle PR/feature?

I think that makes the most sense given the current timeline :+1:

services/orchest-api/app/config.py
fruttasecca: I have renamed the kaniko vpc to `image-builder-cache-vpc", buildkit will now make use of that.

nhaghighat (approved):

PR #790, opened 2022-03-14
k8s services spec changes

fruttasecca:
## Description

This PR introduces some requires changes to a pipeline `services`:
- `ports` are now required, they won't be used to implicitly infer that a service you be exposed outside of Orchest anymore. The role of this property is to define which ports the related k8s service should be reachable through.
- `entrypoint` has been renamed `command` (in line with k8s, also more clear imo)
- `command` has been renamed to `args` (in line with k8s, also more clear imo)
- `exposed` is a new and required property that tells us if a service should be reachable from outside the cluster

@nhaghighat I see that the `requires_authentication` property of a service is not used when setting up ingresses, do all services require authentication by default now? Is it something we can control when creating ingresses at run time?


docs/source/fundamentals/services.rst
```diff
@@ -105,9 +105,10 @@ Non required fields
 **command**
     Service start command, e.g. what process the service will run.
 
-**entrypoint**
-    `command` and `entrypoint` are a 1:1 mapping to Docker, refer to the `Docker docs
-    <https://docs.docker.com/engine/reference/builder/#cmd>`_ for their difference and gotchas.
+**args**
+   Arguments to pass to the command. For example, if the `command` is
+   equal to ``bash``, `args` could be ``-c "echo hello"``. This would
+   be equivalent to ``bash -c "hello"``.
```
yannickperrenet, line 111: I guess it should be `-c "hello"` instead of `-c "hello hello"`

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -924,43 +924,46 @@ def _get_user_service_deployment_service_manifest(
...
-                "pathType": "Prefix",
-            }
+    if service_config["exposed"]:
+        ingress_url = (
+            "service-"
+            + service_config["name"]
+            + "-"
+            + project_uuid.split("-")[0]
+            + "-"
+            + pipeline_uuid.split("-")[0]
```
yannickperrenet, line 934: Not sure whether this should be gated with `if service_config["exposed"]`. Even if the service isn't publicly exposed, the ingress rule is still needed afaik for other Orchest resources to reach it. E.g. code-server wouldn't be publicly (outside of the cluster) exposed but still needs an ingress rule for a user to actually interact with it.

fruttasecca: I am not sure about that, I think setting up an ingress might imply that the service is publicly exposed (for authentication we have `requires_authentication`). @nhaghighat ?

fruttasecca:
Btw, I guess we want `requires_authentication` to be true for basically all existing services templates that are exposed.

services/orchest-api/app/app/core/sessions/_manifests.py
yannickperrenet: I think you are right. Let's see what @nhaghighat has to say :)

services/orchest-webserver/app/app/compat.py
```diff
@@ -36,6 +36,24 @@ def migrate_pipeline(pipeline):
             tmp.add(valid_var)
         service["env_variables_inherit"] = list(tmp)
 
+        # Can't use a get() with a default value because it would imply
+        # a different thing. Migrate from pre-k8s to k8s.
```
yannickperrenet, line 40: Great catch that we need to migrate the spec. Inside the pipeline definition we do store a key (top-level) called `version`. Do you think it makes sense to use that instead? This way we can distinguish between multiple different upgrade paths.

Let's say someone has build an old project in Orchest and we have changed the spec multiple times, then ideally we still know how to migrate (similar to db migrations).

services/orchest-webserver/client/src/components/ServiceTemplatesDialog/content.tsx
```diff
@@ -61,9 +63,10 @@ export const templates: ServiceTemplates = {
       binds: {
         "/project-dir": "/home/coder/code-server",
       },
-      entrypoint: "bash",
-      command:
+      command: "bash",
+      args:
         "-c 'umask 002 && code-server --auth none --bind-addr 0.0.0.0:8080 /home/coder/code-server'",
```
yannickperrenet, line 68: Btw why do we let users take care of the `umask 002`, isn't this an implementation detail we should hide from them? (Not saying we need to change that in this PR)

yannickperrenet (approved):
Overall I think it looks good :+1:

services/orchest-webserver/client/src/components/ServiceTemplatesDialog/content.tsx
fruttasecca: It's indeed an implementation detail, ideally something that we want to take care of, perhaps it's something that can be fixed through the k8s resource manifest?

services/orchest-webserver/app/app/compat.py
fruttasecca: I think a real pipeline schema migration system makes sense (although we rarely change the pipeline specs). The k8s mvp could be a good moment to introduce that.

yannickperrenet: > although we rarely change the pipeline specs

True. 

> The k8s mvp could be a good moment to introduce that.

I agree. Does that mean you are making the changes here?

services/orchest-webserver/client/src/components/ServiceTemplatesDialog/content.tsx
yannickperrenet: Let's make an issue out of it to explore after the mvp.

services/orchest-webserver/app/app/compat.py
fruttasecca: Yeah I'll do those there. A first working version might just be a few lines of code in the `compat.py` module tbh.

services/orchest-webserver/app/app/compat.py
```diff
@@ -45,21 +75,39 @@ def migrate_pipeline(pipeline):
...
-    return pipeline
+# version: (migration function, version to which it's migrated)
+_version_to_migration_function = {
+    "1.0.0": (_migrate_1_0_0, "1.1.0"),
+    "1.1.0": (_migrate_1_1_0, "1.2.0"),
+}
+
+# Make sure no forward version is repeated.
+__to_versions = [item[1] for item in _version_to_migration_function.items()]
+assert len(_version_to_migration_function) == len(__to_versions)
```
yannickperrenet, line 96: Don't think this is correct. You probably want to be using a `set` instead of a list.

fruttasecca: You are absolutely right :)).

nhaghighat:
> @nhaghighat I see that the requires_authentication property of a service is not used when setting up ingresses, do all services require authentication by default now? Is it something we can control when creating ingresses at run time?

I did not add authentication for user-server ingresses. I'll create a small PR for that

services/orchest-api/app/app/core/sessions/_manifests.py
nhaghighat: >  I think setting up an ingress might imply that the service is publicly exposed (for authentication we have requires_authentication)

That is correct, but the authentication is not checked for user services. I'll add that

iannbing (approved):

nhaghighat (approved):

PR #792, opened 2022-03-15
CICD pushes images with version equal to build_tag

fruttasecca:
Leaving the PR here to remind us to update the internal `update-info` repo action.

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -18,8 +18,7 @@ jobs:
     env:
       DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
       DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-      # BUILD_TAG: ${{ github.event.client_payload.build_tag }}
-      BUILD_TAG: "latest"
+      BUILD_TAG: ${{ github.event.client_payload.build_tag }}
 
     strategy:
       matrix:


.github/workflows/docker-ci.yml
```diff
@@ -18,8 +18,7 @@ jobs:
     env:
       DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
       DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-      # BUILD_TAG: ${{ github.event.client_payload.build_tag }}
-      BUILD_TAG: "latest"
+      BUILD_TAG: ${{ github.event.client_payload.build_tag }}
```
yannickperrenet, line 21: Does this value get correctly set on [`release`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) as well? Sounds like it only would on the [`repository_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch). 

https://github.com/orchest/orchest/blob/3cf12a3b1af940e51a996220aee40e69c51ebcd6/.github/workflows/do-release.yml#L26

fruttasecca: Good catch, will take a look later

ricklamers:
LGTM although I find it hard to judge whether you're missing anything

Issue #612, opened 2021-12-23
Add file explorer to the pipeline editor


ricklamers:
**Is your feature request related to a problem? Please describe.**  
Conceptually pipelines in Orchest are pointers to files that get executed. Those files all live in the project. (Which can be seen as a regular git versioned repository.)

It's not always clear that you can simply "drag a file onto the pipeline editor", conceptually speaking. To make this conceptual model more explicit we need to bring the project files closer to the pipeline editor.

**Describe the solution you'd like**
One idea we're currently considering is adding a file explorer directly to the pipeline editor.

The idea is that you can directly drag files from the file explorer onto the pipeline canvas. Or when selecting a file for a pipeline step (in the properties pane), to be able to directly navigate the file project directory in a regular file explorer.

In addition, as a user it's more clear what your "project directory" is without having to look for that context in the JupyterLab editor view.

Furthermore, when inspecting job pipeline runs, currently only the files assigned to pipeline steps can be inspected. Often, when e.g. using library functions in something like a `utils.py` file you want to be able to inspect all the files that were used during the run that are part of the project. If the file explorer is part of the read-only pipeline view it becomes clearer that this run used a specific snapshot of the project files (the snapshot at the time of job creation) (!).

**Describe alternatives you've considered**  
- Pop-up/overlay file explorer in pipeline editor

###Additional context
**Mockup**
![file-explorer-mockup](https://user-images.githubusercontent.com/1309307/147223009-46e9eb95-50d7-4e27-ae09-0dc330687815.png)

PR #795, opened 2022-03-15
[Feature] File Manager

iannbing:
## Description

Add File Manager in Pipeline Editor as the default way of managing files in Orchest.

Fixes: #612 

## Checklist


astrojuanlu:
Looks gorgeous!
Screenshot 2022-03-15 at 18-57-19 Pipeline · Orchest

iannbing:
Figured that it's quite critical to enable drag a file to existing step. Converted this PR to draft again until this part is finished.

iannbing:
Some known minor issues:
- when drag-n-drop from file system, File Manager border highlighting only appears very briefly
- when drag a file onto an existing step, the step should be highlighted

yannickperrenet:
#### Ideas
* Can we make it possible to increase the size of the file manager? Just like we do for the step properties pane. Long names now get cut off.

* Use code annotation when moving files, e.g. `get-data.py` instead of 'get-data.py'. It is already done when trying to move notebooks to the `/data` directory, but not for other "moves".

* Throw an error when users tries to delete the pipeline definition file of the pipeline that is currently opened.

* Completely new idea, not for this PR as it requires removal of the pipelines view (as we have discussed in the past). The idea would be to drag a pipeline (`.orchest` extension) file to the pipeline editor canvas and it draws the pipeline, like we currently do automatically when opening a pipeline.

#### Bugs

Use the same Notebook for multiple steps.

1. Open the quickstart
2. Drag an existing Notebook, like `pre-processing.ipynb` to the `get-data.py` step
3. Get an error
4. Click confirm
5. See that the file assigned to the step has still changed.

Note that this too happens when changing the file path through the step properties pane.

---

Dragging a non-executable file to a step

1. Open the quickstart
2. Drag the `README.md` file onto a step
3. Open the step properties
4. See `Warning: this file wans't found in the project directory.`

So it seems like the `File path` in the properties filters for files with certain extensions. This needs to be consistent with the file-manager. I guess it is reasonable to through an error when dragging an unsupported file to the pipeline editor that it throws an error (similar to how you can't reuse the same notebook).

---

Renaming a file that is in use by the pipeline

1. Open the quickstart
2. Rename `explore-data.ipynb` to `explore.ipynb`
3. See in the file-manager that the file is renamed
4. See in the pipeline that the step still has `explore-data.ipynb` assigned. In addition, when opening the step properties it thinks the `File path` actually exists.

Would be great if renaming a file also changes it in the pipeline definition and the pipeline shows the updated name of the file (for all steps it is assigned to).

However, things are complicated by the fact that JupyterLab could currently have the file open. Going back to JupyterLab, after renaming the file, will say the file no longer exists.

A similar case can be made when deleting a file that is in use in the pipeline (and in JupyterLab).

---

Right click context menu `Duplicate` does not work

1. Right click on a step
2. Press `Duplicate`
3. See that the step is not actually duplicated

---

Creating a Notebook file asks for kernel language

1. Create a new file
2. Get the dialog
3. Enter a `File name` and change the `Kernel language` to `R`
4. Press `+ Create file`
5. Drag the file onto the canvas
6. See that (in the step's properties) the `Kernel language` of the step is set to `Python`

I think we might as well just remove the `Kernel language` option when creating a file.

#### General thoughts

To make the experience top notch we have to start thinking about the interaction between JupyterLab (its kernels and its file-manager) and our new (custom) file-manager. Now that we have our own file-manager we can do additional things (like mentioned above) where renaming a file would also rename usage of it in the pipeline definition (something we couldn't do when renaming through JupyterLab).

Not exactly sure how that would look like, especially given that users could user other editors as well (e.g. code-server) that also have a file manager.

The thing that makes it difficult is that Orchest encapsulates other editors, instead of being e.g. an extension to VSCode or JupyterLab. I am sure we can come up with solutions to give users a great editing experience though! Just keeping in mind what we want to prioritize (@ricklamers)

yannickperrenet:
To add to the above comment:

#### Ideas

* When adding a folder through the file-manager it should be similar to renaming a file in my opinion. The current prompt is not pretty to look at ;)

yannickperrenet:
I just noticed it, but the file-manager as part of pipeline runs of jobs should not be able to change the filesystem.

The pipeline-editor is in read-only mode and so should the file-manager be. The same applies when the pipeline-editor is in read-only mode when the environments are still building (and other possible cases that I currently can't come up with).

iannbing:
> Use code annotation when moving files, e.g. get-data.py instead of 'get-data.py'. It is already done when trying to move notebooks to the /data directory, but not for other "moves".

Fixed.

> Use the same Notebook for multiple steps.

Added a prevention when assign a file to a step via drag-n-drop. However, for the old "typing" behavior, we could only show a warning because every keystroke results in an auto-save. The previous entry has been lost.

> Dragging a non-executable file to a step

Added a prevention with a proper warning message.

> Renaming a file that is in use by the pipeline

Indeed, we can apply the new file path to the current pipeline. Will do. But as you already mentioned, the mapping cannot be perfectly in sync:
- move/rename a file that is used in multiple pipelines.
- move/rename a file in JupytorLab

We could, however, notify user about the steps with missing files (either with an icon / different border color), but then I would prefer to do this with a proper visual design.

> Creating a Notebook file asks for kernel language

When creating a file, the file is not yet assigned to any step. Currently Orchest doesn't attach any metadata to a Notebook file. Selecting a language was meant to assign the corresponding file template. User still need to assign the right kernel language when editing a step. Do you suggest default to Python, and let user change it afterwards in JupyterLab?

> When adding a folder through the file-manager it should be similar to renaming a file

I missed this one. Will do.

yannickperrenet:
> Renaming a file that is in use by the pipeline

Let's make that a feature for after the MVP. Can you create an issue?

> Do you suggest default to Python, and let user change it afterwards in JupyterLab?

When creating a Notebook file we can just default to Python, since the value is not used when assigning the file to a step (so why ask a user for it in the first place). As part of step creation a user then selects it (just the way it currently works pre-filemanager). JupyterLab would break if we don't set the metadata at all.

iannbing:
@yannickperrenet

> Renaming a file that is in use by the pipeline

 issue created #803 

> Creating a Notebook file asks for kernel language

The field is removed. Hard-coded to "python" in the request.

> When adding a folder through the file-manager it should be similar to renaming a file

fixed.

I also fixed a bug: user should not be allowed to move a folder that contains `.ipynb` files to `/data`. If user attempts to do so, the whole batch of operation will be blocked. User needs to make sure that all the files/folders they drag don't contain any Notebook files.

ricklamers:
https://www.tella.tv/video/cl0w7z8ij000209kw4reufk8g/view

Found another issue, relative paths seem to go well when dropping a file onto a step but not when a file is dropped onto the canvas initially.

ricklamers:
Double clicking on a .orchest file should open that pipeline. What do you think?

ricklamers:
This is looking great! Fantastic work @iannbing

ricklamers:
@iannbing what's the status on making the file manager resizable like the properties pane?

iannbing:
> Found another issue, relative paths seem to go well when dropping a file onto a step but not when a file is dropped onto the canvas initially.

Nice catch! Fixed.

iannbing:
> Double clicking on a .orchest file should open that pipeline. What do you think?

Implemented 😁

ricklamers:
> @iannbing what's the status on making the file manager resizable like the properties pane?

I think this is relevant in particular because the current sizing behavior is a bit odd. When you upload a file with a longer name the size becomes a bit bigger. Instead of dynamic sizing I think we should make the size fixed and user-controllable like with the properties pane.

yannickperrenet:
> Double clicking on a .orchest file should open that pipeline. What do you think?

Super curious about this, because from a user standpoint it can also be confusing. Having the session for pipeline A open, then double click `B.orchest` and see that pipeline. Doesn't really make sense to me. If the file is opened in JupyterLab instead, then sure.

At least that is what quickly came to my mind when reading the comment

ricklamers:
Implemented the above.

iannbing:
> @iannbing what's the status on making the file manager resizable like the properties pane?

Added 👍

yannickperrenet:
How about also removing our current file-manager? @iannbing @ricklamers

ricklamers:
> How about also removing our current file-manager? @iannbing @ricklamers[](https://github.com/iannbing)

If you look closely that's gone (at least from the UI)

iannbing:
@yannickperrenet @ricklamers  I also removed `file-manager` service and all the build scripts, workflow, etc. Let me know if I missed anything.

yannickperrenet:
Final comments:

* Absolute great work! :100:  I love the new UX and being able to drag and drop files :))
* The file-manager size can be increased, but I would also love to decrease its size a little more than it currently allows me (maybe until all action icons, e.g. add folder, just barely fit on the screen)
* When dragging a `.ipynb` file from the project files into `/data` I get two warnings. I think the latter can be removed, it feels a bit overkill since I already confirmed on the first.

Total nitpick (incorrect path is shown when pressing `View` in the right-click context menu of files in `/data` when looking at the pipeline run of a job)
1. Open the pipeline run of a job
2. Press `View` in the right-click context menu on a step that is in `/data`
3. See that the path is, e.g. `/data:/get-data.py` instead of `/data/get-data.py`

iannbing:
@yannickperrenet Your feedback items are all addressed. You could check the latest progress of this PR.
Regarding the "nitpick":

> Total nitpick (incorrect path is shown when pressing View in the right-click context menu of files in /data when looking at the pipeline run of a job)

I couldn't reproduce it with my latest changes. It's most likely a job run that you tested earlier while there were still bugs. Could you try again from with a fresh pipeline?

iannbing:
@yannickperrenet  I think I found the issue. Good find! I'll fix it.

services/orchest-webserver/app/app/utils.py
```diff
@@ -830,6 +830,27 @@ def is_valid_project_relative_path(project_uuid, path: str) -> str:
...
 
 
+def resolve_absolute_path(abs_path):
+    """
+    This function resolves absolute paths of pipeline steps
+    to the filesystem layout of the orchest-webserver.
+
+    Currently only /data is supported.
+    """
+    prefix_map = {"/data": "/userdir/data"}
```
yannickperrenet, line 840: Should use the value `_config.USERDIR_DATA`.

services/orchest-webserver/app/app/utils.py
```diff
@@ -830,6 +830,27 @@ def is_valid_project_relative_path(project_uuid, path: str) -> str:
...
+    allowed_prefixes = list(prefix_map.keys())
+    matching_prefixes = [prefix for prefix in allowed_prefixes if abs_path.startswith(prefix)]
```
yannickperrenet: Can be rewritten to

```python
matching_prefixes = [prefix for prefix in prefix_map if abs_path.startswith(prefix)]
```

services/orchest-webserver/app/app/utils.py
```diff
@@ -830,6 +830,27 @@ def is_valid_project_relative_path(project_uuid, path: str) -> str:
...
+    
+    if len(matching_prefixes) == 0:
+        return
+
+    prefix = matching_prefixes[0]
```
yannickperrenet, line 849: Why do we take the first match always (I understand we currently only have one entry in the `prefix_map` but this could then break on the next change to this function)? If we expect there to always be one match, then I think it should be reflected in the implementation.

PR #797, opened 2022-03-15
set the tag names of deployment with env vars

nhaghighat:
## Description

This PR configures the helm chart tag names with envirement variables.
The default environment variable is `latest`. to change all orchest developed services (orchest-api, orchest-webserver, celery-worker, auth-server, update-sidecar and file-manager,node-agent) the tag can be changed by specifying `ORCHEST-DEFAULT_TAG`

orchest-api tag can be changed by following env var:
`ORCHEST_API_TAG`
orchest-webserver tag can be changed by following env var:
`ORCHEST-WEBSERVER_TAG`
orchest-webserver tag can be changed by following env var:
`ORCHEST-WEBSERVER_TAG`
celery-worker tag can be changed by following env var:
`CELERY_WORKER_TAG`
auth-server tag can be changed by following env var:
`AUTH_SERVER_TAG`
update-sidecar tag can be changed by following env var:
`UPDATE_SIDECAR_TAG`
file-manager tag can be changed by following env var:
`FILE_MANAGER_TAG`
postgres tag can be changed by following env var (note: this won't be effected by : `ORCHEST-DEFAULT_TAG`
`POSTGRES_TAG`
rabbitmq tag can be changed by following env var (note: this won't be effected by : `ORCHEST-DEFAULT_TAG`
`RABBITMQ_TAG`

## Checklist


deploy/Makefile
```diff
@@ -132,46 +131,82 @@ NFS_RSC_DEPLOY_CONFIG := nfs-resources \
...
-RABBITMQ_ENABLE_CONFIG := --set rabbitmq.enabled=true
+ORCHEST_WEBSERVER_CONFIG := ${ORCHEST_WEBSERVER_TAG_CONFIG} \
+							${ORCHEST_WEBSERVER_ENABLE_CONFIG}
 
 # update-sidecar service configs
+UPDATE_SIDECAR_TAG := ${if ${UPDATE_SIDECAR_TAG},${UPDATE_SIDECAR_TAG},${ORCHEST_DEFAULT_TAG}}
+UPDATE_SIDECAR_TAG_CONFIG := --set update-sidecar.images.tag=${UPDATE_SIDECAR_TAG}
 UPDATE_SIDECAR_ENABLE_CONFIG := --set update-sidecar.enabled=true
+UPDATE_SIDECAR_CONFIG := ${UPDATE_SIDECAR_TAG_CONFIG} \
+						 ${UPDATE_SIDECAR_ENABLE_CONFIG}
```
fruttasecca, line 173: Is this needed? The `update-sidecar` pod is created at runtime by the `orchest-api`, only a k8s service is deployed through the charts (i.e. no deployments).

fruttasecca (approved):
lgtm aside from that minor comment

deploy/Makefile
```diff
@@ -132,46 +131,79 @@ NFS_RSC_DEPLOY_CONFIG := nfs-resources \
...
-						 ${AUTH_SERVER_ENABLE_CONFIG} \
-						 ${CELERY_WORKER_ENABLE_CONFIG} \
-						 ${FILE_MANAGER_ENABLE_CONFIG} \
-						 ${ORCHEST_WEBSERVER_ENABLE_CONFIG} \
-						 ${POSTGRES_ENABLE_CONFIG} \
-						 ${RABBITMQ_ENABLE_CONFIG} \
-						 ${NODE_AGENT_ENABLE_CONFIG} \
-						 ${UPDATE_SIDECAR_ENABLE_CONFIG}
+ORCHEST_ENABLE_CONFIG := ${} \
+						 ${} \
```
yannickperrenet, line 193: Can't this just be removed now?

deploy/Makefile
```diff
@@ -5,7 +5,7 @@ ROOK_CEPH_NAMESPACE := ${if ${ROOK_CEPH_NAMESPACE},${ROOK_CEPH_NAMESPACE},rook-c
 
 ROOK_NFS_NAMESPACE := ${if ${ROOK_NFS_NAMESPACE},${ROOK_NFS_NAMESPACE},rook-nfs}
 
-DEFAULT_TAG := ${if ${DEFAULT_TAG},${DEFAULT_TAG},latest}
+ORCHEST_DEFAULT_TAG := ${if ${ORCHEST_DEFAULT_TAG},${ORCHEST_DEFAULT_TAG},latest}
```
yannickperrenet, line 8: From a consistency perspective I found it a bit (but only very slightly) confusing that the `ORCHEST_DEFAULT_TAG` is defined here in the `Makefile` but all deployments also have the `tag` set in their `values.yaml` file.

For the `postgres` and `rabbitmq` it is only set in their `values.yaml` file. 

So Orchest deployments and third party deployments have a slightly different configuration route.

yannickperrenet (approved):
Quite the Makefile it has become ^_^

But looks good!

deploy/Makefile
nhaghighat: The postgres and rabbitmq also have their own tag variable, but they are not changed with `ORCHEST_DEFAULT_TAG`, that's why I changed the `DEFAULT_TAG` to `ORCHEST_DEFAULT_TAG`. The default tag is supposed to change all orchest developed services, but each component has it's own tag. for example if we want to try out a different version of orchest-api than others we should try:
`ORCHEST_DEFAULT_TAG=version1 ORCHEST_API_TAG=version2 make orchest`

Then all the orchest developed services will have version1 tag except orchest-api which has version2 tag.

yannickperrenet: But for `orchest-api` and other deployments it would never use the value set in its `values.yaml` file right? Because if `ORCHEST_API_TAG` is not given then it will fall back to `ORCHEST_DEFAULT_TAG`.

However, for `postgres` if `POSTGRES_TAG` is not set then no tag will be set and it uses the tag from its `values.yaml` file.

So that is inconsistent. Or am I wrong with the above description? (Total nitpick though, but it would mean that we have unused code in certain `values.yaml` files)

nhaghighat: Yes, you are right, but it would have made a little bit more complex which I tried to avoid.
for example for orchest-api:
if ORCHEST_API_TAG is present use it, if not, see if ORCHEST_DEFAULT_TAG is present or not, if yes use it, if not fallback to values.yml file.

yannickperrenet: Don't really have a strong opinion here. I will leave it up to you to decide whether you want to make changes or leave it as is :+1:

nhaghighat: I would say let's go with one, we will make it more complex later

PR #799, opened 2022-03-16
Feat/k8s use versioned images

fruttasecca:
## Description

Makes Orchest use versioned images instead of `:latest` as a default. Note: currently not tested since the `orchest-ctl` yaml file is broken, will do so shortly.


fruttasecca:
@yannickperrenet base images need to point to the correct version as well, see https://github.com/orchest/orchest/pull/799/commits/7b3037996d3531a34fd0fbf12a7f0383c1069f46, let me know if your question was answered

PR #801, opened 2022-03-16
FE changes to cope with k8s

iannbing:
## Description

- Hide/disable GPU options for the time being.
- add `Exposed` column to the Service Table 
- only render exposed services in the Services Menu in Pipeline Editor

### DIFF
--- a/services/orchest-webserver/client/src/components/ServiceForm.tsx
+++ b/services/orchest-webserver/client/src/components/ServiceForm.tsx
@@ -384,18 +384,28 @@ const ServiceForm: React.FC<{
                   </FormGroup>
                 </Stack>
                 <Stack direction="column" flex={1}>
-                  <FormSectionTitle title="Require authentication for the exposed service endpoints.">
-                    Authentication required
+                  <FormSectionTitle title="Exposed service endpoints">
+                    Exposed service endpoints
                   </FormSectionTitle>
                   <FormGroup>
+                    <FormControlLabel
+                      label="Exposed"
+                      disabled={props.disabled || hasServiceNameError}
+                      control={
+                        <Checkbox
+                          checked={props.service.exposed}
+                          onChange={(e) => {
+                            handleServiceChange("exposed", e.target.checked);
+                          }}
+                        />
+                      }
+                    />
                     <FormControlLabel
                       label="Authentication required"
                       disabled={props.disabled || hasServiceNameError}
                       control={
                         <Checkbox
-                          checked={
-                            props.service.requires_authentication !== false
-                          }
+                          checked={props.service.requires_authentication}
                           onChange={(e) => {
                             handleServiceChange(
                               "requires_authentication",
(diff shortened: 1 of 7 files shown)


fruttasecca:
I like the fact that it's possible to see if a service is exposed or not through the table. Would it be possible to make the checkbox in the column read-only and have the actual checkbox live near the "Authentication Required" checkbox? This way we can also have a hint near "exposed" down there. 

 Screenshoots for reference
image


Here is where I would put the exposed option (with the authentication required part on the right and exposed on the left)
image

@yannickperrenet keen to know your opinion

fruttasecca:
About the GPU entry being disabled, a nice to have would be to have the image grayed out or something similar
image

yannickperrenet:
### Important

Not sure whether `dev` was already pulled into `feat/k8s`, but this commit https://github.com/orchest/orchest/commit/fbf0a67b9667da0cef61d8e6cf0527c79d490d5e needs to be reverted. We don't want it to be shown anymore.

### `exposed` and `authentication required` options

> Here is where I would put the exposed option (with the authentication required part on the right and exposed on the left)

I think this indeed makes sense.

> I like the fact that it's possible to see if a service is exposed or not through the table. Would it be possible to make the **checkbox in the column read-only**

I like that as well (showing the most important config options without requiring the user to go into another view). I do think that that the checkbox (in the column) should indeed be read-only. Toggling doesn't make sense to me from a user perspective (and I guess from a backend perspective the session would have to be restarted @fruttasecca)

 > I would probably also move Authentication required to one of the table columns.

Sure, why not.

> I guess user might want to toggle a service on/off from time to time, without accidentally change any of the configurations.

Not sure exactly what you mean with this, if you mean disabling a service (so deleting it, without deleting the configuration) then this is currently not supported as the entry for the service would have to be removed from the pipeline definition, yet still stored somewhere. For now out of scope imo.

---

In short, adding both `exposed` and `authentication required` as read-only options to the table sounds good to me. If hovering them can show the tooltip, then that would be great (but not a requirement imo, since you can't toggle them from that view anyways).

iannbing:
@fruttasecca @yannickperrenet I've addressed your feedback. 

In the services table, it shows the status.

Screenshot 2022-03-18 at 08 58 33

In the form, it looks like this

Screenshot 2022-03-18 at 08 58 13

In the Environment Edit View, the GPU option is grayscaled.

Screenshot 2022-03-18 at 09 00 12

Please test it again on your machine to see if anything's still missing.

fruttasecca:
>Toggling doesn't make sense to me from a user perspective (and I guess from a backend perspective the session would have to be restarted

Yes it would have to be restarted

fruttasecca (approved):
Great work, lgtm!

The only thing I would change is the hint of `Exposed service endpoints`. Perhaps it can be more explicative. Maybe @yannickperrenet or @ricklamers would like to pitch in when it comes to the specific content, so I would say that we can merge this so that we can move on with testing and QA'ing, the content of the hint can be committed directly to `feat/k8s` imo. I'll make a note about the hint text.

yannickperrenet:
I do think that that in the `Authentication required` column we need to have an `X` instead of it being empty. Can be done directly on `feat/k8s` @iannbing

iannbing:
@yannickperrenet `X` implies an error. I'm not sure if it's sending the right message. I think `✔` and empty column explains better imo.

yannickperrenet:
> X implies an error. I'm not sure if it's sending the right message. I think ✔ and empty column explains better imo.

@iannbing Let's go for "yes/no" then (or "Yes/No"). An empty column looks strange and doesn't convey any information.

ricklamers:
> @iannbing Let's go for "yes/no" then (or "Yes/No"). An empty column looks strange and doesn't convey any information.

Changed it to this for now, something to potentially revisit in the future since it's now a bit text heavy. Clearer though :)

PR #807, opened 2022-03-17
Feat/enable nfs by default

nhaghighat:
## Description

This PR uses NFS by default

## Checklist

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -15,18 +15,58 @@ else
   ORCHEST_RESOURCES_TARGET := orchest-resources
 endif
 
+DEFINED_STORAGES := 0
+
+# If More than one storage is defined print hout an error and exit.
 ifeq (${ENABLE_ROOK_NFS},TRUE)
-	ifeq (${ENABLE_ROOK_CEPH},TRUE)
-		$(error Both CEPH and NFS can not be enabled at the same time)
-	endif
+	DEFINED_STORAGES := $(shell echo $$(( ${DEFINED_STORAGES} + 1 )))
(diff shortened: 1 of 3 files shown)


deploy/Makefile
```diff
@@ -15,6 +15,19 @@ else
...
+	endif
+endif
+
+ifeq (${ENABLE_ROOK_NFS},)
+	ifeq (${ENABLE_ROOK_CEPH},)
+		# Storage provider is not defined, ROOK_NFS will be used by default
+		ENABLE_ROOK_NFS := TRUE
+	endif
+endif
+
```
fruttasecca, line 30: Just making sure, is it still possible to opt for a hostPath storage provider>

fruttasecca:
Assuming this is part of the `orchest-resources` target, how would users discover such options as of now? I think it currently doesn't matter that much but I would like to either avoid using env vars passing as arguments (it's something that we haven't done until now) or to take a conscious decision to do that, including either docs or help messages etc. Not against it btw.

nhaghighat:
> Assuming this is part of the orchest-resources target, how would users discover such options as of now? I think it currently doesn't matter that much but I would like to either avoid using env vars passing as arguments (it's something that we haven't done until now) or to take a conscious decision to do that, including either docs or help messages etc. Not against it btw, and I don't think there is anything to change about this at the moment.

This is indeed part of orchest-resources target. I really did not expect to makefile grow this far, for future we should drop it and use a sophisticated tool.

fruttasecca (approved):
lgtm :)

PR #811, opened 2022-03-17
Feat/k8s celery parallelism

fruttasecca:
## Description

Allows to control the level of parallelism for the celery workers (job runs and interactive runs)


services/orchest-ctl/app/orchest/_k8s_wrapper.py
```diff
@@ -521,3 +521,13 @@ def get_host_names() -> List[str]:
...
+
+
+def sync_celery_parallelism_from_config() -> None:
+    k8s_apps_api.patch_namespaced_deployment(
+        "celery-worker",
+        "orchest",
+        _get_deployment_container_env_var_patch(
+            "celery-worker", utils.get_celery_parallelism_level_from_config()
+        ),
+    )
```
yannickperrenet, line 533: Would this cause a restart of the Pod? Given the name `sync...` I would expect that it can also be used to actively sync config changes on a running application (although it seems that is currently only used when starting the app.)

yannickperrenet (approved):
:rabbit:

nhaghighat (approved):

services/orchest-ctl/app/orchest/_core.py
```diff
@@ -61,10 +61,10 @@ def _run_helm_with_progress_bar(
...
     # K8S_TODO: remove DISABLE_ROOK?
     env = os.environ.copy()
-    if injected_env_vars is not None:
-        # Put the update before the rest so that these env vars cannot
-        # be overwritten by coincidence.
-        env.update(injected_env_vars)
+    # Put the update before the rest so that these env vars cannot be
+    # overwritten by coincidence.
+    for k, v in injected_env_vars.items():
+        env[k] = str(v)
```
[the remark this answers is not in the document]
fruttasecca: Correct, will fix

services/orchest-ctl/app/orchest/_k8s_wrapper.py
fruttasecca: >Would this cause a restart of the Pod

Yes. So technically speaking we could use this to set the parallelism level on the fly, we could even do that through the `orchest-api`, but it would lead to an uncontrolled restart of the worker.

PR #814, opened 2022-03-18
Add orchest uninstall command

fruttasecca:
Adds a minimal `orchest uninstall` command.

### DIFF
--- a/services/orchest-ctl/app/orchest/_core.py
+++ b/services/orchest-ctl/app/orchest/_core.py
@@ -683,3 +683,19 @@ def _update() -> None:
     k8sw.set_orchest_cluster_version(orchest_version)
 
     utils.echo("Update was successful, Orchest is running.")
+
+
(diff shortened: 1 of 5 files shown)


services/orchest-ctl/app/orchest/_core.py
```diff
@@ -683,3 +683,19 @@ def _update() -> None:
     k8sw.set_orchest_cluster_version(orchest_version)
 
     utils.echo("Update was successful, Orchest is running.")
+
+
+def uninstall():
+    k8sw.abort_if_unsafe()
+    utils.echo("Uninstalling Orchest...", nl=False)
+    k8s_core_api.delete_namespace("orchest")
```
yannickperrenet, line 691: I guess we are still leaving things like the "nginx ingress controller" and the "cert-manager" on the cluster? Removing them could break other applications the users have running on the same cluster.

PR #817, opened 2022-03-18
Redirect to `/login` if not authenticated

yannickperrenet:
## Description

If a user is not authenticated then they should automatically be redirected to `/login` (which hosts the login screen). 

https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#external-authentication


nhaghighat:
looks good

nhaghighat (approved):

PR #822, opened 2022-03-19
add clean targets

nhaghighat:
## Description

This PR adds clean targets to make to allow cleaner uninstall

## Checklist


deploy/Makefile
```diff
@@ -347,13 +399,41 @@ nfs-resources: ./bin/nfs-resources
...
+
+# Dummy clean target
+.PHONY: celan-
+clean-:
+	@echo "noting here fellas"
+
+# clean-all is meant to bring the cluster to its initial state, but currently it 
+# does not delete cert-manager and registry certificates, because the second 
+# cert-manager complains about unknown authority and can not create new certificates.
+.PHONY: celan-all
```
yannickperrenet, line 428: Probably a typo: `celan-all` --> `clean-all`

deploy/Makefile
```diff
@@ -347,13 +399,41 @@ nfs-resources: ./bin/nfs-resources
...
+
+.PHONY: clean-orchest-resources
+clean-orchest-resources:
+	helm uninstall orchest-resources --namespace ${NAMESPACE} --wait
+	rm ./bin/orchest-resources
+
+# Dummy clean target
+.PHONY: celan-
+clean-:
+	@echo "noting here fellas"
```
yannickperrenet, line 423: I guess this can be removed.

deploy/Makefile
```diff
@@ -278,6 +291,11 @@ jupyter-eg: ./bin/jupyter-eg
 	helm install ${JUPYTER_EG_DEPLOY_CONFIG}
 	touch $@
 
+.PHONY: clean-jupyter-eg
+clean-jupyter-eg:
+	helm uninstall jupyter-eg --namespace ${NAMESPACE} --wait
+	rm ./bin/jupyter-eg
```
yannickperrenet, line 297: I don't think we deploy the jupyter-EG using helm right? So I guess the targets `clean-jupyter-eg` and `jupyter-eg` can be removed (just like the `deploy/thirdparty/enterprise-gateway` folder).

fruttasecca: Indeed

deploy/Makefile
```diff
@@ -269,6 +279,9 @@ node-agent:
 	${SHARED_DEPLOY_CONFIG} \
 	${NODE_AGENT_ENABLE_CONFIG} helm
 
+.PHONY: clean-node-agent
+clean-node-agent:
+	helm uninstall node-agent --namespace ${NAMESPACE} --wait
```
fruttasecca, line 281: Isn't this cleaned already in `clean-orchest`? Or is the node-agent deployed separately?

deploy/Makefile
nhaghighat: No, this is here intentionally. once we deploy orchest with host-path enaled, the env var `STORAGE_CTL_TARGET` will be empty, and `make clean-all`, depends on this dummy clean target

deploy/Makefile
nhaghighat: the node-agent is usually deployed with orchest. but for debug purposes it has its own target (eventually all services might need a target for debug), so I also created a clean target just for that

fruttasecca: I see, alright

fruttasecca (approved):
lgtm

PR #826, opened 2022-03-21
remove upgrade targets and unify them with install, rm bin and interm…

nhaghighat:
…ediary bin files.

## Description

This PR unifies upgrade and install targets and removes the bin directory and intermediary files as it seems this is one of the sources of problems,

## Checklist

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -254,7 +254,7 @@ DOCKER_REGISTRY_DEPLOY_CONFIG := registry \
 ###############################################
 .PHONY: orchest
 orchest: initialize | ${ORCHEST_RESOURCES_TARGET}
-	helm install ${ORCHEST_DEPLOY_CONFIG}
+	helm upgrade --install ${ORCHEST_DEPLOY_CONFIG}
 
 .PHONY: clean-orchest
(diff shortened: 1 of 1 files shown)


deploy/Makefile
```diff
@@ -411,25 +394,8 @@ clean-all: clean-${REGISTRY_TARGET}
...
-orchest-upgrade:
-	helm upgrade ${ORCHEST_DEPLOY_CONFIG}
-
-.PHONY: argo-upgrade
-argo-upgrade:
-	helm upgrade ${ARGO_WORKFLOW_DEPLOY_CONFIG}
-
-.PHONY: registry-upgrade
-registry-upgrade:
-	helm upgrade ${DOCKER_REGISTRY_DEPLOY_CONFIG}
```
fruttasecca, line 435: Pretty sure this would break the existing upgrade calls in `orchest-ctl`. In `services/orchest-ctl/app/orchest/_core.py`, line  58.

nhaghighat: the logic of _core.py can be simplified, for the sake of time, I will bring the upgrade target back and we will simplify the _core.py later.

PR #828, opened 2022-03-22
add pvc to docker-registry

nhaghighat:
## Description

This PR adds pvc creation to docker-registry. new env variable is added to Makefile to configure the registry storage class

Fixes: #issue

## Checklist


deploy/Makefile
```diff
@@ -87,14 +86,24 @@ REGISTRY_TARGET := ${if ${DISABLE_REGISTRY},,docker-registry}
 
 SHARED_DEPLOY_CONFIG := ${KUBECONFIG} ${DEBUG} --create-namespace --atomic --timeout 600s
 
+# docker-registry deployment configs
 REGISTRY_SECRET := ${if ${REGISTRY_SECRET}, ${REGISTRY_SECRET},registry-tls-secret}
 REGISTRY_NAME := ${if ${REGISTRY_NAME}, ${REGISTRY_NAME},docker-registry}
 REGISTRY_PORT := ${if ${REGISTRY_PORT}, ${REGISTRY_PORT},443}
-
-# docker-registry deployment configs
+REGISTRY_STORAGE_CLASS := ${if ${REGISTRY_STORAGE_CLASS}, ${REGISTRY_STORAGE_CLASS},standard}
```
fruttasecca, line 93: What are the other options that we can specify? `ceph-filesystem` and `rook-nfs-storage-class`?

fruttasecca:
I guess that with this change we now want to bubble up the registry storage class as a CLI flag for `orchest install`, although we need to make sure that whatever was passed to `orchest-resources` as an env variable needs to be in sync or at least compatible with what is passed to the registry. AFAIK using ceph or nfs for the registry storage if `orchest-resources` wasn't installed with any of those two will not work since the daemons backing them do not exist, correct me if I am wrong.

Essentially we would be looking at always accepting `standard` as a registry storage class but only accepting the other two if those are exactly the one that are being used in the cluster already (e.g. the `userdir-pvc` storage class)

| orchest-resource storage class  | registry storage class | compatible |
| ------------- | ------------- | ------------- |
| standard  | standard  | yes 
| standard  | ceph  | no 
| standard  | nfs  | no
| nfs  | standard  | yes
| nfs  | nfs  | yes
| nfs  | ceph  | no
| ceph  | standard  | yes
| ceph  | nfs  | no
| ceph  | ceph | no

nhaghighat:
> we need to make sure that whatever was passed to orchest-resources as an env variable needs to be in sync or at least compatible with what is passed to the registry

I'm not sure about this. the orchest storage-class can be NFS while the registry storage-class is EBS or ceph-rbd.

>  AFAIK using ceph or nfs for the registry storage if orchest-resources wasn't installed with any of those two will not work since the daemons backing them do not exist, correct me if I am wrong.

The actual ceph-cluster and nfsservers are created by ceph-resources or nfs-resources targets. but If they get enabled if orchest-resources are enabled. do you think we should enable them?

deploy/Makefile
fruttasecca: I see. I think that we can actually look up the list of storage classes to verify the validity of the requested storage class, pretty neat.

fruttasecca:
>I'm not sure about this. the orchest storage-class can be NFS while the registry storage-class is EBS or ceph-rbd.

I see, I guess that what really matters here is if the storage class exists in the cluster, and if that exists we can take for granted that it will work?

nhaghighat:
> I see, I guess that what really matters here is if the storage class exists in the cluster, and if that exists we can take for granted that it will work?

That is true.

fruttasecca:
>The actual ceph-cluster and nfsservers are created by ceph-resources or nfs-resources targets. but If they get enabled if orchest-resources are enabled. do you think we should enable them?

You mean by default? Don't think so no. But we need to provide a good default for:
- orchest installed with standard storage -> it's single node, go with standard storage 
- orchest installed with ceph or nfs -> We could leave it to the user to decide what are its preferences but in a multi-node scenario standard storage will not act nice w.r.t. restarts because the registry might be scheduled on a different node. So we need to make sure it's always scheduled on the same node. Or let users shoot themselves in the foot.

PR #832, opened 2022-03-24
Add development mode (internal use)

fruttasecca:
Adds back `--dev`  mode to Orchest, and provides a number of scripts to make the developer life easier when it comes to testing changes, redeploying images, etc.

`--dev` provides hot reloading for the `orchest-api`, `orchest-webserver` and `auth-server`. To do that, the Orchest repository is mounted into their respective pods, which requires the repository to be mounted in all minikube nodes. To do that, you can start minikube with this command (while being in the repo directory):

`minikube start --memory 16000 --cpus 6 --mount-string="$(pwd):/orchest-dev-repo" --mount`

Once that is done, `--dev` mode can be used as pre k8s, i.e. by starting Orchest with `bash orchest start --dev` and by running the client server for FE files with ` pnpm run dev`.

I have updated the development docs so you can refer to those as well, to summarize the provided scripts:

```bash
# Redeploy a service after building the image using the repo code.
# This is the script that you will likely use the most.
bash scripts/redeploy_orchest_service_on_minikube.sh orchest-api

# Remove an image from minikube. Can be useful to force a pull from a registry.
bash scripts/remove_image_from_minikube.sh orchest/orchest-api

# Build an image with a given tag, on all nodes.
bash scripts/build_image_in_minikube.sh orchest-api v2022.03.7

# Run arbitrary commands on all nodes.
bash scripts/run_in_minikube.sh echo "hello"
```

The first and the third command assume the repository to be mounted in minikube.

@iannbing @ricklamers @astrojuanlu tagging everyone along since these scripts might be of interest


astrojuanlu:
Rendered docs: https://orchest--832.org.readthedocs.build/en/832/development/development_workflow.html

astrojuanlu:
Looks like a bug in minikube https://github.com/kubernetes/minikube/issues/12842

fruttasecca:
@astrojuanlu note that despite the memory limit warning/error you should be able to get minikube with the mounted repo running. The `--memory ...` flag is not really necessary, just an example. The issue is actually here:
>❌  Exiting due to GUEST_MOUNT_CONFLICT: Sorry, docker does not allow mounts to be changed after container creation (previous mount: '', new mount: '/home/juanlu/Projects/Orchest/orchest:/orchest-dev-repo)'

Essentially you have to delete the cluster and create it again by starting minikube with that mount flags. Minikube has other commands that allow mounting post creationg (like `minikube mount $(pwd):/orchest-dev-repo`) but I found them to be a bit unreliable/buggy.

astrojuanlu:
I confirm `minikube start --cpus 2 --mount-string="$(pwd):/orchest-dev-repo" --mount` works 👍🏽

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,112 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
+    # Make use of the in-node docker engine.
+    eval $(minikube -p minikube docker-env)
 
-   # By default the scripts builds all containers in parallel. To learn
-   # more about other options, such as building without cache, check out
-   # the first lines of the script first.
-   head -45 scripts/build_container.sh
+    # Build the desired image.
+    scripts/build_container.sh -i orchest-api \
+        -t <currently deployed orchest version> \
```
astrojuanlu, line 111: Could we clarify here what exactly the user needs to pass? It's not clear to me either.

If it depends on the latest released version, we could perhaps use a [Sphinx substitution](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?highlight=replacement#substitutions) or similar? Otherwise, if it's something custom we could add an example, or if it depends on a previous step, clarify it.

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,112 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
+        -o <currently deployed orchest version>
 
+    # Kill the pods of the orchest-api, so that the new image gets used
+    # when new pods are deployed.
+    kubectl delete pods -n orchest -l "app.kubernetes.io/name=orchest-api"
```
astrojuanlu, line 116: By the time I execute this, I get a message `No resources found`. Aren't we missing `bash orchest install` here?

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,112 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
+when it comes to setting the right tag, label, etc. For this reason, we provide the
+following scripts:
 
 .. code-block:: bash
 
-   cd docs
-   make html
+    # Redeploy a service after building the image using the repo code.
+    # This is the script that you will likely use the most.
+    bash scripts/redeploy_orchest_service_on_minikube.sh orchest-api
```
astrojuanlu, line 126: Getting errors here, unsure if related with the ones above:

```
$ bash scripts/redeploy_orchest_service_on_minikube.sh orchest-api
Error from server (NotFound): namespaces "orchest" not found
Building orchest-api:.
Essentially a wrapper around the build_container.sh script to build
an image on all minikube nodes, so that a pod will use that regardless
on which node it lands on. Requires the orchest repository to be mounted
in all nodes at /orchest-dev-repo.
Usage: build_image_in_minikube.sh <name of orchest service> <desired tag>
build_image_in_minikube.sh orchest-api v2022.03.8. 
Killing all pods belonging to orchest-api, through label app.kubernetes.io/name=orchest-api.
juanlu at valinor in ~/Projects/Orchest/orchest on feat/k8s-dev-mode
$ bash scripts/redeploy_orchest_service_on_minikube.sh orchest-api "v2022.03.7.dev"
This script will build the image of the given service across all
minikube nodes, then kill all the pods related to that orchest service
so that new pods using the newly built image will be created. Requires
the orchest repository to be mounted in all nodes at
/orchest-dev-repo.
Usage: redeploy_orchest_service_on_minikube.sh <name of orchest service>
redeploy_orchest_service_on_minikube.sh orchest-api
```

yannickperrenet:
I still have to look at the PR, but I think it would be a good idea to automatically disable telemetry when running using `--dev`. No need to re-enable it again after imo, the rule can just be: run once with `--dev` --> automatically disable telemetry until manual re-enable again.

fruttasecca:
Might be neat to allow cherry-picking the service that should be run in `--dev` mode, e.g `--dev-orchest-api`, this particular case would allow dev mode without the need to run the FE client process

docs/source/development/development_workflow.rst
fruttasecca: It's the tag of the image used by the deployment. Will clarify!

docs/source/development/development_workflow.rst
fruttasecca: Yes it requires Orchest to be installed first. I believe it's mentioned somewhere at the top of the document (will check). Do you think it needs to be stated again?

docs/source/development/development_workflow.rst
fruttasecca: Guess it's because Orchest was not installed.

docs/source/development/development_workflow.rst
astrojuanlu: OK, it's there:

> Currently, the development scripts/tools assume that you are running Orchest in minikube. If you have managed to successfully install Orchest in minikube now it’s the time to setup your development environment.

Maybe a link to the install instructions that is more easily visible?

astrojuanlu: Also, it's a bit confusing because `bash orchest install` installs the latest released version, no matter on which branch you're on. Am I right?

```
juanlu at valinor in ~/Projects/Orchest/orchest on feat/k8s-dev-mode
$ bash orchest install
...
Installing Orchest v2022.03.8
```

so, for me it looks like I'm mixing Orchest v2022.03.8 with the current git checkout. Related to https://github.com/orchest/orchest/pull/832#discussion_r838328409 ?

docs/source/development/development_workflow.rst
astrojuanlu: By the way, is the `kubectl delete pods -n orchest -l "app.kubernetes.io/name=orchest-api"` step above even necessary? The `bash scripts/redeploy_orchest_service_on_minikube.sh orchest-api` seems to be doing it anyway

docs/source/development/development_workflow.rst
astrojuanlu: Would it work with custom tags? For example:

```
$ scripts/build_container.sh -i orchest-api -t "v2022.03.8.dev" -o "v2022.03.8.dev"
```

astrojuanlu (changes requested):
Left a few comments about the documentation

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,112 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
 
-.. tip::
-   👉 If you didn't follow the :ref:`prerequisites <development prerequisites>`, then make sure
-   you've installed the needed requirements to builds the docs:
+    # Remove an image from minikube. Can be useful to force a pull from
+    # a registry.
+    bash scripts/remove_image_from_minikube.sh orchest/orchest-api
```
astrojuanlu, line 130: Tried this, got:

```
$ bash scripts/remove_image_from_minikube.sh orchest/orchest-api
Untagged: orchest/orchest-api:v2022.03.8.dev
Deleted: sha256:ccf8058b6e5d84dbf183a85909898dd077e1859a21605f4a3501d7d1d0d3d3ee
Deleted: sha256:9ee03c1712fcc194bf2114607c869e94544676b1004d753e70dfa3ce6d42c44f
Deleted: sha256:3d8e1b3a09090c72d07f2a0c2ab80357a3d63341132d639be43f7cd433496165
Deleted: sha256:a3cdc3bbd356b0f3c84c0f91d69ed6ddd9515dbd9f45a92aaac9fbdaf83aaaa3
Deleted: sha256:56355977357c57d64bcee9e949abc2934faf09ba8f636d1d466b96b1762ad449
Deleted: sha256:2eb2cb035c1c835c21a7d499784577e33c05fb03308a05568abdcb5b8477aba1
Deleted: sha256:468405a99d9e4dda1842b87582b745d61a3d3fc762dc0262b8ffa89430e55c6c
Deleted: sha256:39c5fc345b87755c9b737514b70a41ec3b357953f588944f3bfe430ff5ea0cb4
Deleted: sha256:56d6c8bbc5f46fe024ed9791f4ddbd46b2847ecbdc0a93f93b232d115ebcfd3b
Deleted: sha256:419f8b60d80c790e3c9fa07e342fad8f00f7bfd90b3f5fd2118dfc7bb29b5eb8
Deleted: sha256:f15b357abb2f73e88decd9fb01fe375831a1b1c423b8080433d24a64c7d29de4
Deleted: sha256:a44441e907852eb0ceb74b51283658e73d4fa02615eda51f17b3e75f781774bf
Deleted: sha256:9dcd2ed54571105c5cc1bea90753dfe23ff5947ded49bdfa8888226879330210
Deleted: sha256:d9c1aff3e5673434966396c32bbcd3965e8e7d945495d0cb8e0bcc8db046bd16
Deleted: sha256:f49cc9ac53ce139fd9df960dff8f319628465ac18a439519d629a4fb71c40529
Deleted: sha256:9de820bf6824b032486d6bcd7f418b3dd36b9c384946c28e06a80a9d3b4e1291
Deleted: sha256:50c0310829a17f83ab93b1408ee14f841cf2d939b270b7831387bb82e82c1880
Deleted: sha256:bd2c1d465fe7daa90a2dc26f805c66865c76ca4801161386b28c8e3364a9dd52
Deleted: sha256:5f1f17e50fffa26f6339bfbe8785168267209cc198c19827868f62ba99466fb0
Deleted: sha256:0612851c5050c4005f1330c178ef828a72c4c6c071f0ecce7b4d0958600fae3f
Deleted: sha256:3514544d79a0fa91785e030fb12f291c895684f4ee1305316fa2e261db57a1f3
Error response from daemon: conflict: unable to delete e94e9b988038 (cannot be forced) - image is being used by running container 122809aaea53
```

I assume the errors are chained so let's address them in order

docs/source/development/development_workflow.rst
fruttasecca: It would work with any tag you pass, at least the building part. The tag we are passing is essentially the tag of the image the k8s deployment is using, so that when you kill the pod it will use the newly built image.

docs/source/development/development_workflow.rst
fruttasecca: >Also, it's a bit confusing because bash orchest install installs the latest released version, no matter on which branch you're on. Am I right?

Yes you are right. Having `orchest install` install a version based on the branch is currently out of scope, and indeed it's related to that tag.

>Finally, it's worth stating that Orchest needs to be installed and running

I have added this note to the two scripts that have this requirement :+1:

docs/source/development/development_workflow.rst
fruttasecca: >By the way, is the kubectl delete pods -n orchest -l "app.kubernetes.io/name=orchest-api" step above even necessary? 

No it's not, the two code blocks refer to two different cases (single node vs multi-node). The second code block is just there to illustrate different helper scripts, they aren't supposed (or required) to be run one after the other.

docs/source/development/development_workflow.rst
fruttasecca: Unless I am misreading (or have misunderstood the comment), I see only one error. That error is related to
the fact that you are trying to delete an image that is currently in use, not something that this particular helper script should be responsible for imo.

docs/source/development/development_workflow.rst
astrojuanlu: > The tag we are passing is essentially the tag of the image the k8s deployment is using, so that when you kill the pod it will use the newly built image.

Then I understand that I should not be using any tag of my invention, but the one that is shown by

```
kubectl get deployments -n orchest orchest-api -o wide
```

Wondering if there's any `kubectl | sed | awk | grep` command we can use to extract such tag in an env var or something?

docs/source/development/development_workflow.rst
astrojuanlu: Oh, understood. Maybe we could use a tab for this, like https://sphinx-inline-tabs.readthedocs.io/

In any case, I skipped the multi node part for now

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,118 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
 .. _incremental development:
 
-Incremental development
------------------------
-.. warning::
-   🚨 For incremental development to work in WSL2, Docker must be installed within the WSL2
-   environment itself.
+Incremental development (hot reloading)
+---------------------------------------
+Now that you have Orchest and all development dependencies installed you are ready to start Orchest
```
astrojuanlu, line 154: I'm not sure if the "Working with Orchest" section is needed to follow the "Incremental development (hot reloading)" one. The sentence " Now that you have Orchest and all development dependencies installed you are ready to start Orchest" is a bit confusing, since it seems to imply that all the steps above are needed. If it is, maybe this should be a subsection of the former. If it's not, maybe we could rephrase the first line?

docs/source/development/development_workflow.rst
```diff
@@ -89,90 +94,118 @@ note that this won't include all the files defined in the Orchest repo), e.g.:
...
-committing <before committing>` your changes).
+- ``orchest-webserver``
+- ``orchest-api``
+- ``auth-server``
 
 .. code-block:: bash
 
+   # Start minikube with the repository mounted in the required place.
+   # Run this command while you are in the Orchest repository directory.
+   minikube start --memory 16000 --cpus 6 \
```
astrojuanlu, line 168: One cannot launch `minikube start --mount` if there is an existing minikube cluster, so I'm still confused about what the requirements are. If `minikube delete` is needed here, we could be more explicit about it.

astrojuanlu (changes requested):
I'm trying out several combinations of these commands and I'm still not able to make the `--dev` mode run.

As a general summary, what is the audience of this document? If we assume that folks already know how k8s and minikube work, then the docs are probably fine and I need to study more. But if we want to provide some instructions that someone with very rudimentary knowledge can follow, maybe we should be more explicit about the steps. For example, it's not entirely clear to me when `minikube delete` is needed.

An idea could be to add the list of commands (if possible) with minimal explanations at the beginning of the section, then write some more explanations below, then add a "troubleshooting" section. I'm not 100 % sure of how this would look like, but at the moment the instructions are not very linear and so far I haven't been able to make them work.

docs/source/development/development_workflow.rst
fruttasecca: I am fine with adding a line about the delete here, but generally speaking if we also need to cover error cases for external tools we are using the cost of making (and maintaining) docs will get higher, also considering that they will become more verbose. When you attempt to start  an already existing cluster with a mount that wasn't there you get a rather clear error from minikube that says you can't do that and basically you need to create the cluster again.

fruttasecca:
>I'm trying out several combinations of these commands and I'm still not able to make the --dev mode run.

Sounds like an error to me, could you clarify what are the combinations of commands that aren't working?

astrojuanlu:
I just made the `--dev` hot reloading work with these commands:

```
minikube delete  # In case there was an existing cluster
minikube start --cpus 4 --mount-string="$(pwd):/orchest-dev-repo" --mount

# FE dependencies
pnpm i
pnpm run dev &

# Install Orchest
./orchest install

# Only because I'm testing the PR
eval $(minikube -p minikube docker-env)
bash scripts/build_container.sh -i orchest-ctl -o v2022.03.8 -t v2022.03.8  # Or the appropriate tag

# Finally
./orchest stop  # Since `./orchest install` also starts it
./orchest start --dev
```

I think code-wise this is ready, I can address my own comments about the documentation in a separate PR ✔️ Thanks for your patience @fruttasecca !

docs/source/development/development_workflow.rst
astrojuanlu: Addressed in #860

docs/source/development/development_workflow.rst
astrojuanlu: Addressed in #860

docs/source/development/development_workflow.rst
astrojuanlu: Clarified in #860

docs/source/development/development_workflow.rst
astrojuanlu: Addressed in #860

PR #833, opened 2022-03-24
Enable creating or uploading files or folders in sub folders within `Project files` and `/data`

iannbing:
## Description

Currently, user can only upload/create files or folders to `Project files`, but not its sub-folder, or `/data`. 

This PR makes it possible for users to create/upload files or folders to the last selected (sub) folder, within `Project files` and `/data`.

Update: this [PR](https://github.com/orchest/orchest/pull/836) was merged into this one for completeness. Therefore, the changes of the PR also includes:

- FilePicker has access to `/data` folder
- FilePicker uses the new `/async/file-management/browse` endpoint instead of the old file-picker endpoint. 

Fix: #857 

## Checklist


services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -186,3 +196,32 @@ def generate_tree(dir, path_filter="/", allowed_file_extensions=[], depth=3):
...
+
+    if path is not None:
+        if not path.startswith("/"):
+            return (
+                False,
+                "The path query argument should always start with a forward-slash: /",
+            )
+        _, ext = os.path.splitext(path)
+        # path represents a folder
+        if len(ext) == 0 and not path.endswith("/"):
```
fruttasecca, line 218: I think this would lead to an error when passing a file without an extension,  see
```python
import os
path = "/a/file"
_, ext = os.path.splitext(path)
print(ext)
```

services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -24,21 +25,30 @@ def allowed_file(filename):
...
 
     if root not in ALLOWED_ROOTS:
         raise ValueError(f"Received illegal root path: {root}.")
 
+    if root is DATA_DIR_PATH:
+        return _config.USERDIR_DATA
+
+    if root is PROJECT_DIR_PATH and project_uuid is None:
+        raise ValueError(f"project_uuid is required.")
+
```
fruttasecca, line 45: We probably want to stick to the equality operator rather than `is`, which can be quite subtle.

services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -196,3 +196,31 @@ def generate_tree(dir, path_filter="/", allowed_file_extensions=[], depth=3):
...
+    _, ext = os.path.splitext(path)
+
+    # path represents a folder
+    if len(ext) == 0 and not path.endswith("/"):
+        return (
+            False,
+            "The path query argument should end with a forward-slash: / when representing a folder",
+        )
+
+    root_dir_path = construct_root_dir_path(root=root, project_uuid=project_uuid)
```
fruttasecca, line 225: I get the appeal of using a `(bool, str)` tuple, couldn't `validateRequest` and `construct_root_dir_path` basically be merged into the same function, and use a `ValueError` and its message to accomplish the same with less code? I haven't checked the entire PR yet so maybe there are cases where we want to call the latter without the former.

services/orchest-webserver/app/app/views/views.py
```diff
@@ -1025,16 +1025,23 @@ def project_file_exists(project_uuid, pipeline_uuid):
...
-        root_dir_path = filemanager.construct_root_dir_path(
-            request.args.get("root"), project_uuid
+    @app.route("/async/file-management/delete", methods=["POST"])
+    def file_delete():
+        root = request.args.get("root")
+        path = request.args.get("path")
+        project_uuid = request.args.get("project_uuid")
+
+        if path is None:
+            return jsonify({"message": "path is required."}), 400
```
fruttasecca, line 1061: Such checks can probably be offloaded to `validateRequest`

services/orchest-webserver/app/app/core/filemanager.py
iannbing: `ext` would be an empty string in this case according to the [doc](https://docs.python.org/3/library/os.path.html#os.path.splitext). This is also what I observed when testing.

fruttasecca: Wouldn't then this trigger the `if len(ext) == 0 and not path.endswith("/"):` case?

services/orchest-webserver/app/app/views/views.py
```diff
@@ -1051,10 +1076,8 @@ def filemanager_delete(project_uuid):
             return (
                 jsonify(
                     {
-                        "message": (
-                            "It is not allowed to delete roots through "
-                            " the file-manager."
-                        )
+                        "message": "It is not allowed to delete roots "
+                        + "through the file-manager."
```
fruttasecca, line 1083: Just a suggestion, but it might be more handy to use parenthesis and construct a multiline string this way:
```python
 {
     "message":  (
         ...
      )
 }
```

services/orchest-webserver/app/app/views/views.py
```diff
@@ -1099,28 +1129,23 @@ def filemanager_duplicate(project_uuid):
...
+        if path is None:
+            return jsonify({"message": "path is required."}), 400
+
+        is_valid, result = filemanager.validateRequest(
+            root=root, path=path, project_uuid=project_uuid
         )
 
-        if "path" not in request.args:
-            return jsonify({"message": "Missing path query argument."}), 500
+        if is_valid is False:
```
fruttasecca, line 1148: `if not is_valid` might be better form

services/orchest-webserver/app/app/views/views.py
```diff
@@ -1233,37 +1272,35 @@ def filemanager_download(project_uuid):
...
+        if path is None:
+            return jsonify({"message": "path is required."}), 400
+
+        is_valid, result = filemanager.validateRequest(
+            root=root, path=path, project_uuid=project_uuid
         )
 
-        if "path" not in request.args:
-            return jsonify({"message": "No path was passed."}), 500
+        if is_valid is False:
```
fruttasecca, line 1292: same as https://github.com/orchest/orchest/pull/833/files#r835150616

services/orchest-webserver/app/app/utils.py
```diff
@@ -565,20 +565,19 @@ def generate_ipynb_from_template(kernel_name):
     return json.dumps(template_json, indent=4)
 
 
-def create_pipeline_file(file_path, pipeline_directory, kernel_name):
+def create_file_from_template(file_path, kernel_name="python"):
```
fruttasecca, line 568: Now that the scope of this function has changed to arbitrary files it needs to have some more documentation around it and `kernel_name` should probably be  `Optional[str]`. Perhaps in the longer there there should be different functions at different abstraction layers to take care of this, but let's not creep the scope of this PR :P

fruttasecca:
What do you think of moving the file manager views into its own module in `services/orchest-webserver/app/app/views`?

services/orchest-webserver/app/app/views/views.py
```diff
@@ -59,6 +59,8 @@
     serialize_environment_to_disk,
 )
 
+ALLOWED_FILE_EXTENSIONS = ["ipynb", "R", "py", "sh"]
```
fruttasecca, line 62: I see that this is something that was there already, I guess we are missing julia files? The extension should be `jl` AFAIK

services/orchest-webserver/app/app/views/views.py
```diff
@@ -942,6 +942,8 @@ def get_file_tree():
 
         root_dir_path = result
 
+        # no need to check if it's a valid file path,
+        # we only care about the directory after os.path.split
```
fruttasecca, line 946: Comments should start with a capital letter and end with a `.`.  The `orchest-webserver` has some technical debt which includes comments that are not correctly formatted, so I can understand how this can picked up, sorry about that.

services/orchest-webserver/app/app/core/filemanager.py
iannbing: You're right. I was assuming all files should have an extension. Removed this checking.

services/orchest-webserver/app/app/views/views.py
iannbing: Indeed, didn't know this magic  👍

services/orchest-webserver/app/app/utils.py
iannbing: The new name is indeed misleading. I figured that the template was actually a template for metadata of ipynb. Changed it to `create_empty_file` and added some explanation.

services/orchest-webserver/app/app/views/views.py
iannbing: I was a bit hesitant about it. For the endpoint `/async/file-management/browse`, `path` is not required. I just moved this check to `validateRequest` and added `is_path_required` boolean there to handle the special case. I'm still not sure if it's better though. But indeed it's less repetition.

iannbing:
> What do you think of moving the file manager views into its own module in services/orchest-webserver/app/app/views? It would be a great chance to add a top-level docstring that maps out what's happening.

[Blueprint](https://flask.palletsprojects.com/en/2.0.x/tutorial/views/#) seems like the "Flask" way of organizing views. Yet, I probably would do it in another refactoring PR.

Currently, `orchest-webserver` is pretty much UI-driven, e.g. `/async/file-picker-tree`. We probably would want to decouple the endpoint design with the UI. I would give more thought into it since it's hard to choose the right level of abstraction for BFF (because BFF itself doesn't own any resource and is constantly being changed due to the change of FE).

services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -24,21 +25,30 @@ def allowed_file(filename):
...
+        raise ValueError("project_uuid is required.")
+
     project = Project.query.filter(Project.uuid == project_uuid).first()
 
-    if root == PROJECT_DIR_PATH:
-        root = _config.USERDIR_PROJECTS + f"/{project.path}"
-    else:  # We know it is "/data"
-        root = _config.USERDIR_DATA
+    if project is None:
+        raise ValueError("project not found.")
```
yannickperrenet, line 49: Small nitpick, how about making the message: `f"Project with uuid '{project_uuid}' not found."`

services/orchest-webserver/app/app/utils.py
```diff
@@ -558,27 +558,30 @@ def get_ipynb_template(language: str):
...
-def create_pipeline_file(file_path, pipeline_directory, kernel_name):
+def create_empty_file(file_path: str, kernel_name: Optional[str] = "python"):
     """
+    creates an empty file.
+    If the file extension is ".ipynb", kernel_name would be used to
+    configure its metadta
+
     Note: this function does not assume that step['file_path']
     holds the value of file_path!
     """
```
yannickperrenet, line 576: We follow the convention of structuring docstrings like [this](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -186,3 +196,44 @@ def generate_tree(dir, path_filter="/", allowed_file_extensions=[], depth=3):
...
+            depth = int(depth, 10)
+        except Exception:
+            raise ValueError("Invalid value for the depth query argument.")
+
+    # in most cases, path is required,
+    # except for /async/file-management/browse,
+    # where either depth and path should be provided
+    if depth is None and path is None:
+        extra_explanation = (
+            "" if is_path_required else " if the depth query argumentis not present"
```
yannickperrenet, line 227: Typo: `argumentis` --> `argument is`

services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -186,3 +196,44 @@ def generate_tree(dir, path_filter="/", allowed_file_extensions=[], depth=3):
...
     return jsonify(tree)
+
+
+def process_request(
+    root: Optional[str],
+    path: Optional[str],
+    project_uuid: Optional[str],
+    depth: Optional[str] = None,
+    is_path_required: Optional[bool] = True,
+) -> Tuple[str, Optional[int]]:
```
yannickperrenet, line 207: No need to change this now, but I feel like `process_request` is at the wrong level of abstraction. Instead we should use stricter request marshaling (i.e. when a request is made with the wrong schema then it should automatically get the corresponding error).

services/orchest-webserver/app/app/utils.py
iannbing: Changed it. Thanks for the link to an example 👍

services/orchest-webserver/app/app/core/filemanager.py
iannbing: Changed 👍

services/orchest-webserver/app/app/views/views.py
```diff
@@ -1176,94 +1177,86 @@ def filemanager_upload(project_uuid):
...
-                            " start and end with a forward-slash: /"
-                        ),
-                    }
-                ),
-                500,
-            )
+        path_filter = path
 
-        extensions = request.args.get("extensions").split(",")
+        extensions = extensions.split(",")
```
yannickperrenet, line 1253: Strictly speaking we would first need to check whether `extensions is not None`, but that again ties into the request marshalling.

yannickperrenet:
Touching some Python code I see. I love it! 

I made a few additional comments, like about the docstrings, so you can get familiar with how we do things :+1:

services/orchest-webserver/app/app/core/filemanager.py
iannbing: I thought about passing `request` directly to this function. But just like you commented, it feels wrong to have `request` in a function of `filemanager.py`. `request` should stay in `views.py`. So, I think if we can do request marshalling properly. This code smell will be removed together.

ricklamers:
I'm getting some errors related to file extension validation on nested files:

image

So just for clarity, the pipeline sits at the root of the project and I'm dragging in a notebook from `<root>/test/abc.ipynb`.

iannbing:
@ricklamers  Could you be more specific?

> So just for clarity, the pipeline sits at the root of the project and I'm dragging in a notebook from <root>/test/abc.ipynb.

I'm assuming that you're dragging  `<root>/test/abc.ipynb` into "the canvas", and I can reproduce it. If not, please elaborate the issue that you encountered.

> Also got this after dragging that file out

This confuses me. Do you mean dragging a file out to `<root>` ? If yes, then I cannot reproduce it.

ricklamers:
>I'm assuming that you're dragging <root>/test/abc.ipynb into "the canvas", and I can reproduce it, but this issue is already
in dev. If not, please elaborate the issue that you encountered.

Correct.

> This confuses me. Do you mean dragging a file out to <root> ? If yes, then I cannot reproduce it.

I just meant I also got that error after performing the "drag file to canvas" operation. Sorry for the unclear language.

services/orchest-webserver/app/app/core/filemanager.py
yannickperrenet: > So, I think if we can do request marshalling properly. This code smell will be removed together.

Agreed. Marshaling is probably a good candidate to be picked up together with #201 (given that that talks about introducing a public API).

services/orchest-webserver/app/app/views/views.py
yannickperrenet: > Sorry, I'm a big fan of early-return

Hahah yes I noticed. I generally go for readability and merge all if-statements together. Potato potaatooo

ricklamers:
I found one more issue :smile:
https://www.tella.tv/video/cl1gj7xl5000009l6fp4pdiwd/view

Could be a simple line change.

iannbing:
> tldv; when you have selected file in the file manager and you click and drag another file without holding a modifier key instead of just selecting the new file it holds the selection on the previously selected file causing multiple files to become part of the drag operation.

Thanks for catching this :-), I created a separate issue #857 for it.

services/orchest-webserver/app/app/core/filemanager.py
iannbing: @yannickperrenet  I'll then resolve this comment now, and solve this issue later with more systematic approach. Let me know if there's any specifics that can be improved in this PR :-)

iannbing:
@ricklamers  Added the fix for #857 .

Issue #141, opened 2021-02-08
Avoid requiring a session restart to see changes in Project/Pipeline environment variables for kernels


ricklamers:
Currently, when you make changes to project or pipeline environment variables a restart of the session is required.

This is because in the current implementation the environment variables get passed through the Jupyter Enterprise Gateway.

Since we inject environment variables as language native environment variables (e.g. `import os; os.environ['...']` in Python) changing them for running kernels seems hacky and non-desirable.

However, restarting the full pipeline session has some side-effects like clearing the memory server used for passing data. Hence, we could instead make a kernel restart fetch updated environment variables using for example the orchest-api.

This issue will keep track of the technical implementation and the desired properties of such dynamic updating of environment variables for kernels.

PR #839, opened 2022-03-30
Inject user defined environment variables on kernel start

yannickperrenet:
## Description
As a consequence, the session no longer needs to be restarted to get
updated environment variables. Instead, a kernel restart is enough. This
is a better UX for users since a kernel restart is more lightweight than
a complete session restart.

Caveat: With this change, different kernels (of different steps) can now
have different values for the same environment variable. This can be
confusing to users, since environment variables are defined at a higher
level (i.e. project or pipeline level, not step level).

Does the caveat outweigh the improved UX? @ricklamers 

Resolves: #141

### DIFF
--- a/services/orchest-webserver/app/app/res/kernels/launch_kubernetes.py
+++ b/services/orchest-webserver/app/app/res/kernels/launch_kubernetes.py
@@ -50,6 +50,23 @@ def _env_image_name_to_project_environment_uuid(image_name: str) -> Tuple[str, s
     return proj_uuid, env_uuid
 
 
+def _get_user_env_vars(proj_uuid: str, pipe_uuid: str) -> dict:
+    project_resp = requests.get(
+        "http://" + _config.ORCHEST_API_ADDRESS + f"/api/projects/{proj_uuid}"
+    )
+    pipeline_resp = requests.get(
+        "http://"
+        + _config.ORCHEST_API_ADDRESS
+        + f"/api/pipelines/{proj_uuid}/{pipe_uuid}"
+    )
+
+    if project_resp.status_code != 200 or pipeline_resp.status_code != 200:
+        raise RuntimeError("Failed to retrieve user environment variables.")
+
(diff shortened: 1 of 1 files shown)


ricklamers:
>Does the caveat outweigh the improved UX? @ricklamers

I think so. Kernels are "dynamic creatures", I don't see a strong need for environment variable consistency there.

ricklamers:
Clean PR, @yannickperrenet strikes again :fire:!

services/orchest-webserver/app/app/res/kernels/launch_kubernetes.py
```diff
@@ -100,6 +117,18 @@ def _get_kernel_pod_manifest(
...
+    except RuntimeError as e:
+        sys.exit(e)
+    else:
+        # NOTE: This update needs to happen after adding `os.environ`.
+        # Otherwise old env var values (that are present in the
+        # environment of the EG) will overwrite updated values.
+        environment.update(user_env_vars)
     # Let the image PATH be used. Since this is relative to images,
     # we're probably safe.
     environment.pop("PATH")
```
fruttasecca, line 134: Any risk for undesired behaviour here? `PATH` might not be an uncommon env var after all

fruttasecca (approved):
lgtm :+1:  there are a bunch of env vars that we use (e.g. `EG_RESPONSE_ADDRESS`) that might be overwritten, are we okay with that? The risk seems low and I don't want to introduce complexity for no reason, just shooting ideas

services/orchest-webserver/app/app/res/kernels/launch_kubernetes.py
yannickperrenet: I would rather leave it as is given that it has served us fine so far. I also have the feeling that modifying `PATH` could break things with `conda` environments and leading to the wrong environment being used when executing user code.

I will add a quick mention about `PATH` in the docs :+1:

PR #840, opened 2022-03-30
Allow specifying user dir, builder cache and nfs claim size on install.

fruttasecca:
## Description

Makes it so that it's now possible to specify the aforementioned sizes on install, moreover, if no related env var is required it prompts the user for the desired size. This helps in making it clear to the user the size of storage that will be available and explains that it (currently) cannot be altered later.

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -75,6 +75,9 @@ else ifeq (${ROOK_NFS_TARGET},rook-nfs)
   STORAGE_CTL_TARGET := rook-nfs
 endif
 
+USERDIR_PVC_SIZE := ${if ${USERDIR_PVC_SIZE},${USERDIR_PVC_SIZE},50Gi}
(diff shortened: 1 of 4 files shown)


deploy/Makefile
```diff
@@ -146,6 +150,8 @@ ORCHEST_RSC_DEPLOY_CONFIG := orchest-resources \
 							${SHARED_DEPLOY_CONFIG}	\
 							${CLUSTER_RSC_ENABLE_CONFIG} \
 							${ORCHEST_RSC_USERDIR_STORAGE} \
+							--set cluster-resources.orchest.userdir.storage="${USERDIR_PVC_SIZE}" \
```
nhaghighat, line 153: can you set the storage if the env var is not defined? to be able to use the default var in values.

fruttasecca: :+1: will do

nhaghighat (approved):

PR #843, opened 2022-03-30
Lock pip dependencies

astrojuanlu:
## Description

Locks pip dependencies using [pip-tools](https://pypi.org/project/pip-tools/).

For now this is a proof of concept, I only touched `services/auth-server`. Every step is in a separate commit so it's clear how to proceed with the rest (happy to complete the PR if it gets approved).

Quick explanation of how to use pip-tools:

1. Requirements are listed in `requirements.in`, and versions there are generally left unpinned or, at most, include version caps if we know we are incompatible (notice that I unpinned everything *after* locking the dependencies, see the next step)
2. `pip-compile` locks all transitive dependencies into `requirements.txt`
3. Any subsequent call to `pip-compile` will **not** upgrade any dependency
4. To upgrade a given dependency, `pip-compile -P <dep>` will do, or alternatively `pip-compile -U` will try to upgrade everything (not recommended, because then we can have the dependency problems we are willing to avoid 🙃)
5. The `requirements.txt` files produced by pip-tools are perfectly compatible with `pip install -r requirements.txt`, so from the outside nothing changes

The big advantage of this method is that we won't have transitional new versions breaking Orchest from time to time. The disadvantage is that we might forget to upgrade the dependencies from time to time (if that's a bad thing at all). There are two strategies here:

- Just don't upgrade unless necessary (because of security bugs or wanted new features)
- Configure something like https://pyup.io/ to do controlled upgrades

If folks agree with this I will complete the PR for all the rest of the services.

## Checklist


fruttasecca:
nice :+1: , is there a way to sort the compiled dependencies? Perhaps sorting the ones coming from `requirements.in` first. Definitively like the idea tbh, think this also deserves a place in the developer docs.

astrojuanlu:
Notice that direct dependencies could be extracted from the locked `requirements.txt` like this:

```
$ grep requirements.in -B2 requirements.txt | grep -v '#' | grep -v '\--'
-e file:../../lib/python/orchest-internals
eventlet==0.30.2
flask==1.1.1
flask-migrate==2.5.3
flask-script==2.0.6
flask-sqlalchemy==2.4.1
itsdangerous==2.0.1
psycopg2-binary==2.8.6
requests==2.26.0
sqlalchemy-utils==0.37.4
werkzeug==2.0.3
```

astrojuanlu:
Rendered dev docs: https://orchest--843.org.readthedocs.build/en/843/development/development_workflow.html#python-dependencies

fruttasecca (approved):
lgtm :), do you think it makes sense to add a checkbox to the PR template to remind contributors to run `pip-compile` after updating the `requirements.in` file? (No is a perfectly valid answer :P)

astrojuanlu:
Done!

astrojuanlu:
Since this is approved and there are no more comments, I'm updating with `dev` and merging when the tests pass

PR #846, opened 2022-03-30
Set `DEBIAN_FRONTEND=noninteractive` for `sudo` in environments

yannickperrenet:
## Description
Persists `DEBIAN_FRONTEND=noninteractive` from base images.

Adds `env_keep` entry in the sudoers file to ensure the
`DEBIAN_FRONTEND` value of `noninteractive` is persisted. Meaning that
packages, such as `git` can be installed as expected, i.e.:
`sudo apt-get update && sudo apt-get install -y git`

@astrojuanlu Let me know whether installation now works like you would expect :)

### DIFF
--- a/services/base-images/base-kernel-julia/Dockerfile
+++ b/services/base-images/base-kernel-julia/Dockerfile
@@ -13,8 +13,14 @@ USER root
 RUN apt-get update \
     && apt-get install default-libmysqlclient-dev sudo -y
 
-# enable sudo for the NB_USER by default
-RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+RUN passwd -d $NB_USER \
+    # Enable `sudo` for the $NB_USER by default.
+    && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER \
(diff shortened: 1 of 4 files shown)


astrojuanlu:
I'm still struggling a bit understanding how to do a dev install (see my comments at #832 so I'm not sure what would be the best way to test this 😕

services/base-images/base-kernel-julia/Dockerfile
```diff
@@ -13,8 +13,14 @@ USER root
 RUN apt-get update \
     && apt-get install default-libmysqlclient-dev sudo -y
 
-# enable sudo for the NB_USER by default
-RUN passwd -d $NB_USER && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER
+RUN passwd -d $NB_USER \
+    # Enable `sudo` for the $NB_USER by default.
+    && echo "$NB_USER   ALL=(ALL)   NOPASSWD:ALL" | tee /etc/sudoers.d/$NB_USER \
```
fruttasecca, line 18: Would it make sense to add `--append` here as well?

fruttasecca (approved):
lgtm :+1:

services/base-images/base-kernel-julia/Dockerfile
yannickperrenet: The base images are not using the `sudoers.d` files so I like us explicitly calling without `--append`.

Issue #838, opened 2022-03-30
In FileManager, moving a parent item and a child item together throws an error


iannbing:
**Describe the bug**

When user moves a parent item (e.g. `/data/parent`) and a child item (e.g. `/data/parent/child.py`) together, they are treated as two individual items. This causes a problem: since the child already is moved along, it immediately throws an error saying that path is invalid.

**Expected behavior**

No error is thrown. The moving / renaming of the child should not take place.

**To Reproduce**

1. Go to any pipeline.
2. Select a file within a folder
3. Press Ctrl/Cmd, and select the parent folder
4. Move them to another folder
5. Error occurs

PR #847, opened 2022-03-30
Fix the error when moving a folder while its sub folders or files are also selected

iannbing:
## Description

When moving a folder while its files are also selected, it throws an error saying that the selected files do not exist. This is because those files are moved along with the folder already.

Solution: only execute the "moving the folder" operation and skip the operations on moving its offspring nodes.

Fixes: #838 

## Checklist


ricklamers (approved):
Nice catch!

PR #850, opened 2022-03-31
make orchest reachable if ORCHEST_FQDN is not defined via cluster ip(…

nhaghighat:
…:port)

## Description

This PR removes the requirement of ORCHEST_FQDN by default, if it is not specified, the cluster will be reachable via ip:port

Fixes: #issue

## Checklist


nhaghighat:
> What would the port be here? 80?

It will be 80 for nginx ingress, just mentioned it for any user who wants to change it.

fruttasecca:
Looks good to me, since there are been some ingress changes in the yaml files, how will this impact users during an update?

nhaghighat:
> Looks good to me, since there are been some ingress changes in the YAML files, how will this impact users during an update?

I haven't tested the update (will do) but theoretically, helm updates the ingress rules, and the Nginx ingress controller will update Nginx, and then Nginx will start to serve the request based on IP address.

fruttasecca:
Alright sounds good. For testing updates, you should be able to do so by rebuilding `orchest-ctl` (with the right tag, i.e. -t <current deployment version> and -o <current deployment version>) and running the `hidden-update ` command.

fruttasecca (approved):

PR #858, opened 2022-04-01
Compare version correctly

yannickperrenet:
## Description

String comparison of versions breaks because we don't use a padding zero for the
patch segment of our CalVer and SemVer versioning scheme.

## Checklist
- [x] Update `useCheckUpdate.tsx` for update dialog


fruttasecca:
Should we make use of https://pypi.org/project/packaging/? Comparing versions should be about 2 lines of code

astrojuanlu:
> Should we make use of https://pypi.org/project/packaging/?

+1, see https://packaging.pypa.io/en/latest/version.html

yannickperrenet:
> Should we make use of https://pypi.org/project/packaging/? Comparing versions should be about 2 lines of code with this

Given that it is just one function, I don't think it makes sense to add an additional dependency.

fruttasecca (approved):
>Given that it is just one function, I don't think it makes sense to add an additional dependency.

lgtm then

PR #859, opened 2022-04-05
Make dependency pinnings explicit

astrojuanlu:
## Description

After #843, some of the dependencies were not made explicit and accidental upgrades of problematic dependencies (recent releases of werkzeug and jinja2) sneaked in. This fixes it.

The culprit is using an unsupported Flask version by the way. But we can think about upgrading it at some other time.

Fixes: #issue

## Checklist


yannickperrenet (approved):
:+1:

PR #861, opened 2022-04-05
Redirect to Pipelines if user deleted the orchest file of the current pipeline.

iannbing:
## Description

Currently in Pipeline Editor, if user deleted the `.orchest` file of the current pipeline, it throws errors and user gets stuck at `/pipeline`. 

This PR contains the following changes:

- The delete file warning also informs user if the to-be-deleted file/folder paths contain any `.orchest` file. 
- If user deleted the `.orchest` file of the current pipeline, user will be redirect to `/pipelines`.

## Checklist


services/orchest-webserver/client/src/pipeline-view/file-manager/FileManagerLocalContext.tsx
```diff
@@ -155,39 +167,62 @@ export const FileManagerLocalContextProvider: React.FC<{
...
+    const filesToDeleteString =
+      filesToDelete.length > 1
+        ? `${filesToDelete.length} files`
+        : `'${getBaseNameFromContextMenu(filesToDelete[0])}'`;
+
+    const hasPipelineFiles = await Promise.all(
+      filesToDelete.map((pathToDelete) => {
+        if (isFileByExtension(["orchest"], pathToDelete)) return true;
+        let { root, path } = unpackCombinedPath(pathToDelete);
+        return searchFilePathsByExtension({
```
yannickperrenet, line 194: `searchFilePathsByExtension` triggers a rather expensive BE call, which will recursively go over the given path (and all subdirectories) looking for the given extension. Ideally, we minimize the number of traverses of the file system and combine overlapping paths. Right now it seems like the function is called for every file to be deleted, so that can probably be reduced.

yannickperrenet (approved):
Haven't tested myself, but the described changes sound good to me :+1:

services/orchest-webserver/client/src/pipeline-view/file-manager/FileManagerLocalContext.tsx
iannbing: Good point! It has been addressed in the latest commit. Basically, only the ancestor paths will be considered, e.g. given two paths: `/a` and `/a/b/c`, `/a/b/c` will be skipped and only one request will be fired.

ricklamers (approved):

PR #862, opened 2022-04-05
set pvc size of registry to 999Ti

nhaghighat:
## Description

This PR sets the storage size of registry to 999Ti


fruttasecca (approved):
lgtm, I assumed it would have been a more complex change :D

PR #868, opened 2022-04-08
`orchest-api` owns the orchest settings

fruttasecca:
Currently, the orchest settings are a json file that is stored in the `config-pvc`, and which is shared by the `auth-server`, `orchest-wesbserver` and `orchest-ctl`. This PR makes it so that the `orchest-api` is the owner of the settings, and said settings will now be persisted in the `orchest-api` database. Orchest settings are now exposed as API endpoints of the `ctl` API namespace.

This PR is part of the effort to make Orchest be managed by a k8s controller, and it assumes that `orchest-ctl` won't be in charge anymore, in fact, no changes to `orchest-ctl` have been made (it still reads the json from the pvc, etc.)

Note: this PR includes database schema changes.


yannickperrenet:
As a general comment, the [configuration docs](https://docs.orchest.io/en/latest/fundamentals/configuration.html#orchest-settings) need to be updated:
> Orchest stores a global configuration file at ~/.config/orchest/config.json (or at $XDG_CONFIG_HOME/orchest/config.json if defined) that configures the Orchest application.

fruttasecca:
>As a general comment, the [configuration docs](https://docs.orchest.io/en/latest/fundamentals/configuration.html#orchest-settings) need to be updated:

Good catch, will do

services/orchest-api/app/app/utils.py
```diff
@@ -220,3 +225,299 @@ def get_jupyter_server_image_to_use() -> str:
...
+        )
+        stmt = stmt.on_conflict_do_update(
+            index_elements=[models.Setting.name], set_=dict(value=stmt.excluded.value)
+        )
+        db.session.execute(stmt)
+
+        # Delete settings that are not part of the new configuration.
+        models.Setting.query.filter(
+            models.Setting.name.not_in(list(settings_as_dict.keys()))
+        ).delete()
```
yannickperrenet, line 339: Isn't it possible to do `not_in(settings_as_dict)` (given that both implement `__contains__()`)?

fruttasecca: looks like it's not possible  since `notin` only accepts a list of literal values or tuples

yannickperrenet (approved):
:100:

nhaghighat (approved):

PR #869, opened 2022-04-11
Merge  the functionality of the `Pipelines` view into PipelineEditor

iannbing:
## Description

To simplify the workflow, this PR merges the functionality of the `Pipelines` view into `PipelineEditor`, so that user could directly manage pipelines without going back and forth between `Pipelines` and `PipelineEditor`.

Major changes:
- The old "Pipelines" view is removed. Therefore, the navigation item "Pipelines" in the main navigation bar will now lead to PipelineEditor directly.
- When landing on PipelineEditor, it will load the first pipeline in the project. PipelineEditor will try to automatically load the first from the remaining pipelines in the project, for example, when user deletes the currently-open pipeline, the next one will be loaded automatically.
- In PipelineEditor, added a "Sessions" panel for managing Orchest sessions. Added "Create Pipeline" in FileManager for opening the "Create a new pipeline" dialog. Sessions panel is vertically resizable.
- Added the pipeline file path under the pipeline name in the HeaderBar to make it more explicit.

Screenshot 2022-04-11 at 14 08 04

## Checklist


ricklamers:
I think we need a "you have no projects" view for the pipeline page if you have no projects:

image

Perhaps similar to what we have on the other entity pages:

ricklamers:
image

Can we change this to start at the project root with no slash prefix? Like how GitHub shows repo paths:

image

ricklamers:
Another minor issue related to deleting the pipeline you currently have opened:

https://www.tella.tv/video/cl1vv1l3m001509mo54vj6qwo/view

iannbing:
> I think we need a "you have no projects" view for the pipeline page if you have no projects

Good catch! Missed this one. Will do.

> Can we change this to start at the project root with no slash prefix? Like how GitHub shows repo paths

I was trying to make it explicit to users, but I also find it awkward. Good to have your opinion. Will do.

ricklamers:
I've noticed the dialog for deleting a file doesn't receive out focus on the confirm button. So you can't right-click -> delete -> hit return key. Could you add proper focus there? Thank you!

ricklamers:
Should we swap the icons for create file/create folder and upload file/upload folder?

In addition, it might make more sense to put the create file/folder up front (so create file, create folder, upload file, upload folder, refresh, collapse).

image

ricklamers:
I've noticed a couple of times that the log page doesn't load:

image

I don't have a pending network request or error in the console.

ricklamers:
After renaming the pipeline file (after stopping the session) the settings page doesn't update the path:

image

iannbing:
@ricklamers I think all feedback items are addressed. Spent some more time on fixing the failed requests you have experienced. Also fixed an edge case: if user move a folder that contains `.orchest` files of which still have active sessions, prompt user a Confirm dialog, and user can press "Confirm" to kill their sessions.

ricklamers:
Minor, we're missing some padding here: 
image

ricklamers:
I'm still fairly easily able to trigger a stuck loading logs page:

image

The way I can reliably reproduce is by switching projects and than (with a different pipeline open) opening the logs page. Doing that a couple of times break logs page (stuck loading) every time.

ricklamers:
Pretty weird behavior from MUI:

image

The icons are different (first two rounded, last two aren't). Can we fix that? Or use the proposed icons of my earlier comment to avoid having different roundedness styles?

ricklamers:
Note: dragging files onto the canvas doesn't seem to work anymore.

iannbing:
> The icons are different (first two rounded, last two aren't). Can we fix that? Or use the proposed icons of my earlier comment to avoid having different roundedness styles?

MUI seems to tag the icons inconsistently. Solid UploadFile and Solid UploadFolder are not in the "Filled" category and only appeared in "Rounded" category. Fixed.

iannbing:
> The way I can reliably reproduce is by switching projects and then (with a different pipeline open) opening the logs page. Doing that a couple of times break logs page (stuck loading) every time.

This is very strange. I couldn't reproduce it. Are you also running some jobs? If not, then the only thing that could get stuck is Socket.io.

ricklamers:
Fantastic job @iannbing! LGTM

PR #872, opened 2022-04-14
Automatically delete unused environment images from nodes and registry

fruttasecca:
## Description

This PR introduces lifecycle management for environment images. Images that are not considered active will be deleted from the registry and on every node. An image is considered in active and marked for deletion given the following:
- it's not the latest image built for an environment
- it's not in use by a job, interactive session (services, kernels) or interactive run

Deletion from the nodes happens through the `node-agent`, the `orchest-api` will provide which environment images should be on nodes (i.e. active images), the `node-agent` will then proceed to delete environment images that aren't part of the list.

Deletion from the registry happens by having the `celery-worker` interact with the `docker-registry` API, this only deletes references, to have the image layers deleted from the `docker-registry` filesystem the `celery-worker` will initiate the registry garbage collection process.

Things of note:
- the `docker-registry` was pointed to a non existent `redis` cache to make it act as if the metadata cache was always empty, to solve https://github.com/distribution/distribution/issues/1803
- on start, kernels now "lock" a given environment image, that is, images that are used by kernels will be considered active
- the `orchest-api` will periodically run a job that marks images that are to be removed and, if necessary, queues a registry garbage collection task
- we now store the digest of every built environment image to take care of the edge case where different `image:tag`s are backed by the same digest
- the PR contains schema changes

p.s., a lot of lines of code are because of schema migrations, don't be scared to review :p

## Checklist


deploy/Makefile
```diff
@@ -104,6 +104,12 @@ DOCKER_REGISTRY_DEPLOY_CONFIG := registry \
 								${DOCKER_REGISTRY_NAME_CONFIG} \
 								${DOCKER_REGISTRY_PORT_CONFIG} \
 								${DOCKER_REGISTRY_PVC_CONFIG} \
+								--set configData.storage.delete.enabled=true \
+								--set configData.storage.cache.blobdescriptor=redis \
+								--set configData.redis.addr=localhost \
+								--set configData.redis.dialtimeout=0ms \
+								--set configData.redis.readtimeout=0ms \
+								--set configData.redis.writetimeout=0ms \
```
yannickperrenet, line 112: @nhaghighat Most likely something that needs to be added to the operator.

services/orchest-api/app/app/core/environments.py
```diff
@@ -215,3 +219,162 @@ def is_environment_in_use(project_uuid: str, env_uuid: str) -> bool:
...
+        )
+        .label("tag_rank")
+    )
+    latest_imgs = db.session.query(models.EnvironmentImage).add_column(tag_rank)
+    latest_imgs = (
+        latest_imgs.from_self()
+        .filter(tag_rank == 1)
+        .with_entities(models.EnvironmentImage.digest)
+        .subquery()
+    )
```
yannickperrenet: I am not completely sure I understand what is happening here. Let's say you have environments A and B (both in the same project), then all `EnvironmentImage`s resulting from builds from A will have the same `environment_uuid`. The same holds true for B. Now we have `A_1`, `A_2`, `A_3`, etc. where `A_i` means an `EnvironmentImage` for A with tag `i` (similar definition for B).

When doing this `over, partition_by and order_by` don't all `A_i` get the same rank given that we partition using `project_uuid` and `environment_uuid`?

Don't we want all to give rank 1 to the latest tag environment image of all environments? Maybe this is actually happening but I am just misunderstanding the query. On that note, I am not completely sure how to interpret the `order_by` (I thought this was to decide what partition to start the rank count from).

deploy/helm/templates/_cluster-role.tpl
```diff
@@ -28,6 +28,7 @@ rules:
       - pods
       - pods/log
       - pods/attach
+      - pods/exec
```
yannickperrenet, line 31: @nhaghighat Tagging you along for this change as well :)

services/orchest-api/app/app/core/environments.py
fruttasecca: >Don't we want all to give rank 1 to the latest tag environment image of all environments?

Yes that what's happening (AFAIK :P, will verify)

About the partitioning and ranking, what the query is doing is the following: partition those rows by (proj, env), then, at the partition level, rank them, the rank is done (per partition) by tag

fruttasecca: Example
```
           environment_uuid           | tag 
--------------------------------------+-----
 18b59993-23ab-4e40-b291-f78ec273bf3e |   1
 c56ab762-539c-4cce-9b1e-c4b00300ec6f |   1
 c56ab762-539c-4cce-9b1e-c4b00300ec6f |   2
 e0af758d-be45-4355-912a-7ee76ff699dd |   1
 e0af758d-be45-4355-912a-7ee76ff699dd |   2
 e0af758d-be45-4355-912a-7ee76ff699dd |   3
 ```
 
 ```
select project_uuid, environment_uuid, tag,
rank() OVER (
        partition by project_uuid, environment_uuid
        order by tag desc
) rank
from environment_images
order by project_uuid, environment_uuid, tag desc;;
```

result

```
             project_uuid             |           environment_uuid           | tag | rank 
--------------------------------------+--------------------------------------+-----+------
 13006c56-6237-45ff-b1ee-7828a1e1a59f | 18b59993-23ab-4e40-b291-f78ec273bf3e |   1 |    1
 13006c56-6237-45ff-b1ee-7828a1e1a59f | c56ab762-539c-4cce-9b1e-c4b00300ec6f |   2 |    1
 13006c56-6237-45ff-b1ee-7828a1e1a59f | c56ab762-539c-4cce-9b1e-c4b00300ec6f |   1 |    2
 13006c56-6237-45ff-b1ee-7828a1e1a59f | e0af758d-be45-4355-912a-7ee76ff699dd |   3 |    1
 13006c56-6237-45ff-b1ee-7828a1e1a59f | e0af758d-be45-4355-912a-7ee76ff699dd |   2 |    2
 13006c56-6237-45ff-b1ee-7828a1e1a59f | e0af758d-be45-4355-912a-7ee76ff699dd |   1 |    
 ```

**Comment by fruttasecca** [comment]
Btw, if you want to test for the space actually being freed:
- from the registry `du -sh /var/lib/registry/`
- from the node `docker system df`

The node-agent logs will log deletions from the node, the builds worker will log deletion from the registry

File: services/orchest-api/app/app/__init__.py [hunk 4]
```diff
@@ -134,6 +136,13 @@ def create_app(
             args=[app],
         )
 
+        scheduler.add_job(
+            process_images_for_deletion,
```
yannickperrenet, line 140: How about moving this to the `Scheduler` class as well (just like the job that is added a few lines above)?

services/orchest-api/app/app/__init__.py
```diff
@@ -147,6 +156,46 @@ def create_app(
...
+def process_images_for_deletion(app):
+    """Processes built images to find inactive ones.
+
+    Goes through env images and marks the inactive ones for removal,
+    moreover, if necessary, queues the celery task in charge of removing
+    images from the registry.
+
+    Note: this function should probably be moved in the scheduler module
+    to be consistent with the scheduler module of the webserver, said
+    module would need a bit of a refactoring.
```
yannickperrenet, line 168: As per https://github.com/orchest/orchest/pull/872/commits/964dc631fad8394f583b94c5f85a0c31ce088919#r852198108

orchest-sdk/python/RELEASE-CHECKLIST.md
```diff
@@ -2,3 +2,4 @@
 
 - The version (in `_version.py`) is updated according to [SemVer](https://semver.org/)
 - The version compatibility table in its `README.md` and `__init__.py` are updated accordingly
+- No changes that could break existing jobs
```
yannickperrenet, line 5: Might be good to add "backwards compatibility" wording in here.

services/orchest-api/app/app/core/environments.py
yannickperrenet: > the rank is done (per partition) by tag

Then that is indeed what we want. Thank you for verifying and for the great example :heart: 

I like the example so much that I want to put it as a comment near the query (the uuid's can be shortened though to make it fit), e.g.:
```
project_uuid  | environment_uuid | tag | rank 
--------------+------------------+-----+------
 13006c56-... | 18b59993-...     |   1 |    1
 13006c56-... | c56ab762-...     |   2 |    1
 13006c56-... | c56ab762-...     |   1 |    2
 13006c56-... | e0af758d-...     |   3 |    1
 13006c56-... | e0af758d-...     |   2 |    2
 13006c56-... | e0af758d-...     |   1 |    3
```

yannickperrenet:
What happens if a user has other applications running in his/her Kubernetes cluster? Wouldn't we be deleting those Docker images from the nodes as well? I don't remember seeing any filtering for the `maintainer` label that all our images should have.

services/orchest-api/app/app/__init__.py
fruttasecca: That came to my mind but I opted for not increasing the scope of this PR, I think this can be covered in another PR as a dependency for notifications if you agree, given that notifications will likely require such a scheduler job as well.

fruttasecca:
>What happens if a user has other applications running in his/her Kubernetes cluster? Wouldn't we be deleting those Docker images from the nodes as well? I don't remember seeing any filtering for the maintainer label that all our images should have.

When it comes to images on node we are (manually) filtering by name, so that only environment images, custom jupyter images and orchest images are involved. On the node we already have a number of images that are not related to orchest, which are k8s control plane images, so any mistake here would have already likely come up by showing attempts to delete those. I'll try adding a filtering by maintaner as well, should be safer.

deploy/helm/templates/_cluster-role.tpl
nhaghighat: For the sake of simplicity I added * to to the resources.
@fruttasecca could you enlighten me, why orchest-api and celery worker needs exec?

nhaghighat: never-mind, for connecting to docker-registry and executing garbage collect

nhaghighat:
@fruttasecca : A general question regarding garbage collection

What have we done to eliminate this issue:
`You should ensure that the registry is in read-only mode or not running at all. If you were to upload an image while garbage collection is running, there is the risk that the image’s layers are mistakenly deleted leading to a corrupted image`
https://docs.docker.com/registry/garbage-collection/

services/orchest-api/app/app/__init__.py
yannickperrenet: When looking at the current `Scheduler` class, I thought it could just become a method on that class. Not a class in itself. 

If you have "bigger" ideas for scheduling then I would be fine moving that to another PR :+1:

fruttasecca: I think it would require a bit of refactoring to make it consistent and/or par with what we are doing in the webserver, I would prefer doing that in a separate PR to avoid making a release that carries a lot of baggage/risk

yannickperrenet: Sounds good! :+1:

services/orchest-api/app/app/core/registry.py
```diff
@@ -0,0 +1,171 @@
...
+    for its cache to avoid the following issue:
+    https://github.com/distribution/distribution/issues/1803, this
+    makes it possible to run GC without having to restart the registry
+    to clean the cache.
+
+    This function should be called only when having the certainty that
+    no images are being pushed to the registry to avoid race conditions.
+    This is currently accomplished by calling it from a celery task part
+    of the "builds" queue.
+    """
```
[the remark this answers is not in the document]
nhaghighat: I remember this discussion. my question is somehow different.
I'm thinking is it possible to garbage collect a layer while registry is persisting the same layer, in this situation we will end up in a corrupted layers on disk.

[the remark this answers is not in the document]
nhaghighat: Is is possible a `build task` which tries to push a image runs at the same time of another `build task` which tries to garbage collect?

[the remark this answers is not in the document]
fruttasecca: This task goes through the `builds` queue, that queue has a concurrency of one, which gives us a guarantee that no image pushes can be happening because the environment and jupyter image building tasks share the same `builds` queue. There are other factors that are currently limiting the concurrency of the `builds` queue (like the on disk cache of the builder, etc.). If (when) we will need to migrate away from a fixed concurrenvy level of 1 some kind of waiting should solve the situation, given that garbage collection happens relatively quickly.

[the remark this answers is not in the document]
nhaghighat: Thanks, the concurrency setting was the thing I was looking for

PR #873, opened 2022-04-15
Custom Jupyter Server Images Lifecycle

fruttasecca:
## Description

This PR introduces lifecycle management for custom jupyter images. Images that are not considered active will be deleted from the registry and on every node. An image is considered inactive and marked for deletion given the following (OR):
- its version is different from the current Orchest cluster version
- it's not the latest custom jupyter image, and it's digest is different from the latest active custom jupyter image

This PR introduces image versioning for jupyter images. While more lightweight approaches were available it's consistent with how we do environment images versioning and, later, disposal. This fixes an issue where newer custom jupyter images were not pulled on nodes due to the "IfNotPresent" `node-agent` pull policy, by versioning custom jupyter images no coordination when it comes to pull policy is required.

This PR contains schema changes, and it builds on top of https://github.com/orchest/orchest/pull/872.

## Checklist


services/orchest-api/app/app/models.py
```diff
@@ -305,12 +306,48 @@ class JupyterImageBuild(BaseModel):
...
+class JupyterImage(BaseModel):
+    __tablename__ = "jupyter_images"
+
+    # A new image record with a given tag will be created everytime a
+    # jupyter build is started, the tag only increments.
+    tag = db.Column(
+        db.Integer,
+        db.ForeignKey("jupyter_image_builds.image_tag", ondelete="CASCADE"),
+        primary_key=True,
+    )
```
yannickperrenet, line 324: Don't we create an `EnvironmentImage` only on `SUCCESS` of an `EnvironmentImageBuild`? Mentioning as I think it makes sense to be consistent with this for Jupyter builds.

yannickperrenet: Indeed done in the next commit. :+1:

services/orchest-api/app/app/utils.py
```diff
@@ -228,3 +231,54 @@ def get_jupyter_server_image_to_use() -> str:
         return f"{registry_ip}/{_config.JUPYTER_IMAGE_NAME}:{custom_image.tag}"
     else:
         return f"orchest/jupyter-server:{CONFIG_CLASS.ORCHEST_VERSION}"
+
+
+def mark_custom_jupyter_images_to_be_removed() -> str:
```
yannickperrenet, line 242: Return type should be `None` I guess.

services/node-agent/app/image_deleter.py
```diff
@@ -9,14 +9,26 @@
...
+def is_env_image(name: str) -> bool:
+    # Note: due to a k8s_todo, the docker client will return an image
+    # name having the form of: <registry ip>/image:tag
+    return "orchest-env" in name
+
+
+def is_custom_jupyter_image(name: str) -> bool:
+    # Should we add the internal lib to the node agent for this?
+    JUPYTER_IMAGE_NAME = "orchest-jupyter-server-user-configured"
+    return JUPYTER_IMAGE_NAME in name
```
yannickperrenet, line 29: To answer the question in the comment, I think we should default to adding the internal lib to all services except the ones licensed differently (i.e. the `orchest-sdk` and `orchest-cli`).

services/node-agent/app/image_deleter.py
```diff
@@ -27,18 +39,40 @@ async def get_active_environment_images(session: aiohttp.ClientSession) -> Set[s
...
-    """Gets the environment images on the node."""
+async def get_active_custom_jupyter_images(session: aiohttp.ClientSession) -> Set[str]:
+    """Gets the active custom jupyter images."""
+    endpoint = "http://orchest-api/api/ctl/orchest-images-to-pre-pull"
+    async with session.get(endpoint) as response:
+        response_json = await response.json()
+        pre_pull_images = response_json["pre_pull_images"]
+        active_images = [img for img in pre_pull_images if is_custom_jupyter_image(img)]
+    logger.info(f"Found the following active jupyter custom images: {active_images}")
+    return set(active_images)
```
yannickperrenet, line 50: Aren't we forgetting about jupyter images that the user has not customized? They have the format of `"orchest/jupyter-server:{CONFIG_CLASS.ORCHEST_VERSION}"`. 

Those images would have been pre-pulled to the nodes as well, but they can be outdated/marked for removal because the Orchest version has increased.

services/orchest-api/app/app/utils.py
fruttasecca: Indeed

services/orchest-api/app/app/__init__.py
```diff
@@ -282,15 +282,34 @@ def trigger_conditional_jupyter_image_build(app):
...
+        .filter(JupyterImageBuild.status.in_(["PENDING", "STARTED"]))
+        .exists()
+    ).scalar():
+        app.logger.info(
+            "Ongoing custom jupyter image build, no need to trigger a build."
+        )
         return
 
+    # Note: this is not race condition free in case of concurrent APIs
+    # restarting.
```
yannickperrenet, line 310: I guess here we are talking about having multiple `orchest-api` pods running?

yannickperrenet (approved):
LGTM! 

Mostly wondering about https://github.com/orchest/orchest/pull/873#discussion_r851993521, but other than that I can't think of anything else :+1:

services/node-agent/app/image_deleter.py
fruttasecca: I assumed that those images would fall in the same bucket as the images related to Orchest services, which deletion we aren't taking care of as of now, perhaps that could a PR on to of this?

services/orchest-api/app/app/__init__.py
fruttasecca: Indeed

services/node-agent/app/image_deleter.py
yannickperrenet: > same bucket as the images related to Orchest services

You are absolutely right. I was too tunnel visioned on the jupyter images. 

> which deletion we aren't taking care of as of now, perhaps that could a PR on to of this?

I think that would indeed be nice ;)

PR #874, opened 2022-04-18
Allow increasing jobs and interactive runs parallelism without restarting Orchest

fruttasecca:
Allows increasing parallelism levels for jobs and interactive runs without restarting Orchest. Slightly refactors the `OrchestConfig` class.

### DIFF
--- a/services/orchest-api/app/app/utils.py
+++ b/services/orchest-api/app/app/utils.py
@@ -229,13 +229,13 @@ def get_jupyter_server_image_to_use() -> str:
 
 
 def _set_celery_worker_parallelism_at_runtime(
-    worker: str, previous_parallelism: int, new_parallelism: int
+    worker: str, current_parallelism: int, new_parallelism: int
 ) -> bool:
     """Set the parallelism of a celery worker at runtime.
 
     Args:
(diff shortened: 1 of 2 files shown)


services/orchest-api/app/app/utils.py
```diff
@@ -227,6 +228,65 @@ def get_jupyter_server_image_to_use() -> str:
...
 
 
+def _set_celery_worker_parallelism_at_runtime(
+    worker: str, previous_parallelism: int, new_parallelism: int
+) -> bool:
+    """Set the parallelism of a celery worker at runtime.
+
+    Args:
+        worker: Name of the worker.
+        previous_parallelism: Current parallelism level.
```
yannickperrenet, line 238: Guess the variable is better to be called `current_parallelism`. Really like the usage of typing and the docstrings btw :smile:

yannickperrenet (approved):
Looks great! :muscle:

services/orchest-api/app/app/utils.py
fruttasecca: Indeed, naming is hard :P

PR #875, opened 2022-04-18
Automatically delete unused Orchest images from nodes

fruttasecca:
This PR introduces lifecycle management for orchest images. Images that are considered unused, due to their version not matching the current orchest deployment version, will be removed from nodes.

### DIFF
--- a/services/node-agent/Dockerfile
+++ b/services/node-agent/Dockerfile
@@ -12,3 +12,5 @@ RUN pip3 install -r requirements.txt
 COPY . ./
 
 CMD [ "python3", "./app/main.py" ]
+ARG ORCHEST_VERSION
+ENV ORCHEST_VERSION=${ORCHEST_VERSION}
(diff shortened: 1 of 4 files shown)


yannickperrenet:
Couldn't this potentially cause the update to take longer? We don't want the `node-agent` to be deleting e.g. the `orchest-api` before actually pulling it again. Would be good if it is to re-use already pulled layers since for the `orchest-api` many of the layers will remain unchanged.

It seems like the disk size gain is minimal while the possible overhead is greater. What do you think?

fruttasecca:
This is a very good point, we do have a few options:
- don't do this, i.e. cancel the PR
- add more coordination between the controller and the deployment of Orchest, i.e. start the node-agent after everything else has started
- delete an orchest image with an outdated version only if the image with the up to date version (i.e. version == node-agent version) exists on the node
- ...

Point 3 seems like a decent tradeoff, keen to know your opnion

PR #888, opened 2022-04-22
Orchest-api keeps track of job events

fruttasecca:
## Description

Adds minimal support for tracking events happening in the `orchest-api`, and it's part of preliminary work to support notifications.

Currently support event types:
```
                    name                    
--------------------------------------------
 project:cronjob:cancelled
 project:cronjob:created
 project:cronjob:deleted
 project:cronjob:failed
 project:cronjob:paused
 project:cronjob:run:failed
 project:cronjob:run:pipeline-run:cancelled
 project:cronjob:run:pipeline-run:created
 project:cronjob:run:pipeline-run:deleted
 project:cronjob:run:pipeline-run:failed
 project:cronjob:run:pipeline-run:started
 project:cronjob:run:pipeline-run:succeeded
 project:cronjob:run:started
 project:cronjob:run:succeeded
 project:cronjob:started
 project:cronjob:unpaused
 project:job:cancelled
 project:job:created
 project:job:deleted
 project:job:failed
 project:job:pipeline-run:cancelled
 project:job:pipeline-run:created
 project:job:pipeline-run:deleted
 project:job:pipeline-run:failed
 project:job:pipeline-run:started
 project:job:pipeline-run:succeeded
 project:job:started
 project:job:succeeded
 ```
Some remarks:
- a cronjob run is an instance of a cronjob being triggered, and represents a batch of runs
- event types are structured in a way akin to namespacing
- job events have been split up in `project:job:` and `project:cronjob:`

## Checklist

### CODE REVIEW

File: services/orchest-api/app/app/apis/namespace_jobs.py [hunk 1]
```diff
@@ -1298,35 +1312,43 @@ class UpdateJobPipelineRun(TwoPhaseFunction):
...
+
+        if update_status_db(
             status_update,
             model=models.NonInteractivePipelineRun,
             filter_by=filter_by,
-        )
+        ):
+            {
+                "SUCCESS": events.register_job_pipeline_run_succeeded,
+                "FAILED": events.register_job_pipeline_run_failed,
```
yannickperrenet, line 1349: Not sure, but I think this should be `FAILURE` (looking at the if-statement a few lines below).

services/orchest-api/app/app/core/events.py
```diff
@@ -30,24 +48,88 @@ def register_job_deleted(project_uuid: str, job_uuid: str) -> None:
...
+        _register_job_event("project:job:cancelled", project_uuid, job_uuid)
 
 
 def register_job_failed(project_uuid: str, job_uuid: str) -> None:
     """Adds a job failed event to the db, does not commit."""
-    _register_job_event("project:job:failed", project_uuid, job_uuid)
+    if _is_cronjob(job_uuid):
+        _register_cronjob_event("project:cronjob:failed", project_uuid, job_uuid)
+    else:
+        _register_job_event("project:job:failed", project_uuid, job_uuid)
```
yannickperrenet, line 70: I must say that the naming of the function slightly confuses me here. Almost everywhere the distinction is made between a `job` and a `cronjob`, so why does the `register_job...` suddenly take care of `job`s as well as `cronjob`s. As a user of those function it is now easy to get them confused and use the wrong one (without looking at the code of the function, which is of course the goal).

yannickperrenet: But maybe I am just missing why this is clear :see_no_evil:

yannickperrenet (approved):
I think it looks good :thinking: 

Although I wasn't too sure of the naming, in particular `project:job` and `project:cronjob`, it will suit us just fine (is my guess).

How did you do "add more types using a schema migration"?

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: Think you are right, will fix

services/orchest-api/app/app/core/events.py
fruttasecca: You are right, essentially my idea was to avoid higher-level code having to distinguish between job and cronjob although it ends up a bit of confusing as you mention. The confusion (imo) stems from the fact that we are using `job` to name a "normal" job and cronjob for cronjobs. This is definitively a pain point, I guess something should change, although I am not sure if we want to introduce more terms into our glossary.

fruttasecca:
>Although I wasn't too sure of the naming, in particular project:job and project:cronjob, it will suit us just fine (is my guess).

I agree that there is a bit of an impedance mismatch when it comes to our taxonomy, I would be happy if we could solve this once and for all here, although I have the feeling we are in a local optimum

>How did you do "add more types using a schema migration"?

Add an empty revision, e.g. `bash scripts/migration_manager.sh orchest-api revision`, add the statements to add more types to the `event_types` table to the revision

services/orchest-api/app/app/core/events.py
yannickperrenet: > The confusion (imo) stems from the fact that we are using job to name a "normal" job and cronjob for cronjobs. This is definitively a pain point, I guess something should change, although I am not sure if we want to introduce more terms into our glossary.

Totally agree with this point. Why can't we use "one-off-job" and "cron-job"? We can also do a quick brainstorm session :+1:

yannickperrenet:
> I have the feeling we are in a local optimum

I share the feeling, although we haven't thought about naming in quite some time. 

> Add an empty revision, e.g. bash scripts/migration_manager.sh orchest-api revision, add the statements to add more types to the event_types table to the revision

I thought so, could you add this as part of the docstring?

services/orchest-api/app/app/core/events.py
fruttasecca: > Why can't we use "one-off-job" and "cron-job"?

That would be okay with me, I was wary of adding more to our taxonomy and glossary since these events are user-facing, let's indeed have a chat

fruttasecca:
>I thought so, could you add this as part of the docstring?

Will do!

services/orchest-api/app/app/core/events.py
fruttasecca: After an internal discussion we decided to go with `one-off-job`

PR #893, opened 2022-04-25
Add tests for Pipeline Settings

iannbing:
## Description

This PR adds tests for `useFetchPipelineSettings`, especially for two bugs that appear multiple times:

- if user has not yet save the changes they made (e.g. Services), the changes should remain after the browser tab regains  focus.
- if user change the pipeline settings from other places (e.g. change pipeline file path in Pipeline Editor), PipelineSettings should load the latest settings instead of the cached settings.

The fake data for API mocking are dynamically generated (see `mockProjects.mock.tsx`), which can be re-used per test cases in the future.

## Checklist


services/orchest-webserver/client/src/contexts/AppContext.tsx
```diff
@@ -439,22 +438,20 @@ export const AppContextProvider: React.FC = ({ children }) => {
   );
 
   return (
-    <IntercomProvider appId={state.config?.INTERCOM_APP_ID || ""}>
```
iannbing, line 442: Move out the third-party component `IntercomProvider` so we can isolate `AppContext` for mocking / testing.

yannickperrenet (approved):
LGTM!

Taking a 100-feet view here, but what would you guess the maintain burden is on these tests? E.g. we add an additional entry to one of the endpoints, then the mocks have to be updated.

iannbing:
> Taking a 100-feet view here, but what would you guess the maintain burden is on these tests? E.g. we add an additional entry to one of the endpoints, then the mocks have to be updated.

If you're referring to the mock API's, we generate mock data dynamically based on the Typescript type definitions. This means that if the schema of a model changed in the BE, we should do the following:

- update Typescript type definition
- fix all the type error in the mock data generators.
- run the tests to see which functional components are broken due to these changes.
 
If you're asking about the maintenance cost of the whole tests, I would say this kind of tests give use way higher ROI comparing to Cypress.

Hope it answers your question.

PR #897, opened 2022-04-26
Apply useCancelablePromise and useCancelableFetch

iannbing:
## Description

Combine the `makeCancelable` wrapper and `React.useEffect` as `useCancelablePromise`, which cancels the promise when unmounted automatically.

This PR also improves the usage of OverflowListener: call `unobserve` when unmount.

## Explanation

Cancelable fetch cannot be used in all cases. `useCancelablePromise` is a replacement of the old `PromiseManager`.
How you should use `PromiseManager`:
```ts
const promiseManager = React.useMemo(() => new PromiseManager(), [])

promiseManager.appendCancelablePromise(cancelablePromise);

let cancelablePromise = makeCancelable(
      fetch("/async/something"),
      promiseManager
    );

    fetchProjectsPromise.promise
      .then(() => {
        //....do things
      })

React.useEffect(() => {
    return () => promiseManager.cancelCancelablePromises();
}, []);
```
how you should use `useCancelablePromise`:
```ts
const { makeCancelable } = useCancelablePromise()

makeCancelable(fetch("/async/something"))
.then(() => {
 // do things...
})
```
Under the hood, they're the same.  `useCancelablePromise`  has a built-in cancellation effect, so no need to repeat it everywhere (making the code DRYer).

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/hooks/useCancellablePromise.ts
+++ b/services/orchest-webserver/client/src/hooks/useCancellablePromise.ts
@@ -1,0 +1,53 @@
+import { fetcher } from "@orchest/lib-utils";
+import React from "react";
+
+type CancelablePromise<T> = {
+  promise: Promise<T>;
+  cancel: () => void;
+};
+
+// https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html
+function makeCancelable<T>(promise: Promise<T>) {
+  let isCanceled = false;
+  const wrappedPromise = new Promise<T>((resolve, reject) => {
+    promise
+      .then((val) => (isCanceled ? reject({ isCanceled }) : resolve(val)))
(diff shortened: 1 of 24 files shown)


services/orchest-webserver/client/src/hooks/useCancellablePromise.ts
```diff
@@ -0,0 +1,53 @@
...
+  const makeCancellable = React.useCallback(function <T>(p: Promise<T>) {
+    const cPromise = makeCancelable(p);
+    cancelablePromises.current.push(cPromise);
+    return cPromise.promise;
+  }, []);
+
+  return { makeCancellable };
+}
+
+export function useCancalableFetch() {
```
fruttasecca: typo

yannickperrenet (approved):

PR #902, opened 2022-04-27
Add Webhooks logic to `orchest-api`

fruttasecca:
## Description

This is another PR for notifications, it adds logic for webhooks. A couple of things that need to be done:
- an `orchest-api` scheduler and/or celery task that periodically checks for deliveries to sent
- the webhook payload is currently very minimal, it's essentially a dump of the webhook config and the event itself, I had hoped to expand on this today but we will likely need to discuss internally tomorrow what kind of content we expect for the payload, depending on the event type

## Checklist


yannickperrenet (approved):
:fire:

PR #904, opened 2022-04-28
Add recurring task to process deliveries

fruttasecca:
## Description

Part of notifications


yannickperrenet (approved):
Looks good to me!

yannickperrenet:
> Do we want to delete deliveries that have been delivered?

Given that I don't see any use for these types of events after the time they have been sent, I would be okay with that. I mean, I only want to get the notification right when the job fails, I don't see myself looking at the long history of events ever again (for jobs you can even find your failed pipeline runs in its history).

PR #905, opened 2022-04-28
Expand webhook payload content

fruttasecca:
## Description

Part of of the notifications feature. Expands the content of webhook payloads, generally speaking, a payload event will carry the following information if applicable:
- project name and uuid
- job name, pipeline_name, uuid, next_scheduled_time, schedule, status (different properties based on cron-job vs one-off-job)
- cron-job run: run number, status, total number of runs
- job pipeline run: uuid, status, parameters, failed steps if any

Example payload:
```
 {
   "delivered_for": {
     "content_type": "application/json",
    "name": "string",
     "subscriptions": [
       {
         "event_type": "project:one-off-job:created",
         "job_uuid": null,
         "project_uuid": null,
         "subscriber_uuid": "d2dafc8b-fb5e-4c74-af3f-1d02643f3dad",
         "uuid": "b111fd12-6535-4237-b090-da56f27c9db7"
       },
       {
         "event_type": "project:cron-job:run:started",
         "job_uuid": null,
         "project_uuid": null,
         "subscriber_uuid": "d2dafc8b-fb5e-4c74-af3f-1d02643f3dad",
         "uuid": "b111fd12-6535-4237-b090-da56f27c9db6"
       }
     ],
     "type": "webhook",
     "url": "<redacted>",
     "uuid": "d2dafc8b-fb5e-4c74-af3f-1d02643f3dad",
     "verify_ssl": true
   },
   "event": {
     "job": {
       "name": "aa",
       "next_scheduled_time": "2022-04-28 12:55:00+00:00",
       "pipeline_name": "california-housing",
       "run": {
         "number": 83,
         "status": "STARTED",
         "total_pipeline_runs": 1
       },
       "schedule": "* * * * *",
       "status": "STARTED",
       "uuid": "c27e4daa-96ff-4fa8-8cf0-cedb92675b36"
     },
     "project": {
       "name": "Project",
       "uuid": "d4e49111-ce8d-4872-af3e-77e9431bd16e"
     },
     "timestamp": "2022-04-28 12:54:02.163011+00:00",
     "type": "project:cron-job:run:started",
     "uuid": "aac930ab-afc0-4346-8219-c0b4fd2a5ea4"
   }
 }

```

## Checklist

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_notifications.py
+++ b/services/orchest-api/app/app/apis/namespace_notifications.py
@@ -118,6 +118,9 @@ class SendSubscriberTestPingDelivery(Resource):
         ):
             return {"message": "success"}, 200
         else:
+            if response is not None:
+                logger.info(response.status_code)
+                logger.info(response.text)
             return {"message": "failure"}, 500
 
 
--- a/services/orchest-api/app/app/apis/namespace_notifications.py
+++ b/services/orchest-api/app/app/apis/namespace_notifications.py
@@ -29,7 +29,7 @@ class SubscriberList(Resource):
     @api.doc("get_subscribers")
     @api.response(200, "Success", schema.subscribers)
     def get(self):
-        """Gets all subscribers, without subscriptions."""
+        """Gets all subscribers, doesn't include their subscriptions."""
         subscribers = models.Subscriber.query.options(
             noload(models.Subscriber.subscriptions)
         ).all()
@@ -46,7 +46,7 @@ class SubscriberList(Resource):
 class WebhookList(Resource):
     @api.doc("create_webhook")
     @api.expect(schema.webhook_spec, validate=True)
-    @api.response(200, "Success", schema.webhook_with_secret)
+    @api.response(201, "Success", schema.webhook_with_secret)
     def post(self):
         """Creates a webhook with the given subscriptions.
 
--- a/services/orchest-api/app/app/apis/namespace_notifications.py
+++ b/services/orchest-api/app/app/apis/namespace_notifications.py
@@ -118,7 +118,7 @@ class SendSubscriberTestPingDelivery(Resource):
         ):
             return {"message": "success"}, 200
         else:
-            return {"message": "success"}, 500
+            return {"message": "failure"}, 500
 
 
 @api.route("/subscribers/subscribed-to/<string:event_type>")
--- a/services/orchest-api/app/app/apis/namespace_notifications.py
+++ b/services/orchest-api/app/app/apis/namespace_notifications.py
@@ -94,6 +94,33 @@ class Subscriber(Resource):
         return {"message": ""}, 201
 
 
+@api.route("/subscribers/test-ping-delivery/<string:uuid>")
+class SendSubscriberTestPingDelivery(Resource):
+    @api.doc("subscribers/test-ping-delivery")
+    @api.response(200, "Success")
+    @api.response(500, "Failure")
+    def get(self, uuid: str):
+        """Send a test ping delivery to the subscriber.
+
+        This endpoint allows to send a ping event notifications to the
+        subscriber, so that it's possible to test if a given webhook
+        is working end to end, i.e. the deliveree is reachable.
+
+        The endpoint will return a 200 if the response obtained from the
+        deliveree is to be considered successfull, 500 otherwise.
+
+        """
+        response = webhooks.send_test_ping_delivery(uuid)
+        if (
+            response is not None
+            and response.status_code >= 200
+            and response.status_code <= 299
+        ):
+            return {"message": "success"}, 200
+        else:
+            return {"message": "success"}, 500
+
+
 @api.route("/subscribers/subscribed-to/<string:event_type>")
 class SubscribersSubscribedToEvent(Resource):
     @api.doc("get_subscribers_subscribed_to_event")
--- a/services/orchest-api/app/app/apis/namespace_projects.py
+++ b/services/orchest-api/app/app/apis/namespace_projects.py
@@ -40,7 +40,12 @@ class ProjectList(Resource):
         project = request.get_json()
 
         if len(project["name"]) > 255:
-            return {}, 400
+            return {
+                "message": (
+                    "The provided project name exceeds the maximum length of 255 "
+                    "characters."
+                )
+            }, 400
 
         project["env_variables"] = project.get("env_variables", {})
         if not _utils.are_environment_variables_valid(project["env_variables"]):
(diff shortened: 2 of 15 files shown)


yannickperrenet:
Unsure whether we want to include all the `subscriptions` in the payload. Although we could just include them for the time being until someone mentions the payload is to large (e.g. having many many subscriptions). I generally like defaulting to more verbosity :+1:

services/orchest-api/app/app/models.py
```diff
@@ -1177,13 +1213,80 @@ class OneOffJobEvent(JobEvent):
...
+def _prepare_parameters_payload(
+    pipeline_definition: dict, run_parameters: dict
+) -> dict:
+    parameters_payload = {}
+    for k, v in run_parameters.items():
+        if k == "pipeline_parameters":
+            parameters_payload[k] = v
+        else:
+            step_name = pipeline_definition.get("steps").get(k, {}).get("title")
+            parameters_payload[f"step-{step_name}-{k}"] = v
```
yannickperrenet, line 1256: Might want to handle the `None` cases differently, e.g. `step_name` could be `None` here.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -94,6 +94,33 @@ def delete(self, uuid: str):
...
+        """
+        response = webhooks.send_test_ping_delivery(uuid)
+        if (
+            response is not None
+            and response.status_code >= 200
+            and response.status_code <= 299
+        ):
+            return {"message": "success"}, 200
+        else:
+            return {"message": "success"}, 500
```
yannickperrenet: `"success"` -> `"failure"`

services/orchest-api/app/app/apis/namespace_projects.py
```diff
@@ -38,6 +38,10 @@ def get(self):
     def post(self):
         """Create a new project."""
         project = request.get_json()
+
+        if len(project["name"]) > 255:
+            return {}, 400
```
yannickperrenet, line 43: Probably a good idea to include a failure message in the response, e.g. `{"message": "Provided project name exceeds maximum length of 254 characters."}`

services/orchest-api/app/app/apis/namespace_notifications.py
yannickperrenet: Fixed in https://github.com/orchest/orchest/pull/905/commits/e74be70fafefd95267b05890b4726c89305e1d76 :+1:

yannickperrenet:
The entry `"url": "<redacted>"` in the payload, does that refer to the URL of the webhook? Because in case of job events, I think we should add a URL entry linking to the page (`orchest-api` has the `ORCHEST_FQDN`). When setting up a Slack webhook you probably want to be able to quickly open the page in Orchest.

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -85,6 +85,10 @@ def _post_process_payload(payload: dict, webhook: models.Webhook) -> None:
 
     if webhook.is_slack_webhook():
         payload["text"] = json.dumps(payload, sort_keys=True, indent=1)
+    elif webhook.is_discord_webhook():
+        payload["content"] = json.dumps(payload, sort_keys=True, indent=1)
+        # 2000 max allowed characters.
+        payload["content"] = f'{payload["content"][:1997]}...'
```
yannickperrenet, line 91: Instead of sorting and taking the first 1997 characters, how about cherry picking the important items of the payload and send a minimized version (that is still capped at 1997 chars of course)? Yes, the "importance" of items is subjective but I do think we can agree on including on certain items being more important than others.

What do you think? I am not overly convinced yet by going for a special case treatment of discord.

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -232,4 +232,5 @@ def deliver(delivery_uuid: str) -> None:
             delivery.reschedule()
             logger.info(f"Rescheduling {delivery_uuid} at {delivery.scheduled_at}.")
 
+    db.session.delete(delivery)
```
yannickperrenet, line 235: This should probably run in the `else` clause as part of the `try` `except`. Otherwise the `delivery` is deleted, leading to it not existing for the reschedule run.

yannickperrenet (approved):
Another stellar job! :muscle:

services/orchest-api/app/app/core/notifications/webhooks.py
fruttasecca: Indeed, good catch!

services/orchest-api/app/app/apis/namespace_projects.py
fruttasecca: I was banking on the fact that the webserver provides better error messages but adding it here as well won't hurt :).

services/orchest-api/app/app/core/notifications/webhooks.py
fruttasecca: I'll take a look at it for improvements

services/orchest-api/app/app/models.py
fruttasecca: Indeed `step-None-\<uuid\>` isn't really super nice UX, will improve the naming in this case

fruttasecca:
>The entry "url": "<redacted>" in the payload, does that refer to the URL of the webhook? Because in case of job events, I think we should add a URL entry linking to the page (orchest-api has the ORCHEST_FQDN). When setting up a Slack webhook you probably want to be able to quickly open the page in Orchest.

Yes that's the URL of the webhook, which can contain secrets, in a more recent PR I have simply removed the entry from the dictionary instead of going with `redacted`,

>Because in case of job events, I think we should add a URL entry linking to the page (orchest-api has the ORCHEST_FQDN)

Does it? I don't think it always does (FQDN has been made optional AFAIR) and even when defined it doesn't mean that the user will be able to reach the instance through the FQDN. I remember a self hosting user having a setup like this, where the FQDN was defined but not actually used to reach the instance. Given the timing of the controller and the later port to of Orchest cloud to k8s I think we might want to hold on and make it a tiny change in a future PR in a few days, at least the FQDN passing part, perhaps we can just include the path of the URL for now?

yannickperrenet:
> Yes that's the URL of the webhook, which can contain secrets, in a more recent PR I have simply removed the entry from the dictionary instead of going with redacted

I saw :+1: 

> Given the timing of the controller and the later port to of Orchest cloud to k8s I think we might want to hold on and make it a tiny change in a future PR in a few days, at least the FQDN passing part, perhaps we can just include the path of the URL for now?

Good points, providing a 100% correct URL is rather hard. Let's just include the path of the URL for now, I think that already helps users somewhat.

PR #892, opened 2022-04-25
Add project-not-found error

iannbing:
## Description

Currently Orechest switches to an existing project if the given project UUID is invalid. This PR adds an alert if the said case occurs.

Screenshot 2022-04-25 at 13 37 10

## Checklist
- [x] The PR branch is set up to merge into `dev` instead of `master`.

PR #911, opened 2022-04-29
Only redirect if current projectUuid is invalid

iannbing:
## Description

A bug was introduced when implemented #892, ProjectSelector attempts to sync `state.projectUuid` and `project_uuid` in the query args. This is to ensure that redirect user to a proper view when the given query args are no longer valid. For example, if user intends to visit /environment with an invalid environment_uuid, ProjectSelector will redirect user to /environments, so user can start from there. 

The bug is that this redirection occurs at all times, so it forces a redirection to /environments regardless.
 
This PR fixes this issue: only redirect if either condition is met:
1. state.projectUuid is undefined: user is landing on the app. 
2. current query arg `project_uuid` is different from the calculated project_uuid. So it won't trigger chaining redirections.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/components/ProjectSelector.tsx
+++ b/services/orchest-webserver/client/src/components/ProjectSelector.tsx
@@ -133,12 +133,20 @@ export const ProjectSelector = () => {
         ? projectUuidFromRoute
         : state.projects[0]?.uuid;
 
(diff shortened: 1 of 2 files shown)


iannbing:
This bug has been fixed another PR #903. Closing.

Issue #906, opened 2022-04-28
Security Vulnerability Found


porcupineyhairs:
# Absolute Path Traversal due to incorrect use of `send_file` call

A path traversal attack (also known as directory traversal) aims to access files and directories that are stored outside the web root folder. By manipulating variables that reference files with “dot-dot-slash (../)” sequences and its variations or by using absolute file paths, it may be possible to access arbitrary files and directories stored on file system including application source code or configuration and critical system files. This attack is also known as “dot-dot-slash”, “directory traversal”, “directory climbing” and “backtracking”.

## Root Cause Analysis

The `os.path.join` call is unsafe for use with untrusted input. When the `os.path.join` call encounters an absolute path, it ignores all the parameters it has encountered till that point and starts working with the new absolute path.  Please see the example below.
```
>>> import os.path
>>> static = "path/to/mySafeStaticDir"
>>> malicious = "/../../../../../etc/passwd"
>>> os.path.join(t,malicious)
'/../../../../../etc/passwd'
```
Since the "malicious" parameter represents an absolute path, the result of `os.path.join` ignores the static directory completely. Hence, untrusted input is passed via the `os.path.join` call to `flask.send_file` can lead to path traversal attacks. 

In this case, the problems occurs due to the following code :
https://github.com/orchest/orchest/blob/c0fa7cde83de1bd5ceb87855dcd36dba738d21a0/services/orchest-webserver/app/app/views/views.py#L1170

Here, the `path` parameter is attacker controlled. This parameter passes through the unsafe `os.path.join` call making the effective directory and filename passed to the `send_file` call attacker controlled. This leads to a path traversal attack. 

## Proof of Concept

The bug can be verified using a proof of concept similar to the one shown below.

```
curl --path-as-is 'http://<domain>/async/file-management/download?root=/userdir&project_uuid=asd&path=../../../../../../../etc/passwd"'
```
## Remediation

This can be fixed by preventing flow of untrusted data to the vulnerable `send_file` function. In case the application logic necessiates this behaviour, one can either use the `flask.safe_join` to join untrusted paths or replace `flask.send_file` calls with `flask.send_from_directory` calls.

## References
* [OWASP Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
* github/securitylab#669

PR #907, opened 2022-04-28
Fix Path Traversal Vulnerability

porcupineyhairs:
Fixes #906

PR #913, opened 2022-04-29
Fix remove unsafe path joins

fruttasecca:
Related: https://github.com/orchest/orchest/pull/907

Fixes: https://github.com/orchest/orchest/issues/906

Credits to @porcupineyhairs for discovery and fix proposal

There are a couple of regressions on `dev` that make it not possible to test all edge cases, it's something we will have to look into next week.

### DIFF
--- a/services/orchest-webserver/app/app/core/projects.py
+++ b/services/orchest-webserver/app/app/core/projects.py
@@ -6,7 +6,7 @@ import uuid
 from typing import Optional
 
 import requests
-from flask import current_app
+from flask import current_app, safe_join
 
 from _orchest.internals.two_phase_executor import TwoPhaseExecutor, TwoPhaseFunction
 from _orchest.internals.utils import rmtree
@@ -64,9 +64,7 @@ class CreateProject(TwoPhaseFunction):
             FileExistsError:
             NotADirectoryError:
         """
-        full_project_path = os.path.join(
-            current_app.config["PROJECTS_DIR"], project_path
-        )
+        full_project_path = safe_join(current_app.config["PROJECTS_DIR"], project_path)
         # exist_ok=True is there so that this function can be used both
         # when initializing a project that was discovered through the
         # filesystem or initializing a project from scratch.
@@ -74,14 +72,14 @@ class CreateProject(TwoPhaseFunction):
 
         # Create top-level `.gitignore` file with sane defaults, in case
         # it not already exists.
-        root_gitignore = os.path.join(full_project_path, ".gitignore")
+        root_gitignore = safe_join(full_project_path, ".gitignore")
         if not os.path.exists(root_gitignore):
             with open(root_gitignore, "w") as f:
                 f.write("\n".join(current_app.config["GIT_IGNORE_PROJECT_ROOT"]))
 
         # This would actually be created as a collateral effect when
         # populating with default environments, do not rely on that.
-        expected_internal_dir = os.path.join(full_project_path, ".orchest")
+        expected_internal_dir = safe_join(full_project_path, ".orchest")
         if os.path.isfile(expected_internal_dir):
             raise NotADirectoryError(
                 "The expected internal directory (.orchest) is a file."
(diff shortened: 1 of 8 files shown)


services/orchest-webserver/app/app/core/projects.py
```diff
@@ -94,7 +92,7 @@ def _collateral(self, project_uuid: str, project_path: str):
         # `.orchest/` directory because an existing project might be
         # added to Orchest and already contain a root-level
         # `.gitignore`, which we don't want to inject ourselves in.
-        expected_git_ignore_file = os.path.join(
+        expected_git_ignore_file = safe_join(
```
yannickperrenet, line 95: Have you tested this? 

Given that Flask's `safe_join` only takes two arguments and in later versions (see https://github.com/pallets/flask/commit/b7501776a12c15e64b6cdac1e3b903a00df7d285) uses Werkzeug's `safe_join` that does take multiple arguments, and us running `Flask=1.1.2`, I don't think this will actually work.

services/orchest-webserver/app/app/core/projects.py
```diff
@@ -107,7 +105,7 @@ def _collateral(self, project_uuid: str, project_path: str):
 
         # Initialize with default environments only if the project has
         # no environments directory.
-        expected_env_dir = os.path.join(full_project_path, ".orchest", "environments")
+        expected_env_dir = safe_join(full_project_path, ".orchest", "environments")
```
yannickperrenet, line 108: Here are also more than 2 arguments, so need to verify. As per https://github.com/orchest/orchest/pull/913/files#r861897137

services/orchest-webserver/app/app/core/projects.py
fruttasecca: Yes I have tested every change AFAIK, I've also just rebuilt the webserver image and run the following in a python repl inside it
```
>>> import flask
>>> flask.__version__
'1.1.2'
>>> from flask import safe_join
>>> safe_join("1", "2", "3")
'1/2/3'
>>>
```

yannickperrenet (approved):
:100: 

Thank you @porcupineyhairs!

services/orchest-webserver/app/app/utils.py
```diff
@@ -629,7 +623,7 @@ def pipeline_set_notebook_kernels(pipeline_json, pipeline_directory, project_uui
 
         if "ipynb" == step["file_path"].split(".")[-1]:
 
-            notebook_path = os.path.join(pipeline_directory, step["file_path"])
+            notebook_path = safe_join(pipeline_directory, step["file_path"])
```
ricklamers, line 627: I expect this to cause problems because file paths can be relative and point to files above the pipeline directory. We should find all similar resolves of paths that use the pipeline step file path in particular.

fruttasecca: Good catch, will take a look

services/orchest-webserver/app/app/utils.py
```diff
@@ -749,6 +751,23 @@ def is_valid_project_relative_path(project_uuid, path: str) -> str:
...
+    )
+    pipeline_path = get_pipeline_path(pipeline_uuid, project_uuid)
+    abs_path = os.path.abspath(
+        os.path.normpath(os.path.join(os.path.split(pipeline_path)[0], path))
+    )
+    return abs_path.startswith(project_path)
+
+
+def is_valid_data_path(path: str):
+    return os.path.abspath(os.path.normpath(path)).startswith("/data")
```
iannbing, line 768: @fruttasecca  Isn't this line always returning `false`? `os.path.abspath(os.path.normpath(path))` most likely would start with `/userdir`?

iannbing: I propose changing it to https://github.com/orchest/orchest/blob/6f66315c05b8b9a3c2061c59ab7279c4510125c0/services/orchest-webserver/app/app/utils.py#L768 in PR #914. Please review the change and discuss in the said PR. :-)

PR #914, opened 2022-05-02
Fix: allow assigning files in the `/data/` folder to pipeline steps

iannbing:
## Description

Previously, only the project files can be assigned to pipeline steps. To encourage reusing generic scripts across projects, this PR allows user assign files in the `/data` folder to pipeline steps.

This PR also fixes the following bugs:

- When user uses the dropdown of the file picker to specify the file for pipeline steps, the value is the path should be relative to the pipeline file, instead of the project root.
- When user moves `.orchest` files from `Project files` to `/data`, it should not throw errors and the corresponding pipeline should be removed (i.e. disappear from the sessions panel).
- When user moves `.orchest` files from `/data` to `Project files`, the `.orchest` files should be recognized and added as pipelines (i.e. visible in the sessions panel).

@fruttasecca  Tag you here because there are some minor BE code changes.

## Checklist

### DIFF
--- a/services/orchest-webserver/app/app/views/views.py
+++ b/services/orchest-webserver/app/app/views/views.py
@@ -844,12 +844,12 @@ def register_views(app, db):
 
             # Normalize relative paths.
             for step in pipeline_json["steps"].values():
-                if not is_valid_pipeline_relative_path(
-                    project_uuid, pipeline_uuid, step["file_path"]
-                ):
-                    raise app_error.OutOfProjectError(
-                        "Step path points outside of the project directory."
-                    )
+                extensions = list(
+                    map(lambda x: f".{x.lower()}", ["ipynb", "py", "R", "sh", "jl"])
+                )
+
+                if not step["file_path"].endswith(tuple(extensions)):
+                    raise ValueError("Unsupported pipeline step file type.")
 
                 if not step["file_path"].startswith("/"):
                     step["file_path"] = normalize_project_relative_path(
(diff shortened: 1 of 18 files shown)


services/orchest-webserver/app/app/views/views.py
```diff
@@ -844,12 +844,12 @@ def pipelines_json(project_uuid, pipeline_uuid):
...
-                ):
-                    raise app_error.OutOfProjectError(
-                        "Step path points outside of the project directory."
-                    )
+                extensions = list(
+                    map(lambda x: f".{x.lower()}", ["ipynb", "py", "R", "sh", "jl"])
+                )
+
+                if not step["file_path"].endswith(tuple(extensions)):
+                    raise ValueError("Unsupported pipeline step file type.")
```
fruttasecca, line 852: Was `is_valid_pipeline_relative_path` leading to any issues? :thinking: That's needed to avoid file system traversal exploits

fruttasecca: Modifying `is_valid_pipeline_relative_path` is probably the right call assuming we ar supporting files in `/data` as steps

fruttasecca:
I wouldn't have bet on the fact that interactive runs and job runs would "just work" with steps coming from `/data`, is it just working? :thinking:

services/orchest-webserver/app/app/views/views.py
iannbing: Added the check.

fruttasecca (approved):
lgtm aside from a minor fix related to pipeline relative paths checks

iannbing:
> I wouldn't have bet on the fact that interactive runs and job runs would "just work" with steps coming from /data, is it just working? 🤔

I did a "hello world" testing by creating a `py` file with `print("something")`, and it just works. Not sure about complex cases though.

Issue #866, opened 2022-04-08
K8s installation into pre-existing namespace?


HWiese1980:
I'd like to be able to install Orchest into an existing K8s cluster and an already existing namespace. We use OpenShift and have a pre-defined OpenShift project -> K8s namespace which I'd like to deploy and run Orchest in.

I think the current limitation to Minikube and a fixed K8s namespace is clunky and makes things more complicated than necessary. It helps to install Orchest locally (on Mac and Linux; Windows not so much). However, it does not help to set up Orchest in a quasi-productive environment.

ricklamers:
@HWiese1980 you’re absolutely right. You caught us early. We’re in the process of wrapping up our operator that will allow for more flexible/idiomatic deployments. I’ll make a note of making it possible to define the namespace you want to deploy in.

@nhaghighat please make note ✌🏻

nhaghighat:
@HWiese1980 : Thanks for pointing this out. As @ricklamers mentioned we are in the process of developing orchest controller (operator), which manages `OrchestCluster` CRD. `OrchestCluster` CRD will be a namespaced resource and operator will install orchest in the namespace of the `OrchestCluster` CRD.

mysticaltech:
Glad to hear about the operator. That is the ideal way of managing/installing such a multiple-moving-parts system.

HWiese1980:
Oh yeah, that sounds awesome, looking forward to it! I've managed to run Orchest in a local minikube cluster on a MacBook so far, however I've had some issues with it in minikube on Windows and in an OpenShift namespace.

yannickperrenet:
I'd like to give an update when it comes to installing Orchest in a custom namespace. 

With #966 it becomes possible to install Orchest in a custom namespace but only on a clean cluster (so nothing else is installed on it), installation can be done with:
```bash
pip install orchest-cli
orchest install --namespace=mynamespace
```
where the namespace will default to `orchest` if not specified.

To make it possible to install Orchest in a custom namespace on an existing cluster we need to prevent collisions between third-party software that we install on the cluster (e.g. Argo) and the software already installed on the cluster, or we need a way to deal with those collisions. Essentially, if Argo is already installed then it has created cluster-level resources which we would (in the current state of things) try to create as well. The next question to be answered is "who manages Argo now?"

Not sure exactly what will be the timeline on this, but we are working on it ;)

@nhaghighat Please feel free to correct me if I missed anything or didn't explain it properly.

PR #916, opened 2022-05-04
Introduce Orchest Kubernetes Operator

yannickperrenet:
## Description

This PR introduces the `orchest-controller` and the CRD that respectively manage and define `orchestclusters` resources. 

Big step towards a better solution for #866 and improving the management of the Orchest Cluster, e.g. installing it.

#### Deploying the `orchest-controller`
1. Create the Orchest namespace, e.g. `kubectl create ns orchest`. The namespace is required, given that the operator will be deployed in it.
2. Build the controller's `Dockerfile` on the `minikube` node, using:
    ```
    scripts/build_container.sh -v -i orchest-controller -t "latest"
    ``` 
    (It is listed in the below Todo section to migrate away from the `"latest"` tag.)
3. Deploy the `orchest-controller` and CRD:
   ```
   kubectl -f apply services/orchest-controller/deploy-controller
   ```
   
**Deploying an updated `orchest-controller`** (e.g. after changes):
1. Remove the CR Object, which will also remove the corresponding row in etcd, so that we can start from scratch when deploying a new controller. Go to `orchestcluster` in k9s, or using `kubectl`: `kubectl -n orchest delete orchestcluster cluster-1`
2. Build the new `orchest-controller`
3. Remove the `orchest-controller` pod. This will immediately create a new pod (since it is a deployment)

#### Setting up the CLI

1. Install the `orchest-cli` in its own virtualenv
    ```
    python3 -m venv venv && source venv/bin/active
    # Make sure you are in the `orchest-cli/` directory 
    pip install -e .
    ```
2. You should now be able to run commands like `orchest install`.

#### Set up Orchest for testing

You will need to build the minimal set of containers required to run Orchest with the tag `v2022.04.0` (given that that is currently the version that `orchest install` will get you):
```
eval $(minikube docker-env)
scripts/build_container.sh -m -t "v2022.04.0" -o "v2022.04.0"
```

Now you can run `orchest install` followed by any other command you would like to test.

**Testing `orchest update`**:
1. Build the images with another tag (this should be super quick given you have already build them before). I recommend using `v2022.04.1` so you can test update multiple times
2. Run `orchest update --version=v2022.04.1` (otherwise it will get the latest available version from our update-info endpoint)

## Todo
Here follows an extensive list of items that need to be done before we can merge:
- [ ] Merge in `dev` with needed DB migration fixes.
- [ ] Check with `orchest-ctl` whether we have all the functionality we need.
- [ ] GitHub Action workflow files to publish CLI on release.
- [ ] Release checklist similar to SDK to ensure the `_version.py` is updated. Needs to be included
  in the PR template as well.
- [ ] Installation docs (and how to self-host as you might need nginx reverse-proxy)
  - [ ] We will recommend to install Orchest on a clean cluster (a non-existing cluster) because it is hard to play well with other software already installed on the cluster, e.g. ingress, argo, etc.
  - [ ] Open an issue to fix the above point.
  - [ ] Show usage of CLI and direct yaml apply. The controller will get a service and ingress rule
        as well.
- [ ] `update` and `restart` to work in the GUI
  - [ ] `orchest-controller` needs to get service and ingress rules so that its endpoints is reachable from outside the cluster. These files are to be added to the `deploy-controller` directory.
- [ ] Seems to be an issue with `dev` mode? The orchest-api doesn't start correctly and thus the webserver also can't query it for the config settings.
- [ ] Seems like the `cleanup` pod fails?
- [ ] Whenever the CRD is changed, users need to `kubectl apply` its changes. Otherwise the controller might use fields that aren't defined yet. How do we plan to tackle this? Do we want to incorporate anything in the CLI?
- [ ] Removal of `orchest-ctl` from codebase (and required changes to `namespace_ctl.py`)
- [ ] `orchest-controller` needs to be versioned as well instead of using `"latest"` tags. This includes changing the `deployment.yaml` template.
  - [ ] `orchest-controller` to contain the right default version. Otherwise it will only ever install the hardcoded default version of `v2022.04.0`

Items we can postpone for after the release:
- [ ] `uninstall` command to be added. Delete CR Object and then delete `orchest` namespace. Alternatively, we can also tell users to delete their cluster if they were using it solely for Orchest.
- [ ] Development docs around the operator
  - [ ] Note that the `docker-registry` is managed through Helm and not the `orchest-controller`.
  - [ ] How to run the operator outside of the cluster for easy debugging


scripts/build_container.sh
```diff
@@ -152,8 +147,9 @@ run_build () {
                 cp $DIR/../$pnpm_file $build_ctx/pnpm_files 2>/dev/null
             done
         fi
-        if containsElement "${image}" "${HELM_IMAGES[@]}" ; then
+        if containsElement "${image}" "${CLI_IMAGES[@]}" ; then
             cp -r $DIR/../deploy $build_ctx/deploy 2>/dev/null
```
yannickperrenet, line 151: Should be able to remove this line.

scripts/build_container.sh
```diff
@@ -194,8 +190,9 @@ function cleanup() {
             if containsElement "${image}" "${PNPM_IMAGES[@]}" ; then
                 rm -rf $i/pnpm_files 2>/dev/null
             fi
-            if containsElement "${image}" "${HELM_IMAGES[@]}" ; then
+            if containsElement "${image}" "${CLI_IMAGES[@]}" ; then
                 rm -rf $i/deploy 2>/dev/null
```
yannickperrenet, line 194: Removal of this line as well.

.dockerignore
```diff
@@ -1,25 +0,0 @@
-# Python
```
yannickperrenet, line 1: Not sure why this file got deleted, we should probably check/verify this.

services/orchest-controller/deploy-controller/crds.yaml
```diff
@@ -0,0 +1,851 @@
+apiVersion: apiextensions.k8s.io/v1
```
yannickperrenet, line 1: Mentioning for reviewers, this file is auto-generated based on the schema in `types.go`. So no need to review this file.

services/orchest-controller/hack/tools.go
```diff
@@ -0,0 +1,23 @@
+//go:build tools
+// +build tools
+
+/*
+Copyright 2019 The Kubernetes Authors.
```
yannickperrenet, line 5: @ricklamers I guess we need to change this header?

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
```diff
@@ -0,0 +1,752 @@
...
+	InCluster                      bool
+	DefaultPause                   bool
+}
+
+func NewDefaultControllerConfig() ControllerConfig {
+	return ControllerConfig{
+		DeployDir:                 "/deploy",
+		PostgresDefaultImage:      "postgres:13.1",
+		RabbitmqDefaultImage:      "rabbitmq:3",
+		OrchestDefaultVersion:     "v2022.05.0",
```
yannickperrenet, line 79: @nhaghighat @fruttasecca We need to decide how the default version is added to the controller.

My favorite is using the `ORCHEST_VERSION` env var value that is inserted through the build script. The release of the controller needs to be inline with the release of Orchest.

services/orchest-controller/pkg/server/logs.go
```diff
@@ -0,0 +1,34 @@
+package server
```
yannickperrenet, line 1: @nhaghighat Want to rename this file to `status.go` given that the endpoint is no longer `/logs` either?

services/orchest-controller/pkg/utils/deepcopy.go
```diff
@@ -0,0 +1,11 @@
...
+
+func DeepCopy(src map[string]string) map[string]string {
+	dst := make(map[string]string, len(src))
+
+	for k, v := range src {
+		dst[k] = v
+	}
+
+	return dst
+}
```
yannickperrenet, line 11: Strictly speaking, this is not a "deepcopy" right? I would argue this is more like a shallow copy.

services/orchest-controller/pkg/utils/k8s_utils.go
```diff
@@ -0,0 +1,314 @@
...
+	mapEnvVars := make(map[string]string, length)
+
+	for _, envVars := range envVarLists {
+		for _, envVar := range envVars {
+			mapEnvVars[envVar.Name] = envVar.Value
+		}
+	}
+
+	return GetEnvVarFromMap(mapEnvVars)
+}
```
yannickperrenet, line 231: I haven't checked it again in the codebase, but we need to make sure that the merging of the environment variables is done correctly given that this implementation is giving precedence to arguments later in the list. Meaning `merge(A, B, C)` then `C` env vars would take precedence when overlapping with `A` and/or `B`.

@nhaghighat

yannickperrenet (commented):
A large part of this PR contains moved `deploy/` files and auto-generated files, my guess is only ~70 files out of the (currently) 319 files need to be reviewed.

services/orchest-controller/pkg/utils/deepcopy.go
yannickperrenet: @nhaghighat

fruttasecca:
Couple of  very minor things:
- we should probably mention that this is a breaking change in the PR description
- `scripts/build_container.sh -v -i orchest-controller -t "latest"` , might want to add `-o "latest"`

docs/source/development/troubleshooting.rst
```diff
@@ -77,3 +77,50 @@ Breaking schema changes
...
+       scripts/build_container.sh -i orcehst-api -t "v2022.04.0" -o "v2022.04.0"
+
+       # Edit the orchest-api deployment again to make sure to not
+       # run the sleep command anymore.
+       kubectl -n orchest edit deploy orchest-api
+
+
+**Context**
+    Alembic creates revision files to do migrations. When two different branches have done schema
+    migrations than the head will diverge, similar to git now having two different branches which
```
fruttasecca, line 124:
```suggestion
    migrations then the head will diverge, similar to git now having two different branches which
```

orchest-cli/setup.py
```diff
@@ -0,0 +1,48 @@
...
+exec(version, about)
+
+setuptools.setup(
+    name="orchest-cli",
+    description="CLI for Orchest",
+    keywords="orchest",
+    version=about["__version__"],
+    license="Apache 2.0",
+    author="Rick Lamers",
+    author_email="rick@orchest.io",
```
fruttasecca, line 19: Guess the author should be @yannickperrenet ?

services/orchest-controller/hack/tools.go
nhaghighat: This file is copy pasted from k8s repo and changed a little bit. I think it's safe to just add a line that some part is edited by us.

orchest-cli/setup.py
```diff
@@ -0,0 +1,48 @@
...
+        "Intended Audience :: Science/Research",
+        "Intended Audience :: System Administrators",
+        "License :: OSI Approved :: Apache Software License",
+        "Topic :: Utilities",
+    ],
+    packages=setuptools.find_packages(),
+    install_requires=[
+        "click",
+        "kubernetes>=21.7.0",
+        "requests",  # required by kubernetes as well
```
fruttasecca, line 38: Is the lack of pinning for `click` and `requests` intentional?

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
nhaghighat: injecting it during build time is indeed better.

services/orchest-api/app/app/__init__.py
```diff
@@ -156,6 +163,28 @@ def create_app(
...
 
+def create_required_directories() -> None:
+    """Creates required directories by Orchest services.
+
+    Should work fine when running multiple gunicorn workers.
+
+    Note:
+        It is very important, that this function is backwards compatible
+        meaning that new directories can be added but no old directories
+        can be (re)moved.
```
fruttasecca, line 174: Might be good to add some info about why this is the case, i.e. what are we trying to avoid

fruttasecca:
Could it be that `kubectl -f apply services/orchest-controller/deploy-controller` is meant to be `kubectl apply -f services/orchest-controller/deploy-controller` in the PR description?

services/orchest-controller/pkg/utils/deepcopy.go
nhaghighat: The Deepcopy of CRD objects are auto generated by kubebuilder, this is a copy for a map. (I guess it is you are right, it is a shallow copy, but it's fine as we don't use it to change the cache)

services/orchest-controller/pkg/utils/k8s_utils.go
nhaghighat: it indeed gives precedence to the arguments in the later list. we should add unit tests for it.

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+    Note:
+        Will do nothing in case the current CLI command is invoked with
+        the `--json` flag.
+
+    """
+    click_ctx = click.get_current_context(silent=True)
+
+    if click_ctx is None:
+        return click.echo(*args, **kwargs)
+    elif click_ctx.params.get("json_flag") is True:
```
fruttasecca, line 72: What's the reason for comparing like this in this particular case?

services/orchest-controller/pkg/utils/k8s_utils.go
yannickperrenet: FYI I check the code and the usage of the function is expected, i.e. important ENV vars being passed last.

orchest-cli/orchestcli/cli.py
```diff
@@ -0,0 +1,437 @@
...
+)
+def stop(watch: bool, **common_options) -> None:
+    """Stop Orchest.
+
+    All underlying Orchest deployments will scaled to zero replicas.
+
+    \b
+    Equivalent `kubectl` command:
+        kubectl -n orchest patch orchestclusters cluster-1 --type='merge' -p='{"spec": {"orchest": {"pause": true}}}'
+    """
```
fruttasecca, line 327: Does the same happen for daemonsets? If so, do we want to mention that here? (And also updated the equivalent kubectl command accordingly?)

orchest-cli/setup.py
yannickperrenet: I've gone ahead and made @ricklamers authors given that he is also listed as the author for the `orchest-sdk`. We should probably make this into a "company author" and "company email address".

orchest-cli/setup.py
yannickperrenet: For `requests` yes, given that it is required by `kubernetes` as well. We do also use `requests` ourselves but don't use any version specific features.

For `click` I still have to check, will add it to the list of Todos. Will test some other major versions :+1:

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+import time
+import typing as t
+from functools import partial
+
+import click
+import requests
+from kubernetes import client, config, stream, watch
+
+if t.TYPE_CHECKING:
+    from multiprocessing.pool import AsyncResult
```
fruttasecca, line 22: A comment explaining what's happening and/or why migh help

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+
+
+# NOTE: Schema to be kept in sync with `OrchestClusterPhase` from:
+# `services/orchest-controller/pkg/apis/orchest/v1alpha1/types.go`
+class ClusterStatus(enum.Enum):
+    INITIALIZING = "Initializing"
+    DEPLOYINGTHIRDPARTIES = "Deploying Third Parties"
+    DEPLOYEDTHIRDPARTIES = "Deployed Third Parties"
+    DEPLOYINGORCHEST = "Deploying Orchest Control Plane"
+    DEPLOYEDORCHEST = "Deployed Orchest Control Plane"
```
fruttasecca, line 111: Are these "substates" of the installation or initialization process?

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+) -> None:
+    """Updates Orchest.
+
+    Note:
+        The arguments `controller_deploy_name` and
+        `controller_pod_label_selector` need to be explicitely provided
+        given that everyone could have applied a custom
+        orchest-controller deployment.
+
+    """
```
fruttasecca, line 200: Is there a way to avoid this? Perhaps by looking if a controller with the default label/name exists when not specified? We could also prompt the user to make sure in that case, or something like that.

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+
+        In other words, returns whether `new` is a newer version than
+        `old`.
+
+        Note:
+            Both `old` and `new` must follow the CalVer versioning
+            scheme, e.g. "v2022.03.8".
+
+        """
+        old, new = old[1:], new[1:]
```
fruttasecca, line 213: Verifying that it indeed starts with `v` before removing the first character might avoid some edge cases in the future

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+        version = fetch_latest_available_version(curr_version)
+        if version is None:
+            echo("Failed to fetch latest available version to update to.", err=True)
+            sys.exit(1)
+    else:
+        # Verify user input.
+        try:
+            year, month, patch = version.split(".")
+            if len(year) != 5 or len(month) != 2 or len(patch) == 0:
+                raise
```
fruttasecca, line 271: Might be worth just having a function to verify the format to solve both this and https://github.com/orchest/orchest/pull/916/files#r864621177

.dockerignore
yannickperrenet: Got accidentally deleted in da1d3c02b816856d93922b5f00b4127897a2d8a8. I've put it back.

scripts/build_container.sh
yannickperrenet: Done.

scripts/build_container.sh
yannickperrenet: Done.

orchest-cli/orchestcli/cmds.py
yannickperrenet: In case this function was indirectly invoked through the CLI, e.g. `orchest install`, instead of through a Python call the `--json` flag should be honored. Meaning that when doing `orchest version --json` there is allowed to be only one output and that is the dictionary containing the output. If there are also printed lines it would break. 

By putting the check in this function we prevent this from happening even if higher level functions don't follow the convention. The `click` docs also mentioned using the `Context` to change `echo` behavior.

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+                continue
+
+            if isinstance(spec, dict):
+                convert_to_strategic_merge_patch(patch_obj[key], obj[key])
+            elif isinstance(spec, list):
+                # Strategically merge it.
+                patch_items = set(d["name"] for d in spec)
+                for obj_item in obj[key]:
+                    if obj_item["name"] not in patch_items:
+                        spec.append(obj_item)
```
fruttasecca, line 397: Just to make sure I get this right, the docstring says

>Precedence is given to `patch_obj`, i.e. if a key only exists in `patch_obj` then that value is preserved.
        
But that's not the only case, it would give precedence to `patch_obj` also if the `obj_item` exists in both `patch_obj` and `obj`, right? I.e. we are only adding new items, not modifying anything that already exists

orchest-cli/orchestcli/cli.py
yannickperrenet: It is actually not happening for our `node-agent` currently:
https://github.com/orchest/orchest/blob/367973de0bfeff7a5ba860975ca8a9ac638fe6fe/services/orchest-controller/pkg/controller/orchestcluster/orchest_reconciler.go#L55-L61

@nhaghighat The `node-agent` isn't getting deployed at all currently (as far as I can tell). Can you look into this?

orchest-cli/orchestcli/cmds.py
yannickperrenet: Yes they are. The `orchest-controller` will change the Status of the CR Object depending on the changes it is making to the Orchest Cluster. The CLI just wraps all `Status -> State` possibilities and outputs them to the user. This way there is one place of control, which is the `orchest-controller`.

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+    """Stops Orchest."""
+    ns, cluster_name = kwargs["namespace"], kwargs["cluster_name"]
+
+    echo("Stopping the Orchest Cluster.")
+    try:
+        patch_namespaced_custom_object(
+            name=cluster_name,
+            namespace=ns,
+            body={"spec": {"orchest": {"pause": True}}},
+        )
```
fruttasecca, line 573: What would happen if this is issued during, let's say, an update? Would the controller ignore it?

orchest-cli/orchestcli/cmds.py
yannickperrenet: Added in ca21a0f4871cec8b0bbaf7e6b124f0c1569ac233.

orchest-cli/orchestcli/cmds.py
yannickperrenet: Solved in ca21a0f4871cec8b0bbaf7e6b124f0c1569ac233.

orchest-cli/orchestcli/cmds.py
```diff
@@ -0,0 +1,1008 @@
...
+        "authServer": {
+            "env": [env for env in [env_var_dev, env_var_cloud] if env is not None],
+        },
+        "orchestWebServer": {
+            "env": [env for env in [env_var_dev, env_var_cloud] if env is not None],
+        },
+        "orchestApi": {
+            "env": [env for env in [env_var_dev] if env is not None],
+        },
+        "env": [env for env in [env_var_log_level] if env is not None],
```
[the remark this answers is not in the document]
yannickperrenet: Example usage (this `"env"` is the `orchest.Spec.Orchest.Env`): 
https://github.com/orchest/orchest/blob/367973de0bfeff7a5ba860975ca8a9ac638fe6fe/services/orchest-controller/pkg/controller/orchestcluster/auth_server.go#L25-L30

As far as I know these ENV vars are merged with all deployments that we do but is also used to store general environment variables such as the `ORCHEST_FQDN` (that are used in other places, e.g. ingress rules).

@nhaghighat Please correct me if I am wrong or missing something.

orchest-cli/orchestcli/cmds.py
yannickperrenet: This should be a concern of the `orchest-controller` and not of the CLI, given that this is a "management command" meaning that direct changes to the CR Object could be made to mimic the CLI commands behavior (e.g. a user uses `kubectl` to do this).

To answer how the `orchest-controller` handles this, see:
https://github.com/orchest/orchest/blob/367973de0bfeff7a5ba860975ca8a9ac638fe6fe/services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go#L290-L307
From what I understand the `orchest-controller` gets informed of cluster changes through "informers" (watching the k8s api-server, see [this code](https://github.com/orchest/orchest/blob/367973de0bfeff7a5ba860975ca8a9ac638fe6fe/services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go#L181)) that have "handlers" attached to them for certain events, e.g. Deployment ADDED. These events are then added to a queue that the `OrchestClusterController` operates on with a given `Threadiness` (which is currently set to `1`). This means that messages from the queue are handled in a sequential manner.

Thus to answer your question, the `pause` would be issued after the `update`. 

@nhaghighat Again, please correct me if needed ;)

orchest-cli/orchestcli/cmds.py
yannickperrenet: Addressed in c9a1514bea32cc12cce691ff221cd0eeb6146111.

orchest-cli/orchestcli/cmds.py
yannickperrenet: You are right, the language in the docstring is wrong. Will fix. :100:

orchest-cli/orchestcli/cmds.py
fruttasecca: My bad, I only wanted to select `elif click_ctx.params.get("json_flag") is True:` by ended up selecting more, what I meant to say is why aren't we comparing with `elif click_ctx.params.get("json_flag"):`

orchest-cli/orchestcli/cmds.py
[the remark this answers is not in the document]
fruttasecca: I see, alright, if that's the case we could change the name to something more indicative if you guys agree

orchest-cli/orchestcli/cmds.py
yannickperrenet: The value of `click_ctx.params.get("json_flag")` could be `None`, `True` or `False`, so I wanted to explicitly check whether the value `is True`. 

I see in https://google.github.io/styleguide/pyguide.html#2144-decision:
> Never compare a boolean variable to False using ==. Use if not x: instead. If you need to distinguish False from None then chain the expressions, such as if not x and x is not None:

So I will change it to that instead :+1:

yannickperrenet: Done: 612dc8c935ceefbdfdeacae1ac512e33a92b3b5d.

orchest-cli/orchestcli/cmds.py
fruttasecca: I see, makes sense. Does this mean there could be a race condition when waiting for `display_spinner(ClusterStatus.RUNNING, ClusterStatus.PAUSED)` in `orchest-cli/orchestcli/cmds.py`? I.e.:
- wait for it to be PAUSED
- it gets PAUSED
- some other change in the queue moves the state from PAUSED, e.g. STARTING
- we spinner never see the `PAUSED` status

I think in this particular case it doesn't matter, just good to keep in mind if situations like this can happen

orchest-cli/orchestcli/cmds.py
yannickperrenet: Those arguments are filled in with default values when invoked through the CLI. So a user only needs to specify them in case they deployed the `orchest-controller` with other values.

When it comes to invoking `update` through the UI, we call this Python function and pass the arguments. Note that in this process we can't ask the user to provide us with the values. Unless we add a pop-up in the UI but that feels like a poor user experience.

Not sure how we can solve this issue, given that users could have used any label/name for the controller deployment.

orchest-cli/orchestcli/cli.py
nhaghighat: added that.

nhaghighat: daemonset is removed and created again on pause and unpase

orchest-cli/orchestcli/cmds.py
yannickperrenet: When it comes to `controller_pod_label_selector` we should be able to infer that based on the `orchest-controller` deployment name, so we can at least remove that one. But then again, a user might have also changed that convention.

@nhaghighat Any other ideas?

fruttasecca:
Given that we are introducing `go` in the codebase by implementing a k8s controller it might be worth improving the docs of the controller. In particular:
- a README that explains how things work at the high level (e.g. why does it have an API, why does it have helm, etc.)
- the `go` equivalent of module level docstring for some key modules, e.g. `orchest_controller.go`, `orchest_reconciler.go`
- docstring for key functions

orchest-cli/orchestcli/cmds.py
nhaghighat: @yannickperrenet : you explained it perfectly

orchest-cli/orchestcli/cli.py
yannickperrenet: @nhaghighat Added the `node-agent` in ba21c595eda38125ac436c0aa2548b25074d9f11.

> I would be fine leaving it out

Same for me. So I guess we can resolve this discussion?

nhaghighat:
@fruttasecca :

> a README that explains how things work at the high level (e.g. why does it have an API, why does it have helm, etc.)

Will add the README and docs, thanks for the heads up.

fruttasecca:
@nhaghighat I think it might also be worth documenting somewhere how the controller avoids race conditions and/or the logic around the controller queue

yannickperrenet:
> a README that explains how things work at the high level (e.g. why does it have an API, why does it have helm, etc.)

I also added this to the Todo list. In my opinion we can better put in a development section of the `docs` so that we stick to the a single source for all documentation instead of it being scattered across READMEs.

orchest-cli/orchestcli/cli.py
fruttasecca: indeed

orchest-cli/orchestcli/cmds.py
nhaghighat: I think we can have a default name for controller deployment, and if the user changed it, ask him/her to provide the new name.

fruttasecca:
Given that `services/node-agent/app/image_deleter.py` takes care of deleting outdated Orchest images, which will include the controller, do you foresee any issue? We first update the controller to a new image and later update the rest of the cluster, if the `node-agent` is running it will see a controller image on the node which version differs from his. I guess the node-agent should do a `lte` comparison rather than an equality comparison?

fruttasecca:
Is the migration of existing configurations handled already? I couldn't find any reference to the `/orchest-settings` endpoint in the controller but I might be looking in the wrong place. With migration I mean the fact that we are moving from a `config.json` that lives in a pvc to having settings that are controlled by the `orchest-api`, and we would like to maintain the existing values. If not I wonder if we actually want to invest in such effort given that this is  breaking change already and we could simply inform users that they have to readjust the - very limited - settings in the settings panel.

nhaghighat:
> Given that services/node-agent/app/image_deleter.py takes care of deleting outdated Orchest images, which will include the controller, do you foresee any issue? We first update the controller to a new image and later update the rest of the cluster, if the node-agent is running it will see a controller image on the node which version differs from its own. I guess the node-agent should do a lte comparison rather than an equality comparison?

It indeed should not remove the latest image

fruttasecca:
I guess we want to update the info endpoint (`/api/orchest/update-info/v2`) to not provide a new version for pre controller versions of Orchest. I can see an upgrade working but then some functionality which depends on the controller existing wouldn't work, like the GUI update.

nhaghighat:
> Is the migration of existing orchest settings handled already? I couldn't find any reference to the /orchest-settings endpoint in the controller but I might be looking in the wrong place. With migration I mean the fact that we are moving from a config.json that lives in a pvc to having settings that are controlled by the orchest-api, and we would like to maintain the existing values. If not I wonder if we actually want to invest in such effort given that this is breaking change already and we could simply inform users that they have to readjust the - very limited - settings in the settings panel.

As you already mentioned this is a breaking change, so as part of that we can include the setting changes also

fruttasecca:
Btw, we might want to check how the controller defines the `Running` state, I had installed a version of the controlled which tried to pull `v2022.05.0 ` images, and failed, according to the status Orchest is running, but it's not.

fruttasecca:
Stopping Orchest doesn't stop argo and the registry, is this intentional?

nhaghighat:
> Btw, we might want to check how the controller defines the Running state, I had installed a version of the controller which tried to pull v2022.05.0 images, and failed, according to the status Orchest is running, but it's not.

Good catch, will check

fruttasecca:
I think I got into some error state, and `start`, `stop`, `restart` etc. aren't really working. They report success but
the expected changes do not take place. I am seeing this in the controller logs:
```
orchest_controller.go:334] failed to update orchest with default values  "cluster-1": Operation cannot be fulfilled on orchestclusters.orchest.io "cluster-1": the object has been modified; please apply your changes to the latest version and try again
```

nhaghighat:
> orchest_controller.go:334] failed to update orchest with default values  "cluster-1": Operation cannot be fulfilled on orchestclusters.orchest.io "cluster-1": the object has been modified; please apply your changes to the latest version and try again

This error state is known issue, but your cluster should come to stable phase even with these errors

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
```diff
@@ -0,0 +1,758 @@
...
+
+	return true
+}
+
+func (controller *OrchestClusterController) handleErr(err error, key interface{}) {
+	if err == nil {
+		controller.queue.Forget(key)
+		return
+	}
+	klog.Warningf("dropping orchest csluter %q out of the queue: %v", key, err)
```
fruttasecca, line 332:
```suggestion
	klog.Warningf("dropping orchest cluster %q out of the queue: %v", key, err)
```

orchest-cli/orchestcli/cmds.py
yannickperrenet: > I think in this particular case it doesn't matter, just good to keep in mind if situations like this can happen

That case could indeed happen (even more likely if the k8s api is slow to respond, e.g. user has high ping). Although I expect the case to be minimal and I'd say we can just postpone any improvements around this.

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
fruttasecca: agree

orchest-cli/orchestcli/cmds.py
fruttasecca: Can the controller pass its name to the webserver responsible for initiating the UI update? e.g. as an ENV var. Not super nice because if one gets updated the env var needs to be changed etc.

orchest-cli/orchestcli/cmds.py
fruttasecca: FYI I am getting this case quite often when we have back to back `_display_spinner` calls because we expect multiple state changes, like in `patch`, `restart`, `update`.

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
yannickperrenet: I've added the env var to the `orchest-controller` in 895537adf8d4dcde5e02df4a70824763b701dee0 and added the usage of that value for the default to the todo list in the PR description.

services/orchest-controller/hack/tools.go
ricklamers: @nhaghighat can you link to the source code?

yannickperrenet: Seems to me it is: https://github.com/kubernetes/kube-aggregator/blob/master/hack/tools.go

services/orchest-controller/pkg/controller/orchestcluster/orchest_controller.go
yannickperrenet: Done in 9e78fdb7801a7df0f3dc91c6a6799e9ce3033fc6 and c873e922ced790d96c4bd4384772b7ef29ce7168.

PR #917, opened 2022-05-04
Add tests to the hooks that load projects and pipelines

iannbing:
## Description

This PR adds FE integration tests to the hooks that load projects and pipelines. 

**Projects**
Extract hooks from `ProjectSelector` and implement tests on `useProjectSelector`. `ProjectSelector` is moved into `src/project-selector` so the hooks could be grouped in the same folder.

**Pipelines**
There are two parts in the codebase that help load valid pipelines: some effect hooks in `ProjectsContext` and `useEnsureValidPipeline`.  Extract the effects in `ProjectsContext` to a separate hook `useAutoFetchPipelines` and import it into `useEnsureValidPipeline` for better testability.

Some changes on configurations were needed:

The `component` property in `siteMap` object was extracted and moved to `src/Routes.tsx`. This is needed in order to import `siteMap` without importing other components. Otherwise, when importing `siteMap` into jest tests, jest also evaluates the code from the imports where it shouldn't. Normally it would work though, but if the components imports some third-party libraries (e.g. `xterm-for-react`), tests would break unexpectedly.

`toQueryString` function is also moved to `src/utils` so that it's not mixed in the routing configurations.

## Checklist


ricklamers:
Nice job and LGTM! I like how for testability you cleaned up things by separating them out better.

ricklamers (approved):

PR #918, opened 2022-05-04
Move analytics module to the shared internal library

fruttasecca:
## Description

Move analytics module to the shared internal library. This is a first step towards introducing analytics to the `orchest-api`.

Might need some more work after the k8s operator (https://github.com/orchest/orchest/pull/916) is merged to dev.


lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -14,10 +14,16 @@
 
 ALLOWED_FILE_EXTENSIONS = ["ipynb", "py", "R", "sh", "jl"]
 
+POSTHOG_API_KEY = "c3l6aU4waEhweEhBQnQ0UHRyT0FxRm1iX25wLXYwanRDNElIanZCZ1pwMA=="
```
astrojuanlu, line 17: Is this supposed to be here?

fruttasecca: Do you mean as part of the public repo? If so yes, it was previously in the webserver config.

yannickperrenet (approved):
:+1:

Issue #589, opened 2021-12-13
Add a right-click context-menu for actions in the pipeline editor


ricklamers:
**Is your feature request related to a problem? Please describe.**  
Add a right-click context-menu for actions in the pipeline editor

**Describe the solution you'd like**  
Normal right-click menu override that looks consistent with the rest of the MUI components. It can handle actions like:
- New step
- Duplicate step
- Delete step
- Run step
- Run all incoming steps
- Edit in JupyterLab

PR #931, opened 2022-05-09
Feat/canvas context menu

BrunoOrchest:
## Description

Adds context menu to pipeline editor for common actions for both steps and general canvas.

Menu flicker on Safari only is a MUI bug, is addressed here https://github.com/mui/material-ui/issues/31849 and should be fixed in the next/a future version of MUI.

Fixes: #589 

## Checklist

- [x] I haven't introduced breaking changes that would disrupt existing jobs.
- [x] In case I changed the dependencies in any `requirements.in` I have run `pip-compile`

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
@@ -27,7 +27,7 @@ import {
   fetcher,
   hasValue,
 } from "@orchest/lib-utils";
-import React, { useState } from "react";
+import React from "react";
 import { BackToJobButton } from "./BackToJobButton";
 import {
   getNodeCenter,
@@ -395,7 +395,7 @@ export const PipelineEditor = () => {
     setPipelineJson((value) => value, true);
   }, [setPipelineJson]);
 
-  const contextMenuState = useState(false);
+  const isContextMenuOpenState = React.useState(false);
 
   const autoLayoutPipeline = () => {
     const spacingFactor = 0.7;
@@ -654,7 +654,7 @@ export const PipelineEditor = () => {
           canvasFuncRef={canvasFuncRef}
           executeRun={executeRun}
           autoLayoutPipeline={autoLayoutPipeline}
-          contextMenuState={contextMenuState}
+          isContextMenuOpenState={isContextMenuOpenState}
         >
           {connections.map((connection) => {
             if (!connection) return null;
@@ -778,7 +778,7 @@ export const PipelineEditor = () => {
                 interactiveConnections={interactiveConnections}
                 onDoubleClick={onDoubleClickStep}
                 getPosition={getPosition}
-                contextMenuState={contextMenuState}
+                isContextMenuOpenState={isContextMenuOpenState}
               >
                 <ConnectionDot
                   incoming
--- a/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
@@ -27,7 +27,7 @@ import {
   fetcher,
   hasValue,
 } from "@orchest/lib-utils";
-import React from "react";
+import React, { useState } from "react";
 import { BackToJobButton } from "./BackToJobButton";
 import {
   getNodeCenter,
@@ -395,6 +395,8 @@ export const PipelineEditor = () => {
     setPipelineJson((value) => value, true);
   }, [setPipelineJson]);
 
+  const contextMenuState = useState(false);
+
   const autoLayoutPipeline = () => {
     const spacingFactor = 0.7;
     const gridMargin = 20;
@@ -650,6 +652,9 @@ export const PipelineEditor = () => {
         <PipelineViewport
           ref={pipelineViewportRef}
           canvasFuncRef={canvasFuncRef}
+          executeRun={executeRun}
+          autoLayoutPipeline={autoLayoutPipeline}
+          contextMenuState={contextMenuState}
         >
           {connections.map((connection) => {
             if (!connection) return null;
@@ -760,6 +765,9 @@ export const PipelineEditor = () => {
               <PipelineStep
                 key={`${step.uuid}-${hash.current}`}
                 data={step}
+                executeRun={executeRun}
+                onOpenFilePreviewView={onOpenFilePreviewView}
+                onOpenNotebook={onOpenNotebook}
                 selected={selected}
                 savePositions={savePositions}
                 movedToTop={movedToTop}
@@ -770,6 +778,7 @@ export const PipelineEditor = () => {
                 interactiveConnections={interactiveConnections}
                 onDoubleClick={onDoubleClickStep}
                 getPosition={getPosition}
+                contextMenuState={contextMenuState}
               >
                 <ConnectionDot
                   incoming
--- a/services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
@@ -19,7 +19,6 @@ import { getFilePathForRelativeToProject } from "./file-manager/common";
 import { useFileManagerContext } from "./file-manager/FileManagerContext";
 import { useValidateFilesOnSteps } from "./file-manager/useValidateFilesOnSteps";
 import { MenuItem, useContextMenu } from "./hooks/useContextMenu";
-import { useEventVars } from "./hooks/useEventVars";
 import { RunStepsType } from "./hooks/useInteractiveRuns";
 import { useUpdateZIndex } from "./hooks/useZIndexMax";
 import { InteractiveConnection } from "./pipeline-connection/InteractiveConnection";
@@ -127,7 +126,10 @@ const PipelineStepComponent = React.forwardRef<
     onDoubleClick: (stepUUID: string) => void;
     interactiveConnections: Connection[];
     getPosition: (node: HTMLElement | undefined | null) => Position | null;
-    contextMenuState: [boolean, React.Dispatch<React.SetStateAction<boolean>>];
+    isContextMenuOpenState: [
+      boolean,
+      React.Dispatch<React.SetStateAction<boolean>>
+    ];
     children: React.ReactNode;
   }
 >(function PipelineStep(
@@ -144,7 +146,7 @@ const PipelineStepComponent = React.forwardRef<
     // the cursor-controlled step also renders all the interactive connections, to ensure the precision of the positions
     interactiveConnections,
     getPosition,
-    contextMenuState,
+    isContextMenuOpenState,
     children, // expose children, so that children doesn't re-render when step is being dragged
   },
   ref
@@ -287,7 +289,7 @@ const PipelineStepComponent = React.forwardRef<
   const onMouseDown = React.useCallback(
     (e: React.MouseEvent) => {
       // user is panning the canvas or context menu is open
-      if (keysDown.has("Space") || contextMenuState[0]) return;
+      if (keysDown.has("Space") || isContextMenuOpenState[0]) return;
 
       e.stopPropagation();
       e.preventDefault();
@@ -296,17 +298,12 @@ const PipelineStepComponent = React.forwardRef<
         forceUpdate();
       }
     },
-    [forceUpdate, keysDown]
+    [forceUpdate, keysDown, isContextMenuOpenState]
   );
-
-  const [contextMenu, setContextMenu] = React.useState<{
-    mouseX: number;
-    mouseY: number;
-  } | null>(null);
 
   const onClick = async (e: React.MouseEvent) => {
     // user is panning the canvas or context menu is open
-    if (keysDown.has("Space") || contextMenuState[0]) return;
+    if (keysDown.has("Space") || isContextMenuOpenState[0]) return;
 
     e.stopPropagation();
     e.preventDefault();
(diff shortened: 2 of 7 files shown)


services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +454,78 @@ const PipelineStepComponent = React.forwardRef<
...
+      title: "Properties",
+      action: (e: React.MouseEvent, itemId: string) => {
+        dispatch({ type: "SELECT_SUB_VIEW", payload: 0 });
+      },
+    },
+    {
+      type: "item",
+      title: "Open in JupyterLab",
+      action: (e: React.MouseEvent, itemId: string) => {
+        onOpenNotebook(e);
```
iannbing, line 483: I think this only works if the step is "opened", because `onOpenNotebook` passed in from `PipelineEditor` depends on `eventVars.openedStep`. So I guess make `onOpenNotebook` accept step UUID would help.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +454,78 @@ const PipelineStepComponent = React.forwardRef<
...
+      title: "Delete",
+      action: (e: React.MouseEvent, itemId: string) => {
+        dispatch({ type: "REMOVE_STEPS", payload: [itemId] });
+      },
+    },
+    {
+      type: "item",
+      title: "Properties",
+      action: (e: React.MouseEvent, itemId: string) => {
+        dispatch({ type: "SELECT_SUB_VIEW", payload: 0 });
```
iannbing, line 476: I think `SET_OPENED_STEP` is the action that opens the Property panel.

services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
```diff
@@ -260,6 +269,80 @@ export const PipelineViewport = React.forwardRef<
...
+      action: () => {
+        dispatch({
+          type: "SELECT_STEPS",
+          payload: { uuids: Object.keys(eventVars.steps) },
+        });
+      },
+    },
+    {
+      type: "item",
+      title: "Run selected steps",
```
iannbing, line 297: Just nitpicking: would this option still useful if there's no step selected?

services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
```diff
@@ -260,6 +269,80 @@ export const PipelineViewport = React.forwardRef<
...
+      type: "item",
+      title: "Zoom out",
+      action: () => {
+        dispatch({
+          type: "SET_SCALE_FACTOR",
+          payload: eventVars.scaleFactor - 0.1,
+        });
+      },
+    },
+  ] as MenuItem[];
```
iannbing, line 339: As mentioned elsewhere, I'd recommend annotation over assertion if possible 😄.

iannbing (commented):
Good job! Some things to iron out but overall it looks nice! I like `useContextMenu`. Very simple to understand 👍.

I also left some open questions regarding UX. Curious about your thoughts 🙂

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +454,78 @@ const PipelineStepComponent = React.forwardRef<
...
+      title: "Duplicate",
+      action: (e: React.MouseEvent, itemId: string) => {
+        dispatch({ type: "DUPLICATE_STEPS", payload: [itemId] });
+      },
+    },
+    {
+      type: "item",
+      title: "Delete",
+      action: (e: React.MouseEvent, itemId: string) => {
+        dispatch({ type: "REMOVE_STEPS", payload: [itemId] });
```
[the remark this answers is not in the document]
BrunoOrchest: I started with the `selectedSteps` first (also for duplicate, etc) but this didn't make much sense since the 'context menu' doesn't select the step per se, so having an action affect the selected steps seemed weird. If we want it to also select it, then it will make sense. If we want, we can also have a context menu for not selected step (which always apply to the step only) and one for when a context menu is invoked on a step that is selected (and actions apply to all the selected steps)

services/orchest-webserver/client/src/pipeline-view/hooks/useEventVars.tsx
```diff
@@ -385,6 +389,34 @@ export const useEventVars = () => {
           });
         }
 
+        case "DUPLICATE_STEPS": {
```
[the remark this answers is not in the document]
BrunoOrchest: Hmm if we are duplicating a step and can't use the same file, should we 'reset' the file but keep all other properties?

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
BrunoOrchest: Hmm but we need the properties, probably will need to create a new action that opens it and selects the properties tab (since a fix before made it so the current tab was persisted between selection)

services/orchest-webserver/client/src/pipeline-view/hooks/useEventVars.tsx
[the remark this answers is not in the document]
iannbing: I think it makes sense :-). To recap, clean up file_path property if it ends with `.ipynb`.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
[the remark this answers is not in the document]
iannbing: I think your proposal makes sense. It's also how MacOS's files system works. Feels natural to me. 👍

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -452,18 +541,16 @@ const PipelineStepComponent = React.forwardRef<
           isStartNodeOfNewConnection && "creating-connection"
         )}
         style={{ transform, zIndex }}
+        onContextMenu={handleContextMenu}
```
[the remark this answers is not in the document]
iannbing: The current way of working in Orchest always starts with "selecting" the item (then the operational buttons will appear in the canvas.). 

Following the same logic, I would expect the operational buttons in the context menu would be effective if the targeted item is selected. Otherwise, consider the following scenario  :
- Step A, B, C are selected (the DELETE button appears on the bottom-right conner)
- right click on Step D, there's a DELETE in the context menu.

These two DELETE buttons have different behaviors, which is quite confusing.

It also applies to run steps. I would vote for right-clicking also selects the item for consistency 😄

services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
```diff
@@ -27,7 +27,7 @@ import {
   fetcher,
   hasValue,
 } from "@orchest/lib-utils";
-import React from "react";
+import React, { useState } from "react";
```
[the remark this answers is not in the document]
BrunoOrchest: I usually import all, but we have big files with a lot of things going on on the same file, so may be better to keep using 'namespaces' to avoid conflicts and make code easier to reason with

services/orchest-webserver/client/src/pipeline-view/hooks/useEventVars.tsx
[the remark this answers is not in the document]
BrunoOrchest: I have been thinking about this a bit, I think we shouldn't clean the file. IT should be an error to reuse the file but if I am duplicating steps, I might want to keep some files in the originals, some change. As a user, I would expect a duplicate to make a copy and then edit any issues that may occur then getting somethings that I setup manually removed, specially if it only happens in a single case  (notebooks) but works for others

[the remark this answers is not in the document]
iannbing: Indeed. However we are not there yet. Ideally, Orchest should accept a "not perfect" pipeline definition (i.e. `pipelineJson`, and doesn't allow user to run the pipeline if there's an error in the pipeline). => (this would be a great improvement. 😄)

At the moment, PipelineEditor only accepts a valid pipelineJson, and this means that if a `.ipynb` file is re-used, all the other valid operation will also be blocked (e.g. move the position of a step) and the "disallow re-using notebook file" alert will pop up again and again.

For the time-being, we probably could choose to:
- roll back the changes when errors occur (only duplicate the allowed steps, but not for the ones with `.ipynb` files.)
- disallow the operation if conditions are not met (stop if seeing a step with `.ipynb` file)
- tweak a bit to let user proceed and learn the rules along the way (duplicate the steps but leave out the file_path if it's a `.ipynb` file.)

Like I said, there're many workaround, but none of them is perfect 😅. But I think we could make user less frustrated with some nice touch.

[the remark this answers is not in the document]
BrunoOrchest: Went with removing the file path for now. We may want to have some kind of popup in the future or better support, but for now, per your suggestion, just remove the file if it is a notebook file

iannbing (approved):
LGTM! A long-awaited feature!

yannickperrenet:
I see it has already been merged but I will still put some comments here. Might be a good idea to revert the merge in my opinion.

---
Right-clicking on a step doesn't just give the context menu but also opens the step's properties. I expect a context menu not to open to properties though (unless that option in the context menu is pressed).

--- 
Duplicating a step with a Notebook assigned to it, doesn't work correctly. The file isn't correctly assigned to it. Reason is probably that a Notebook is not allowed to be used by two steps at the same time, normally we raise a warning but the context menu action doesn't

---
Confusing behavior where having multiple steps selected and opening the context menu of a step (that is inside the selection) and pressing delete does not operate on the entire selection. Not sure what we want to do here?

--- 
The zoom level (when zooming in/out) that is triggered through the context menu is different from the zoom trigger buttons in the button left corner of the editor.

---
The "run incoming and this step" option in the context menu doesn't seem to be working for me. I also don't think there is a BE endpoint to trigger this behavior (but I might be mistaken) so I guess the option needs to be removed.

iannbing:
> Duplicating a step with a Notebook assigned to it, doesn't work correctly. The file isn't correctly assigned to it. Reason is probably that a Notebook is not allowed to be used by two steps at the same time, normally we raise a warning but the context menu action doesn't

@yannickperrenet  We had a thread regarding this. https://github.com/orchest/orchest/pull/931#discussion_r880291384. Would you prefer the other two options?

- roll back the changes when errors occur (only duplicate the allowed steps, but not for the ones with .ipynb files.)
- disallow the operation if conditions are not met (disallow duplicating if seeing a step with .ipynb file)

yannickperrenet:
> disallow the operation if conditions are not met (raise the alert, don't perform any operation if seeing a step with .ipynb file)

I think this would be best. Why treat the "duplicate step" through the context menu different from drag-drop through the file-manager. In my opinion they should all behave similarly (for consistency reasons).

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +462,90 @@ const PipelineStepComponent = React.forwardRef<
...
+        }
+      },
+    },
+    {
+      type: "item",
+      title: "Run incoming and this step",
+      action: (e: React.MouseEvent, itemId?: string) => {
+        if (itemId) {
+          executeRun([itemId], "incoming");
+          executeRun([itemId], "selection");
```
iannbing, line 537: `executeRun` will set a `runUuid` (see `executePipelineSteps` in `useInteractiveRuns`, and then it will do polling to check the status of the interactive pipeline run. This implementation will only monitor the run of the latest call of `executeRun`. To achieve "Run incoming and this step", it should have a queue of runUuid's and poll the status until all runs are finished.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -285,15 +298,27 @@ const PipelineStepComponent = React.forwardRef<
...
-    [forceUpdate, keysDown]
+    [forceUpdate, keysDown, isContextMenuOpenState]
   );
 
+  const onContextMenu = async (e: React.MouseEvent) => {
+    const ctrlKeyPressed = e.ctrlKey || e.metaKey;
+    if (!selected) {
+      dispatch({
+        type: "SELECT_STEPS",
+        payload: { uuids: [uuid], inclusive: ctrlKeyPressed },
```
iannbing, line 309: Might need an additional flag to determine if the Properties panel should open or not.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
yannickperrenet: > To achieve "Run incoming and this step", it should have a queue of runUuid's and poll the status until all runs are finished.

I'd say let's just remove "run incoming and this step" for now. Even easier :P

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
iannbing: I think this is not yet covered. Currently the Properties panel will open if user open the context menu of a step.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
iannbing: > I'd say let's just remove "run incoming and this step" for now. Even easier :P

This is probably the most pragmatic decision for now 😅.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +462,90 @@ const PipelineStepComponent = React.forwardRef<
...
+          dispatch({ type: "DUPLICATE_STEPS", payload: [itemId] });
+        }
+      },
+    },
+    {
+      type: "item",
+      title: "Delete",
+      action: (e: React.MouseEvent, itemId?: string) => {
+        if (itemId) {
+          dispatch({ type: "REMOVE_STEPS", payload: [itemId] });
```
iannbing, line 480: Should be `eventVars.selectedSteps`.

services/orchest-webserver/client/src/pipeline-view/PipelineStep.tsx
```diff
@@ -437,6 +462,90 @@ const PipelineStepComponent = React.forwardRef<
...
+      type: "item",
+      title: "Open in File Viewer",
+      action: (e: React.MouseEvent, itemId?: string) => {
+        if (itemId) {
+          onOpenFilePreviewView(e, itemId);
+        }
+      },
+    },
+    {
+      type: "separator",
```
iannbing, line 511: Everything above the separator should be hidden when `isReadOnly` is true.

services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
```diff
@@ -260,6 +265,81 @@ export const PipelineViewport = React.forwardRef<
...
+      type: "item",
+      title: "Zoom out",
+      action: () => {
+        dispatch({
+          type: "SET_SCALE_FACTOR",
+          payload: eventVars.scaleFactor - 0.1,
+        });
+      },
+    },
+  ];
```
iannbing, line 336: Only keep "Center view", "Zoom in", and "Zoom out" when when `isReadOnly` is true.

PR #941, opened 2022-05-10
Fix requires restart alert not triggered issue.

iannbing:
## Description

If user changes Orchest configs at `/settigns` and the changes requires restart to take effects, an alert should be shown.
This PR fixes this bug and also implemented jest tests to prevent future regressions. `SettingsView.tsx` was moved to `/settings-view` folder in order to extract hooks for testing.

## Checklist


.gitignore
```diff
@@ -14,6 +14,7 @@ orchest.code-workspace
 #  .gitignore to avoid confusion
 services/*/.dockerignore
 services/*/lib
+services/orchest-webserver/app/core
```
fruttasecca, line 17: Not sure I understand the rationale behind this

services/orchest-webserver/client/src/settings-view/common.tsx
```diff
@@ -0,0 +1,46 @@
+import { OrchestConfig, OrchestUserConfig } from "@/types";
+import "codemirror/mode/javascript/javascript";
+import cloneDeep from "lodash.clonedeep";
+
+export const configToVisibleConfig = (
+  orchestConfig: OrchestConfig | undefined,
+  configJSON: Partial<OrchestUserConfig>
+): Partial<OrchestUserConfig> => {
+  if (!orchestConfig) return {};
+
```
fruttasecca, line 10: Is this because the response containing the Orchest config might still be in progress? Just making sure I understand the context

services/orchest-webserver/app/app/views/views.py
```diff
@@ -252,7 +252,8 @@ def user_config():
 
         if request.method == "POST":
             # Updated config, from client.
-            config = request.form.get("config")
+            response = request.get_json()
```
fruttasecca, line 255: Not sure `response` is the right name for this variable

.gitignore
iannbing: It's a binary file popped up in my repo (maybe because I'm running `dev` mode?). This started to happen since the latest release. It shouldn't be part of the source code.

services/orchest-webserver/client/src/settings-view/common.tsx
iannbing: Yes, I made it so that the app is not waiting for orchestConfig to start preparing rendering. But indeed the undefined check shouldn't be here. It should be done by the caller of this function. Pushed the change.

services/orchest-webserver/app/app/views/views.py
iannbing: Indeed. Changed to `request_body`.

fruttasecca (approved):
lgtm :), I really like the incremental tests additions

PR #943, opened 2022-05-10
Remove unnecessary side effects in  Step details

iannbing:
## Description

The `StepDetailsProperties` contains some side effects that saves pipeline steps multiple times, which could result in losing data (new data was overwritten by old data). This PR removes these side effects, and also fixes a bug in FilePicker.

## Checklist


ricklamers:
Found something potentially unrelated to this PR, but when I switch pipelines the headerbar updates but the canvas doesn't. (Two pipelines in the same project)

ricklamers:
I have a folder structure like this:

```
project/
  sub
    main.orchest
explore-data.ipynb
```

When double clicking on a step that points to `<project>/explore-data.ipynb` (which in the step is encoded as `../explore-data.ipynb`, as paths of the steps are relative to the pipeline file) the path that's passed to JupyterLab is wrong:

"Can't find path `xplore-data.ipynb`". So I think there's something going wrong with path resolution when going from relative path (in the step properties) to absolute path (which is being passed to JupyterLab "doc open" call).

ricklamers:
When switching the environment of a notebook step while JupyterLab has that notebook open with an active kernel there should be a prompt to change the kernel in JupyterLab. This is not being triggered for me.

iannbing:
> "Can't find path xplore-data.ipynb". So I think there's something going wrong with path resolution when going from relative path (in the step properties) to absolute path (which is being passed to JupyterLab "doc open" call).

I can reproduce it. Will fix this in this PR.

ricklamers:
With the same structure:

```
project/
  sub
    main.orchest
explore-data.ipynb
get-data.py
```

`View file` fails to load ("Failed to load file. Make sure the path of the pipeline step is correct.").

BrunoOrchest (approved):

iannbing:
@ricklamers 

> when I switch pipelines the headerbar updates but the canvas doesn't. (Two pipelines in the same project)

Fixed.

> "Can't find path xplore-data.ipynb". So I think there's something going wrong with path resolution when going from relative path (in the step properties) to absolute path (which is being passed to JupyterLab "doc open" call).

Fixed.

> View file fails to load ("Failed to load file. Make sure the path of the pipeline step is correct.").

This failed at `Flask` `safe_join` (the security fix we had earlier). I'll wait for the input from @fruttasecca.

services/orchest-webserver/client/src/pipeline-view/file-manager/FileTree.tsx
```diff
@@ -372,28 +376,41 @@ export const FileTree = React.memo(function FileTreeComponent({
...
+            const payload = shouldRemovePipeline
+              ? currentPipelines.filter((pipeline) => {
+                  return pipeline.uuid !== pipelineUuid;
+                })
+              : currentPipelines.map((pipeline) => {
+                  return pipeline.uuid === pipelineUuid
+                    ? { ...pipeline, path: pipelineFilePath }
+                    : pipeline;
+                });
+
```
fruttasecca, line 395: Since it's possible to move multiple files at a time, would this approach work in that case? I am not familiar with the details and I haven't found a reason for it not to work aside from the fact that we are comparing with only one `pipelineUuid`, just making sure this edge case is covered.

fruttasecca: Btw, seems to work properly on my side.

fruttasecca:
One minor quirk of the file manager that I've noticed, likely not related to the PR, the selection is preserved when moving files from the project to `/data`, but not viceversa. See https://www.loom.com/share/1d209e33cd8642ffa2431cac48fdc6d8

fruttasecca:
Tangentially related to the PR but it seems that pipeline parameters might suffer of a similar behaviour, see https://www.loom.com/share/c9ba0d2018414f0ca40189aa58b3264b

fruttasecca:
Likely unrelated to the PR but there might be some state issues during a session start, essentially creating new pipelines doesn't properly increment the names if a name clashes (i.e. main_1, main_2, etc.), See https://www.loom.com/share/d815c456587b48caa216d93cdd6bfe01

fruttasecca:
Might be worth splitting the fix for those issues I've found in separate PRs to avoid the scope of this one increasing too much, if you agree.

iannbing:
> View file fails to load ("Failed to load file. Make sure the path of the pipeline step is correct."). This failed at Flask safe_join (the security fix we had earlier). I'll wait for the input from @fruttasecca.

This is also fixed thanks to @fruttasecca 's help.

iannbing:
> Tangentially related to the PR but it seems that pipeline parameters might suffer of a similar behaviour, see

This is fixed in PR #946.

ricklamers:
When switching between two pipelines using back/forward the pipeline fails to load.

ricklamers:
> When switching the environment of a notebook step while JupyterLab has that notebook open with an active kernel there should be a prompt to change the kernel in JupyterLab. This is not being triggered for me.

This works but the dialog submit button needs a "busy" state where confirm becomes disabled. The dialog does not disappear immediately and when you click the button multiple times (there is no feedback on first click) it will fail (I get an error in JupyterLab). If you carefully just click once it does work for me.

iannbing:
> This works but the dialog submit button needs a "busy" state where confirm becomes disabled. The dialog does not disappear immediately and when you click the button multiple times (there is no feedback on first click) it will fail (I get an error in JupyterLab). If you carefully just click once it does work for me.

Removed the `await` to close the dialog immediately.

iannbing:
@fruttasecca The weird session-management issues were caused by redirect forth and back (after creating a pipeline, Pipeline Editor attempts to open it but then circle back to the current pipeline). Then the session was auto-started without you knowing.

@ricklamers also found some issues with routing. 

I have a new PR #949  that touches the code for routing, so I think it's better to fix these in that PR. The rest of the issues should already be covered. This PR is growing out of hand and becoming hard to get properly reviewed.

fruttasecca (approved):
lgtm :+1:

PR #946, opened 2022-05-12
Use `order` as the serial number for services

iannbing:
## Description

Currently services use user-defined `name` as the key for BE, and with this design, FE generates a FE-only property: `order` to ensure the consistency of the UI. This PR (re-)uses this property as the serial number of service.

This PR also fixes an unable-to-save issue in Pipeline Settings, and added more test cases.

## Checklist

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -4,7 +4,7 @@ import logging
 import os
 import re
 import subprocess
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Set, Tuple
 
 import requests
 from werkzeug.serving import is_running_from_reloader as _irfr
@@ -141,26 +141,14 @@ def is_service_name_valid(service_name: str) -> bool:
     return bool(re.match(r"^[0-9a-zA-Z\-]{1,36}$", service_name))
 
 
-def is_service_order_valid(
-    service_order: Any, existing_service_orders: list[int]
-) -> bool:
-    # service_order acts as a serial number of a service
-    if isinstance(service_order, int) and service_order not in existing_service_orders:
-        existing_service_orders.append(service_order)
-        return True
-    return False
-
-
-def is_service_definition_valid(
-    service: Dict[str, Any], existing_service_orders: list[int]
-) -> bool:
+def is_service_definition_valid(service: Dict[str, Any]) -> bool:
     return (
         isinstance(service, dict)
         and is_service_name_valid(service["name"])
         and isinstance(service["image"], str)
         and service["image"]
         and isinstance(service["scope"], list)
-        and is_service_order_valid(service.get("order", 0), existing_service_orders)
+        and isinstance(service.get("order"), int)
         and isinstance(service.get("preserve_base_path", False), bool)
         and
         # Allowed scopes.
@@ -194,14 +182,21 @@ def is_service_definition_valid(
 
 
 def is_services_definition_valid(services: Dict[str, Dict[str, Any]]) -> bool:
-    existing_service_orders = []
-    return isinstance(services, dict) and all(
-        [
-            is_service_definition_valid(service, existing_service_orders)
-            and sname == service["name"]
-            for sname, service in services.items()
-        ]
-    )
+    if not isinstance(services, dict):
+        return False
+
+    existing_orders: Set[int] = set()
+
+    for sname, service in services.items():
+        if (
+            not is_service_definition_valid(service)
+            or sname != service["name"]
+            or service["order"] in existing_orders
+        ):
+            return False
+
+        existing_orders.add(service["order"])
+        return True
 
 
 def rmtree(path, ignore_errors=False) -> None:
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -141,13 +141,26 @@ def is_service_name_valid(service_name: str) -> bool:
     return bool(re.match(r"^[0-9a-zA-Z\-]{1,36}$", service_name))
 
 
-def is_service_definition_valid(service: Dict[str, Any]) -> bool:
+def is_service_order_valid(
+    service_order: Any, existing_service_orders: list[int]
+) -> bool:
+    # service_order acts as a serial number of a service
+    if isinstance(service_order, int) and service_order not in existing_service_orders:
+        existing_service_orders.append(service_order)
+        return True
+    return False
+
+
+def is_service_definition_valid(
+    service: Dict[str, Any], existing_service_orders: list[int]
+) -> bool:
     return (
         isinstance(service, dict)
         and is_service_name_valid(service["name"])
         and isinstance(service["image"], str)
         and service["image"]
         and isinstance(service["scope"], list)
+        and is_service_order_valid(service.get("order", 0), existing_service_orders)
         and isinstance(service.get("preserve_base_path", False), bool)
         and
         # Allowed scopes.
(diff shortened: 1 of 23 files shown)


services/orchest-webserver/client/src/hooks/useFetchPipelineJson.ts
```diff
@@ -56,9 +55,21 @@ export const fetchPipelineJson = (
...
-    // Augment services with order key
-    for (let service in pipelineObj.services) {
-      pipelineObj.services[service].order = getOrderValue();
+    let maxOrder = 0;
+    const sortedServices = Object.entries(pipelineObj.services).sort((a, b) => {
+      if (!hasValue(a[1].order) && !hasValue(b[1].order))
+        return a[1].name.localeCompare(b[1].name); // If both services have no order value, sort them by name.
+      if (!hasValue(a[1].order)) return -1;
+      if (!hasValue(b[1].order)) return 1;
+      maxOrder = Math.max(maxOrder, a[1].order, b[1].order);
```
fruttasecca: Could it be that by returning early in a case where one of the order values is not defined but the other is we would actually skip a possible update to maxOrder, thus getting to the wrong maxOrder value?

iannbing: @fruttasecca Good catch!. Could you refresh? This issue has been solved in newer commits. :-)

services/orchest-webserver/client/src/hooks/useFetchPipelineJson.ts
```diff
@@ -56,9 +56,29 @@ export const fetchPipelineJson = (
...
+    // Ensure that order value is unique, and assign a valid value to `order` if it's undefined
+    let maxOrder = -1;
+    for (let sorted of sortedServices) {
+      const targetServiceOrder = pipelineObj.services[sorted[0]].order;
+      if (hasValue(targetServiceOrder)) {
+        const orderValue =
+          maxOrder === targetServiceOrder // Order value is duplicated.
+            ? targetServiceOrder + 1
+            : targetServiceOrder;
+        pipelineObj.services[sorted[0]].order = orderValue;
```
fruttasecca, line 75: When could such a duplication take place?  Also, is `pipelineObj.services[sorted[0]].order = orderValue;` only required because of said duplication?

fruttasecca:
Minor thing, since the args of `ensure_unique_order`, `sort_service_key_function`, `fill_missing_order` have typing, could you add typing to the return value as well?

It might also make sense to mark them as private by prefixing them with an underscore since we don't expect those to be used outside of the `compat` module.

About `ensure_unique_order` and `fill_missing_order`, they both return their argument and change it in place, it might be better to do away with returning their arg. The return value of `fill_missing_order` is not actually used anyway.

services/orchest-webserver/app/app/compat.py
```diff
@@ -113,3 +167,5 @@ def migrate_pipeline(pipeline: dict):
         ]
         migration_func(pipeline)
         pipeline["version"] = migrate_to_version
+
+    fill_missing_order(pipeline.get("services"))
```
fruttasecca, line 171: Could you change this migration to follow the migration "system" which is in place? I.e. adding it as a migration here 
```
_version_to_migration_function = {
    "1.0.0": (_migrate_1_0_0, "1.1.0"),
    "1.1.0": (_migrate_1_1_0, "1.2.0"),
}
```,
perhaps migrating to 1.2.1?

File: services/orchest-webserver/app/app/compat.py [hunk 4]
```diff
@@ -101,6 +103,58 @@ def _migrate_1_1_0(pipeline: dict) -> None:
 assert len(_version_to_migration_function) == len(__migration_functions)
 
 
+def ensure_unique_order(sorted_service_list: Tuple[str, Dict[str, Any]]):
+    anchor = -1
```
fruttasecca, line 107: Is anchor actually just a `current_max`? Perhaps the naming can be improved to something similar to what's used in the FE?

services/orchest-webserver/app/app/compat.py
```diff
@@ -101,6 +103,58 @@ def _migrate_1_1_0(pipeline: dict) -> None:
...
+    for key, service in sorted_service_list:
+        service_order = service.get("order", -1)
+        if service_order == -1:
+            continue
```
fruttasecca, line 111: I think you can use `.get("order")` and then compare with not as in `if service_order is None:` -> return

services/orchest-webserver/app/app/compat.py
```diff
@@ -101,6 +103,58 @@ def _migrate_1_1_0(pipeline: dict) -> None:
...
+def fill_missing_order(services: Dict[str, Dict[str, Any]]):
+    service_list = services.items()
+    ordered_dict: Dict[str, int] = {}
+
+    service_list = sorted(
+        service_list,
+        key=lambda service: sort_service_key_function(service[1], ordered_dict),
+    )
+
+    service_list, max_order = ensure_unique_order(service_list)
```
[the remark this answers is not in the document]
fruttasecca: Given this the FE could be simplified as well since it can assume that the BE returns a json where all services have an order, and it would just need to set the order on newly created services before posting them

services/orchest-webserver/client/src/hooks/useFetchPipelineJson.ts
iannbing: The current implementation is using localstorage to store the order value, so it's hard to say there won't be any duplicates. This precaution is here just in case issues occur after data migration.

>  is pipelineObj.services[sorted[0]].order = orderValue; only required because of said duplication?

Yes, however, `orderValue` could be still the same as `targetServiceOrder`. Any potential risk you saw?

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -141,13 +141,26 @@ def is_service_name_valid(service_name: str) -> bool:
...
+def is_service_definition_valid(
+    service: Dict[str, Any], existing_service_orders: list[int]
+) -> bool:
     return (
         isinstance(service, dict)
         and is_service_name_valid(service["name"])
         and isinstance(service["image"], str)
         and service["image"]
         and isinstance(service["scope"], list)
+        and is_service_order_valid(service.get("order", 0), existing_service_orders)
```
fruttasecca, line 163: In the event of multiple services with no defined `order` this would define the first one as valid because of the default to `0`, the second as invalid because there is already a 0 in `existing_service_orders`.

Given the more "global" nature of checking if there are no duplicates among services orders, what do you think of the following?
- have `is_service_definition_valid` make sure that `order` exists and it's an `int`
- have ` is_services_definition_valid` verify that there are no overlaps in service orders after verifying that all services are valid

fruttasecca:
This is only tangentially related, but I think we might have to add a call to `compat.migrate_pipeline` in `services/orchest-webserver/app/app/core/pipelines.py` right after loading the pipeline def at line 333, to avoid a rare edge case of a pipeline being moved without first having been loaded (thus migrated) by the BE and then stored again. Could you take care of it?

services/orchest-webserver/client/src/hooks/useFetchPipelineJson.ts
fruttasecca: >The current implementation is using localstorage to store the order value, so it's hard to say there won't be any duplicates. This precaution is here just in case issues occur after data migration.

I see, then I think we can indeed leave it here for a while (could you add a comment explaining why such duplicate checking is there?), but the BE can do away with this I guess.

>Yes, however, orderValue could be still the same as targetServiceOrder. Any potential risk you saw?

No, looks good, was just getting a more info on the context :))

services/orchest-webserver/app/app/compat.py
iannbing: Not sure if this is correct?

```
_version_to_migration_function = {
    "1.0.0": (_migrate_1_0_0, "1.1.0"),
    "1.1.0": (_migrate_1_1_0, "1.2.0"),
    "1.2.0": (_migrate_1_1_0, "1.2.1"),
}
```

services/orchest-webserver/app/app/compat.py
iannbing: Changed to `max_order`

services/orchest-webserver/app/app/compat.py
iannbing: Fixed.

ricklamers:
LGTM! Nice job. Tested everything manually too.

PR #947, opened 2022-05-12
Give orchest api access to the analytics module

fruttasecca:
## Description

This PR gives access to the `analytics` module to the `orchest-api` by moving it in the shared library. Analytics events that were previously sent by the webserver and that exist in the `orchest-api` are now responsibility of the `orchest-api` since this allows to leverage a number of things like asynchronous delivery, retries, access to the `orchest-api` db, etc.

The `orchest-api` analytics back-end is implemented as a subscriber subscribed to all events.

Now that the `analytics` event is in the shared library I think the interface/contract between the caller and the callee should be stronger. In particular, sending an event through the `analytics` module works by doing something akin to `analytics.send_event(event_type, event_properties)` , where event_type is an `Enum` and `event_properties` is a dictionary. It might be better to have a defined type (a class, data class, or typed dict) and make sending an event have the form of `analytics.send_event(my_event_instance)`. This would reduce coupling between the `analytics` module and its callers, along with providing better guarantees about what is being sent, since currently we rely on the caller to act correctly w.r.t. the content of `event_properties`. @yannickperrenet Keen to know your opinion.

This PR changes the name of the `analytics` events, bringing them in line with how events are named in the `orchest-api`, more details will be provided once other changesets to `improv/orchest-api-analytics-base` are done, in particular, the following PRs:
- moving more analytics events from the `orchest-webserver` to the `orchest-api` for the aforementioned reasons
- expanding the types of `orchest-api` events, thus expanding `analytics`
- possibly, a PR that refactors the interface of `send_event` like explained in the previous paragraph 

The base of this PR was started pre-controller and I ended up not being able to split this into two smaller PRs for easier review, sorry about that :).


fruttasecca:
>since currently we rely on the caller to act correctly w.r.t. the content of event_properties

Adding to this, this is not wrong per se, but imo it's wrong that the caller provides the content while the callee provides the content transformation (anonymization) while having such a weak interface.

yannickperrenet:
Let me share my general thoughts on the comments made in this PR before I start reviewing the current implementation.

> imo it's wrong that the caller provides the content while the callee provides the content transformation (anonymization) while having such a weak interface.

I agree that it is a weak interface that could easily lead to breakage. Sounds like we want to do some marshalling to assert whether the provided content is correct. Given that incorrect/missing data makes the analysis a lot complexer, I think it makes sense to fail at runtime (e.g. using something like assert statements) on missing/incorrect data so that developers make sure the content is correct.

> To add to this, I am not sure having the event_properties and derived_properties split provides much.

My reasoning behind going with this split was that `event_properties` contains the raw object as it lives within Orchest (with certain elements removed for anonimization) and `derived_properties` some information about the removed (due to anonimization) elements if possible. In this way the `event_properties` would stay consistent much longer whereas the `derived_properties` could get new items every now and then, which I felt would ease analysis. In addition, I think that having this split allows for an easier mental model during analysis (without the split you have no idea whether it is a derived property or not). Happy to change this though and keen to know your opinion @astrojuanlu 

> I think the interface/contract between the caller and the callee should be stronger. (...) It might be better to have a defined type (a class, data class, or typed dict) and make sending an event have the form of analytics.send_event(my_event_instance)

I don't yet see the benefit of moving the two arguments (`event_type` and `event_properties`) into one argument, may it be a class (`my_event_instance`) or whatever that encapsulates the current arguments. But (like I explained earlier in this comment) I agree that the interface could/should be improved. Maybe we can re-use the `types.py` or `schema.py` or do something similar.

services/orchest-api/app/app/core/notifications/analytics.py
```diff
@@ -0,0 +1,163 @@
...
+        models.AnalyticsSubscriber.uuid == delivery.deliveree
+    ).first()
+    if deliveree is None:
+        raise ValueError("Deliveree of delivery isn't of type AnalyticsSubscriber.")
+
+    try:
+        # TODO: this is needed because a depency of this PR actually
+        # lives in the controller branch, remove this.
+        current_app.config["TELEMETRY_DISABLED"] = False
+        current_app.config["TELEMETRY_UUID"] = "e2e2abb5-3c54-4209-a245-e9fa37810824"
```
yannickperrenet: Just highlighting as to not forget about this.

services/orchest-api/app/app/core/notifications/__init__.py
```diff
@@ -127,3 +128,23 @@ def process_notifications_deliveries_task() -> None:
...
+    )
+
+    logger.info(f"Found {len(analytics_deliveries)} analytics deliveries to deliver.")
+
+    for delivery in analytics_deliveries:
+        try:
+            analytics.deliver(delivery.uuid)
+        # Don't let failures affect other deliveries.
+        except Exception as e:
+            logger.error(e)
```
yannickperrenet, line 150: Would code similar to this be required whenever a new `Subscriber` model is defined?

services/orchest-api/app/app/core/events.py
```diff
@@ -33,11 +34,16 @@ def _register_event(ev: models.Event) -> None:
             f"Scheduling delivery for event {ev.uuid}, event type: {ev.type} for "
             f"deliveree {sub.uuid}."
         )
+        if isinstance(sub, models.AnalyticsSubscriber):
+            notification_payload = api_analytics.generate_payload_for_analytics(ev)
+        else:
+            notification_payload = ev.to_notification_payload()
```
yannickperrenet, line 43: Maybe for readability (will require a small change in `generate_payload_for_analytics` as well):

```python
notification_payload = ev.to_notification_payload()
if isinstance(sub, models.AnalyticsSubscriber):
    notification_payload = api_analytics.generate_payload_for_analytics(notification_payload)
```
makes it clearer that the "normal" `notification_payload` is modified to make it compatible for the analytics subscriber.

services/orchest-api/app/app/core/events.py
```diff
@@ -4,7 +4,9 @@
 accordingly based on any subscribers subscribed to the event type that
 happened.
 """
-from app import models, utils
+from app import models
+from app import utils
+from app import utils as app_utils
```
yannickperrenet, line 9: One of the two should be removed.

services/orchest-api/app/app/core/events.py
```diff
@@ -30,15 +32,21 @@ def _register_event(ev: models.Event) -> None:
...
-        _logger.info(
-            f"Scheduling delivery for event {ev.uuid}, event type: {ev.type} for "
-            f"deliveree {sub.uuid}."
-        )
         if isinstance(sub, models.AnalyticsSubscriber):
+            if app_utils.OrchestSettings()["TELEMETRY_DISABLED"]:
+                _logger.info(
+                    "Telemetry is disabled, skipping event delivery to analytics."
+                )
+                continue
```
yannickperrenet, line 40: How about moving the check for `TELEMETRY_DISABLED` into the `AnalyticsSubscriber`, feels like it is something the subscriber should own, e.g.: "Am I active?".

services/orchest-api/app/app/core/notifications/analytics.py
yannickperrenet: Has been removed in https://github.com/orchest/orchest/pull/947/commits/75906a449f268c83a87618fbe63967abffed6ab3

lib/python/orchest-internals/_orchest/internals/analytics.py
```diff
@@ -166,7 +166,7 @@ def send_event(
         sent to the telemetry service. False otherwise.
 
     """
-    if app.config["TELEMETRY_DISABLED"]:
+    if app.config.get("TELEMETRY_DISABLED", True):
```
yannickperrenet, line 169: I guess this is needed given that it is put in the internal library?

yannickperrenet (approved):
:+1: I think it looks great! I thought creating an Analytics subscriber was a clever way to implement things ;)

Apart from what comes out of the discussions on this PR, I think it can be merged.

services/orchest-api/app/app/core/notifications/__init__.py
fruttasecca: No the code can be made to be more generic, something I avoided to save time and because the current way (imo) makes it harder to run into bugs for the time being while we don't have tests.

services/orchest-api/app/app/core/events.py
fruttasecca: You mean at the database level? Or just under the umbrella of the `AnalyticsSubscriber` logic? Note that `models.py` won't be able to access `OrchestSettings()` because said settings use the `models.py` module, but we might be able to fetch the `TELEMETRY_DISABLED` value through `current_app`, although I am not too happy to haves `models.py` access the `current_app`, feels a bit peculiar.

lib/python/orchest-internals/_orchest/internals/analytics.py
fruttasecca: Indeed, overall it's just precautionary

fruttasecca:
>My reasoning behind going with this split was that event_properties contains the raw object as it lives within Orchest (with certain elements removed for anonimization) and derived_properties some information about the removed (due to anonimization) elements if possible.[...]

I see, I think it's very reasonable from the application POV, I think we can leave it as it is, let's see what's @astrojuanlu opinion.

services/orchest-api/app/app/core/notifications/__init__.py
yannickperrenet: > No the code can be made to be more generic

Indeed what I was hinting at. 

> something I avoided to save time and because the current way (imo) makes it harder to run into bugs for the time being while tests are not enabled.

I like it, lets leave it as is then :+1:

services/orchest-api/app/app/core/events.py
yannickperrenet: > Or just under the umbrella of the AnalyticsSubscriber logic?

This. I agree that putting it inside the `models.py` feels wrong.

fruttasecca:
>I don't yet see the benefit of moving the two arguments (event_type and event_properties) into one argument, may it be a class (my_event_instance) or whatever that encapsulates the current arguments. 

About the two arguments, if we are to create ad hoc classes or `types.py` entries I don't see the reason for having to pass the event type, given that said class would already encapsulate the information provided by the first argument, but we (the users of the module, and the implementor) would be in charge of verifying that what's passed is correct.

>But (like I explained earlier in this comment) I agree that the interface could/should be improved. Maybe we can re-use the types.py or schema.py or do something similar.

About the use of a class as the contract, I think one of the main "issues" (for lack of a better word) is that the `analytics` module of the internal library is shared, which means it has no knowledge of the types of single services. If we want to provide a contract we  have a few alternatives:
- allow single services to pass already anonymized events to be sent, every service is responsible for what it's sent, this allows the `analytics` module to ignore the structure of the event itself, only define the `event_types` that it accepts. 
- typing of the arguments to be passed through a class, typeddict, etc., the `analytics` module ends up having to define the hierarcy/structure of events (likely redundant w.r.t. what the single services have defined)
- move `types.py` to the shared lib, I am not sure we are ready for that, it implies we are moving to shared "global" types across all services


About point 2:
- current approach: callers have to make sure to not add fields that would be missed by the `analytics` anonymization logic, it's a weak interface
- `TypedDict`: can't share behaviour through inheritance/methods, overall not too big of an big issue. `TypedDicts` have no default values, which makes for worse ergonomics when it comes to some pre-defined fields (like the full name of the event type). The `types.py` module belongs to the `orchest-api`, making it shared would require more work and I am not sure we want to do that, regardless of that, the shortcomings of `TypedDicts` remain
- data classes (https://docs.python.org/3/library/dataclasses.html): can reuse behaviour, have a bunch of initialization nice to haves like `frozen` or `kw_only`, default values, etc. A lot of implementation overhead of "normal" classes it taken away.

If I were to synthesize:
- "bring your own anonymized event": no redundancy when it comes to schema/typing (I kinda like this because it plays very well with an implementation similar to `to_notification_payload` through the class hierarchy of events)
- "pass an event to analytics": either error-prone (weak interface) or leads to redundant and maybe harder to maintain code due to all extra typings required
- "define global types used across services": quite a task imo

services/orchest-api/app/app/core/events.py
fruttasecca: I see. `AnalyticsSubscriber` is a model so we are in a bit of a rut given the lack of extension methods in python (and I'd rather not monkey patch), nevertheless I'll look for improvements.

yannickperrenet:
> "bring your own anonymized event"

I think I like this best (although not completely convinced about it being an improvement). One of the main reasons I went with the current "weak interface" is that I didn't want to pollute a bunch of webserver proxy endpoints to do the anonymization. Given that this constraint is now no longer valid (although might translate to the orchest-api) the "bring your own" seems like a valid proposal.

In the end, to share anonymization functions across endpoints it feels like pretty much the same logic is repeated as it is currently. Just in a different place.

Ideally, the `send_event` call is super easy to use as to encourage usage of it throughout the codebase. Having to import the correct anonymization functions to anonymize the data you want to send doesn't sound like an improvement.

fruttasecca:
After an internal chat we decided to go for a "bring your own anonymized event" approach.

services/orchest-api/app/app/core/events.py
fruttasecca: Not relevant anymore since we are moving to a "bring your own anonymized event" approach

astrojuanlu:
👍🏽 to having `event_properties` with the original, raw information as much as possible (modulo anonymization of course) and splitting  `derived_properties`!

PR #953, opened 2022-05-16
Add  interactive pipeline run events to `orchest-api` (notifications)

fruttasecca:
## Description

Adds support for interactive pipeline run events in the `orchest-api`, these events can now be subscribed to and are now part of the analytics module.

## Checklist


fruttasecca:
>From a user perspective I don't think notifications for interactive runs makes a ton of sense, but happy to include it given that you already implemented it +1

Mostly for analytics

PR #961, opened 2022-05-18
Move job_update event to the `orchest-api` (notifications)

fruttasecca:
## Description

Moves the "job updated" analytics event to the `orchest-api` as an event that can be subscribed to. Part of the larger effort to add more events to the `orchest-api` and analytics.

## Checklist

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_jobs.py
+++ b/services/orchest-api/app/app/apis/namespace_jobs.py
@@ -16,6 +16,7 @@ from _orchest.internals import config as _config
 from _orchest.internals import utils as _utils
 from _orchest.internals.two_phase_executor import TwoPhaseExecutor, TwoPhaseFunction
 from app import schema
+from app import types as app_types
 from app.apis.namespace_runs import AbortPipelineRun
 from app.celery_app import make_celery
 from app.connections import db
@@ -1074,6 +1075,7 @@ class UpdateJob(TwoPhaseFunction):
         confirm_draft,
     ):
         job = models.Job.query.with_for_update().filter_by(uuid=job_uuid).one()
+        changes = []
 
         if name is not None:
             job.name = name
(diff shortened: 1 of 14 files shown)


services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -1225,13 +1226,58 @@ def _transaction(
...
+        call to this function will catch invalid updates.
+        """
+        changes = []
+        changes.extend(
+            get_env_vars_update(old_job["env_variables"], new_job["env_variables"])
+        )
+        # (field name, if values should be recorded) for notifications.
+        # Don't record sensitive values.
+        to_compare = [
+            ("name", False),
```
yannickperrenet, line 1256: Why do we not include the `name` in the notification payload? Given that it is send to a user configured endpoint I feel like it can be, especially because `name` seems like an important attribute to the user.

yannickperrenet: Let's say a user gets an update notification that the job has been updated, but then can't find the job in the UI because the new name is not included in the message.

yannickperrenet (approved):
:muscle:

services/orchest-api/app/app/apis/namespace_jobs.py
fruttasecca: The project and job name would already be contained in the notification payload, but it would not be contained in the changes list, like you noted. The reason is that the list of changes can be quite heterogeneous so defining a contract between its content and analytics could be quite difficult, so for now I opted for an approach where an `EntityUpdate` is sent as is to both notifcations and analytics (pasting code since it's easier):

```python
    # Changes that the event led to, i.e. a project PUT. It's a
    # dictionary given the nature of the content and the fact that other
    # kind of events might have such a column as well.
    # Note that the event payload will contain the changes AS IS, i.e.
    # no transformation is done to the content of the column, so no
    # sensitive data should be added. The update must follow the
    # "EntityUpdate" TypedDict schema.
    @declared_attr
    def update(cls):
        return Event.__table__.c.get("update", db.Column(JSONB, nullable=True))
```

Highly related to the issue we are discussing in https://github.com/orchest/orchest/pull/947#issuecomment-1129835113, imo when the content is potentially so heterogeneous not shifting the responsibility of the content to the logic that produced it might be very costly

fruttasecca: After we have decided to go for the "bring your own anonymized event" approach this behaviour might change.

PR #964, opened 2022-05-19
Expand 'orchest-api' events and analytics capabilities 

fruttasecca:
## Description

This PR expands the `orchest-api` scope when it comes to what events are covered by notifications and telemetry, to do so, it refactors the `analytics` module and moves it to the internal shared library. `Event`s models now both supply a notification payload and a telemetry payload, which is anonymized. The responsibility of anonymizing a payload has been shifted from the `analytics` module to callers of said module. The reason is that callers, which are the producers of the payload, have more context about the actual content of the payload, thus better information about what to anonymize. 

The newly added docs explain how to add event (and a telemetry event) to the `orchest-api`, same for the `orchest-webserver` and the FE client.

This PR acts as a "base" PR for a collection of PRs that made this change possible, each of those PRs description can help in adding more context:
- https://github.com/orchest/orchest/pull/918
- https://github.com/orchest/orchest/pull/956
- https://github.com/orchest/orchest/pull/962
- https://github.com/orchest/orchest/pull/961
- https://github.com/orchest/orchest/pull/953
- https://github.com/orchest/orchest/pull/950
- https://github.com/orchest/orchest/pull/947
- https://github.com/orchest/orchest/pull/980
- https://github.com/orchest/orchest/pull/963
- https://github.com/orchest/orchest/pull/975
- https://github.com/orchest/orchest/pull/974
- https://github.com/orchest/orchest/pull/970 

## Checklist

- [x] The documentation reflects the changes.
- [X] The PR branch is set up to merge into `dev` instead of `master`.
- [X] In case I changed one of the services’ `models.py` I have performed the appropriate database
      migrations (refer to `scripts/migration_manager.sh`).
- [X] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards
      compatibility is maintained.

Todos:
- [x] merge `dev` back this branch and solve conflicts, to be done after https://github.com/orchest/orchest/pull/1008 is merged
- [x] check for branching schema migrations (which will be there)
- [ ] more QA and analytics backend checks before merging back to `dev`

PR #970, opened 2022-05-19
Shift anonymization responsibility to analytics module callers, part 1

fruttasecca:
## Description

First part of the ongoing effort to shift the anonymization responsibility from the shared `analytics` module to the callers of said module. The purpose of this refactor is to move the logic of anonymization and the logic that maintains the analytics payload compatible with the old analytics schema (i.e. deprecated stuff) closer to where said payload is generated.

The first commit of the PR splits the `models.py` `orchest-api` module into sub files because it was getting too big, so don't get scared by the count of LOC changes :p.


lib/python/orchest-internals/_orchest/internals/analytics.py
```diff
@@ -171,38 +146,58 @@ def _initialize_posthog() -> None:
...
         True if the event (including its information) was successfully
         sent to the telemetry service. False otherwise.
 
     """
+    if event_data is not None and (
+        "event_properties" not in event_data
+        or "derived_properties" not in event_data
+        or len(event_data) != 2
+    ):
+        raise ValueError("event_data should be of type TelemetryData")
```
yannickperrenet, line 199: I don't think we are already using `pydantic`, but this would be a stellar use case for it. Especially the `len(event_data) != 2` is bound to break ("hidden") if we ever make changes in the `TelemetryData` definition.

services/orchest-webserver/app/app/views/analytics.py
```diff
@@ -15,7 +15,12 @@ def analytics_send_event():
             )
             return "Invalid analytics event name.", 500
 
-        success = analytics.send_event(app, analytics_event, request.json["properties"])
+        data = analytics.TelemetryData(
+            event_properties=request.json["properties"],
+            # TODO: will the FE ever have derived properties?
+            derived_properties={},
+        )
```
yannickperrenet, line 22: Good question. I don't think the FE ever will. Guess only future will tell :P

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -252,11 +252,16 @@ def catch_api_proxy_jobs_duplicate():
...
+            analytics.TelemetryData(
+                event_properties={
+                    "duplicate_from": json_obj["job_uuid"],
+                    # Deprecated fields, kept to not break the analytics
+                    # BE schema.
+                    "job_definition": None,
+                    "snapshot_size": None,
+                },
+                derived_properties={},
+            ),
```
yannickperrenet, line 264: Why are these fields deprecated? Can't the `job_spec` be changed such that the previous `job_definition` in the telemetry data is "outdated"?

yannickperrenet:
I guess for the `@staticmethod` functions `current_layer_notification_data(event)` we need to make sure that they never return a reference to the `event` or modify it because that could lead to `to_notification_payload(self)` and `to_telemetry_payload(self)` to change the same object (`self` is passed as `event`). Possibly leading to notification data slipping into telemetry data (if `to_telemetry_payload` is called before `to_notification_payload` and then the `telemetry_payload` being `send_event`).

yannickperrenet:
To check whether my understanding of the changes is correct. Let's say the user creates a new job, then:

1. The `orchest-api` creates a new job and `events.register_job_created`
2. Whenever an event is registered it is checked what subscribers are subscribed to this event
3. For all of these subscribers the correct payload is constructed. This is done based on the type of the subscriber, e.g. `AnalyticsSubscriber`, and the type of the event, e.g. `CronJobEvent`. 
4. The payload is added to the DB under `models.Delivery`
5. The delivery of the payload is then handled by the `celery-worker`

So to send a notification (may it be a webhook one or an analytics one), all you have to do is call the right `events.register_...` function in the right place?

On one hand feels like a very nice abstraction, whereas on the other hand whenever you want to add a new type of event (which is the common case imo since e.g. job creation will not suddenly be added to another place) you have to create a lot of functions in the right places. Definitely a huge improvement on the previous "analytics system" though, handing of the different responsibilities (e.g. creating the payload, delivering the payload) instead of having to do everything in one place.

:muscle:

lib/python/orchest-internals/_orchest/internals/analytics.py
fruttasecca: It would indeed be great if we were to start using `pydantic`, so far I've put it off because it felt like committing to something that should be decided when solving https://github.com/orchest/orchest/issues/391, which we put on hold for a long time, but I am not agains the idea tbh (quite the opposite)

services/orchest-webserver/app/app/views/orchest_api.py
fruttasecca: >Why are these fields deprecated?

They are being passed "properly" when creating the payload. `snapshot_size` in particular can't be here because of the synchronous nature of the API call and the unbound time it could take to get the result

> Can't the job_spec be changed such that the previous job_definition in the telemetry data is "outdated"? 

Not sure I follow here, could you clarify?

fruttasecca:
>To check whether my understanding of the changes is correct. Let's say the user creates a new job, then:
>So to send a notification (may it be a webhook one or an analytics one), all you have to do is call the right events.register_... function in the right place?

Yes pretty much. Of course you will have to create the model and/or add the required `event_type` if we are talking about new events

>On one hand feels like a very nice abstraction, whereas on the other hand whenever you want to add a new type of event (which is the common case imo since e.g. job creation will not suddenly be added to another place) you have to create a lot of functions in the right places. 

Yes that's true. Not sure if I would define it as a lot but the asynchronous delivery and the overall abstraction comes indeed at a price. About having to place `events.register_...` calls, I had thought about using hooks but DB level ones imo are a no go given that we will find ourselves mixing the codebase between python and SQL, sqalchemy hooks had some gotchas that I was afraid were going to leak the abstraction when it comes to how a user is saving events, so I opted for this more naive approach (I also didn't want to add more and more to the PR ahah :p)

lib/python/orchest-internals/_orchest/internals/analytics.py
yannickperrenet: Indeed feels a bit like jumping the gun, but we have (like you mentioned) waited on it for a long time. I'd say let's just start using it.

In the case where we had to revert the decision, it is pretty easy given that usage can be found using a `grep`, whereas otherwise (not using it) we would have to search extensively where to start using the alternative.

services/orchest-webserver/app/app/views/orchest_api.py
yannickperrenet: > They are being passed "properly" when creating the payload.

I think this answers both my questions :+1:

yannickperrenet:
> Given the logic and the fields of the Events models I am not sure this is a big risk, where most risk is at (imo) is that often current_layer_telemetry_data calls current_layer_notification_data to be DRY, so there is the chance of letting things slip through here and is part of the reason to move the creation of the payload and its anonymization closer ( i.e. the "Shift anonymization responsibility" PRs)

Makes sense. Just wanted to have a short discussion about it. All good!

yannickperrenet (approved):
LGTM! I will leave it up to you to decide whether you want to start using `pydantic` here (and we can then on an ongoing basis start using it elsewhere as well).

fruttasecca:
>LGTM! I will leave it up to you to decide whether you want to start using pydantic here (and we can then on an ongoing basis start using it elsewhere as well).

Might be preferable to merge the PR, so that the "base" PR (https://github.com/orchest/orchest/pull/964) can be prepared (e.g. fleshing out the description, merging dev back in, taking care of merge conflicts). Will probably wait for https://github.com/orchest/orchest/pull/1008 to be merged to `dev` since expanding the `orchest-api` analytics scope has introduced a lot of changes and I would like to do some more end to end QA. I might add a small PR on top of this to introduce `pydantic` depending on our schedule with other tasks, if not, I'll make sure to introduce `pydantic` the next time we find ourselves in such a spot; of course, open to discussion.

PR #971, opened 2022-05-19
Remove swr in useCheckUpdate

iannbing:
## Description

This is one of the PR's that removes the usage of swr (see #973).

Edit:
This ends up a rather-big refactoring. `useCheckUpdate` was moved to a new context `AppInnerContext`. The reason is that if `useCheckUpdate` lives inside a view, the polling is most likely won't be useful (user needs to stay in the same view for one hour to actually trigger the next check). To meet the requirement of the actual use case, we need to move `useCheckUpdate` to an app-wide context.

However, the existing `AppContext` cannot be used for this is that `useCheckUpdate` needs to detect route changes, so it needs to be wrapped as any child of `<Route />`, but `AppContext` must be a parent of `<Route /`>. So we run into a circular dependency.

Therefore, `AppInnerContext` was created. It's an app-wide context that has access to `react-router` context.

Ideally, I would rename `AppConext` to `GlobalContext`, and `AppInnderContext` to `AppContext`. But I think we could change it later in a separate PR.

### DIFF
--- a/services/orchest-webserver/client/src/App.tsx
+++ b/services/orchest-webserver/client/src/App.tsx
@@ -13,6 +13,7 @@ import { OnboardingDialog } from "./components/Layout/OnboardingDialog";
 import { AppDrawer } from "./components/MainDrawer";
 import { SystemDialog } from "./components/SystemDialog";
 import { useAppContext } from "./contexts/AppContext";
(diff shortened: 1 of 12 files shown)


yannickperrenet (approved):
LGTM!

Thank you for the elaborate PR description, really helps :)

PR #972, opened 2022-05-19
Remove swr in useSessionPoller

iannbing:
## Description

This is one of the PR's that removes the usage of swr (see #973).


iannbing:
Close this PR since #986 is merged. Any remark / bug in this PR should be resolve in a separate PR so that we could keep git history clean.

Issue #849, opened 2022-03-31
Improved dev process for making `base-image` changes


yannickperrenet:
**Describe the problem this improvement solves**
Whenever you make changes to the `base-images`, e.g. `base-kernel-py`, then testing the changes is rather hard. Because the builder pulls from DockerHub the new image won't actually be used.

In order to test your changes you need to:
1. Point the builder to the local registry
2. Push the new image to the local registry

**Describe the solution you'd like**
Not sure what a better solution would look like, but definitely removing some steps from this process. In addition, I totally forgot that the builder pulls from DockerHub and lost some time testing my changes (my guess is that others could also easily forget about this behavior).

astrojuanlu:
In case we don't immediately tackle this issue, how would one "Point the builder to the local registry"?

fruttasecca:
The process is quite convoluted atm:
- modify `services/orchest-api/app/app/core/image_utils.py`  by prefixing the base image name in the workflow manifest (see `def _get_*_image_build_workflow_manifest` ) with the registry ip, a search through the codebase for `registry_ip` will yield examples.
- build & redeploy the celery-worker
- port forward the `docker-registry` pod
- from now on it's the same process as pushing any image to any registry, i.e. build the desired image and push it to the registry, e.g. build with with a name like `localhost:5000/repo/image:tag` and then `docker push` it

PR #966, opened 2022-05-19
Improve robustness of Orchest Operator and update mechanism

nhaghighat:
## Description

This PR fixes order of deployment by introducing new internal CRD named `OrchestComponent`. The `OrchestCluster` will be created by the user or the `orchest-cli`, then `OrchestCluster` controller creates different `OrchestComponent` for each service, then a dedicated component controller for each component, controls the status of the underlying objects of `OrchestComponent`

Fixes: #952, #991

## Testing the PR
```sh
minikube addons enable ingress
eval $(minikube docker-env)
scripts/build_container.sh -M -t "v2022.05.3" -o "v2022.05.3"
pip install -e orchest-cli

# yes, this is ALL that is needed now to install Orchest
orchest install --dev
```

#### Testing `orchest update` through the CLI
```bash
orchest uninstall
scripts/build_container.sh -M -t "v2022.04.4" -o "v2022.04.4"
orchest install --dev
scripts/build_container.sh -M -t "v2022.04.5" -o "v2022.04.5"
orchest update --dev --version=v2022.04.5
```

#### Testing `orchest update` through the UI
NOTE: You need to have created the minikube with the orchest-dev-repo mount

```bash
orchest uninstall
scripts/build_container.sh -M -t "v2022.04.4" -o "v2022.04.4"
orchest install --dev
orchest patch --dev
pnpm run dev
scripts/build_container.sh -M -t "v2022.04.5" -o "v2022.04.5"
INVOKE THROUGH UI (go to http://localorchest.io/update)
scripts/build_container.sh -M -t "v2022.04.6" -o "v2022.04.6"
INVOKE THROUGH UI (go to http://localorchest.io/update)
... (as often as you like)
```

## Checklist

- [x] The documentation reflects the changes.
- [x] The PR branch is set up to merge into `dev` instead of `master`.
- [x] In case I changed code in the `orchest-cli` I followed its [release
      checklist](https://github.com/orchest/orchest/blob/master/orchest-cli/RELEASE-CHECKLIST.md)
- [x] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards
      compatibility is maintained.
- [x] In case I changed the dependencies in any `requirements.in` I have run `pip-compile` to update
      the corresponding `requirements.txt`
- [x] Start reports that Orchest is successfully started, but some deployments still have to start.
- [x] webserver hangs then restarts after a while on start if it's started concurrently w.r.t. to the orchest-api.
- [x] Celery fails to boot on fresh install
[...]

PR #985, opened 2022-05-24
Improve base image development process

fruttasecca:
## Description

Makes it so that it's possible to build environments and customize jupyter from base images that have been built locally on the node, e.g.:

- `orchest patch --dev`
- `eval $(minikube -p minikube docker-env)`
- `bash scripts/build_container.sh -i base-kernel-py -o v2022.05.3 -t v2022.05.3`
- specify `registry:docker-daemon:orchest/base-kernel-py` as the custom base image and build

Fixes: #849

## Todos:
- [x] changes that would break after https://github.com/orchest/orchest/pull/966 is merged, like passing `dev` mode to the `celery-worker`

## Checklist


fruttasecca:
About https://github.com/orchest/orchest/pull/985#discussion_r880591377, we could actually detect if the base image starts with `docker-daemon:` and mount the docker socket dynamically but I am afraid of potential security implications, also I like the separation that `--dev` gives.

Having to run `--dev` mode just for that is a bit annoying, so maybe a `--dev-image-builder` could be nice, altought I am not sure about the ROI, we would have to reason about what to do with `--dev`, `--no-dev`, etc.

orchest-cli/orchestcli/cmds.py
```diff
@@ -703,6 +703,9 @@ def disable_telemetry() -> None:
         "orchestApi": {
             "env": [env for env in [env_var_dev, env_var_cloud] if env is not None],
         },
+        "celeryWorker": {
+            "env": [env for env in [env_var_dev] if env is not None],
+        },
```
yannickperrenet, line 708: This will require an increase in the `orchest-cli` `_version.py`

docs/source/development/development_workflow.rst
```diff
@@ -456,3 +456,54 @@ branches respectively. Therefore, we require PRs to be merged into ``dev`` inste
...
+
+
+Testing jupyter base image changes
+----------------------------------
+
+Required reading: :ref:`testing environment base image changes <environment base images changes>`.
+
+Orchest does not allow to specify custom jupyter base images, which makes it slightly more difficult
+to tell the back-end that you want to use an image which was built locally. To do so, simply add
+``# LOCAL`` to the first line of the custom build script.
```
yannickperrenet, line 498: Why not take this approach for the environment images as well? I can definitely remember the `# LOCAL` but the exact `registry:docker-daemon:` prefix I would probably have to look up every time.

services/orchest-api/app/app/core/jupyter_image_builds.py
```diff
@@ -114,31 +119,41 @@ def prepare_build_context(task_uuid):
...
+        # Create empty shell script if no setup_script exists.
+        os.system(f'touch "{snapshot_setup_script_path}"')
+
+    base_image = f"orchest/jupyter-server:{CONFIG_CLASS.ORCHEST_VERSION}"
+    if CONFIG_CLASS.DEV_MODE:
+        # Use image from local daemon if instructed to do so.
+        with open(snapshot_setup_script_path, "r") as script_file:
+            first_line = script_file.readline()
+            if "# LOCAL" in first_line:
+                base_image = f"registry:docker-daemon:{base_image}"
```
yannickperrenet, line 139: So the `registry:docker-daemon:` prefix needs to be added before `{base_image}` both by the Jupyter build and the environment builds. However, then I don't understand why in `write_jupyter_dockerfile` the `registry:` prefix is stripped away but for the environments builds, this isn't done.

docs/source/development/development_workflow.rst
fruttasecca: `registry:docker-daemon` felt cleaner, but we can indeed go with the `#LOCAL#` approach for both, definitively more consistent

services/orchest-api/app/app/core/jupyter_image_builds.py
fruttasecca: The `custom_registry_prefix = "registry:"` used in `write_jupyter_dockerfile`and were `write_environment_dockerfile` were introduced to support registries that are different from `docker.io`, when no registry is specified we assume it's `docker.io`. Essentially we are telling the lower layer to not use `docker.io`

yannickperrenet (approved):
LGTM!!

Issue #856, opened 2022-04-01
Allow keyboard navigation of the select file dropdown for the file path of a pipeline step property


ricklamers:
**Describe the problem this improvement solves**
It's a bit annoying to have to use the mouse.

**Describe the solution you'd like**
Allow keyboard navigation of the select file dropdown for the file path of a pipeline step property

**Screenshots**
![image](https://user-images.githubusercontent.com/1309307/161274389-c46743bd-4588-42c0-831c-bb608de7cb51.png)

PR #992, opened 2022-05-26
Feat/file picker keyboard navigation

BrunoOrchest:
## Description

Adds keyboard navigation to file picker. When the file path text field has focus, pressing arrow key down, moves the focus to the file browser, making it keyboard navigable.

Fixes: #856 

## Checklist

      compatibility is maintained.


iannbing:
Super nice that you found this solution! Currently it works when user clicks on the TextField, however, once user presses enter (i.e. the MunuItem's are re-rendered), the new first MenuItem doesn't gain the focus as expected. Is it possible to re-assign `menuFirstItemRef.current` when the menu items are changes (i.e. `options` is updated). And by looking at the implementation, when `isRootNode` is true, it also won't work.

On top of that, as a user, I would also expect to press Escape to close the dropdown.

It would be super nice if user could use FilePicker fully with keyboard. Could you try it?

iannbing:
Super nice that you found this solution! Currently it works when user clicks on the TextField, however, once user presses enter (i.e. the MunuItem's are re-rendered), the new first MenuItem doesn't gain the focus as expected. Is it possible to re-assign `menuFirstItemRef.current` when the menu items are changes (i.e. `options` is updated). And by looking at the implementation, when `isRootNode` is true, it also won't work.

On top of that, as a user, I would also expect to press Escape to close the dropdown.

It would be super nice if user could use FilePicker fully with keyboard. Could you try it?

iannbing (approved):
Nice! All cases are covered!

PR #994, opened 2022-05-26
Fix useAsync setData causing indefinite re-rendering bug

iannbing:
## Description

setData in useAsync depends on `data`, which will cause indefinite re-rendering if the consumer put `setData` and `data` in the deps of React.useEffect.

This PR fixes it by remove `data` from the deps of `setData`.
The type ReducerActionWithCallback is to simplify typing of this pattern.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/hooks/useAsync.ts
+++ b/services/orchest-webserver/client/src/hooks/useAsync.ts
@@ -1,3 +1,4 @@
+import { ReducerActionWithCallback } from "@/types";
 import React from "react";
 import { useCancelablePromise } from "./useCancelablePromise";
 
@@ -23,10 +24,16 @@ type State<T, E> = {
   error: E | undefined;
 };
 
+type AsyncReducerAction<T, E> = ReducerActionWithCallback<
+  State<T, E>,
+  Action<T, E>
+>;
+
 const asyncReducer = <T, E>(
   state: State<T, E>,
-  action: Action<T, E>
+  _action: AsyncReducerAction<T, E>
 ): State<T, E> => {
+  const action = _action instanceof Function ? _action(state) : _action;
   switch (action.type) {
     case "PENDING": {
       const payload = action.caching ? state.data : undefined;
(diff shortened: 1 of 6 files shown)


services/orchest-webserver/client/src/hooks/useAsync.ts
```diff
@@ -23,10 +24,16 @@ type State<T, E> = {
...
 
+type AsyncReducerAction<T, E> = ReducerActionWithCallback<
+  State<T, E>,
+  Action<T, E>
+>;
+
 const asyncReducer = <T, E>(
   state: State<T, E>,
-  action: Action<T, E>
+  _action: AsyncReducerAction<T, E>
```
BrunoOrchest, line 34: Why the underscore? Usually (eslint and normal TS conventions) use a leading underscore for unused variables

BrunoOrchest (approved):

services/orchest-webserver/client/src/hooks/useAsync.ts
yannickperrenet: I guess to prevent shadowing of the variable (usage in 2 lines below this comment).

iannbing: Ah, interesting to know. I think in some other conventions it means a private variable. For unused variables, currently we need to explicitly put `//eslint-disable-line @typescript-eslint/no-unused-vars` to remove the warning, rather than replying on the naming of the variable itself. 🙂

iannbing: I'll merge this PR for now since it's a quite critical fix. We could continue the discussion on the naming convention. 🙂

BrunoOrchest: Sounds good. For reference: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals

Issue #998, opened 2022-05-27
Do not allow user to delete own account


yannickperrenet:
**Describe the bug**
When authentication is enabled and the user deletes himself/herself then the user can't do any more actions in the UI. As a consequence the user can't access the instance anymore (ever).

**Expected behavior**
A user shouldn't be able to delete himself/herself.

**To Reproduce**

1. Go to settings > manage users
2. Create two users (only one is needed but otherwise you are stuck at a very annoying state)
3. Enable authentication and restart
4. Log in to one of the two users
5. Delete the current user
6. Try to add a new user, which you can't do. Meaning that you can never recover from the mistake you just made, leaving the instance "locked"

PR #999, opened 2022-05-27
Prevent logged in user from deleting own account

yannickperrenet:
## Description

Fixes: #998

### DIFF
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -198,7 +198,7 @@ def register_views(app):
                 if user.is_admin:
                     return jsonify({"error": "Admins cannot be deleted."}), 500
                 elif self_username == to_delete_username:
-                    return jsonify({"error": "Deleting own user is not allowed."}), 500
+                    return jsonify({"error": "Deleting own user is not allowed."}), 405
                 else:
                     db.session.delete(user)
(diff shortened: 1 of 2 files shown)


services/auth-server/app/app/views.py
```diff
@@ -189,13 +189,16 @@ def delete_user():
...
-            username = request.form.get("username")
+            to_delete_username = request.form.get("username")
 
-            user = User.query.filter(User.username == username).first()
+            user = User.query.filter(User.username == to_delete_username).first()
             if user is not None:
                 if user.is_admin:
                     return jsonify({"error": "Admins cannot be deleted."}), 500
+                elif self_username == to_delete_username:
+                    return jsonify({"error": "Deleting own user is not allowed."}), 500
```
iannbing, line 201: Would 401 Unauthorized makes more sense than 500?

yannickperrenet: I think [405](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405) would be even better actually.

iannbing: Hmm, 405 doesn't seem 100% correct to me. The method DELETE  is actually allowed. I'd think it in a way like "users are not authorized to delete their own account".

Issue #867, opened 2022-04-08
Orchest configuration might lose state about requiring a restart if a pod is restarted


fruttasecca:
Currently, the `OrchestGlobalConfig.save()` method return value will tell if Orchest needs to be restarted and, if so, which changed settings required the restart. Said logic is based on comparing the flask application settings with the current settings.

If a pod is restarted, for whatever reason, the application will be initialized again, leading to the flask application settings being updated with the content of the `OrchestGlobalConfig`, then, according to the logic, no restart would be required, but any change affecting entities outside of the flask application (like celery parallelism) won't actually have been applied.

yannickperrenet:
Nice catch! I hope this is the only one that slipped through the cracks during the k8s port ;)

PR #1001, opened 2022-05-27
Fix/dont lose settings restart required state on restart

fruttasecca:
## Description

Makes it so that the fact that a setting requires a restart is persisted in the database.

Couple of things I am not too happy about:
- using the cleanup pod to register the fact that Orchest is being stopped, although I am not sure if it makes sense to explore more fleshed out solutions, i.e. doesn't seem too much of a priority
- the fact that we are using the flask app config as storage for the user Orchest settings, which seems to be only required because we are passing the flask app to the analytics module (which I am not too happy about either). This isn't something introduced by the PR but it might be a good occasion to discuss the matter and create an issue to solve it in the future

Fixes: #867 

## Checklist

### DIFF
--- a/services/orchest-api/app/app/__init__.py
+++ b/services/orchest-api/app/app/__init__.py
@@ -102,7 +102,7 @@ def create_app(
 
         with app.app_context():
             settings = utils.OrchestSettings()
-            settings.save()
+            settings.save(app)
             app.config.update(settings.as_dict())
 
     # Create a background scheduler (in a daemon thread) for every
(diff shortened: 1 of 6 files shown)


yannickperrenet:
> the fact that we are using the flask app config as storage for the user Orchest settings, which seems to be only required because we are passing the flask app to the analytics module (which I am not too happy about either)

I agree. I don't think we still need the flask app to store the configuration values. However, not "caching" the values in the Flask app means that we would have to query the `orchest-api` every time we want to send an event (given that that requires the `TELEMETRY_DISABLED` value, among others) or that we would have to cache the value in some other place.

yannickperrenet (approved):
LGTM!

Issue #954, opened 2022-05-17
`Ctrl+K` not captured while on JupyterLab


astrojuanlu:
**Describe the bug**
`Ctrl+K` is not captured while on JupyterLab.

**Expected behavior**
`Ctrl+K` works consistently everywhere in Orchest.

**To Reproduce**

1. Go to the pipeline editor
2. Hit `Ctrl+K`, see command palette open
3. Enter JupyterLab
4. Hit `Ctrl+K`, nothing happens or the browser captures the event (on Firefox, address bar gains focus)

**Environment**

- OS: Linux Mint
- Browser: Firefox
- Orchest's version (in the settings page): v2022.03.6

yannickperrenet:
@ricklamers @iannbing Anything we can do here given that JupyterLab lives in an iframe? I guess we can simply catch certain commands (in this case `Ctrl+K`) before sending it to the iframe?

PR #1004, opened 2022-05-27
Fix issue: Command Palette is not callable in JupyterLab view

iannbing:
## Description

Fixes: #954 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/App.tsx
+++ b/services/orchest-webserver/client/src/App.tsx
@@ -7,7 +7,7 @@ import React from "react";
 import { BrowserRouter as Router, Prompt } from "react-router-dom";
 import { useIntercom } from "react-use-intercom";
 import BuildPendingDialog from "./components/BuildPendingDialog";
-import CommandPalette from "./components/CommandPalette";
+import { CommandPalette } from "./components/CommandPalette";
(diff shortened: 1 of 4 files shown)


astrojuanlu (changes requested):
I tested this locally and I'm observing a weird behavior:

- After clicking on the top bar, `Ctrl+K` opens the command palette ✔️  (this was happening already before)
- After clicking on the JupyterLab area, now the command palette opens ✔️  (which wasn't happening before) but ⚠️ the address bar gains focus and breaks something so I can't just close the command palette anymore

Video:

https://user-images.githubusercontent.com/316517/171133694-c99d9cc8-9891-4a0a-a4dd-84e75e445444.mp4

iannbing:
> the address bar gains focus and breaks something so I can't just close the command palette anymore

I'm afraid that's the browser built-in behavior ([see Firefox for example](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly)). This is something we could not control. We could consider changing the key combination though.

iannbing:
@astrojuanlu  Could you try again with the new commit? And which browser are you using? I tested it on Chrome Firefox, Brave, all seem to work as expected.

astrojuanlu (approved):

Issue #878, opened 2022-04-19
Make PostgreSQL service template use mount by default


astrojuanlu:
**Describe the problem this improvement solves**
As stated in the documentation, services are stateless ("To ease development, services have a lifetime that spans across the entire [session](https://docs.orchest.io/en/stable/fundamentals/glossary.html#interactive-session"). However, this caught me by surprise after using the built-in PostgreSQL template and realizing that my example data was gone after a session restart.

**Describe the solution you'd like**
Have the built-in PostgreSQL template be configured with a `/data` disk mount by default, so the data persists across session restarts.

Even though it's easy to set this up after the fact, this default is probably more helpful.

PR #1005, opened 2022-05-30
Hide postgres service

astrojuanlu:
Since Orchest services are supposed to be stateless, recommending PostgreSQL (or any other database) can be misleading to users. They can always set it up themselves, but hopefully this will avoid typical data loss.

## Description

This hides the PostgreSQL service from the list of options.

WIP: Docs

Fixes: #878

## Checklist

- [ ] In case I changed one of the services’ `models.py` I have performed the appropriate database
- [ ] In case I changed code in the `orchest-cli` I followed its [release
- [ ] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards
      compatibility is maintained.
- [ ] In case I changed the dependencies in any `requirements.in` I have run `pip-compile` to update


astrojuanlu:
Rendered docs: https://docs.orchest.io/en/stable/fundamentals/services.html (just missing a new screenshot)

iannbing (approved):
No concerns on the FE side. Yet I'm curious about how we could inform user about this: 

> Since Orchest services are supposed to be stateless

astrojuanlu:
It's mentioned in the first paragraph:

> To ease development, services have a lifetime that spans across the entire [session](https://orchest--1005.org.readthedocs.build/en/1005/fundamentals/glossary.html#interactive-session).

But perhaps it's worth adding a more visible indication.

fruttasecca:
Wouldn't changing the `postgres` service template to make use of the data or project dir mount also solve the issue? I.e. the solution that's proposed in https://github.com/orchest/orchest/issues/878.

astrojuanlu:
That would be another solution yes, but after some discussion we decided against it:

> I think until we have services that don’t share the lifecycle with sessions (eg start and stop with the session, ie in jobs they get killed when the pipeline run ends and for sessions when the session is stopped), we should recommend using external services for stateful/persistent services (like Postgres on RDS or Streamlit apps/(Fast)API endpoints on Heroku/Render/Northflank).

fruttasecca:
Alright sounds good!

Issue #120, opened 2021-01-25
Notifications when builds and pipelines have finished executing


yannickperrenet:
Notifications should be made optional.

Without notifications the user always needs to check the pipeline or build in order to know whether or not it is done building.

Notifications can be browser based (when inside the application) but possible also via integrations such as email or Telegram.

yannickperrenet:
Would be great if we have some form of webhooks. This way users can specify a URL to which we will send the appropriate requests to indicate applications status, e.g. pipeline run failure.

yannickperrenet:
> also via integrations such as email

or Slack! I would really like to see this feature added to be honest :smile:

astrojuanlu:
Also for job failures - I find myself realizing that some runs have failed after a few hours or days.

I think webhooks are the way - folks can then connect that to any system they like.

liflovs:
Upvoting this. Impossible to keep track of failures at the moment, unfortunately.

hamisB:
Upvoting this too. Just found out some runs have failed a few hours ago. This notification feature could be a game changer :)

sawtooth9:
Upvoting this as well, for job failures in particular. It isn't hard to monitor when you only run a few jobs or you are using it for your own purposes. However, we are frequently running into this problem now that we have many jobs run every night, and many people within our organization rely on the data that flows with these job runs.

fruttasecca:
Hi @liflovs @hamisB @sawtooth9, we are going to start working on this in a few days, we are exploring an array of options and collecting preferences and brainstorming a bit. Would webhooks be a strong preference here, or is some kind of jobs dashboard (https://github.com/orchest/orchest/issues/485) more likely to suit your workflow? Solutions aren't exclusive, so feel free to propose.

liflovs:
@fruttasecca dashboard is "nice" to have but webhooks is "must" to have imho, because we need to know sometimes immediately when something fails and react. In some way we can already check on jobs if we want to, but we have to do it proactively

sawtooth9:
@fruttasecca The ideal workflow for us would likely be something similar to what is described in #865 so we could create/add to a report that is alongside all of our other monitoring in the dashboard we use day-to-day. That said, a built-in dashboard or webhook would be extremely helpful as well. Most of our jobs run overnight so the immediacy of the webhook isn't necessary for us, but being able to send out a slack notification when a job fails or collecting the events from a webhook would be a great feature.

yannickperrenet:
[...]

PR #1008, opened 2022-05-31
Sending notifications when jobs fail.

iannbing:
## Description

This PR exposes the BE functionality of sending notifications to desired channels (e.g. Slack) when jobs fail.

Closes: #120 

## Todo

- [x] Items from https://github.com/orchest/orchest/pull/1008#issuecomment-1143731359

## Checklist

      compatibility is maintained.


fruttasecca:
Currently, we are subscribing users to 
```
 project:one-off-job:failed
 project:cron-job:run:failed
 ```
 A one off job is registered as failed after all its pipeline runs have entered an end state, and there is at least 1 failed run.
 A cron job is registered as failed only if, pretty much, fails to enter the `STARTED` state, which is something that happens after "confirming" the job from a draft.
 
 We might want to subscribe users to the following events (along with the previous two): 
 - `project:one-off-job:pipeline-run:failed`: a pipeline run of a job has failed, note that it implies that, if there are many runs that fail, many notifications will be delivered
 - `project:cron-job:run:failed`: a cron job run is a batch of pipeline runs created based on the cron job schedule. This event can be triggered in two different moments: if the BE is not able to queue the batch of pipeline runs, or if a pipeline run part of the batch fails
- `project:cron-job:run:pipeline-run:failed`: a pipeline run has failed, note that it implies that, if there are many runs that fail, many notifications will be delivered

fruttasecca:
I am not entirely sure showing events in this raw form
image

fruttasecca:
About the webhook docs, which should probably decide who's doing what

iannbing:
> Given the design, what happens if we have existing users with existing webhooks, and we decide to change what events we are subscribing users to by default? I think it implies either that we need to update existing webhooks in the database or that the view will have to change to accommodate for the fact that different webhooks might be subscribed to different events?

If you meant users who are already using the endpoints without GUI, indeed it's then not correctly presented in the UI. 

I think, though, we have to allow user to select the events to subscribe before changing any default event types (then we could show a toggle per event).

And also, the list of webhooks in Notification settings would be app-wide, meaning that its `subscriptions` would be empty. We have to think about how to allow user subscribe to jobs with a specific UUID.

yannickperrenet:
> A one off job is registered as failed after all its pipeline runs have entered an end state, and there is at least 1 failed run.
A cron job is registered as failed only if, pretty much, fails to enter the STARTED state, which is something that happens after "confirming" the job from a draft.

I would argue that users are interested in getting notified whenever a pipeline run fails. Reasoning about the failure of an encapsulating one-off job or cronjob (meaning that it has many pipeline runs due to parameters) I don't think makes much sense. I would want to get a notification the moment something fails so that I can expect it and fix things. 

Let's say you have two pipeline runs X and Y as part of a job, then X and Y could run the same code but on different data sources. In that case it is very important that both X and Y run, so the moment one fails we should notify the user.

> `project:cron-job:run:pipeline-run:failed`: a pipeline run has failed, note that it implies that, if there are many runs that fail, many notifications will be delivered

This I like ^_^. We could later add an option where you can limit the number of notifications per job, if that is what users want.

> I am not entirely sure showing events in this raw form

I totally agree. I would rather give them more appropriate names.

yannickperrenet:
### Overview of nitpicks/issues I encountered

#### 1. Alignment is of banner is off

The text in the banner (in the jobs list view) is not vertically aligned in the center. Left side is too high and right side is too low.

image

#### 2. Alternative text proposal for notification settings

I would change the text to: "Get notified about specific events in Orchest." The fact that we use webhooks is something they can see in the actual settings themselves.

image

#### 3. Unclear "error" indication when testing webhook

I used https://webhook.site/ to quickly test out the _TEST_ button when creating a webhook. The payload got successfully POST'ed but in our UI we show an :exclamation:. Not sure how to interpret it.

image

#### 4. Icon alignment in webhooks table is off

After clicking the _TEST_ in the Connection column I felt like the :white_check_mark: was not aligned properly (vertically speaking).

image

#### 5. Connection column in webhooks table

In the above screenshot you can see that we offer a _TEST_ functionality in the webhooks table. When I first saw this I immediately thought: "Why doesn't Orchest check it for me? Why do I have to press that button?" 

In addition, after I pressed _TEST_ and added another webhook the :white_check_mark: from the first test disappeared. 

#### 6. Unclear / Unconfigurable webhook events vs channels

From the UI it seems like every webhook will be subscribed to all events. I don't think this is a common use case though, I would expect different channels (if more than 1 at all) to be subscribed to different events. Why don't we allow what events a webhook channel is subscribed to? Moreover, then the channel should also show the events it is subscribed to.
image

#### 7. Can't edit existing webhook 

In the above screenshot you can see the table listing the webhook channels. How can I edit an existing webhook? I tried clicking on the row in the table but that didn't work. I would have liked to add a name.

iannbing:
@yannickperrenet 

> The text in the banner (in the jobs list view) is not vertically aligned in the center. Left side is too high and right side is too low.

> I would change the text to: "Get notified about specific events in Orchest." The fact that we use webhooks is something they can see in the actual settings themselves.

Changed accordingly.

> I used https://webhook.site/ to quickly test out the TEST button when creating a webhook. The payload got successfully POST'ed but in our UI we show an ❗. Not sure how to interpret it.

You probably were testing an older commit. @fruttasecca helped out with a new endpoint a bit later. I just tested with `https://webhook.site/` and it works.

> After clicking the TEST in the Connection column I felt like the ✅ was not aligned properly (vertically speaking).

Could you share which browser you're using? I couldn't reproduce it on Safari, Chrome, and Brave on Mac. 

> In the above screenshot you can see that we offer a TEST functionality in the webhooks table. When I first saw this I immediately thought: "Why doesn't Orchest check it for me? Why do I have to press that button?"

Note that this "test" is a POST request that sends an actual notification, not a silent GET health check. It'd be better letting user fire it for testing instead of firing unsolicited notifications every time this view is rendered.

> From the UI it seems like every webhook will be subscribed to all events. I don't think this is a common use case though, I would expect different channels (if more than 1 at all) to be subscribed to different events. Why don't we allow what events a webhook channel is subscribed to? Moreover, then the channel should also show the events it is subscribed to.

> In the above screenshot you can see the table listing the webhook channels. How can I edit an existing webhook? I tried clicking on the row in the table but that didn't work. I would have liked to add a name.

AFAIK the agreed scope of this PR is a MVP: send notifications when job runs fail. I think the said scenarios are 100% valid, but design and endpoints are not yet in place (POST, GET, but no PUT).

iannbing:
> In addition, after I pressed TEST and added another webhook the ✅ from the first test disappeared.

Fixed 🙂

yannickperrenet:
Would it make sense to change the `Events` title in the notifications settings to `Subscribed events`? 

To me `Events` feels like a list of all possible events and not perse "events that I subscribed to". Given that the dialog for setting up a webhook doesn't allow for configuration of events, I think we should make clear that all events are automatically subscribed to.

ncspost:
> Would it make sense to change the Events title in the notifications settings to Subscribed events?

It makes more sense when the table headings + switches are there...

CleanShot 2022-06-01 at 17 01 48@2x

These switches can appear disabled for the MVP

ncspost:
CleanShot 2022-06-01 at 17 19 47@2x

It might be cleaner to structure events like this

fruttasecca:
Setting an invalid URL is not handled correctly, as a user I am allowed to save it, the request will fail and the FE will show an entry regardles.

iannbing:
> Setting an invalid URL is not handled correctly, as a user I am allowed to save it, the POST request to the BE will fail and the FE will show an entry regardless.

I intentionally made it so. Orchest could not be hold responsible for ensuring the receiver service is active. Even if the TEST is passing while user is creating a webhook in the dialog, the integration could still fail anytime.

BrunoOrchest (approved):

fruttasecca (approved):
lgtm :+1:

iannbing:
@ncspost 
I've addressed your feedback as much as possible. Some questions about the remainders:

- Why do we need to capticalize webhook? I think it's a common noun, not something like "Internet". See [wekipedia](https://en.wikipedia.org/wiki/Webhook) for example.
- I couldn't reproduce the functionality failures that you encountered. Could you deploy the latest commit and try again? It could be that you were testing on an old commit 🙂

ncspost:
CleanShot 2022-06-06 at 14 20 42@2x

Switches should be off and disabled until a webhook is created

iannbing:
> Switches should be off and disabled until a webhook is created

Fixed.

ncspost (approved):
LGTM! Great work ⭐

yannickperrenet (approved):
Super sick! I love it :1st_place_medal:

PR #1014, opened 2022-06-01
Add the PUT endpoint for webhook

iannbing:
## Description

This PR allows to update a webhook.


services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -60,6 +60,32 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+        subs = notification_utils.subscription_specs_to_subscriptions(
+            webhook.uuid, mutation["subscriptions"]
+        )
+        db.session.bulk_save_objects(subs)
+
+    for key in mutation:
+        if key != "subscriptions" and webhook[key] is not None:
+            webhook[key] = mutation[key]
+
+    db.session.commit()
```
yannickperrenet, line 84: To be in line with the `create_webhook` function, we should probably not be committing to the DB here and instead do that in the `PUT` endpoint.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -136,6 +136,25 @@ def get(self, uuid: str):
...
+            models.Subscriber.query.options(joinedload(models.Subscriber.subscriptions))
+            .filter(models.Subscriber.uuid == uuid)
+            .first()
+        )
+        if subscriber is None:
+            return {"message": f"Subscriber {uuid} does not exist."}, 404
+
+        if isinstance(subscriber, models.Webhook):
+            subscriber = marshal(subscriber, schema.webhook)
+            subscriber = webhooks.update_webhook(subscriber, request.get_json())
```
yannickperrenet, line 151: The `marshal` should probably happen last as it makes sure the return object is marshaled/as expected.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -71,6 +71,32 @@ def post(self):
...
+        changed. The original value will remain unchanged if not
+        mentioned in the mutation.
+        """
+        subscriber = (
+            models.Subscriber.query.options(joinedload(models.Subscriber.subscriptions))
+            .filter(models.Subscriber.uuid == uuid)
+            .first()
+        )
+        if subscriber is None or not isinstance(subscriber, models.Webhook):
+            return {"message": f"Webhook {uuid} does not exist."}, 404
```
fruttasecca, line 90: I think we can use the Webhook model here to make sure we are only fetching webhooks, and using `one_or_none()` after the filter might be preferable.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -71,6 +71,32 @@ def post(self):
         db.session.commit()
         return marshal(webhook, schema.webhook), 201
 
+    @api.doc("update_webhook")
+    @api.expect(schema.webhook_mutation, validate=True)
+    @api.response(200, "Success", schema.webhook)
+    def put(self, uuid: str):
```
fruttasecca, line 77: The put shouldn't be part of the `WebhookList` but it should be a `Resource` of its own so that we can use a URL like `/notifications/subscribers/webhooks/<uuid>` to PUT to.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -71,6 +71,32 @@ def post(self):
...
+        if subscriber is None or not isinstance(subscriber, models.Webhook):
+            return {"message": f"Webhook {uuid} does not exist."}, 404
+
+        try:
+            subscriber = webhooks.update_webhook(subscriber, request.get_json())
+        except Exception as e:
+            return {"message": f"Invalid webhook changes: {e}"}, 400
+
+        db.session.commit()
+        return marshal(subscriber, schema.webhook), 200
```
fruttasecca, line 98: Note that here you are returning the `subscriber` object that you fetched earlier, and given how sqlalchemy works this object might not contain (i.e. I don't remember how this would work) the updated subscriptions that you grabbed with `.options(joinedload(models.Subscriber.subscriptions))`, could you verify?

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -71,6 +71,32 @@ def post(self):
...
+            .filter(models.Subscriber.uuid == uuid)
+            .first()
+        )
+        if subscriber is None or not isinstance(subscriber, models.Webhook):
+            return {"message": f"Webhook {uuid} does not exist."}, 404
+
+        try:
+            subscriber = webhooks.update_webhook(subscriber, request.get_json())
+        except Exception as e:
+            return {"message": f"Invalid webhook changes: {e}"}, 400
```
fruttasecca, line 95: Might be preferable to be more specific with the exception types (maybe something like `except (ValueError, sqlalchemy.exc.IntegrityError) as e:`?) so that actual errors will bubble up at 500's

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -60,6 +60,32 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
     return webhook
 
 
+def update_webhook(webhook: models.Webhook, mutation: dict) -> models.Webhook:
```
fruttasecca, line 63: This comes down to personal preference but I am not sure we need to return the webhook given we are performing an in-place update and not creating a new object

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -72,6 +72,36 @@ def post(self):
...
+        mentioned in the mutation.
+        """
+        try:
+            webhook = (
+                models.Webhook.query.options(joinedload(models.Webhook.subscriptions))
+                .filter(models.Webhook.uuid == uuid)
+                .one_or_none()
+            )
+        except Exception as e:
+            return {"message": f"Webhook {uuid} does not exist. {e}"}, 404
```
yannickperrenet, line 94: Given the [`one_or_none()`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none) we should explicitly check whether `webhook is not None`, e.g.

```python
try:
    ...
except Exception as e:
    return {"message": f"Multiple webhooks with UUID {uuid} exist: {e}"}, 404
else:
    if webhook is None:
        return {"message": f"Webhook {uuid} does not exist. {e}"}, 404

File: services/orchest-api/app/app/core/notifications/webhooks.py (hunk 7 above)
Comment (yannickperrenet) [reply services/orchest-api/app/app/core/notifications/webhooks.py:63]: Afaik convention says that if an object is returned in-place you need to return `-> None` to indicate that an in-place change has taken place. So I agree with Jacopo here :)

File: services/orchest-api/app/app/core/notifications/webhooks.py [hunk 9]
```diff
@@ -60,6 +60,32 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+            raise ValueError(f"Invalid property: {key}")
+
+        if key != "subscriptions":
+            webhook[key] = mutation[key]
+        else:
+            subs = notification_utils.subscription_specs_to_subscriptions(
+                webhook.uuid, mutation[key]
+            )
+            db.session.bulk_save_objects(subs)
+
```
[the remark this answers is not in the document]
fruttasecca: Sounds good, @iannbing you are probably looking for something like partial unique indexes (maybe this can help? https://dba.stackexchange.com/questions/9759/postgresql-multi-column-unique-constraint-and-null-values) given that a subscription is characterized by an event type and some optional filters that can be NULL like the project uuid and job uuid. The partial part of the index is to avoid ("my event type", project_uuid, NULL) clashing with ("my event type", NULL, NULL) given how postgres treats NULLs in unique indexes.

services/orchest-api/app/app/apis/namespace_notifications.py
iannbing: I moved the logic into `update_webhook` function in `webhooks` file, and in the put function in `namespace_notifications` it simply catches the thrown error without specifying the specific type of the error. Could you confirm if it's a reasonable implementation?

services/orchest-api/app/app/core/notifications/webhooks.py
yannickperrenet: > I thought in Python, a function will return None if there's no explicit return. And -> None means that the function shouldn't return anything except None

Totally correct! 

> The return type does not indicate if a new object is created or it's an in-place change?

Strictly speaking you can `-> True` (for example) but still modify the passed object in-place. I would say it is just a convention some people stick with, i.e. when you see `-> None` you can guess that the object is changed in place (based on the function name). 

Hope that answers your questions.

services/orchest-api/app/app/apis/namespace_notifications.py
yannickperrenet: Yea I would say the changes in https://github.com/orchest/orchest/pull/1014/commits/1879e181ee490d5b040c8328f10a28318e2a3788 look good. Generally speaking it is better to be as strict as possible when it comes to exceptions, but in this case its fine to just catch `Exception` (which will catch almost everything in Python - [Python docs](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)).

services/orchest-api/app/app/apis/namespace_notifications.py
iannbing: I've adopted your suggestion by calling `models.Webhook.query` to get webhooks directly. So this shouldn't matter any more.

services/orchest-api/app/app/apis/namespace_notifications.py
```diff
@@ -72,6 +72,27 @@ def post(self):
...
+        """Updates a webhook, including its subscriptions.
+
+        The mutation only contains the values of a webhook to be
+        changed. The original value will remain unchanged if not
+        mentioned in the mutation.
+        """
+        try:
+            webhooks.update_webhook(uuid, request.get_json())
+        except Exception as e:
+            return {"message": f"Failed to update webhook. {e}"}, 400
```
fruttasecca, line 90: I am still of the opinion that we should distinguish different exceptions types to return 404s, 500s,properly, this can make debugging harder

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,93 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+        mutation: Same as webhook_spec, but all fields are optional.
+
+    """
+
+    try:
+        webhook = (
+            models.Webhook.query.options(joinedload(models.Webhook.subscriptions))
+            .filter(models.Webhook.uuid == uuid)
+            .one_or_none()
+        )
```
fruttasecca, line 82: Related to https://github.com/orchest/orchest/pull/1014/files#r888653702 , we could make use of `.one()` to get a `sqlalchemy.orm.exc.NoResultFound`

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,93 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+    except Exception as e:
+        raise ValueError(f"Multiple webhooks with UUID {uuid} exist: {e}")
+
+    if webhook is None:
+        raise ValueError(f"Webhook with UUID {uuid} not found.")
+
+    webhook.with_for_update()
+
```
fruttasecca, line 90: Would this actually work? In any case it's preferable to make it part of the query directly

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,93 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+    marshalled_mutation = marshal(mutation, schema.webhook_mutation)
+
+    for key in marshalled_mutation:
+        value = marshalled_mutation.get(key)
+
```
fruttasecca, line 95: No need to use `get` here since we already know the key is defined given we are iterating on those keys

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,93 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+                    if sub is not None:
+                        continue
+
+                    sub_specs_to_add.append(sub_spec)
+
+            # Add new subscriptions.
+            new_subs = notification_utils.subscription_specs_to_subscriptions(
+                webhook.uuid, sub_specs_to_add
+            )
```
fruttasecca, line 151: I am going to use this comment for different remarks about the logic.

```python
            subs = models.Subscription.query.filter_by(
                models.Subscription.subscriber_uuid == uuid
            )
            project_specific_subs = models.ProjectSpecificSubscription.query.filter_by(
                models.ProjectSpecificSubscription.subscriber_uuid == uuid
            )
```

This is not necessary due to sqlalchemy polymorphism, i.e. you can just query for subscribers and you will get a heterogeneous list of "normal" subscriptions, project level subs and job subs.

One possible issue is that, imo, while `subscription_specs_to_subscriptions()`  has to work only with subscription specs, here you are juggling with subscription specs and actual subscription entries. My suggestion would be to use `.subscription_specs_to_subscriptions` to reuse its logic (and it's checks for input correctness) to get a list of subscription entries, so that you can operate at a layer where you only have to think about those, and not specs.

The loop makes a query for every subscription in the list, this is unnecessary work, imo the logic can be simplified by:
- transforming the sub specs into sub entries
- getting all existing sub-entries for the subscriber
- now it's a matter of comparison between these two sets to find out which new sub entries should be persisted to the db

There are other alternatives to makthing things simplier, for example, assuming this https://github.com/orchest/orchest/pull/1014#discussion_r887767222 is implemented, you wouldn't need to find duplicate entries by managing what happens in case of a conflict throug a `on_conflict_do_update` (ctrl-f for examples in the codebase), while for deletion a single query (but a bit more complex imo) could take care of it. Given that the number of event types is low my suggestion would be to go for the first approach to keeps things simple.

Might also be better to encapsulate this logic it its own "private" function in the module

fruttasecca: Another alternative would be to delete all subscriptions of the subscriber and add the new sub specs. This is definitively the simplest solution, I am not too much into it due to the following:
- there might be, in the future, entities that depend on the subscription through a FK, and such behaviour would lead to hard to debug deletions later, 
- the subscriptions entries returned with a subscriber by the API contain a `uuid`, so replacing subs in this way would make it so that the `uuid` of a subscription that hasn't effectively changed becomes different. This is okaysh tbh, as long as we don't expose those to users.

With this said, atm this is definitively doable, @yannickperrenet what do you think?

yannickperrenet: > the subscription entries returned with a subscriber by the API contain a uuid, so replacing subs in this way would make it so that the uuid of a subscription that hasn't effectively changed becomes different. This is okayish tbh, as long as we don't expose those to users.

I guess this case will be hit as well when doing: (1) remove subscription (2) add subscription again. Leading to the `uuid` being changed, i.e. not the same as the original `uuid`. So I don't think this is much of an issue.

> there might be, in the future, entities that depend on the subscription through a FK, and such behaviour would lead to hard to debug deletions later

This is a valid point. 

> With this said, atm this is definitively doable, @yannickperrenet what do you think?

I agree it is the simplest solution, but it couldn't definitely bite us later on and doing right this first time doesn't seem too hard (if I understand it correctly). So I am leaning towards doing a proper update instead of a combined delete + insert.

services/orchest-api/app/app/core/notifications/webhooks.py
yannickperrenet: I would also guess (haven't checked it!) that it would need to be part of the actual query.

iannbing: I realized that after reading the doc of sqlalchemy. My bad 😅. Pushed the change.

services/orchest-api/app/app/schema.py
```diff
@@ -904,6 +904,28 @@ def _session_base_url(s) -> str:
...
+            required=False,
+            description="Content type of the payload, e.g. json, urlencoded, etc.",
+            enum=[
+                models.Webhook.ContentType.JSON.value,
+                models.Webhook.ContentType.URLENCODED.value,
+            ],
+        ),
+    },
+)
+
```
iannbing, line 928: @fruttasecca  This is basically a clone of `webhook_spec` but all attributes are set as optional. Is there a way to achieve this without copy & paste?

yannickperrenet: I wouldn't be surprised if such functionality isn't supported in `flask_restx`.

yannickperrenet (approved):
LGTM! Mr. Fullstack at it again :wink: 

(Probably a good idea to wait on @fruttasecca's approval as well)

services/orchest-api/app/app/schema.py
fruttasecca: don't think so unfortunately

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,91 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
+
+    try:
+        webhook = (
+            models.Webhook.query.with_for_update()
+            .filter(models.Webhook.uuid == uuid)
+            .one()
+        )
+        if mutation.get("subscriptions") is not None:
+            # Query subscriptions separately because with_for_update
+            # does not support joinload
```
fruttasecca, line 100:
```suggestion
            # does not support joinload.
```

services/orchest-api/app/app/core/notifications/webhooks.py
```diff
@@ -65,6 +65,91 @@ def create_webhook(webhook_spec: dict) -> models.Webhook:
...
 
 
+def _get_subscription_comparison_key(subscription: object) -> str:
+    event_type = getattr(subscription, "event_type")
+    project_uuid = getattr(subscription, "project_uuid", "")
+    job_uuid = getattr(subscription, "job_uuid", "")
+    return f"{event_type}|{project_uuid}|{job_uuid}"
+
+
+def _get_subscription_comparison_key_set(subscriptions: List[dict]) -> Set[str]:
```
fruttasecca, line 75: Both functions signature typing can be improved imo, since we are passing a `models.Subscription`

services/orchest-api/app/app/core/notifications/webhooks.py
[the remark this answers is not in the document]
iannbing: @fruttasecca Is [this change](https://github.com/orchest/orchest/pull/1014/commits/8fcb398bd15b62b7e5bca5459529c783ff9179b4) suffice? Is there any other place I need to update accordingly?

services/orchest-api/app/app/models.py
```diff
@@ -1627,6 +1633,12 @@ def check_constraints(mapper, connection, target):
             )
 
 
+UniqueConstraint(
+    ProjectJobSpecificSubscription.event_type,
+    ProjectJobSpecificSubscription.project_uuid,
+    ProjectJobSpecificSubscription.job_uuid,
+)
+
```
fruttasecca, line 1641: Was this tested? Because this might not work imo, and we have to use partial unique indexes as in https://github.com/orchest/orchest/pull/1014#discussion_r887767222. 

Two main issues: 
- even if this was a single model, the unique constraint wouldn't work since no subscriptions to the same event (and project uuid) from different subscribers could exist, i.e. you need to add the `subscriber_uuid` into the mix
- `ProjectJobSpecificSubscription` would lead to a constraint error when a `ProjectSpecificSubscription` is defined for the same project. This can be explained from two different points of view, 1) we are using single table inheritance so those constraints are applied to the same table, 2)  `ProjectJobSpecificSubscription` ISA `ProjectSpecificSubscription`. Essentially a `ProjectJobSpecificSubscription` with type X and project Y would clash with an existing `ProjectSpecificSubscription` of the same type and project.

Now that I've taken a look at the models again we might have an even simpler alternative to partial indexes, since to implement the single table inheritance there is a `type` field that allows to distinguish between "plain" subs, project subs and job subs, so we might get away with expressions constraints like:
- unique subscriber ID, subscription `type`, `event_type` on the `Subscriber` model. This gives us the uniqueness of a "plain" subscription
- for the ` ProjectSpecificSubscription` , same as above plus adding the `project_uuid` into the constraint defined on the ` ProjectSpecificSubscription` model
- for the ` ProjectJobSpecificSubscription` , same as above plus adding the ]`job_uuid` into the constraint defined on the ` ProjectJobSpecificSubscription` model

Also, after defining the constraints you will have to create a schema migration file, which in in most cases is completely automated, just follow https://docs.orchest.io/en/latest/development/development_workflow.html#database-schema-migrations

fruttasecca:
Great job on simplifying the overall logic btw :+1:

services/orchest-api/app/app/models.py
fruttasecca: See https://github.com/orchest/orchest/pull/1014/commits/bbef94c5e1937cad262f309d5b97e16e88d6f668

PR #1016, opened 2022-06-02
Allow Orchest installation without installing Argo

nhaghighat:
## Description

These changes let you specify whether or not the `orchest-controller` should install and manage an Argo Workflows deployment. Thanks to this option, users that already have Argo Workflows deployed in their cluster, can now also install Orchest. Previously, Orchest would still install Argo Workflows leading to problems due to cluster level resources such as the custom resources.

Fixes: #866

Through the CLI it now becomes possible to install Orchest without it also installing Argo:
```bash
orchest install --no-argo
```
This would remove the last "third-party" cluster-level resource that Orchest manages.

Installing into a custom namespace would be (under the assumption you don't have Argo installed on the cluster yet):
```bash
orchest install --namespace=myns
```

## Checklist

      compatibility is maintained.


services/orchest-controller/pkg/addons/addons.go
```diff
@@ -0,0 +1,106 @@
...
+	"k8s.io/klog/v2"
+)
+
+var (
+	// list of all addons
+	ArgoWorkflow   = "argo-workflow"
+	DockerRegistry = "docker-regitry"
+
+	// default orchest-system namespace
+	Namespace = "orchest-system"
```
yannickperrenet, line 17: I would rather put it in the same namespace that the `orchest-controller` is deployed in. So Orchest in its entirety is installed in one namespace.

nhaghighat: You mean orchest namespace?

nhaghighat:
CLI changes still has to be added.

services/orchest-controller/pkg/addons/addons.go
yannickperrenet: We can't hardcode `orchest` as the namespace, since the user might specify a custom namespace. That's why I meant "the same namespace that `orchest-controller` is deployed in".

nhaghighat: There should be some default namespace, in case controller is outside of the cluster?

nhaghighat: I'm leaning to have one namespace for controller and argo, in case we want to move to multi-tenancy.

yannickperrenet: Makes sense, in that case I would go with `orchest` to make sure it is the same one as in the generated `orchest-controller.yaml` file. Ideally the testing environment is as similar as possible as the prod environment (i.e. having all services in the same namespace)

nhaghighat: Sounds good, will do

orchest-cli/orchestcli/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.2.1"
+__version__ = "0.3.1"
```
yannickperrenet, line 1: This should be `0.3.0` not `0.3.1` right?

orchest-cli/orchestcli/cli.py
```diff
@@ -162,6 +162,15 @@ def cli():
     hidden=True,
     help="Run install in dev mode.",
 )
+@click.option(
+    "--disable-argo",
+    "disable_argo",  # name for arg
+    is_flag=True,
+    default=False,
+    show_default=True,
+    hidden=True,
```
yannickperrenet, line 171: This line should be removed. If `hidden=True` then the CLI option doesn't show up in the help menu.

orchest-cli/orchestcli/cli.py
```diff
@@ -162,6 +162,15 @@ def cli():
...
+    help="Disable Argo Workflow during controller deployment or not.",
```
yannickperrenet, line 172: I would rather go for something like: `"Disable deploying Argo as part of Orchest."`

orchest-cli/orchestcli/cli.py
```diff
@@ -210,11 +223,21 @@ def uninstall(**common_options) -> None:
...
+)
```
yannickperrenet, line 234: What is the reasoning behind adding `--disable_argo` to `orchest update`? Doesn't that break updating Orchest through the UI (given that users can't specify how to update Orchest, i.e. with or without Argo)?

orchest-cli/orchestcli/cmds.py
```diff
@@ -156,13 +158,21 @@ def install(cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **kwargs) -> Non
         if e.reason == "Conflict":
             echo(f"Installing into existing namespace: {ns}.")
 
+    if disable_argo:
+        echo(
+            "ArgoWorkflow installation is disabled, Orchest Controller assumes"
+            "ArgoWorkflow is already installed on your cluster"
+        )
```
yannickperrenet, line 165: This would miss a space when printed, it should instead be:
```suggestion
        echo(
            "ArgoWorkflow installation is disabled, Orchest Controller assumes"
            " ArgoWorkflow is already installed on your cluster."
        )
```
In addition, I would like to either print a message explaining the permissions that Argo needs to be compatible with Orchest or have a link to the docs. I think I prefer the latter.

orchest-cli/orchestcli/cmds.py
```diff
@@ -156,13 +158,21 @@ def install(cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **kwargs) -> Non
...
+
+        manifest_file_name = "orchest-controller-disable-argo.yaml"
```
yannickperrenet, line 167: I don't think this file (`orchest-controller-disable-argo.yaml`) is added to the release assets yet, or is it? In addition, it isn't added to the `services/orchest-controller/Makefile` either.

orchest-cli/orchestcli/cmds.py
```diff
@@ -469,17 +482,27 @@ def lte(old: str, new: str) -> bool:
         )
         sys.exit(1)
 
+    if disable_argo:
+        echo(
+            "ArgoWorkflow installation is disabled, Orchest Controller assumes"
+            "ArgoWorkflow is already installed on your cluster"
+        )
```
yannickperrenet, line 489: This message needs to be updated as well. As per https://github.com/orchest/orchest/pull/1016/commits/81e661371b0c46b335e78357ab7c8fb2caba7aae#r889919959.

If we choose to include the option on `orchest update` as well.

orchest-cli/orchestcli/cli.py
nhaghighat: Well, we need to know it to realize which manifest should be applied, 
I can also add an annotation to orchest-controller deployment, then cli can read it and determine if argo is disabled or not, but then the question would be how user can change the setting of argo deployment in update, for example it accidentally installed orchest in the cluster without argo and now wants to update it but do wants argo?

fruttasecca: Do we support changing from argo to non argo? Perhaps it should be part of `orchest patch` rather than `orchest update`?

orchest-cli/orchestcli/cli.py
```diff
@@ -170,10 +179,14 @@ def cli():
 )
 @cli.command(cls=ClickCommonOptionsCmd)
 def install(
-    cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **common_options
+    cloud: bool,
+    dev_mode: bool,
+    disable_argo: bool,
```
fruttasecca, line 184: Purely a personal POV, but as a user I would find `--no-argo` clearer than `--disable-argo` as an installation flag

orchest-cli/orchestcli/cli.py
fruttasecca: This approach might pay off dividends in the future if/when we allow doing the same, for example, with the docker registry.

services/orchest-controller/cmd/controller/cmd/controller.go
```diff
@@ -94,6 +93,12 @@ func NewControllerCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&serverConfig.Endpoint,
 		"endpoint", serverConfig.Endpoint, "The endpoint of Http Server")
 
+	cmd.PersistentFlags().StringSliceVar(&addonsConfig.Addons, "addons", addonsConfig.Addons,
+		"the default addons to enable on controller deployment")
+
+	cmd.PersistentFlags().StringVar(&addonsConfig.AssetDir,
+		"assetDir", addonsConfig.AssetDir, "The directory of assets")
+
```
fruttasecca, line 101: Strictly informative, what directories are important to the controller? `addons`, `assetDir` and?

services/orchest-controller/pkg/addons/addons.go
```diff
@@ -0,0 +1,106 @@
...
+	"path"
+
+	orchestv1alpha1 "github.com/orchest/orchest/services/orchest-controller/pkg/apis/orchest/v1alpha1"
+	"k8s.io/klog/v2"
+)
+
+var (
+	// list of all addons
+	ArgoWorkflow   = "argo-workflow"
+	DockerRegistry = "docker-regitry"
```
fruttasecca, line 14:
```suggestion
	DockerRegistry = "docker-registry"
```

fruttasecca:
Really like the addons approach :)!

orchest-cli/orchestcli/cli.py
nhaghighat: > Do we support changing from an argo deployed in "our" namespace to one controlled by users and viceversa? Perhaps it should be part of orchest patch rather than orchest update? Might be less prone to errors and easier to test and revert back. I don't feel like such changes belong to an update procedure

sounds good, simpler and more reliable.

fruttasecca:
Are there any plans to add user-facing docs about the concept of addons or to highlight the fact that deploying with no argo is possible?

orchest-cli/orchestcli/cli.py
nhaghighat: I like `--no-argo` better, will do that.

services/orchest-controller/cmd/controller/cmd/controller.go
nhaghighat: I think only `assetDirs`

nhaghighat:
> Are there any plans to add user-facing docs about the concept of addons or to highlight the fact that deploying with no argo is possible?

Sure, will add that

services/orchest-controller/cmd/controller/cmd/controller.go
fruttasecca: I see, thank you!

fruttasecca:
LGTM assuming changes w.r.t. https://github.com/orchest/orchest/pull/1016#discussion_r889963365

fruttasecca (approved):

orchest-cli/orchestcli/cli.py
yannickperrenet: I would say we only have to support the following two paths:
1. User does not want to install with Argo and we assume the user never wants us to install Argo. 
2. User wants to install Argo and we assume the user always wants us to update it.

These can be summarized as: `orchest update` should "update" the current cluster configuration (not change it). Thus Argo not installed? Then it is also not updated.

> Well, we need to know it to realize which manifest should be applied,

So then why do we let the `orchest-cli` create the manifest? Can't we just specify it as part of the CRD, e.g. `manageArgo`?

orchest-cli/orchestcli/cli.py
yannickperrenet: I agree :+1:

orchest-cli/orchestcli/cli.py
nhaghighat: > I do like this. Sounds like that means we need managing/disabling Argo to be part of the CRD instead then.

I don't think having it as part of CRD would be a good idea. the orchest CRD should hold the information of the one orchest deployment, but argo is like a cluster component. if we add it to crd, and we supported multi tenancy, then what should controller do if it is enabled in 1 CR and disabled in another?

nhaghighat: This is the only outstanding point of this PR.

although I agree that during update it should preserve the setting of argo-workflow deployment, but the question is how:
one approach would be that orchest-cli reads the current deployment and detects if it has argo or not, and based on that apply the appropriate manifest, this means hard-coding the name of orchest deployment in `orchest-cli `which is not very nice.

orchest-cli/orchestcli/cli.py
```diff
@@ -210,11 +222,21 @@ def uninstall(**common_options) -> None:
     hidden=True,
     help="Run update in dev mode.",
 )
+@click.option(
+    "--no-argo",
+    "no_argo",  # name for arg
+    is_flag=True,
+    default=False,
+    show_default=True,
+    hidden=True,
```
yannickperrenet, line 231: This should also not be `hidden=True`.

services/orchest-controller/pkg/addons/addons.go
yannickperrenet: I don't think this has been resolved yet. The typo is still there.

nhaghighat:
> I can't run the `scripts/build_container.sh` script anymore, because the `services/orchest-controller/Makefile` fails to generate the manifests.
> 
> ```
> make: Entering directory '/home/orchest/services/orchest-controller'
> Updating CRD manifests
> /home/orchest/.cache/tools/controller-gen crd paths=./pkg/apis/... output:dir=deploy/helm/templates
> Generating orchest-controller.yaml manifests
> Error: template: orchest-controller/templates/deployment.yaml:16:14: executing "orchest-controller/templates/deployment.yaml" at <eq .Values.addons.enableArgo "true">: error calling eq: incompatible types for comparison
> 
> Use --debug flag to render out invalid YAML
> make: *** [Makefile:59: manifestgen] Error 1
> make: Leaving directory '/home/orchest/services/orchest-controller'
> ```

@yannickperrenet : fixed

orchest-cli/orchestcli/cli.py
nhaghighat: I know, but I postponed changing it once we decide the approach on `orchest update`

orchest-cli/orchestcli/cli.py
yannickperrenet: > how to apply the correct configuration in update?

By not making it the responsibility of `orchest update`. E.g. putting it in the CRD, given that we don't support multiple OrchestClusters anyways. Later we could introduce another CRD that contains all configuration options just for the `orchest-controller` if we ever happen to have many of them.

orchest-cli/orchestcli/cmds.py
```diff
@@ -240,6 +244,26 @@ def install(cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **kwargs) -> Non
         )
         sys.exit(1)
 
+    applications = [{"name": "docker-registry", "config": {}}]
+
+    if no_argo:
+        echo(
+            "ArgoWorkflow installation is disabled, orchest will be "
+            "deployed without it, make sure ArgoWorkflow is already installed "
+            "on your cluster otherwise orchest will not act properly."
```
yannickperrenet, line 253: "will not act properly" --> "won't work correctly".

In addition, we should point them to our docs so they make sure that their `ArgoWorkflow` has the correct privileges set.

orchest-cli/orchestcli/cmds.py
```diff
@@ -240,6 +244,26 @@ def install(cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **kwargs) -> Non
...
+        )
+    else:
+        applications.append(
+            {
+                "name": "argo-workflow",
+                "config": {
+                    "helm": {
+                        "parameters": [{"name": "singleNamespace", "value": "true"}]
```
yannickperrenet, line 261: What does the `singleNamespace` parameter mean?

services/orchest-controller/deploy/helm/values.yaml
```diff
@@ -2,8 +2,12 @@
...
-images:
+image:
   pullPolicy: IfNotPresent
   registry: docker.io
   name: orchest/orchest-controller
   tag: latest
+
+
+addons:
+  enableArgo: false
```
yannickperrenet, line 13: I think we want to have `enableArgo` set to `true` by default.

nhaghighat: This is for enabling argo as a cluster level component. the default behavior would be to disable it as cluster level component and install argo in the namespace of orchest as a namespace level component.

docs/source/getting_started/installation.rst
```diff
@@ -89,6 +86,15 @@ Next, make Orchest reachable through your FQDN reachable from the browser:
    # /etc/hosts so that you can reach Orchest locally.
    echo "$(minikube ip)\tlocalorchest.io" >> /etc/hosts
 
+Installing without ``argo-workflow``
```
yannickperrenet, line 89: Can you add something about the privileges that need to be set for Argo to work with Orchest?

orchest-cli/orchestcli/cmds.py
nhaghighat: It deploys argo as a namespace level component, so it only schedules workflows in it's own namespace.

services/orchest-controller/deploy/helm/values.yaml
yannickperrenet: Awesome! Makes sense :+1: 

> This is for enabling argo as a cluster level component.

Maybe the name of the parameter should better reflect that? What do you think?

docs/source/getting_started/installation.rst
nhaghighat: I can point them to the argo installation doc.

services/orchest-controller/deploy/helm/values.yaml
nhaghighat: makes sense, something like `clusterAddons`?

yannickperrenet: > makes sense, something like clusterAddons?

Yea I think that would be better, or maybe even the more verbose `clusterLevelAddons`? Whichever you prefer really (no strong opinion from my end) :)

Issue #556, opened 2021-11-24
Deletion of a single pipeline step using keyboard


yannickperrenet:
**Current problem**  
It is possible to delete pipeline steps using `<del>` and `<backspace>`, however, it is quite difficult to use them on a single step.

1. Create a pipeline
2. Select a single step
3. Press `<del>` or `<backspace>` to delete the step

Note how the step properties pane (on the right) open automatically and focuses the title text field. Because the text field is focused, the `<del>` and `<backspace>` don't operate on the step but on the text field.

**Describe the solution you'd like**  
So the solution should make the `<del>` and `<backspace>` work without having to deselect the text field. Note how deletion for multiple steps is straightforward because it doesn't have this behavior.

yannickperrenet:
To get to a working solution, we decided on:

- If you create the step, then auto-focus the title textfield. --> keyboard deletion on step doesn't work without un-focussing the title text field.
- If you select an existing step, then it doesn't auto-focus the title textfield. --> keyboard deletion on step works

PR #1017, opened 2022-06-03
Allow deleting steps using the keyboard

BrunoOrchest:
## Description

When user selects a step, the `Title` field is no longer auto focused to allow the delete/backspace keys to delete the step. In the case the step was just added, the `Title` field is focused so the user can start writing normally

Fixes: #556

## Checklist

- [ ] In case I changed one of the services’ `models.py` I have performed the appropriate database
- [ ] In case I changed code in the `orchest-cli` I followed its [release
- [ ] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards
      compatibility is maintained.
- [ ] In case I changed the dependencies in any `requirements.in` I have run `pip-compile` to update

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/hooks/useEventVars.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/hooks/useEventVars.tsx
@@ -39,6 +39,7 @@ export type EventVars = {
     y2: number;
     active: boolean;
   };
+  shouldAutoFocus: boolean;
   error?: string | null;
   timestamp: number | undefined;
   subViewIndex: number;
@@ -393,6 +394,7 @@ export const useEventVars = () => {
             ...updated,
             openedStep: newStep.uuid,
             subViewIndex: 0,
+            shouldAutoFocus: true,
             ...selectSteps([newStep.uuid]),
           });
         }
@@ -424,6 +426,7 @@ export const useEventVars = () => {
           return withTimestamp({
             ...state,
             ...updated,
+            shouldAutoFocus: true,
             ...selectSteps(newSteps.map((s) => s.uuid)),
           });
         }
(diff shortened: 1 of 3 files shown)


yannickperrenet (approved):
Works like a charm! :+1: 

I quite like this behavior, feels "natural".

iannbing:
Currently it seems that we have to carefully set the value of `shouldAutoFocus` back to `false` case by case. Is there a simpler way, for example, always reset `shouldAutoFocus` to `false` for all actions except for`CREATE_STEP` or `DUPLICATE_STEPS`?

BrunoOrchest:
> Currently it seems that we have to carefully set the value of `shouldAutoFocus` back to `false` case by case. Is there a simpler way, for example, always reset `shouldAutoFocus` to `false` for all actions except for`CREATE_STEP` or `DUPLICATE_STEPS`?

Not necessary since the `SELECT_STEPS` and related will take care of that. Other actions don't trigger the properties visibility so no need to be 'spamming' the code all over since the selection/deselection will take care of it always.

iannbing:
@BrunoOrchest  You could set `shouldAutoFocus` as false before the switch instead of "spamming" in each case. 

```
const state  = { ...currentState, shouldAutoFocus: false };

switch (action.type) {

     case "CREATE_STEP": {
          return {
                ...
                shouldAutoFocus: true,
           }
     } 

     // No need to set shouldAutoFocus to false in any other case.
}


```

The current implementation could lead to missing cases. For example:

- there's an existing step `A`
- create a new step
- Immediately select step `A`: the title field is auto-focused. (because in case `SET_OPENED_STEP`,  `shouldAutoFocus` is not set to `false`. Was it intended?).

Issue #451, opened 2021-08-31
Mass copy of parameters / environment variables


ricklamers:
**Describe the solution you'd like**  
Mass copy of parameters / environment variables

**What does your solution aim to solve?**  
It's often tedious to copy a set of environment variables / parameters from one project to another. One might do this for various reasons (e.g. when importing someone else's version of a project).

There are several views for which this should be implemented. E.g. the project environment variables view, pipeline environment variables view, job edit view, job view.

PR #1024, opened 2022-06-06
Add support for loading job pipeline parameters with JSON files

ricklamers:
## Description

Add support for loading job pipeline parameters with JSON files.

image

The problem this solves is when you need to run a large number configurations of your pipeline through parameterization. Ideally you'd be able to programmatically specify the set of parameters you'd like to run.

E.g. if you have a list of databases that a pipeline should run for (passed as a pipeline parameter, "database-1", ..., "database-n") you might want to generate this file with code or in an editor outside of Orchest **instead of tediously entering it through the GUI** per step/pipeline parameter.

## Todos
- [x] Add a way to generate a populated JSON file from the currently configured step/pipeline parameters (as a starting point, or as a reference for how to generate the file programmatically)

## Checklist

      compatibility is maintained.

## Bug fixes
This also contains two bug fixes:
- A notebook step with a pipeline assigned from `/data/...` would prevent saving the pipeline json
- Dragging out a file from the file manager to the canvas if your pipeline doesn't sit in the project root would not adjust the paths relative to the pipeline file correctly


docs/source/fundamentals/jobs.rst
```diff
@@ -94,6 +94,34 @@ Then inside the pipeline step you can access the parameters as follows:
...
+.. code-block:: json
+
+   {
+      "pipeline_parameters": {
+         "some_key": ["a", "list", "of", "values"]
+      },
+      "62a62810-336c-44c4-af6a-35228e8f2028": {
+         "some_key": [1, 2, 3],
+         "another_key": [1]
+      }
```
yannickperrenet, line 120: I don't really like having to refer to a step using its UUID. From a user perspective this is not a great experience. Given that this file is not auto-generated but created by a user, we need to focus on this user experience.

yannickperrenet:
Could you explain (and also changing the PR description) what problem this PR solves? Additionally, could you add a better title and label to the PR so that it can be included in the release notes. To users skimming of the release notes the current title doesn't really reflect what was done in this PR.

If I remember correctly @ncspost already created a design to solve #451. Doesn't that already solve this problem as well?

docs/source/fundamentals/jobs.rst
```diff
@@ -94,6 +94,34 @@ Then inside the pipeline step you can access the parameters as follows:
...
+You can easily run a pipeline for multiple parameter configurations by creating
+a parameter file.
+
+If you place the file in the same location as your pipeline file the parameter file will automatically be detected when creating a job.
+
+For a pipeline called ``main.orchest`` the parameter file should be name ``main.parameters.json``, and be put in the same location as the pipeline file (both in the project directory).
+
+You can also select a file manually when creating a job.
+
+The JSON file should be formatted as below. **Note that wrapping the values in a list is required, event if you're assigning just one parameter value to a key.** It is allowed to omit keys you don't want to specify.
```
yannickperrenet, line 109: `event if` -> `even if`

docs/source/fundamentals/jobs.rst
```diff
@@ -94,6 +94,34 @@ Then inside the pipeline step you can access the parameters as follows:
...
 
+Specify job parameters with a file
+----------------------------------
+
+You can easily run a pipeline for multiple parameter configurations by creating
+a parameter file.
+
+If you place the file in the same location as your pipeline file the parameter file will automatically be detected when creating a job.
+
+For a pipeline called ``main.orchest`` the parameter file should be name ``main.parameters.json``, and be put in the same location as the pipeline file (both in the project directory).
```
yannickperrenet, line 105: `should be name` -> `should be named`

Also, what does it mean "in the same location"? Does it work if we have `/project-dir/pipeline.orchest` and `/project-dir/subdir/pipeline.parameters.json`? And I guess we let users themselves rename the parameters file after renaming the pipeline?

ricklamers: > Does it work if we have /project-dir/pipeline.orchest and /project-dir/subdir/pipeline.parameters.json?

That won't work. I should probably say "needs to be in the same directory".

ricklamers:
Great feedback!

1. That's odd. https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText shows this should be supported in chromium based browsers. Are your specific browser settings/extensions aggressively blocking clipboard access?

E.g. on this random StackOverflow page (scroll down, slick share, click "Copy link")

The idea behind this feature is that users can _programatically_ generate their own JSON files. The "Generate parameter file" is just a way to check the schema. At least, that's how I envisioned it.

2. Moved docs link to the pipeline settings page below the button. Discussed with @fruttasecca but couldn't come up with a better place to put it. Since the file is part of editing your project files I feel like it needed to be in the pipeline editor or in the pipeline settings. Those are the two places where GUI actions can modify your project files.

3. The idea is of course to not let people discover the concept of parameter files at the point where they're already editing the job. The envisioned flow is: you create parameters, you might create a parameter file if you have many parameters you want to specify for all the jobs that you create for this pipeline, then you go and create an actual job that makes use of the parameter file.

4. Like I mentioned, the value is to simplify communicating the "schema" to the user. That includes in this case the step UUIDs that might be annoying to copy and paste. Note to match the UUID string with the step the user still needs to look in the pipeline file.

fruttasecca:
It keeps saying that the file can't be found in the given path, despite automatically initializing the parameters with the file.
image

fruttasecca:
I am no UX expert but imo this part (see picture) should be behind an "advanced" button or something. I am afraid there is a risk that the fact that you can click on a parameter value and fill in parameters through the GUI has become less evident, the "LOAD PARAMETERS" button does capture the attention. In a sense, those are two competing ways of setting the parameters of a job, and we aren't explaining to users when they should pick one or the other, imo.
image

yannickperrenet:
> Like I mentioned, the value is to simplify communicating the "schema" to the user.

If that is the case, then maybe we should show the schema instead? Calling it "generate parameters file" sounds like the purpose of clicking it is different from "communicating the schema".

> I am afraid there is a risk that the fact that you can click on a parameter value and fill in parameters through the GUI has become less evident, the "LOAD PARAMETERS" button does capture the attention.

I hadn't thought of it this way yet, but I definitely agree with this sentiment.

services/orchest-webserver/client/src/edit-job-view/EditJobView.tsx
```diff
@@ -267,48 +235,184 @@ const EditJobView: React.FC = () => {
...
+            </span>
+          );
+          resolve();
+        } catch (error) {
+          if (error.status !== 404) {
+            console.error(error);
+          }
+          reject();
+        }
+      });
```
iannbing, line 340: This pattern is quite new to me, because async function already returns a promise.

```
new Promise<void>(async (resolve, reject) => {})
```

From the code I think the sole purpose is to `console.log` when the response is not 404? And actually the following should have the same effect, but we don't need to mix using `new Promise` and `async`:

```
async () => {
      try {
            await somePromise()
      }catch(error){
          if (error.status !== 404) {
            console.error(error);
          }
          return Promise.reject(); // Reject this promise intentionally.
      }
}

```

The difference is that we don't need to use `new Promise`

ricklamers: Let's discuss it over a call, it's not about the console.log but about the function explicitly triggering the catch clause for the caller to make use of (hence returning a Promise to make it explicit that this async return can be handled for both success and failure case). Mostly style v.s. using a "regular" async function.

Issue #876, opened 2022-04-19
Allow deletion of project from project settings


ricklamers:
**Describe the problem this improvement solves**
It's a bit annoying to have to search through a long list of projects to delete it.

**Describe the solution you'd like**
Allow deletion of project from project settings

**Screenshots**
Somewhere on this page:

![image](https://user-images.githubusercontent.com/1309307/164003568-8c696112-8e0b-4eba-96eb-89d443170157.png)

yannickperrenet:
Added in #1060.

PR #1027, opened 2022-06-07
Add delete project to Project Settings

BrunoOrchest:
## Description

Description of what your PR solves or adds.

Fixes: #876

## Checklist

      compatibility is maintained.

### DIFF
--- a/services/orchest-webserver/client/src/contexts/AppContext.tsx
+++ b/services/orchest-webserver/client/src/contexts/AppContext.tsx
@@ -46,6 +46,7 @@ export type Alert = {
 export type Confirm = {
   type: "confirm";
   title: string;
+  warning?: boolean;
   content: string | React.ReactElement | JSX.Element[];
   onConfirm: () => Promise<boolean> | boolean; // if it's confirm type, something needs to happen. Otherwise, it could have been an alert.
   onCancel?: () => Promise<boolean | void> | void | boolean;
@@ -64,6 +65,7 @@ type AlertConverter = (props: {
 
 type ConfirmConverter = (props: {
   title: string;
+  warning?: boolean;
   content: string | React.ReactElement | JSX.Element[];
   confirmHandler: () => Promise<boolean> | boolean;
   cancelHandler?: () => Promise<boolean | void> | void | boolean;
@@ -132,7 +134,8 @@ export type ConfirmDispatcher = (
         onCancel?: CancelHandler;
         confirmLabel?: string;
         cancelLabel?: string;
-      }
+      },
+  warning?: boolean
 ) => Promise<boolean>;
 
 type PromptMessageDispatcher<T extends PromptMessage> = T extends Alert
@@ -231,7 +234,8 @@ const withPromptMessageDispatcher = function <T extends PromptMessage>(
           onCancel?: CancelHandler;
           confirmLabel?: string;
           cancelLabel?: string;
-        }
+        },
+    warning?: boolean
   ) => {
     // NOTE: consumer could either provide a callback function for onConfirm (for most use cases), or provide an object for more detailed config
     return new Promise<boolean>((resolve) => {
@@ -263,6 +267,7 @@ const withPromptMessageDispatcher = function <T extends PromptMessage>(
       const message = convert({
         title,
         content,
+        warning,
         confirmHandler,
         cancelHandler,
         confirmLabel,
@@ -298,6 +303,7 @@ const convertAlert: PromptMessageConverter<Alert> = ({
 
 const convertConfirm: PromptMessageConverter<Confirm> = ({
   title,
+  warning,
   content,
   confirmHandler,
   cancelHandler,
@@ -307,6 +313,7 @@ const convertConfirm: PromptMessageConverter<Confirm> = ({
   return {
     type: "confirm",
     title,
+    warning,
     content: contentParser(content),
     onConfirm: confirmHandler,
     onCancel: cancelHandler,
--- a/services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
+++ b/services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
@@ -161,7 +161,8 @@ const ProjectSettingsView: React.FC = () => {
           setIsDeletingProject(true);
           return true;
         }
-      }
+      },
+      true
     );
   };
 
--- a/services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
+++ b/services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
@@ -18,10 +18,10 @@ import {
 import ArrowBackIcon from "@mui/icons-material/ArrowBack";
 import DeleteIcon from "@mui/icons-material/Delete";
 import SaveIcon from "@mui/icons-material/Save";
+import { Dialog } from "@mui/material";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import LinearProgress from "@mui/material/LinearProgress";
-import Modal from "@mui/material/Modal";
 import { fetcher, HEADER } from "@orchest/lib-utils";
 import React from "react";
 import { Link } from "react-router-dom";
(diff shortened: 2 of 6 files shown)


services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -126,6 +131,33 @@ const ProjectSettingsView: React.FC = () => {
     fetchSettings();
   }, []);
 
+  const deleteProject = () => {
+    let deletePromise = makeRequest("DELETE", "/async/projects", {
```
iannbing, line 135: I looked into the current implementation of deleting a project, and found two appearances:
- `deleteProject` in `ImportDialog`, using `fetcher`
- `deleteProjectRequest` in `ProjectsView`, using `makeRequest`

We are deprecating `makeRequest` and use `fetcher` instead.

```
import { fetcher } from "@orchest/lib-utils";

```

It'd be even better if we clean up a bit by moving `deleteProject` to a separate file and import it in `ImportDialog`,  `ProjectsView` and `ProjectSettingsView`.

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -138,6 +170,25 @@ const ProjectSettingsView: React.FC = () => {
 
   return (
     <Layout>
+      <Modal open={isDeletingProject}>
```
iannbing, line 173: Is there any reason you chose `Modal` over `Dialog`? Quote from [MUI](https://mui.com/material-ui/react-modal/)

> If you are creating a modal dialog, you probably want to use the [Dialog](https://mui.com/material-ui/react-dialog/) component rather than directly using Modal. Modal is a lower-level construct that is leveraged by the following components:

If we must create an exception here, it would be nice to leave a comment as a reference 🙂.

ncspost:
CleanShot 2022-06-08 at 13 33 00@2x
 
I'd even potentially hide this behind a dropdown to add a bit more friction

BrunoOrchest:
> CleanShot 2022-06-08 at 13 26 00@2x

This seems better but to implement, it would be part of the overall redesign of the page. Should we just try to find a quick solution for the button for now, and then focus on the re-design?

ncspost:
> This seems better but to implement, it would be part of the overall redesign of the page. Should we just try to find a quick solution for the button for now, and then focus on the re-design?

Sorry - forget to say ignore the other bits.

The only stuff that's in scope is what I listed above.

BrunoOrchest:
Adding a confirmation text field Ala Gitlab will be tricky without re-designing the warning dialog as it doesn't really support validation for the Confirm button (afaik). Will this be sufficient for now until the redesign @ncspost ?

Screenshot 2022-06-09 at 10 15 30

ncspost:
That's fine @BrunoOrchest - let's update the copy and styles though...

- Heading: Delete [project name]?
- Body: "Warning: Deleting a Project is permanent. All associated Jobs and resources will be deleted and unrecoverable."
- Cancel button: Blue text
- Confirm: Change to Red filled button with "Delete Project" copy

BrunoOrchest:
> - Cancel button: Blue text
> - Confirm: Change to Red filled button with "Delete Project" copy

Changing these will change also the colours on other dialogs throughout. (We can make it specific for this only I think, but then it will be inconsistent with other dialogs we have throughout the app for other destructive and non-destructive actions)

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -138,6 +180,25 @@ const ProjectSettingsView: React.FC = () => {
...
+        <Box
+          sx={{
+            position: "absolute",
+            top: "50%",
+            left: "50%",
+            transform: "translate(-50%, -50%)",
+            width: "20%",
+            bgcolor: "background.paper",
+            boxShadow: 24,
+            p: 4,
```
iannbing, line 193: I don't think this styling is working.

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -138,6 +180,25 @@ const ProjectSettingsView: React.FC = () => {
 
   return (
     <Layout>
+      <Dialog open={isDeletingProject}>
```
iannbing, line 183: This dialog is probably never rendered? `setIsDeletingProject(true);` is run in `finally`

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -126,6 +135,39 @@ const ProjectSettingsView: React.FC = () => {
...
+      return;
+    }
+    setConfirm(
+      "Warning",
+      "Are you certain that you want to delete this project? This will kill all associated resources and also delete all corresponding jobs. (This cannot be undone.)",
+      async (resolve) => {
+        try {
+          if (projectsState.projectUuid === projectUuid) {
+            dispatch({ type: "SET_PROJECT", payload: undefined });
+          }
```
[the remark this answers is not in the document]
BrunoOrchest: Kept the same logic as in other place where this is set before resolving the promise. Changing.

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
BrunoOrchest: Refactoring error, it should be set to true before doing the call, and then to false in finally.

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
iannbing: There are a few complementary components you could use with `Dialog`. See [MUI doc](https://mui.com/material-ui/react-dialog/#main-content) or check some implementation in our codebase, for example [SystemDialog](https://github.com/orchest/orchest/blob/master/services/orchest-webserver/client/src/components/SystemDialog.tsx).

Screenshot 2022-06-09 at 15 00 40

ncspost:
Looks great! Just need to change copy of the `confirm` button to `Delete Project` please!

services/orchest-webserver/client/src/contexts/AppContext.tsx
```diff
@@ -51,6 +51,7 @@ export type Alert = {
 export type Confirm = {
   type: "confirm";
   title: string;
+  warning?: boolean;
```
iannbing, line 54: Not sure if it's a good idea to add a workaround to a very generic implementation that is used across the whole app. In general, I would propose limiting a workaround within a component, and be more strict on things like `AppContext` and `SystemDialog`.

What about properly add a more usable property to `callbackOrParams`, for example: 

```

callbackOrParams?:
    | ConfirmHandler
    | {
        onConfirm: ConfirmHandler;
        onCancel?: CancelHandler;
        confirmLabel?: string;   // you can use this to change "DELETE" to "DELETE PROJECT"
        cancelLabel?: string;
        // extra property here, something like `confirmButtonColor`?
      }

```

If it's too much work for this specific task, implementing a Dialog directly in `ProjectSettingsView` could also be a good option imo. We can improve `setConfirm` later.

services/orchest-webserver/client/src/views/ProjectSettingsView.tsx
```diff
@@ -126,6 +135,41 @@ const ProjectSettingsView: React.FC = () => {
...
 
+  const { setConfirm } = useAppContext();
+  const deleteCurrentProject = () => {
+    if (!projectUuid) {
+      return;
+    }
+    setConfirm(
+      `Delete ${state.path}?`,
+      "Warning: Deleting a Project is permanent. All associated Jobs and resources will be deleted and unrecoverable.",
+      async (resolve) => {
```
iannbing, line 146: To specify Confirm button label, you could change the third arg an object `callbackOrParams`, see [useCheckUpdate](https://github.com/orchest/orchest/blob/7671975e58277ba104b6533955293e4e12acf86d/services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx#L138) for example.

yannickperrenet (changes requested):
Seems like there are still a few open comments on the PR, but other than that I think it looks good :+1:

iannbing:
@yannickperrenet @mausworks  This UX issue #876 would most likely be gone (or very trivial) in the new design. I'm not sure if it's still worth it to work on this PR.

mausworks:
Discussed this with @iannbing, I'm closing this PR as the settings-page redesign replaces this.

PR #1033, opened 2022-06-07
Update environment_variables.rst

ncspost:
## Description

Clean up copy

Fixes: #issue

## Checklist

### DIFF
--- a/docs/source/fundamentals/environment_variables.rst
+++ b/docs/source/fundamentals/environment_variables.rst
@@ -2,72 +2,57 @@
 
 Environment variables
 =====================
-Defining environment variables can be useful when you want to avoid persisting sensitive data in your
-versioning system. These environment variables will be injected in your pipeline steps, and can be
-retrieved using the native way of your language of choice, for example in Python:
+
+Environment variables let you exclude sensitive data from your versioning system. 
+
+Environment variables are injected in your pipeline steps, and can be natively retrieved with your chosen language. For example, in Python:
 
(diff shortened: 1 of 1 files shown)


astrojuanlu (approved):
🚀

docs/source/fundamentals/environment_variables.rst
```diff
@@ -2,72 +2,57 @@
...
 
 .. code-block:: python
 
    import os
    secret = os.environ["MY_VAR"]
 
-It is possible to define environment variables for projects, pipelines and jobs. The environment
-variables set for these different entities have different priorities: environment variables with the
-same name in jobs take precedence over pipelines, take precedence over projects.
+You can define Environment variables for Projects, Pipelines and Jobs.
```
yannickperrenet, line 15: Why did you choose to change capitalization and make it `Environment` instead of `environment` (same for projects, etc.)? So far we have followed the "rule" of no special capitalization.

yannickperrenet:
Thanks for cleaning it up! Definitely easier to read now :)

docs/source/fundamentals/environment_variables.rst
ncspost: Which in this case might not make sense. Since environment variables are general concepts across multiple tools.

yannickperrenet: > capitalising primary concepts in the UI

I would be fine with that :+1:

PR #1039, opened 2022-06-09
Improv/docs hover boxes

astrojuanlu:
## Description

Description of what your PR solves or adds.

Fixes: #issue

## Checklist

- [ ] The documentation reflects the changes.
- [ ] The PR branch is set up to merge into `dev` instead of `master`.
- [ ] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards compatibility is maintained.
- [ ] In case I changed the dependencies in any `requirements.in` I have run `pip-compile` to update the corresponding `requirements.txt`.
- [ ] In case I changed one of the services' `models.py` I have performed the appropriate database migrations (refer to the [DB migration docs](https://docs.orchest.io/en/stable/development/development_workflow.html#database-schema-migrations)).
- [ ] In case I changed code in the `orchest-sdk` I followed its [release checklist](https://github.com/orchest/orchest/blob/master/orchest-sdk/python/RELEASE-CHECKLIST.md).
- [ ] In case I changed code in the `orchest-cli` I followed its [release checklist](https://github.com/orchest/orchest/blob/master/orchest-cli/RELEASE-CHECKLIST.md).

PR #1041, opened 2022-06-10
Docs: enable `sphinx-hoverxref` on all references

humitos:
## Description

This is a continuation of the work from #1039 since it was reported that there were some tooltips that were showing empty content.

This commit will enable hoverxref on all the references of the documentation so we can debug it deeply and understand why there some cases where the content comes empty.

## Checklist

- [ ] In case I changed the dependencies in any `requirements.in` I have run `pip-compile` to update the corresponding `requirements.txt`.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 sphinx
 sphinx_press_theme
 myst-parser
-docutils==0.16
+docutils
 sphinx-hoverxref==1.1.1
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 sphinx
 sphinx_press_theme
 myst-parser
-docutils
+docutils==0.16
 sphinx-hoverxref==1.1.1
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,3 +2,4 @@ sphinx
 sphinx_press_theme
 myst-parser
 docutils==0.16
+sphinx-hoverxref==1.1.1
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -42,7 +42,10 @@ extensions = [
     "myst_parser",
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
+    "hoverxref.extension",
 ]
+
+hoverxref_auto_ref = True
 
 myst_enable_extensions = [
     "deflist",


humitos:
@astrojuanlu I think I found the reason. It only happens on the `quickstart` link from the home page. If you take a look at the HTML source page of the target, there is a `<span id="quickstart"></span>` which has no content. That's exactly what the API is returning. See https://readthedocs.org/api/v3/embed/?doctool=sphinx&doctoolversion=5.0.1&url=https%3A%2F%2Forchest--1041.org.readthedocs.build%2Fen%2F1041%2Fgetting_started%2Fquickstart.html%23quickstart

However, if the reference would point to `quickstart-tutorial`, the API would return the correct content. See https://readthedocs.org/api/v3/embed/?doctool=sphinx&doctoolversion=5.0.1&url=https%3A%2F%2Forchest--1041.org.readthedocs.build%2Fen%2F1041%2Fgetting_started%2Fquickstart.html%23quickstart-tutorial

It seems this happens because you are using regular references manually defined instead of `sphinx.ext.autosectionlabel`. I'm not sure if there is a way the backend could "auto-detect" this case and return the content of the upper HTML node that always works fine. It seems it could check that:

* it's a `span`
* has no content at all
* the upper HTML node is a `<div class="section">`

That could give the backend some clue about this. Also, it is worth mentioning that changes between Sphinx/docutils versions, and this should be considered in the backend.

humitos:
Another issue that I noticed is that some images from the tooltips are not being rendered for some reason. It seems it's using a relative path `../_images/services.png` when it should be using the absolute path `/_images/services.png`

Screenshot_2022-06-10_12-06-17
 
The reST code for this is located at https://github.com/humitos/orchest/blob/625a0f6929f5b7862a0dca4010d55615b8c7e740/docs/source/fundamentals/services.rst#L24

astrojuanlu:
Hi @humitos , thanks a lot for helping me with the debugging!

> It seems this happens because you are using regular references manually defined instead of `sphinx.ext.autosectionlabel`.

Is it worth tracking this use case in https://github.com/readthedocs/sphinx-hoverxref/issues ? 

> Another issue that I noticed is that some images from the tooltips are not being rendered for some reason. It seems it's using a relative path ../_images/services.png when it should be using the absolute path /_images/services.png

https://github.com/orchest/orchest/blob/5ed6e9951e9266c7a8b90ca41a3db2a7ef1b2a4c/docs/source/fundamentals/services.rst?plain=1#L24

We could work around the paths issue by changing them to absolute then. But similarly, I'm wondering if this is something you want to track upstream?

humitos:
@astrojuanlu yes, this is definitely something we need to track, debug and hopefully make it work in a generic way on sphinx-hoverxref. It's not a problem with your documentation set. If anything, it's a current limitation on the extension itself.

These are two different issues that we will need to open upstream, yeah. I commented on here since I found them here, so it was a way to keep you all updated 👍🏼

ricklamers:
@humitos just wanted to quickly for contributing to better docs! People who care about the docs are the unsung heroes of our time 🙏

humitos:
Thanks you both! I opened https://github.com/readthedocs/sphinx-hoverxref/issues/199 and https://github.com/readthedocs/sphinx-hoverxref/issues/200 copying the comments from this PR. We will be tracking these issues there. Subscribe to them if you are interested in following the news 😄 

I'm closing this PR for now, we can come back to it when the issues are fixed upstream.

humitos:
I did a quick test with a newer version of `docutils`, that's why I re-opened this PR. However, that didn't work. I finally solved the issue about the empty tooltip and merged the PR in the backed. It will be deployed to production next week and I'm happy to receive feedback.

astrojuanlu:
Thanks again @humitos for working on this!

> It will be deployed to production next week and I'm happy to receive feedback.

Excellent, I will re-run after the next deploy (still happening on Tuesdays I assume? 😊)

humitos:
> Excellent, I will re-run after the next deploy (still happening on Tuesdays I assume? blush)

Yes 😄

humitos:
@astrojuanlu @ricklamers I just tested this after releasing a new version of the backend, and it works great! Let me know if there is anything else on this PR that you want me to do.

ricklamers:
@humitos could you sign the CLA? Would be much appreciated. We need it before we can release your contribution 🙏

humitos:
@ricklamers I'm not too happy with that process and with the _required_ data it asked me, but alas. It's done.

Issue #519, opened 2021-10-18
Add a way to see pipeline editor shortcuts


yannickperrenet:
**Is your feature request related to a problem? Please describe.**  
It is currently unclear what shortcuts the pipeline editor supports. For example, many don't even know how to drag the canvas.

The solution could very well be a general shortcuts screen/view for the entire Orchest application.

**Describe the solution you'd like**  
Some way to view the pipeline editor shortcuts.

Would also be great if they can be customized: #429

yannickperrenet:
We need to make sure that the command palette added in #564 is also visible in the shortcut overview, even though (strictly speaking) it is an app wide shortcut and not only a pipeline editor shortcut.

PR #1042, opened 2022-06-10
Add keyboard shortcuts overlay

BrunoOrchest:
## Description

Add a stopgap keyboard overlay allowing the user to see all keyboard shortcuts. Will be sunset with new design.

Fixes: #519 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/hooks/useKeyboardShortcutsOverlay.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/hooks/useKeyboardShortcutsOverlay.tsx
@@ -1,0 +1,144 @@
+import { Box, Dialog, Grid, Stack, styled, Typography } from "@mui/material";
+import Paper from "@mui/material/Paper";
+import React from "react";
+
+type KeyboardShortcut = {
+  keys: string[];
+  description: string;
+};
+
+type KeyboardGroupShortcut = {
+  shortcuts: KeyboardShortcut[];
+  description: string;
+};
+
+export const keyboardShortcuts: KeyboardGroupShortcut[] = [
+  {
+    description: "Pipeline editor",
+    shortcuts: [
+      {
+        keys: ["Space", "_click_", "_drag_"],
+        description: "Pan canvas*",
+      },
+      {
+        keys: ["Ctrl", "_click_"],
+        description: "Select multiple steps",
+      },
+      {
+        keys: ["Ctrl", "A"],
+        description: "Select all steps*",
+      },
+      {
+        keys: ["Ctrl", "Enter"],
+        description: "Run selected steps*",
+      },
+      {
+        keys: ["H"],
+        description: "Center view and reset zoom",
+      },
+      {
+        keys: ["Delete/Backspace"],
+        description: "Delete selected step(s)",
+      },
+      {
+        keys: ["_Double click a step_"],
+        description: "Open file in JupyterLab",
+      },
+    ],
+  },
+  {
+    description: "Command palette",
+    shortcuts: [
+      {
+        keys: ["Control/Command", "K"],
+        description: "Open command palette",
+      },
+      {
+        keys: ["↑/↓"],
+        description: "Navigate command palette commands",
+      },
+      {
+        keys: ["PageUp/PageDown"],
+        description: "Navigate command palette commands",
+      },
+      {
+        keys: ["Escape"],
+        description: "Dismiss command palette",
+      },
+    ],
+  },
+];
+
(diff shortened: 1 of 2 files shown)


yannickperrenet:
Finally some keyboard shortcuts overview! :muscle: 

Some comments that I had:
* Increasing the width of the dialog would be great to make sure everything fits on the screen so that no scrolling/scrollbar is needed
* Can we make the `*` at the end of the explanations red instead? And I think the `* Requires mouse to hover the canvas` should be moved further down so that it is easier to distinguish it from the rest of the text. @ncspost Thoughts?

## Screenshot
image

services/orchest-webserver/client/src/pipeline-view/hooks/useKeyboardShortcutsOverlay.tsx
```diff
@@ -0,0 +1,144 @@
...
+              </Stack>
+            </div>
+          ))}
+        </Stack>
+        <Typography variant="body2">
+          * Requires mouse to hover the canvas
+        </Typography>
+      </Box>
+    </Dialog>
+  );
```
iannbing, line 139: In our codebase, we have a convention on separation of concerns: hooks only contain logic and data; all JSX code should only live in components. This makes it easy to test and troubleshoot. Moreover, declaring components within a component is quite a waste of resource. For example, `Key` could totally be a component itself. Could you clean it up? 😄

iannbing (changes requested):
Apart from the nice UI implementations, it would be great to modify the PR to adhere to our code convention.

services/orchest-webserver/client/src/pipeline-view/hooks/useKeyboardShortcutsOverlay.tsx
yannickperrenet: @iannbing How about adding this code convention to https://docs.orchest.io/en/stable/development/implementation_best_practices.html :)

ncspost:
Great work @BrunoOrchest - just an FYI, it's useful to ping me on all UI-related PRs from the start. My notes...

### Canvas

- Keyboard button hidden behind delete button when multiple steps selected or single step select and properties panel is dismissed
- Use keyboard icon from outlined icon library rather than filled icon library
- Move button into bottom left canvas control button group (last item in list)
- Keyboarding to this button should happen after "organise canvas" button

### Dialog

- Increase space between list items from 4px → 8px
- I can't seem to keyboard into the dialog
- Change font size to "body 2"
- Change Keyboard box bg colour to "grey 200"
- Change keyboard box text colour to "text primary"
- Change text (e.g. + click) next to keyboard boxes to "body 2" and "text secondary"
- Include a close icon in top right corner for user to click out of
- I updated how some of the content is presented in [Figma](https://www.figma.com/file/wpI1sYmTo37qyWg3fEGSJm/Help-centre?node-id=653%3A18409) - please refer there and update content (e.g. + click + drag is redundant)
- There seems to be inconsistent padding between different keyboard boxes (e.g. space and escape)
- Alignment between text and pluses are off... the should be aligned by the middle rather than the bottom.
- Reduce space between plusses, keyboard boxes and text from 8px → 4px.

### Questions

- Does this take OS into account and dynamically change keyboard symbols?

> Increasing the width of the dialog would be great to make sure everything fits on the screen so that no scrolling/scrollbar is needed

@yannickperrenet - I don't think we need to increase dialog width if we reduce text size and neaten up some of the content

> Can we make the * at the end of the explanations red instead? And I think the * Requires mouse to hover the canvas should be moved further down so that it is easier to distinguish it from the rest of the text. @ncspost Thoughts?

@yannickperrenet @BrunoOrchest - I think we should move asterisk directly underneath pipeline section and increase spacing between this and other list items to 16px. I think we should differentiate the colour as grey (text secondary) rather than red. Copy should read "While mouse hovers canvas".

services/orchest-webserver/client/src/pipeline-view/hooks/useKeyboardShortcutsOverlay.tsx
iannbing: Some [Stakeoverflow answers](https://stackoverflow.com/questions/62186513/confusion-about-this-react-custom-hook-usage) resonate with my concerns. If one really wants to bundle the component inside of a hook, using Context API to pay the performance penalty could be an option, see [this article](https://blog.bitsrc.io/new-react-design-pattern-return-component-from-hooks-79215c3eac00). All in all, this usage of hooks is quite uncommon. I would suggest sticking with the proven path.

PR #1046, opened 2022-06-17
Retry on controller installation if cluster is not ready yet

yannickperrenet:
## Description

Instead of failing if the cluster is not ready, we now retry `4` times (which leads to a maximum waiting time of `75s`) to install the `orchest-controller`. If the cluster is just created (e.g. through our `scripts/convenience_install.sh`) then the ingress might not yet be ready, thus leading to the `orchest-controller` not installing.

Moreover, I chose to go with a retry mechanism given that the resources in the manifest (the `orchest-controller.yaml` that is added to the release assets) can change in future versions. Thus we need to dynamically decide whether the cluster is "ready" (which would lead to a race-condition anyways given that the cluster could change state after the check). 

## Checklist


ricklamers:
Looks good 👍🏻 Why `create_from_dict` the first time and `update_from_dict` for the retry loop?

ricklamers:
A minor nitpick is that the method could be a bit more DRY, I feel like the pre-retry logic is quite similar to what happens in the retry bit. But I’d be OK with leaving it as is.

yannickperrenet:
> Why create_from_dict the first time and replace_from_dict for the retry loop?

As per the comment, `replace_from_dict` is used the second time to account for `"List"` kinds. And using `create_from_dict` in the initial try allows us to distinguish a special case (e.g. someone runs `orchest install` a second time or already has an `OrchestCluster` CR Object defined). 

> the install function could be a bit more DRY

Yeah I had the same feeling, but chose to leave it as is given that the difference of using `create_...` vs `replace_...` would introduce an abstraction that could be harder to understand than just writing it out (in addition, the exception handling is slightly different).

ricklamers:
> would introduce an abstraction that could be harder to understand than just writing it out

Good reason! This can be merged then 👍

ricklamers (approved):

PR #1048, opened 2022-06-20
Support `containerd` as the container runtime

nhaghighat:
## Description

This PR enables working with containerd runtime by introducing an init container to pull images, the controller detects the runtime and configures orchest accordingly”

In order to test this PR you need to have a cluster with contained runtime, microk8s is suggested, after microk8s is installed ([here](https://microk8s.io/)), the following addons need to be enabled.
```sh
microk8s enable hostpath-storage \
  && microk8s enable dns \
  && microk8s enable ingress
```

In order to be able to push images to microk8s node, after rebuilding all the images with some valid tag (for example `v2022.06.4`), you can save them to a tar file via following command:
```sh
docker save $(docker images | awk '{if ($1 ~ /^orchest\//) new_var=sprintf("%s:%s", $1, $2); print new_var}' | grep v2022.06.4 | sort | uniq) -o orchest-images.tar
```

Then this tar file can be shipped to microk8s node via scp.
`scp ./orchest-images.tar {your_user}@${microk8s node ip}:~/`

then inside the microk8s node, you can import the images via following command (note `ctr` has to be installed, binaries can be found [here](https://github.com/containerd/containerd/releases))
```sh
sudo ctr -n k8s.io -a /var/snap/microk8s/common/run/containerd.sock i import orchest-images.tar

# Or use microk8s ctr
microk8s ctr --namespace k8s.io --address /var/snap/microk8s/common/run/containerd.sock image import orchest-images.tar
```

then orchest can be installed with orchest-cli, with following command
```sh
orchest install --socket-path=/var/snap/microk8s/common/run/containerd.sock --dev
```

# Note:
 the manifests must be generated via `make manifestgen` in the orchest-controller directory.

```sh
TAGNAME=v2022.06.4 make -C ./services/orchest-controller manifestgen
```

## Checklist

- [x] The newly added image-puller has to be pushed to DockerHub on release. So they need to be added to the correct .github/workflow/... file.
- [x] add document about installing in microk8s
- [x] merge both cli init containers into one
- [x] ~~update thirdparties on update~~


nhaghighat:
It is tested with microk8s and minikube with containerd. k3s will be tested

orchest-cli/orchestcli/cmds.py
```diff
@@ -240,6 +244,26 @@ def install(cloud: bool, dev_mode: bool, fqdn: t.Optional[str], **kwargs) -> Non
...
         sys.exit(1)
 
+    applications = [{"name": "docker-registry", "config": {}}]
+
+    if no_argo:
+        echo(
+            "ArgoWorkflow installation is disabled, orchest will be "
+            "deployed without it, make sure ArgoWorkflow is already installed "
+            "on your cluster otherwise orchest will not act properly."
+        )
```
fruttasecca, line 254: Bit of a random question that came up thinking about the fact that when we create a custom object representing a workflow we specify the namespace, example:
```python
  k8s_custom_obj_api.create_namespaced_custom_object(
            "argoproj.io", "v1alpha1", ns, "workflows", body=manifest
        )
```
Will an argo deployment simply pick this up regardless of the namespace?

orchest-cli/orchestcli/cmds.py
```diff
@@ -1341,10 +1368,12 @@ def _fetch_latest_available_version(
         return None
 
 
-def _fetch_orchest_controller_manifests(version: t.Optional[str]) -> str:
+def _fetch_orchest_controller_manifests(
+    version: t.Optional[str], manifest_file_name: str
+) -> str:
```
fruttasecca, line 1373: Guess we should provide a default value for `version` since it's optional

fruttasecca:
Not sure I agree with the deletion of the node agent:
- Orchest loses the ability to delete images from the node
- pre-pulling is still an important part of the UX imo, not only we pull env images but a bunch of other things

I guess part of the reason for this is the same reason behind the fact that we are using different images to interface with different CRIs? https://github.com/orchest/orchest/pull/1048#discussion_r901773195

Is the requirement for the cluster nodes to have a homogeneous runtime dictated by a lack of time (i.e. to keep scope down?) or a technical requirement?

orchest-cli/orchestcli/cmds.py
nhaghighat: Argo usually deployed to pick up any workflow object regardless of the namespace, but in #866 we restricted the default deployment, so it only picks up the workflow object in it's own namepspace

scripts/build_container.sh
```diff
@@ -81,6 +81,8 @@ if [ ${#IMGS[@]} -eq 0 ]; then
         "auth-server"
         "node-agent"
         "orchest-controller"
+        "docker-cli"
+        "containerd-cli"
```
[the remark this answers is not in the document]
nhaghighat: To have a smaller images and minimize startup time.

services/orchest-controller/deploy/helm/values.yaml
```diff
@@ -2,8 +2,12 @@
 service:
   port: 80
 
-images:
+image:
   pullPolicy: IfNotPresent
   registry: docker.io
   name: orchest/orchest-controller
   tag: latest
```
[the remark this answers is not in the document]
nhaghighat: No, as update don't use this, update uses the generated files of the helm charts which is not changed

nhaghighat:
> Not sure I agree with the deletion of the node agent.

I agree, this has this downside, but if we want to use node-agent we need to have some mechanism to delay pod deployment, until node-agent informed us about pulling.

> Is the requirement for the cluster nodes to have a homogeneous runtime dictated by a lack of time (i.e. to keep scope down?) or a technical requirement?

It is a technical requirement, we don't know on which node the pod will end up, so in not-homogeneous cluster, orchest does not know which image to use.

scripts/build_container.sh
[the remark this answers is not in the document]
fruttasecca: I see, alright

fruttasecca:
>I agree, this has this downside, but if we want to use node-agent we need to have some mechanism to delay pod deployment, until node-agent informed us about pulling.

This would only affect the "puller" part of the node-agent, so I think we should at least retain the "deleter" part of it and aim to keep the puller part of it if possible. To be more precise, it only affects environment images and custom jupyter images (essentially all images that are pushed to the internal registry),  so other images that we are pre-pulling are not part of the issue and would still benefit from being pre-pulled.

I think we have two architectural issues/questions here to solve:
- how do different container runtimes handle concurrent pull calls? AFAIK that's fine but I haven't been able to fine explicit docs about it for docker and/or containerd
- now that we support `contaienrd` the node agent should cover that as well
- this PR provides multiple images to interface with different container runtimes, what do we do with the node-agent? One image of the node agent for every container runtime?


Note that, assuming we keep the node-agent around, the new images introduced by this PR should be part of the set of images that the node puller and/or deleter considers.

fruttasecca:
Note that since users can create custom jupyter images that can be customized and that are pushed to the internal registry, so the same kind of pre-pull behaviour should take place (you should look into the sessions `_core.py` module AFAIR). Also the `/orchest-images-to-pre-pull` endpoint in `namespace_ctl` might need to be looked into (not sure)

fruttasecca:
Would jupyter kernels actually work? I think `_get_kernel_pod_manifest` in `services/orchest-webserver/app/app/res/kernels/launch_kubernetes.py` might need to be changed to include the init containers?

fruttasecca:
Another thing that might need looking into, user services allow you to use an environment image built in Orchest as a service, the logic happens at `_get_user_service_deployment_service_manifest` in `services/orchest-api/app/app/core/sessions/_manifests.py`. I guess we need to have init containers there are as well when the service image is an Orchest environment image.

fruttasecca:
Another thing I would like to mention is the path of the init container directories, e.g. `services/tools/docker/`. Can we find a name that fits what the tool is doing and then improve the directory structure? I think this might have benefits in the long run, e.g. `services/tools/init-container-puller/docker/`, etc., discussing it in this PR feels a bit off already since I keep coming up with (probably wrong) names for it. This would also affect how to refer to this particular tool in the codebase, i.e. `RUNTIME_IMAGE`

lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -39,6 +39,11 @@
 REGISTRY = "docker-registry"
 REGISTRY_FQDN = f"docker-registry.{ORCHEST_NAMESPACE}.svc.cluster.local"
 
+# Runtime configs.
+RUNTIME = os.environ.get("RUNTIME")
+RUNTIME_IMAGE = os.environ.get("RUNTIME_IMAGE")
+RUNTIME_SOCKET = os.environ.get("RUNTIME_SOCKET")
```
yannickperrenet, line 45: How about prefacing these variables with `CONTAINER_`? Otherwise it might be/become unclear what "runtime" we are talking about here.

orchest-cli/orchestcli/cli.py
```diff
@@ -178,16 +178,24 @@ def cli():
     show_default=True,
     help="Fully Qualified Domain Name that Orchest listens on.",
 )
+@click.option(
+    "--socket-path",
+    "socket_path",  # name for arg
+    default=None,
+    show_default=True,
+    help="The path of the container runtime socket location.",
```
yannickperrenet, line 186: I would change this help text to: `The absolute path of the container runtime socket.`

orchest-cli/orchestcli/cmds.py
```diff
@@ -795,6 +809,13 @@ def disable_telemetry() -> None:
...
+    if socket_path is not None:
+        annotations = {"orchest.io/container-runtime-socket": socket_path}
+        convert_to_strategic_merge_patch(
+            annotations,
+            custom_object["metadata"]["annotations"],
+        )
+
     try:
         patch_namespaced_custom_object(
             name=cluster_name,
```
yannickperrenet, line 821: Doesn't seem like these annotations would be picked up. They need to be passed as part of the `body` in `patch_namespaced_custom_object`.

lib/python/orchest-internals/_orchest/internals/config.py
nhaghighat: Sounds good, will do

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -229,3 +231,62 @@ func getOrchestComponent(name, hash string,
 	}
 
 }
+
+// detectContainerRuntime detects the container runtime of the cluster and the socket path
+// returns error if the container runtime is not supported or the cluster is not homogeneous
+// If the socket path is not set or not deteted, it will use the default socket path based on
```
yannickperrenet, line 237: `deteted` -> `detected`

yannickperrenet:
Some questions I have (in no particular order, but I figured the numbers could help to reference resp. question):

(1) I am not so sure about the deletion of the `node-agent` (pretty much what @fruttasecca mentioned in https://github.com/orchest/orchest/pull/1048#issuecomment-1160599091):

* We need to keep deletion as otherwise all nodes in the cluster would slowly start to get a full disk due to the images.
* Pre-pulling seems like a good idea still (to me) given that otherwise the init container would have to do the pulling, meaning that (from a UX standpoint) the user has to wait longer for e.g. a session to start.
* "One image of the node agent for every container runtime?" (from https://github.com/orchest/orchest/pull/1048#issuecomment-1161386423 by @fruttasecca) -> I think we could use environment variables as well here (just like it is done for the `orchest-api`) that are added by the `orchest-controller`. I would rather have one bigger `node-agent` image that works for both container runtimes instead of having one for each. Given that it has a Python base, I don't think there is much gain anyways (in a relative sense).

(2) @nhaghighat You said in https://github.com/orchest/orchest/pull/1048#issuecomment-1160103387 that you tested `microk8s`. Did it work out of the box or did you have to use the newly added `--socket-path` CLI option? In case it did not work by default then I think we should add a section to the docs that explains how to run Orchest on `microk8s`. 

(3) The newly added `docker-cli` and `containerd-cli` need to be pushed to DockerHub on release. So they need to be added to the correct `.github/workflow/...` file.

* In https://github.com/orchest/orchest/pull/1048#discussion_r901955851 it was discussed that having one image for each would save their size. What size benefits are we talking about here? In case it is super minimal, then we might as well ship one slightly larger image as to not have to distinguish between container runtimes. We can just first try to pull using Docker, and if that does not exist, then we can try to pull using containerd. 

(4) As mentioned in https://github.com/orchest/orchest/pull/1048#issuecomment-1161410010, we might want to change the path from `services/tools/...` to `tools/...`. I don't really see the `docker-cli`/`containerd-cli` containers as services (at least not in the same way as the `orchest-api` etc.)

(5) Indeed the init container needs to be added to more places:

* jupyter-server (https://github.com/orchest/orchest/pull/1048#issuecomment-1161389561)
* kernels (https://github.com/orchest/orchest/pull/1048#issuecomment-1161393286)
* user services (https://github.com/orchest/orchest/pull/1048#issuecomment-1161396587)

nhaghighat:
> how do different container runtimes handle concurrent pull calls? AFAIK that's fine but I haven't been able to find explicit docs about it for docker and/or containerd

@fruttasecca : You mean concurrent pull of different images or same image?

> now that we support contaienrd the node agent should cover that as well

> Pre-pulling seems like a good idea still (to me) given that otherwise the init container would have to do the pulling, meaning that (from a UX standpoint) the user has to wait longer for e.g. a session to start.

I also like the idea of pre-pulling, but I think at the moment it does not bring us much, most of our users have limited amount of nodes and the images will be pulled by the first step container, so the for the rest of the steps the image is probably already present on the node.
Also it fills all the nodes with the images that might not be needed at all.
For example in a 3 node cluster it pulls jupyter-eg and jupyter-server on all of them, but the pod is spawned in only one the nodes.

> this PR provides multiple images to interface with different container runtimes, what do we do with the node-agent? One image of the node agent for every container runtime?

> I would rather have one bigger node-agent image that works for both container runtimes instead of having one for each. Given that it has a Python base, I don't think there is much gain anyways (in a relative sense).

If we decide to keep node-agent that would also be my preference.

> We need to keep deletion as otherwise all nodes in the cluster would slowly start to get a full disk due to the images.

> I think we should at least retain the "deleter" part of it and aim to keep the puller part of it if possible

Kubelet has a functionality to remove unused, un-tagged images from the nodes, regularly.

> In case it did not work by default then I think we should add a section to the docs that explains how to run Orchest on microk8s.

No, I have changed the OrchestCluster CRD, will add an example and a section to the doc about it.

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
```diff
@@ -164,13 +164,13 @@ controller:
 
   # -- Specifies the container runtime interface to use (one of: `docker`, `kubelet`, `k8sapi`, `pns`, `emissary`)
   ## Ref: https://argoproj.github.io/argo-workflows/workflow-executors/
-  containerRuntimeExecutor: docker
+  containerRuntimeExecutor: emissary
```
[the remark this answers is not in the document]
nhaghighat: I think not, the third parties only checked on installation time

nhaghighat:
> We can just first try to pull using Docker, and if that does not exist, then we can try to pull using containerd. This would also solve the need to check for a homogeneous cluster 

The image size will not grow that much if we have both in the same image.now it is like 50mb each, then it would be 100mb.
But it still does not solve deployment in not homogeneous cluster, we still need to detect the socket location beforehand and it is different for different container run-times.

fruttasecca:
>@fruttasecca : You mean concurrent pull of different images or same image?

same, was thinking about race conditions etc.

>I also like the idea of pre-pulling, but I think at the moment it does not bring us much, most of our users have limited amount of nodes and the images will be pulled by the first step container, so for the rest of the steps the image is probably already present on the node.

I don't feel very strongly about the need for a pre-puller like the current state of the node-agent, although I think it still brings a significant UX improvement on first use(s) and first impressions matter, although I don't have anything quantifiable. About pulling images where they are not needed, I get the sentiment but any multi node (serious) use of Orchest will eventually lead to all "types" of pods being scheduled on different nodes, so I am not sure we should consider this point very much.
Imo if we keep the node-agent, let's say, for deletion, (and thus implement/make it work with different container runtimes) the adjustments needed to bring the puller logic up to speed should be so minimal that it wouldn't make sense to renounce the UX improvement

>I would rather have one bigger node-agent image that works for both container runtimes...
>If we decide to keep node-agent that would also be my preference.

same :)

>Kubelet has a functionality to remove unused, un-tagged images from the nodes, regularly.

I think it might help us in reducing complexity by offloading more logic to k8s, I've skimmed through the docs very quickly and I think it has some downsides:
- it requires setting and/or changing the kubelet configuration, I am not sure we can do that on third-party clusters or what's the overhead to implement this on different flavours of k8s, and asking users to do that would get in the way of a "it just works" experience imo
- as far as I could see it's controlled by a periodic check configured through `--minimum-image-ttl-duration duration`, which means that freeing space is not immediate and you end up having delays at random times due to the image being re-pulled, the latter is a minor point imo, especially if the logic that considers an image "unused" is smart (which I don't know tho)
- thinking about the cloud, if we require users to have at least X free gigs before performing an update we know that the outdated images will be swiftly cleaned up after said update, a periodic check with a long period wouldn't give us that, and we would have to complicate the logic further to clean up unused images
- generally speaking, and especially thinking about the cloud, freeing up space quickly might make a great difference when it comes to UX

So imo even if this kubelet functionality would have came up during the implementation/review of the node-deleter I am not sure we would have gone with that

>Solving this problem now seems like a pre-mature optimization, most f the clusters are homogeneous.

Agree

fruttasecca:
>I think the ultimate solution to solve all the mentioned issues is to eventually have our own scheduler.

I agree, or at least I agree on the fact that the current solution is not the final one, but when it comes to managing scope/roadmap/timelines I think this init container "hack" might be the right call in order to save time

nhaghighat:
> I agree, or at least I agree on the fact that the current solution is not the final one, but when it comes to managing scope/roadmap/timelines I think this init container "hack" might be the right call in order to save time

agree, it was the right call.

nhaghighat:
> it requires setting and/or changing the kubelet configuration

garbage collecting of images and containers is the default behavior of kubelet, so it is safe to assume users have not disabled it.

> as far as I could see it's controlled by a periodic check configured through...

There are different ttl and threshold configurations which I think the default one is fine for us.

> end up having delays at random times due to the image being re-pulled

This is true, but it can still happen with `node-agent`. imagine this scenario: kubelet deletes an image and then before node-agent pulls the image a pod is configured to run on that node.

For example there is `image-gc-high-threshold` which is triggered if the disk usage reaches the threshold (ofcourse it checks the disk usage periodically)

more information about it can be found here:
https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/cluster-administration/kubelet-garbage-collection/

fruttasecca:
>garbage collecting of images and containers is the default behavior of kubelet, so it is safe to assume users have not disabled it.

that's true, what I meant to say is that we are not in control of parameters so we risk being at the mercy of coincidence

>There are different ttl and threshold configurations which I think the default one is fine for us.
>This is true, but it can still happen with node-agent. imagine this scenario: 

noted

nhaghighat:
> that's true, what I meant to say is that we are not in control of parameters so we risk being at the mercy of coincidence

kubelet garbage collection is always there and it is working even when node-agent is deleting unused images. so we are always in an environment with un-controlled parameters, whether node-agent is there or not. by having node-agent we might decrease the chance of disk pressure in a very rare scenario. (and we might even increase the chance, because node-agent fills nodes with lot's of unused images)

I still don't get the benefit of having node-agent any more, but as @yannickperrenet mentioned I'll add the functionality to node-agent

nhaghighat:
> But what about the use case of making sure space is cleaned up quickly or improved UX of pre-pulling? Given that the node is both a puller and a deleter, do you think both of those functionalities are now redundant?

Puller is indeed helps improving UX, but at the same time it fills up the nodes, and there are scnarios where kubelet deletes the image and then a pod get scheduled on this node.

> That's true, but if we need to get those images deleted earlier we can do that with an agent we control

That is true, not sure if it will become a bottleneck or not, as in multi node cluster, k8s scheduler considers the disk space left on the nodes before scheduling the pods, so it chooses nodes with available disks.

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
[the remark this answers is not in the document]
fruttasecca: I feel the update should indeed take care of updating third parties as much as possible

orchest-cli/orchestcli/cli.py
```diff
@@ -178,16 +178,24 @@ def cli():
...
+)
```
fruttasecca, line 187: Some questions:
- should we also allow to specify the container runtime directly? i.e. why allow specifying the socket and not the runtime? Can k8s "fail" to tell us one but not the other?
- this particular flag is, imo, at a way lower abstraction level that anything else part of the command. I think the command description needs to be fleshed out some more and should probably state that as a "normal" user you are very likely to not need this

orchest-cli/orchestcli/cli.py
```diff
@@ -289,7 +289,7 @@ def update(
     "socket_path",  # name for arg
     default=None,
     show_default=True,
-    help="The path of the container runtime socket location.",
+    help="The absolute path of the container runtime socket on the host.",
```
fruttasecca, line 292: Might make sense to hint at the fact that we are talking about the node host, since our users (currently) deploy Orchest through a k8s flavour like minikube on their own host

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -421,6 +433,23 @@ def _get_jupyter_server_deployment_service_manifest(
...
+                                {
+                                    "name": "IMAGE_TO_PULL",
+                                    "value": utils.get_jupyter_server_image_to_use(),
+                                }
+                            ],
+                            "command": ["/pull_image.sh"],
+                            "volumeMounts": [
+                                volume_mounts_dict["container-runtime-socket"],
+                            ],
+                        },
```
fruttasecca, line 452: `utils.get_jupyter_server_image_to_use(` might return `orchest/jupyter-server...` or the custom jupyter server image that the user has built through Orchest, the former lives on dockerhub, while the latter is pushed to the internal registry. So the init container is only needed for the custom image, and since we are pointing to the socket in `pull_image.sh` the "orchest" jupyter server image pull will fail

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -715,6 +744,25 @@ def _get_jupyter_enterprise_gateway_deployment_service_manifest(
...
+                                        + CONFIG_CLASS.ORCHEST_VERSION
+                                    ),
+                                }
+                            ],
+                            "command": ["/pull_image.sh"],
+                            "volumeMounts": [
+                                volume_mounts_dict["container-runtime-socket"],
+                            ],
+                        },
+                    ],
```
fruttasecca, line 765: No need to pull with the init container here, since this image never lives in the internal registry

services/orchest-api/app/app/core/sessions/_manifests.py
nhaghighat: > since we are pointing to the socket in pull_image.sh the "orchest" jupyter server image pull will fail

could you elaborate?

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -46,15 +46,24 @@ def get_step_and_kernel_volumes_and_volume_mounts(
...
+            The pipeline file.
         container_project_dir:
+            The container project directory.
         container_pipeline_file:
+            The container pipeline file.
+        container_runtime_socket:
+            The socket to use for the container runtime.
+
 
     Returns:
```
fruttasecca, line 68: The modified implementation does not respect what the function doc state in the `Returns:` section. If it did it would be a security hazard since the risk of mounting the socket into a container running user code would be very real, I actually thought that was happening until I read the implementation.

Given that all services that need create a manifest have access to the internal lib it would be nice to delegate the construction of the socket volume and volume mount to a single function (extremely similar to `get_step_and_kernel_volumes_and_volume_mounts`) and always use those return values instead of writing mounts inline. The intent would be to have a single place to lookup/grep to know when and where such socket is being mounted. Same for `_get_common_volumes_and_volume_mounts`, I really feel like we need to treat this particular volume and mount as a special case given the security implications

fruttasecca: What I meant to comment:
```python
Returns:
        A pair of lists, the first element is a list of volumes, the
        second a list of volume_mounts, valid in k8s pod manifest. The
        two lists are coupled, each volume mount is related to a volume.
```

orchest-cli/orchestcli/cli.py
fruttasecca: I'd prefer `node` but I am not feeling strongly about it really

lib/python/orchest-internals/_orchest/internals/utils.py
nhaghighat: I agree with you, it would be nice to have this volume generated in one place, but to do that we also need to change all volumes and volume-mounts, in the code. they are created multiple places in the code and sometimes in-lined, I avoided to do that to lower the scope.

services/orchest-api/app/app/core/sessions/_manifests.py
nhaghighat: I know, but I think in order to have manifests in a same convention, it makes sense to also have it enterprise-gateway.
besides it does not increase the time for spawning sessions that much.

lib/python/orchest-internals/_orchest/internals/utils.py
nhaghighat: I'm not sure if the doc were correct before.
it implies there is a 1 to 1 mapping between volumes and volume_mount which is not always true.

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca: Mmh I think what I wrote was not entirely correct, sorry, I thought this was a bug but it's a rather a minor thing, I'll clarify:

- `get_jupyter_server_image_to_use` might return an image to be pulled from the registry or from dockerhub 
- we always use the init container to pull the image regardless of the case
- given an image pointing to dockerhub, the init container `pull_image.sh` will try to pull it with flags that may not be compatible with the fact that we are pulling from a public registry, I don't think this is the case

Probably not an issue, although I'd rather avoid the init container when it's not necessary

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: >I'm not sure if the doc were correct before.
>it implies there is a 1 to 1 mapping between volumes and volume_mount which is not always true.

I see what you mean, the `"each volume mount is related to a volume"` is a bit confusing :p

>I agree with you, it would be nice to have this volume generated in one place, but to do that we also need to change all volumes and volume-mounts, in the code. they are created multiple places in the code and sometimes in-lined, I avoided to do that to lower the scope.

Not sure I exactly follow here, but I would be fine with just making the socket volume and mount creation the responsibility of a single function, and leave the rest as is.

fruttasecca:
Do you think it would make sense to encapsulate the creation/injection of the `initContainer` definition into a function to be DRY? Seems like the only thing that's really changing is the image

services/orchest-api/app/app/core/sessions/_manifests.py
nhaghighat: > Have we documented the reason for this init container approach, ssl skip etc. somewhere in the codebase? No idea where we would put it but it would make sense to have it written somewhere (although some of our internal docs/discussion may persist it already

I'll make sure the design doc is updated

lib/python/orchest-internals/_orchest/internals/utils.py
nhaghighat: what I meant was that if we want to encapsulate this volume into a single location, we should first encapsulate all the volumes into a single place, then this goes with them.

nhaghighat:
> Coming back to the the new images, CONTAINER_RUNTIME_IMAGE is still confusing imo, the image actually contains a thin interaction layer with the runtime, so it should reflect that better 

A better name could be `CONTAINER_RUNTIME_CLI_IMAGE`

orchest-cli/orchestcli/cli.py
yannickperrenet: The arguments/options here would just be used to construct the Custom Resource Object that describe the OrchestCluster. I would be fine leaving it as is, but can definitely be improved once we have more "unlikely to be used" arguments.

yannickperrenet:
> A better name could be CONTAINER_RUNTIME_CLI_IMAGE

I thought the idea was to ship just one image? So merging the `docker-cli` and `containerd-cli` into one, alleviating the problem.

fruttasecca:
>@yannickperrenet : You proposed it, but I guess we did not reach the consensus or did we?

Imo this could be a good idea, the controller wouldn't have to pass the image as a plus

nhaghighat:
> Imo this could be a good idea, the controller wouldn't have to pass the image as a plus

Controller needs to detect the socket location any way and it determines the socket location based on the type, so it has to pass the sokcet location to puller, besides detecting container runtime is easy and robust.
on the other hand if we rely on first trying docker and if it failed, then try containerd, we increase the session start up couple of seconds, because the first try might waits for couple of second then times out and moves to the next container runtime.

lib/python/orchest-internals/_orchest/internals/utils.py
fruttasecca: >then this goes with them

My idea was to avoid the socket volume and volume mount be nearby any logic related to "normal" volumes.  I.e. not have `get_step_and_kernel_volumes_and_volume_mounts` (et. simila) deal with those volumes or the inline definition of a socket mount like it happens in `services/orchest-api/app/app/core/pipelines.py` at around line 604. Imo it makes it harder to introduce said volume or mount in the wrong place by mistake, and, for example, makes the expectantions around functions like `get_step_and_kernel_volumes_and_volume_mounts` more consistent.

For example, the volumes returned by `get_step_and_kernel_volumes_and_volume_mounts` include the socket volume, while the mounts returned by the same functions do not include the socket volume mount, which is then inlined in the dictionary. Imo at this point we would be better off not considering the socket volume & mount belonging to "step and kernel volumes and volume mounts" or to any set of volume/mounts that we would consider part of the "normal" flow.

As far as I understand the init container is the only one needing said mount, so generating the init container "manifest" (i.e. the dictionary) through a function would allow us to kill two birds with one stone already since it would allow us to be DRY and to know that that socket is only mounted there

fruttasecca: >what I meant was that if we want to encapsulate this volume into a single location, we should first encapsulate all the volumes into a single place, then this goes with them.

I think the current encapsulation of existing volumes is sort of okayish by the fact that they are generated through functions like `get_step_and_kernel_volumes_and_volume_mounts`, but I might have misunderstood what you meant, so feel free to clarify

yannickperrenet:
> on the other hand if we rely on first trying docker and if it failed, then try containerd, we increase the session start up couple of seconds

Given that there will only be one container runtime running at any giving time we might as well make both the pull commands in parallel. We know one will always fail but that is fine. Just an idea.

I just like simplicity of having one container (that we use as an init container) instead of having to maintain multiple (and having to decide which one to use). I don't have a strong preference, but think the simplicity of going with one container will pay of in the long run. 

> I didn't mean to get rid of any information passing from the controller to the orchest-api in favour of an approach that tries to interact with all runtimes.

Same here. I also like passing the information given that we already have it.

nhaghighat:
> Same here. I also like passing the information given that we already have it.

I think the information should be passed because of things mentioned in my previous comment. since I don't see this as a long term solution I don't mind having two container. but will change it 1 container.

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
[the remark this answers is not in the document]
nhaghighat: I spent quite some time on it, it is quite complex to solve, as we use OrchestCluster hash to detect if the cluster needs updating or not, and this part is not part of OrchestCluster. so the way we have to detect upgrading needs to be changed, I implemented that part but I'm afraid it might take more time.

as it does not affect our current users, I suggest creating a ticket for it to properly fix it later.

orchest-cli/orchestcli/cmds.py
yannickperrenet: Resolved in https://github.com/orchest/orchest/pull/1048/commits/b570aa85ffa071172c2e8ea02ec024192972c859

yannickperrenet:
So regarding https://github.com/orchest/orchest/pull/1048#discussion_r909459902 on `orchest update` we are currently not updating any changes we make to third-parties? E.g. for new version X we make a change to the manifest of Argo (this has happened in the past where we added an environment variable, if I am not mistaken @fruttasecca), then this change won't actually be reflected when an existing user updates/upgrades to version X? That is not acceptable imo, but since it is not introduced in this PR I am happy to have it fixed in another PR.

Thus, for the time being, I guess we should not make changes to third-party helm charts as existing users won't get the update. Luckily adding this functionality to the `orchest-controller` should not be a breaking change.

services/orchest-controller/deploy/examples/example-orchestcluster-micrik8s.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: orchest.io/v1alpha1
```
yannickperrenet, line 1: The name of this file contains a typo: `micrik8s` -> `microk8s`

nhaghighat:
> then this change won't actually be reflected when an existing user updates/upgrades to version X

That is true, this change is not reflected on existing users

> That is not acceptable imo

I agree, but that requires detecting changes of the thrird-parties, which is not that easy, as we use in container values file for that

PR #1049, opened 2022-06-21
remove helm releated labels

nhaghighat:
## Description

Description of what your PR solves or adds.

Fixes: #issue

## Checklist


yannickperrenet:
Just to be safe, how about also changing the `.github/workflows/do-release.yaml` so that `https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3` is actually using a specific version instead of `main`? This way we know for sure that the same helm version is used to generate the yaml file for every release.

yannickperrenet (approved):
Looks good to me!

Issue #178, opened 2021-03-15
Improve saving indicator such that pipeline autosaving is clear


ricklamers:
The current indicator only shows when loading takes longer than 100ms. But it isn't clear enough the pipeline definition is being saved as changes happen.

E.g. moving the pipeline step or editing pipeline properties.

Issue #1009, opened 2022-05-31
Dragging a file when there are no pipelines gives no feedback


astrojuanlu:
**Describe the bug**
I was confused trying to drag a file to the pipeline editor and nothing was happening, to realize that I needed a pipeline first.

**Expected behavior**
The UI gives an error message or tells me what to do.

**To Reproduce**

1. Create a new project
2. Go to the pipeline editor
3. Create a new file
4. Try to drag it

**Screenshots**

https://user-images.githubusercontent.com/316517/171226937-e8bafd22-362a-40ea-bb5f-9d3887416554.mp4

**Environment**

- OS: Linux Mint
- Browser: Firefox 100
- Orchest's version (in the settings page): v2022.05.4

astrojuanlu:
I think this has been discussed before somewhere else, but it would be cool to hand hold the user a bit more with their first pipeline. Some proposed flows:

- When the user creates a project, add a pipeline there. A bit intrusive.
- When the user opens an empty project, a dialog asks "Create a pipeline? Yes/No" With a checkbox "Don't show this again".
- A big call-to-action in the middle of the canvas saying "create a pipeline to start working" or similar

yannickperrenet:
Since we now automatically create a Pipeline when a Project is created (#1261), it is now much less likely that someone will hit this case. In addition, as it stands currently, users have a hard time discovering the ability to drag-and-drop steps. Thus I would say that this issue has low priority.

astrojuanlu:
>  In addition, as it stands currently, users have a hard time discovering the ability to drag-and-drop steps.

I agree with this, are we tracking that somewhere? Would be cool to have a more visible onboarding in that aspect

yannickperrenet:
> are we tracking that somewhere?

We are. It is part of some UX studies that @ncspost is running :raised_hands:

Issue #959, opened 2022-05-17
Switch between pipelines using sessions list


fruttasecca:
**Describe the problem this improvement solves**
Currently, both the toggle button (`1` in the picture) and the pipeline name (`2`) toggle the pipeline
session when clicked.

**Describe the solution you'd like**
 My suggestion would be to have `2` switch between pipelines, and leaving 1 unchanged.

**Screenshots**
![image](https://user-images.githubusercontent.com/19429509/168811292-ed260289-9bc6-4933-86ac-f21f9c590ca6.png)

yannickperrenet:
@ncspost @ricklamers What do you think?

Personally, I feel that the proposed behavior is somewhat confusing given that it is part of the `Sessions` "block". Generally not only the toggle but also the name tend to toggle and I would rather stick with such conventions (if it really is one). 

If we want to make switching between pipelines easier, instead of having to find them within all the files in your project then I think that is a valid point but I don't think this is the way to do it.

ricklamers:
We'll name block "Pipelines" and add a "Session column" to make the block's purpose clear and still get the benefit of the utility of Orchest "finding" all the pipelines in the project for you, and allowing for easy switching.

Issue #948, opened 2022-05-13
On load of project without pipeline prompt to create a pipeline automatically


ricklamers:
**Describe the problem this improvement solves**
It's not immediately clear how to proceed when creating an empty project as a first time new user when you see this:

**Screenshots**
![image](https://user-images.githubusercontent.com/1309307/168325093-f4b78cbc-f75a-4586-bdba-bd6ac0fc8beb.png)

**Describe the solution you'd like**
"Pre-click" the create pipeline button on pipeline editor load if the project contains no pipelines.

astrojuanlu:
Or display the "create pipeline" more prominently the first time?

yannickperrenet:
I am also in favor of prominently adding an action to "create pipeline", for example in the middle of the pipeline canvas. Creating a pipeline right of the bat feels somewhat intrusive. New users might just want to check out the view instead of being forced into an action. Moreover, let's say they "cancel" out of the option then I think it is good practice to still prominently point them towards the right action, i.e. creating a pipeline.

PR #1060, opened 2022-06-27
New UI design

iannbing:
## Description

New design can be roughly divided into the following items:

- [x] Global navigation
- [x] Project list
- [x] Pipeline Editor
- [x] Job list
- [x] Environment List

Fixes: #178, #591, #821, #959, #1197

## Known issues: 

- [x] switching projects results in functionality breakage due to inconsistent states between ContextAPI and `zustand`. For example, when viewing environments in project A, switch to project B it still shows the environment list of project A.

- [x] jobs is still incorrect when switching projects. Fixed. It was introduced by #1260 .
- [x] Environment build status shows old status.

- [ ] if an environment of a pipeline is deleted and user attempts to create a job, it does nothing. It should at least prompt an alert.

- [x] File Manager context menu position is off 
Screenshot 2022-09-14 at 20 51 39

- [ ] After deleting the last pipeline, the `.orchest` file is still visible in File Manager (the pipeline is correctly deleted, it was just that FE should refresh).

- [ ] Job run inspection header styling issue 
Screenshot 2022-09-19 at 10 52 04

## Todo:

- [x] ensure zustand stores are reset when switching projects.
- [ ] Pipeline editor indefinite canvas grid (as per #94) 
- [ ] Implement a custom pagination to replace the built-in `TablePagination`
- [x] Add confirmation if user attempts to cancel a job

If time allows it, then it would be great to also resolve the following issues:
- [x] https://github.com/orchest/orchest/issues/959
- [ ] https://github.com/orchest/orchest/issues/1009
- [x] https://github.com/orchest/orchest/issues/736
- [x] https://github.com/orchest/orchest/issues/568
- [x] https://github.com/orchest/orchest/issues/948


yannickperrenet:
(cc @ncspost @ricklamers @iannbing @mausworks)

image

Feels a bit strange to me that we say "everything unverified can contain malicious code" and then only have 1 project in the list that is verified. We basically give the users a bunch of templates that they shouldn't use because it could be a security hazard...
@ricklamers Can't we do something as to make all projects listed in `orchest-examples` verified? (We check them before including them, we just have to make sure we add a commit hash or something so it can't be updated).

----

image

What happened with the old thought of "shopping for examples/templates"? Thus being able to either view the imported project or shop for more examples.

----

image

- When renaming a project, the page needs to be refreshed in order to see the name change. As a user I expected the rename to not go through.
- In addition, the menu that opens when clicking the dots remains on screen until I click somewhere else (and not when it loses focus). Is this intended? Because now I need to click twice on another button in the UI for the action to take effect (first click closes menu, second click clicks button). @ncspost 

----

image

Instead of just pointing to the docs, should be point the user to creating a project as well? Either by adding an arrow to the button or adding an additional button? @ncspost 

---- 

image

I was looking for a right-click context menu on the project item, expecting to find the same actions I have in the project page, i.e. "delete project", "rename project", "project settings". I felt like I was lost for a while until I found the "projects page" button at the bottom.

yannickperrenet:
image

- [ ] When changing the zoom level by changing the value in the field, e.g. from 100 to 200, I expect the zoom to be applied when I press `<enter>` but it only works if I click some place else. This means that if I don't like the zoom-level I have to move my mouse again in order to change the zoom-level vs just clicking again to activate the field.

----

https://user-images.githubusercontent.com/26223174/184153364-26b2925c-daf5-4bac-9300-6c4c3b60d97c.mp4

- [ ] Step jumps to original create position when entering the path name

----

https://user-images.githubusercontent.com/26223174/184154309-dea449ff-2ac9-4a89-99f4-911e6007ac0c.mp4

- [ ] Hard to discover how to create a file. The empty state talks about no "pipeline steps" and when one is created, you can't actually do anything because there is no file assigned to it. This way we make it mandatory for users to find the small "create file" button in order to do anything in the editor. @ncspost 
    In addition, when a user leaves the step on the canvas without creating the file, then running the pipeline (e.g. through "run all") will always lead to the step not executing. So I feel like we need to handle the case of file creation/assignment better.

----

https://user-images.githubusercontent.com/26223174/184156391-dc2059b6-7192-4d8c-a7e7-570ba173cd0a.mp4

- [ ] On mouse release the connection should be removed when it is not connected yet. Edge case where this doesn't happen if trying to connect to the output of another step

----

https://user-images.githubusercontent.com/26223174/184157245-0e842d57-c410-47ca-8b55-fa077bbb760c.mp4

- [ ] Mouse wheel behavior seems ill-defined. Sometimes using the mouse-wheel will scroll up and down and sometimes it will start zooming in and out. This only happens for me when using the "up-and-down-mouse-wheel" not the "side-ways-mouse-wheel". 
- [ ] Talking about mouse and trackpads. I can't drag the canvas using my trackpad + holding space bar. I guess this is because trackpad users can just use the trackpad to navigate the canvas? But if that is the case, then scrolling up/down on the trackpad shouldn't change the zoom-level but should instead pan the canvas right?

mausworks:
I mean, it's right there, where we always put the "primary buttons" in the new design:

image

mausworks:
> When renaming a project, the page needs to be refreshed in order to see the name change. As a user I expected the rename to not go through.

This is fixed, or I can't reproduce.

mausworks:
>  Mouse wheel behavior seems ill-defined. Sometimes using the mouse-wheel will scroll up and down and sometimes it will start zooming in and out. This only happens for me when using the "up-and-down-mouse-wheel" not the "side-ways-mouse-wheel".

This has been improved now, but there's still some jankyness, I'll look further at it.

> Talking about mouse and trackpads. I can't drag the canvas using my trackpad + holding space bar. I guess this is because trackpad users can just use the trackpad to navigate the canvas? But if that is the case, then scrolling up/down on the trackpad shouldn't change the zoom-level but should instead pan the canvas right?

I think that for trackpad users with multi-touch enabled, two-finger panning is the recommended approach. To zoom you should pinch (that's how it works on MacOS at least). I'll do some testing on my ThinkPad as well to ensure that the behavior is consistent, but I think it should be improved after https://github.com/orchest/orchest/pull/1060/commits/614fc56a41e642b81522337601000c7add344ba2 at least.

iannbing:
> Hard to discover how to create a file. The empty state talks about no "pipeline steps" and when one is created, you can't actually do anything because there is no file assigned to it. This way we make it mandatory for users to find the small "create file" button in order to do anything in the editor. 

@yannickperrenet @ncspost 

Currently user can do the following to create a file:
- Right click on the step to open context menu
- create file button in File Manger.

How about adding the following paths:  
- click on the step to open the Properties panel, add a "+" button next to the file picker, which opens create file dialog
- double click on a step without a file (i.e. file_path is "")  -> open create file dialog

ncspost:
Good points @yannickperrenet @iannbing 

> click on the step to open the Properties panel, add a "+" button next to the file picker, which opens create file dialog

I'm a maybe on this 

> double click on a step without a file (i.e. file_path is "") -> open create file dialog

I'm not so sure about this one. Very divergent behaviour from a double click

-------

Some of my ideas:

- Slightly update empty state 
   - New copy
   - Add arrow to show where buttons are
   - Add buttons to create

CleanShot 2022-08-17 at 09 40 47@2x

- Include file creation in walkthrough: https://linear.app/orchest/issue/ORC-620/adoption-tutorial-walkthrough-tooltips
- Right-click side panel context shows file & folder creation/import options
- Drag&drop file/folder from computer file manager onto canvas or file manager imports
- Change `new pipeline` button into a generic new dropdown with all the options.

CleanShot 2022-08-17 at 09 48 17@2x

WDYT?

yannickperrenet:
> Slightly update empty state

I like this. The arrow definitely helps to point the user to where they should be looking.

> Change new pipeline button into a generic new dropdown with all the options.

I think I like this as well (maybe tweaking the listed options).

> click on the step to open the Properties panel, add a "+" button next to the file picker, which opens create file dialog

This is actually on our roadmap:
image

mausworks:
When the step sidebar is open at the same time as the settings, there's a z-index issue.

image

mausworks:
When trying to create a new job while environments are building, you get the following error:

```
Uncaught (in promise) Error: Some referenced environments do not exist in the project. The following environments do not exist: ['c56ab762-539c-4cce-9b1e-c4b00300ec6f'].

Please make sure all pipeline steps and services are assigned an environment that exists in the project.
    at fetcher.ts:29:7
    at Generator.next (<anonymous>)
    at fulfilled (main.js?hash=1663154647:74:26)
```

ricklamers:
> On load of project without pipeline prompt to create a pipeline automatically #948

I think this is definitely worth adding. @iannbing should be as simple as creating that file in the Python create project endpoint. Could you look into that?

yannickperrenet:
**To provide a good overview of my review comments I will keep updating this comment as I go.**

----

image

Creating a new Project as a new user and being taken to the Pipeline editor, I must say I would be a bit lost. My reasoning behind this:
- [ ] I highly doubt I would read the full help message to reach: "Drag & drop files from the file manager to get started.". In addition, when creating a new Project I don't even have any files to drag & drop... so basically I have to create a file first, which ... only has a very small icon in the file manager. So basically I have a very hard time to get started.
- [ ] The message reads `No Pipeline steps` which indicates (given that it is the biggest title/text on the page) that I should go ahead and create a Step. However (if I can even find the `new step` button), when I press `new step` I have to select a file from my Project, but then again... I don't have a file in my project --> I have to create a file first.

In short, wrong call to action in the empty state. We should people to create files if they don't exist in my opinion.

Then some nitpicks:
- [ ] `No Pipeline steps` --> `No Pipeline Steps` (it is also capitalized in the help text)
- [ ] `an executable file` --> Strictly speaking (yes, we used to have this in our docs -- we might even still have it) Notebooks and Python scripts are by default not executable, i.e. you have to `chmod +x` them. Thus this language isn't fully correct. Although I don't see a quite way to change the language to be correct and still convey the message of "file is executed in an Environment".

ricklamers:
image

Potentially a known issue, but this is what I'm seeing when switching to the "jobs" page through the header nav.

ricklamers:
image
image

Comparison of Figma to what I see. Note I'm on a high DPI screen:
image

PR #1061, opened 2022-06-27
Celery worker queries `orchest-api` for parallelism setting on start

fruttasecca:
## Description

Fixes a bug where the job and interactive runs parallelism level wouldn't be applied to the celery-worker.


yannickperrenet (approved):
LGTM!

PR #1069, opened 2022-06-27
New UI design part 1: Global navigation

iannbing:
## Description

Global navigation of the new design.

Note that:
- The endpoint of fetching projects doesn't support pagination yet. So currently it shows all projects in one go, and thus the search is actually client-side filtering.
- It's still unclear how we want to promote "examples" as we discussed in the Figma file.

### DIFF
--- a/scripts/redeploy_branch.sh
+++ b/scripts/redeploy_branch.sh
@@ -54,7 +54,7 @@ bash "${DIR}/build_container.sh" -m -t ${TAG} -o ${TAG}
 echo "Installing the orchest-cli from this branch..."
 pip install "${DIR}/../orchest-cli" > /dev/null
 
-orchest install
+orchest install --dev
 
 echo "Restoring auth users..."
 # Needed because some rows contain dollar symbols, perhaps we should
--- a/services/auth-server/app/app/models.py
+++ b/services/auth-server/app/app/models.py
@@ -10,15 +10,13 @@ class User(db.Model):
     uuid = db.Column(
         db.String(36),
         primary_key=True,
-        # required to be referenced as a foreign key by the Token table,
-        # since postgres does not accept foreign keys referencing non
-        # unique fields
         unique=True,
     )
 
     username = db.Column(
         db.String(255),
         primary_key=True,
+        unique=True,
     )
 
     password_hash = db.Column(db.String(255), nullable=False)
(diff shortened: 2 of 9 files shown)


ricklamers:
JupyterLab has this issue where it doesn't render properly when it's not "visible on screen" during certain phases in its app loading phase. This happens a lot because we <iframe> embed it.

It looks like this: 
image

The fix I had working before was we would detect when the visual rendering had glitched and reload itself (you can check the code in Jupyter.js but basically it would check the size of Jupyter elements against the iframe viewport size).

Not sure if introduced in this PR specifically, but it's now breaking for me. Worth figuring out what the issue is and fixing it. @iannbing if you get stuck with this please Slack DM me and we'll pair program on it, I have some context that might help you move faster.

ncspost:
Small screen behaviour is a bit off

CleanShot 2022-06-30 at 10 35 58@2x

Could the tab collapse to a menu when width is too narrow?

CleanShot 2022-06-30 at 10 45 46@2x

We could potentially bucket settings, help and signout there too.

ncspost:
Add 8px upper and lower padding

CleanShot 2022-06-30 at 10 52 33@2x

ncspost:
Can we include a "No Projects found"

CleanShot 2022-06-30 at 10 55 23@2x

ncspost:
Can we have settings and help tooltips please

CleanShot 2022-06-30 at 10 57 00@2x

ncspost:
Can we add a slide out/slide in animation when the menu is clicked please?

I'm not completely convinced, but maybe we should have a chevron arrow on the button too...

CleanShot 2022-06-30 at 11 10 24@2x

Top = menu closed
Bottom = menu open

ncspost:
Search projects → Search Projects - Capitalise primary objects

iannbing:
> Can we add a slide out/slide in animation when the menu is clicked please?

Not sure if it's an improvement. A small heads-up: this PR is just the first of many PR's of the overhaul (note that this PR is **not** merging into `dev`). Maybe we could focus on making the MVP first and tweak other details later (i.e. animation, responsiveness for screen size smaller than 1052px)? We might have better ideas after putting things together. 

> When keyboarding with tabs...

Added. However, "search" -> "new" -> "import" is quite counter-intuitive. I made it as follows:
Tab 1. search **(if user press "arrow up", it goes to "new")**
Tab 2. project 1
Tab 3. .....
Tab 4. last project
Tab 5. view all projects
Tab 6. new
Tab 7. import

You could check the latest commit to play with it. 😄

iannbing:
@ricklamers The JupyterLab issue is not related to this PR, and this PR has involved changes to more than 60 files. I think it would be better to fix the said issue in another PR, see [issue](https://github.com/orchest/orchest/issues/1075).

Issue #570, opened 2021-12-01
Support multiple client sessions per user


ricklamers:
**Describe the solution you'd like**  
Support multiple user sessions per user

**What does your solution aim to solve?**  
Right now a user is logged out of their browser client when the same user is logged in someplace else. We ideally want to support multiple client sessions per user.

yannickperrenet:
After I quick look, it is probably caused by how we do authentication based on the token in the cookie:

https://github.com/orchest/orchest/blob/e082a136524bea29968a47729ea38e7ca3a3e541/services/auth-server/app/app/views.py#L60-L64

and

https://github.com/orchest/orchest/blob/e082a136524bea29968a47729ea38e7ca3a3e541/services/auth-server/app/app/views.py#L170-L176

We instead need the ability to have multiple tokens for one user.

PR #1073, opened 2022-06-29
Support multiple client sessions per user

fruttasecca:
## Description

Description (including a screenshot/video if applicable) of what your PR solves or adds.

Fixes: #570 

## Checklist

### DIFF
--- a/services/auth-server/app/app/views.py
+++ b/services/auth-server/app/app/views.py
@@ -52,29 +52,19 @@ def register_views(app):
         cookie_token = request.cookies.get("auth_token")
         username = request.cookies.get("auth_username")
 
-        user = User.query.filter(User.username == username).first()
-
-        if user is None:
-            return False
-
-        token = (
-            Token.query.filter(Token.token == cookie_token)
-            .filter(Token.user == user.uuid)
-            .first()
+        token_creation_limit = datetime.datetime.utcnow() - datetime.timedelta(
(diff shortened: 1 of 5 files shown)


services/auth-server/app/app/views.py
```diff
@@ -144,7 +145,7 @@ def handle_login(redirect_type="client"):
             # Check whether the given user exists.
             user = User.query.filter(User.username == username).first()
 
-            invalid_login_msg = "Username password combination does not exist."
+            invalid_login_msg = "Username does not exist."
```
yannickperrenet, line 154: I would prefer sticking with the older message as it is more secure to always give the same message. Otherwise an attacker could infer, based on the message, whether or not the user exists.

This change would revert: https://github.com/orchest/orchest/pull/704

yannickperrenet (approved):
Haven't tested, but looks good to me!

services/auth-server/app/app/views.py
fruttasecca: Makes sense, will do.

fruttasecca: https://github.com/orchest/orchest/pull/1073/commits/91703706614052219c2cab47b95d3c8423a0b0e0

yannickperrenet (approved):

ricklamers:
- Increase token duration to 14 days
- Explicit check for username existence to not rely on broader integrity exception (which could mean something else)

Otherwise, LGTM!

PR #1074, opened 2022-06-29
Periodically cleanup builder cache

fruttasecca:
## Description

Adds an `orchest-api` level scheduler job that periodically cleanups the builder cache since it's (AFAIK) the only storage that user actions can't affect.

Opinion on adding an `orchest-api` endpoint that queues such a task? Even without a GUI button would be fine, just to know that we have at least a temporary solution for "power" users that need/want it to be cleaned up asap, might also prove useful for the cloud.

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_ctl.py
+++ b/services/orchest-api/app/app/apis/namespace_ctl.py
@@ -9,6 +9,7 @@ from orchestcli import cmds
 
 from _orchest.internals import config as _config
 from app import schema, utils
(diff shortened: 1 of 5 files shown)


yannickperrenet:
> the builder cache since it's (AFAIK) the only storage that user actions can't affect.

How about the `/pip-cache`?

> Opinion on adding an orchest-api endpoint that queues such a task?

Sounds like a good idea :+1:

yannickperrenet (approved):

PR #1090, opened 2022-07-06
Feat/UI overhaul project list

iannbing:
NOTE: this PR actually starts from [this commit ](https://github.com/orchest/orchest/tree/99e0f6a5f7cc0c82bdb7685b4d93ecf35f04cca5), but GH includes many past commits, which result in 55 files changes. You only need to review the changes within `/services/orchest-webserver/client`. Sorry for the annoyance.

## Description

The new Project List view.

Some explanations on not implementing the details in the design:
- The drop-files-to-screen-to-create-project  functionality is removed due to unclear value and usability concerns (now Examples Tab resides in the same view, it's confusing, for example, a user might want to drag-n-drop files to submit an example). 
- The pagination of Project List in the design didn't consider "specify the number of project per page", we need further discussion on this. At the moment the original `TablePagination` is kept.
- I think the "Sorting" in the Example List needs more discussion: 1) it's a new feature 2) its location is far away from the list in the design, might have some usability concerns (normally the filter/sorting option should directly on top of the list). So, I decided to postpone the implementation.

Screenshot 2022-07-06 at 11 35 41
Screenshot 2022-07-06 at 11 35 19
Screenshot 2022-07-06 at 11 35 07


services/orchest-webserver/client/src/projects-view/ProjectList.tsx
```diff
@@ -0,0 +1,214 @@
...
+  | "job_count"
+  | "environment_count"
+> & {
+  settings: string;
+};
+
+export const ProjectList = ({
+  refetch,
+}: {
+  refetch: () => void | Promise<Project[]>;
```
mausworks, line 39: Unless there is a good reason for this approach, I think it would be better if the parent component manages the state instead:

```ts
type ProjectListProps = {
  projects: Project[];
  onDelete(project: Project): Promise<void>;
}
```

This makes the rendering logic of the component dead simple, and we don't have to worry about "refetching" in the component itself, we just have to worry about calling the delete callback and we can then easily change how that is implemented (instead of using a refetch, we could filter out the project for example). So the props will then describe the functionality that the component exposes—and not implementation details about how that is done.

services/orchest-webserver/client/src/projects-view/ProjectsTabs.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const selectTab = React.useCallback(
+    (tabIndex: number) => {
+      setProjectTabIndex(tabIndex);
+      // This is to persist the nice MUI Tabs transition.
+      window.setTimeout(() => {
+        history.replace(`${siteMap.projects.path}?tab=${tabIndex}`);
+      }, 500);
+    },
+    [setProjectTabIndex, history]
+    // [setProjectTabIndex]
```
mausworks, line 27: Nit: Please remove commented-out code. :)

mausworks (changes requested):
A solid lift overall! But I have left a few comments on some things I have noticed, but no biggies. 👍 

Once I get a better understanding of the code base I will be able to do more in-depth reviews.

services/orchest-webserver/client/src/components/DataTable.tsx
```diff
@@ -270,6 +270,8 @@ function Row<T, C>({
             ? { cursor: "pointer" }
             : null),
           height: data.details ? rowHeight - 1 : rowHeight,
+          // color: (theme) => theme.palette.action.disabled,
+          // disabled ? theme.palette.action.disabled : undefined,
```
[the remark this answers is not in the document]
iannbing: Totally missed this. Will do. Thanks 👍

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
```diff
@@ -16,7 +17,8 @@ export type IProjectsContextState = {
   pipelineSaveStatus: "saved" | "saving";
   pipelines: PipelineMetaData[] | undefined;
   pipeline?: PipelineMetaData | undefined;
-  projects: Project[];
+  projects: Project[] | undefined;
+  examples: Example[] | undefined;
```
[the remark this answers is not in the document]
iannbing: Sometimes we want to set a value as `undefined` intentionally. Then `?` doesn't allow you to do so. Therefore, we need to distinguish the following cases:
- a variable is optional.
- a variable is required but accepts `undefined` as value.

Although in most cases they are interchangeable. 😄 

By the way, in our code base, we tend to use `undefined` over `null` to prevent falling for the infamous "null is an object" issue. And I guess because of this, `| undefined` is more useful than `?` in our convention.

services/orchest-webserver/client/src/pipeline-view/pipeline-connection/common.tsx
```diff
@@ -11,7 +11,7 @@ export const curvedHorizontal = (
   x2: number,
   y2: number
 ) => {
-  let line = [];
+  let line: (number | string)[] = [];
   let mx = x1 + (x2 - x1) / 2;
```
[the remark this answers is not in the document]
iannbing: Will do. Thanks!

services/orchest-webserver/client/src/project-selector/ProjectSelectorMenu.tsx
```diff
@@ -27,21 +28,26 @@ export const ProjectSelectorMenu = ({
...
 
   const customNavigation = React.useCallback(
-    (path: string) => {
+    (targetTab: PROJECT_TAB) => {
       onClose();
-      if (location.pathname !== path) navigateTo(path);
+      if (
+        location.pathname !== siteMap.projects.path ||
+        (location.pathname === siteMap.projects.path &&
+          tab !== targetTab.toString())
```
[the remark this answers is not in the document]
iannbing: Good one! Will do.

services/orchest-webserver/client/src/projects-view/ProjectsTabs.tsx
iannbing: Will do. Thanks!

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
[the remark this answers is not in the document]
mausworks: You are allowed to set a property to undefined explicitly, even if you use the `foo?:` syntax, see [this playground for various examples](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvKAoK2oEMD8AXFAM7ABOAlgHYDmANJjlAEbFlV1QA+UArtQAmEAGY0IgzAF8A3BgwB6BVAAqAC0okomqGOoRCGAMZxqZKAA9i8RCnS5iA4XomM2-IaPGTZi5ep0dFyhgBEMTM2AoEGsEZDRWR08XHzklVQ0tPjASCGAtYDVoFXBoAAU4ABtKIwg3CCNcPlyoAAMWVu0tajgomhDMkNLw03MAL1jbBIcPZ28oKT8MnVxKkkRs3PyBiABbEcioNUn49CkgA). It's the opposite that doesn't work: `project:  Project[] | undefined` does not allow you to omit the property, you must specify as undefined.

But from your examples I new understand your distinction between the two. As a side note, I personally I prefer `null` in this case, since it's an explicit value and not a "default value". 

If you see `null`, you know that `null` was intentionally set, if you see `undefined`, it can be for at least two reasons: it was explicitly set, or somebody/something forgot to set a value. But I'm no evangelist here! I understand that you want to avoid the `typeof foo === 'object'` trap. :)

services/orchest-webserver/client/src/projects-view/ExampleList.tsx
```diff
@@ -0,0 +1,219 @@
...
+};
+
+const ENTRIES_PER_PAGE = 10;
+
+export const ExampleList = ({
+  data = [],
+  imporProject,
+}: {
+  data: Example[] | undefined;
+  imporProject: (url: string) => void;
```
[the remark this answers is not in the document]
mausworks: I highly recommend a spell checker, I have used [Code Spell Checker for VSCode](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) for years, and I could never go back!

There is probably something similar available if you use a different IDE, a really good productivity tool IMO. :)

services/orchest-webserver/client/src/projects-view/ProjectList.tsx
mausworks: >  If the concern is about the complexity of this component, indeed we could extract 'DataTable' as another "dumb" component so this component is only responsible for processing data from useProjectsContext.

I think this sounds like a great idea. 👍 

> Now since you raise the concern, should we always do filter and solve the multi-users scenario properly (e.g. polling or websocket) later systematically?

I think that (for now) we can do it the way we have been doing, it works and it's likely not a concern, and it's not a concern from my side at all. If we want more interactiveness in the future, I think we should look at a new approach for that.

The reason I mentioned it was just as an example for why it would be nice to separate the concern of "action of deleting the project" away from "how we do that".

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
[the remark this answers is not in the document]
iannbing: Thanks for the example 👍.

> If you see null, you know that null was intentionally set, if you see undefined

I used to follow this convention as well :-). Then I found it hard to hold (it can only be maintained by engineers, and easily be overlooked during code review). Then I fall back to either "has value" or "has no value" (yes, [we have a `hasValue` helper function](https://github.com/orchest/orchest/blob/e76376941d2d7465b3e837dd3cebbc8035cd9214/lib/javascript/utils/src/typed.ts#L272) 😆).

Maybe not in this PR, but if you find valid use case, feel free to apply this rule again 🙂.

services/orchest-webserver/client/src/projects-view/ProjectList.tsx
iannbing: I have refactored it and extract `ProjectsTable` out of `ProjectList`. And also cleaned up the use of `useProjectsContext`.

services/orchest-webserver/client/src/projects-view/ProjectsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  | "session_count"
+  | "job_count"
+  | "environment_count"
+> & {
+  settings: string;
+};
+
+export const ProjectsTable = ({
+  projects,
+  projectBeingDeleted,
```
mausworks, line 31: Nit: No action needed, just a thought. We could instead add `beingDeleted: boolean` to the props of each `ProjectRow`. Even though it might not be possible today, there could be a future where we support multiple projects pending deletion at the same time, so I guess it could be nice? 🤷

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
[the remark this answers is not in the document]
mausworks: It's super rare that I do any other checks than `if (!foo)` or similar to check for "nullish values", so for all intents and purposes it doesn't _really_ matter. If the convention here is to use "explicit `undefined`", then I will follow that convention. :)

mausworks (approved):
I think the state of this is good now. Let's ship it and then iterate on it in the future if we need to. 👍 

I have left a few comments, but they are all just questions/ideas. Nothing blocking. :)

services/orchest-webserver/client/src/project-selector/useProjectSelector.tsx
```diff
@@ -11,7 +11,7 @@ export const useProjectSelector = (
   targetRoutePath: string | undefined,
   customNavigateTo: (projectUuid: string, path: string | undefined) => void
 ) => {
-  const projects = useFetchProjectsForSelector();
+  useFetchProjectsForSelector();
```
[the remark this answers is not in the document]
iannbing: The returned value was meant for passing along to the components, but since the children could get `projects` from `useProjectsContext`. The return value here is no longer needed.

services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -48,6 +48,13 @@ def get(self):
         jobs = models.Job.query
         if "project_uuid" in request.args:
             jobs = jobs.filter_by(project_uuid=request.args["project_uuid"])
+        if "active" in request.args:
+            should_select_active = request.args["active"] == "true"
+            jobs = (
+                jobs.filter(models.Job.status == "STARTED")
+                if should_select_active
+                else jobs.filter(models.Job.status != "STARTED")
+            )
```
yannickperrenet, line 57: Could probably change this into something like:

```python
if request.args.get("active", "false") == "true":
    jobs = jobs.filter(models.Job.status == "STARTED")
```
given that only when `active=true` we want to do additional filtering.

Arguably, I would say the status should be in `["STARTED", "PENDING"]` because one-off jobs will be `PENDING` until they are started. Given that cron-jobs will be `STARTED` even though they are not executing at the moment the one-off jobs should also be considered active in a similar state (even though we call it differently from a BE perspective).

iannbing: My intention was to cover the case: when `active` is `false`, return jobs with stale states, which seems to me more precise with the given arg. But indeed the code can be improved. Pushed my changes.

> I would say the status should be in ["STARTED", "PENDING"]

Thanks for the explanation. Added it.

ncspost:
It's okay to replace this with GitHub widget actually... it's a better indicator of what this is/doe
CleanShot 2022-07-08 at 11 04 45@2x
s

ncspost:
CleanShot 2022-07-08 at 11 07 04@2x

Can this go straight to this dialog?

CleanShot 2022-07-08 at 11 07 46@2x

ncspost:
CleanShot 2022-07-08 at 11 09 58@2x

Title of this "New Project"

ncspost:
CleanShot 2022-07-08 at 11 11 28@2x

Can we change this to the non-table type of pagination please?

ncspost:
CleanShot 2022-07-08 at 11 12 57@2x

Reduce to 0px

ncspost:
Reduce by 16px

CleanShot 2022-07-08 at 11 16 02@2x

ncspost:
CleanShot 2022-07-08 at 11 17 59@2x

Capitalise "Delete Project"

ncspost:
Could we change the text button for all dialogs to be blue?

CleanShot 2022-07-08 at 11 18 41@2x

ncspost:
Change to "more options"

CleanShot 2022-07-08 at 11 19 44@2x

ncspost:
Please use this version illustration!

Project

ncspost:
This is super annoying overlap... could you move intercom to bottom right side of screen please?

CleanShot 2022-07-08 at 11 26 39@2x

ncspost:
CleanShot 2022-07-08 at 11 27 50@2x

Could we match with https://github.com/orchest/orchest/pull/1027#issuecomment-1151116909

Also could we update text button on both dialogs to say "Keep Project"?

services/orchest-api/app/app/apis/namespace_jobs.py
yannickperrenet: > My intention was to cover the case: when active is false, return jobs with stale states, which seems to me more precise with the given arg.

Totally agree that that is better than what I proposed :+1:

iannbing: Pushed my changes.

iannbing:
@ncspost All your remarks are addressed. Here are some explanations.

> It's okay to replace this with GitHub widget actually... it's a better indicator of what this is/does

Actually I found a technical bottleneck: every example will fire a request to GitHub, and user could pretty easily hit rate limits with an IP (i.e. the numbers will stop showing after a few visits to the Examples View.). 

> Can we change this to the non-table type of pagination please?

The fact that it is a MUI `Table` under the hood and `TablePagination` just works out of box. And the same Table component is re-used in many places. If we want to plug `Pagination` to `Table` component, it'd be better to do this in the mega PR, where we could capture all use cases.


> Can this go straight to this dialog?

Good idea. I'll come back to this later.

iannbing:
> Can "click Import of an example" go straight to this dialog?

@ncspost Just pushed my changes. You could test it.

iannbing:
> Please use this version illustration!
> 
> Project

@ncspost  I don't think this version is the same as the one in the final Figma file. https://www.figma.com/file/o9S8qtPAx3Vd1HIVtsQfYv/ORC-573-Navigation%2C-layout-%26-IA-(solution)?node-id=1445%3A60215
Could you update the Figma file since it's the single source of truth for implementation? Thanks.

ncspost:
CleanShot 2022-07-11 at 09 24 12@2x

Caught a typo here.

I see what you mean about it being a little weird with the tooltip @iannbing - we can make this better by using a disabled cursor and extending the text to "First create a Project to navigate here"

ncspost:
CleanShot 2022-07-11 at 10 06 19@2x

I keep on getting this error message when trying to import quickstart

iannbing:
@ncspost 

> we can make this better by using a disabled cursor and extending the text to "First create a Project to navigate here"

Implemented and pushed.

> I keep on getting this error message when trying to import quickstart

I couldn't reproduce this on the latest commit. Could you deploy the latest commit and test it again?

PR #1093, opened 2022-07-07
Set correct permissions on `/pip-cache`

nhaghighat:
## Description

This PR fixes the `pip-cache` warning during environment image building by `chown`ing the pip-cache to 1000/100 which is jovyan/users. The warning:

> WARNING: The directory '/home/jovyan/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.

The warning was observed when running on GKE.

## Checklist

### DIFF
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -150,16 +150,17 @@ def _get_buildah_image_build_workflow_manifest(
                         "command": ["/bin/sh", "-c"],
                         "args": [
                             (
-                                # Change the UID/GID to jovyan/users
-                                "chown -R 1000:100 /pip-cache; "
+                                # Creating the cache directories
+                                "mkdir -p -m 777 /builder-pvc/cache/pip && "
+                                "mkdir -p -m 777 /builder-pvc/cache/conda && "
                                 # Build
                                 f"buildah build -f {dockerfile_path} --layers=true "
                                 # Package managers caches.
                                 # jovyan is the user of the base image.
-                                "-v /pip-cache:/home/jovyan/.cache/pip "
+                                "-v /builder-pvc/cache/pip:/home/jovyan/.cache/pip "
                                 # Obtained by running "conda info". Note
                                 # that this cache is also used by mamba.
-                                "-v /conda-cache:/opt/conda/pkgs "
+                                "-v /builder-pvc/cache/conda:/opt/conda/pkgs "
                                 # https://github.com/containers/buildah/issues/2741
                                 "--format docker "
                                 "--force-rm=true "
@@ -192,13 +193,8 @@ def _get_buildah_image_build_workflow_manifest(
                             },
                             {
                                 "name": "image-builder-cache-pvc",
-                                "subPath": "pip-cache",
-                                "mountPath": "/pip-cache",
-                            },
-                            {
-                                "name": "image-builder-cache-pvc",
-                                "subPath": "conda-cache",
-                                "mountPath": "/conda-cache",
+                                "subPath": "cache",
+                                "mountPath": "/builder-pvc/cache",
                             },
                             {
                                 "name": "image-builder-cache-pvc",
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -150,6 +150,8 @@ def _get_buildah_image_build_workflow_manifest(
                         "command": ["/bin/sh", "-c"],
                         "args": [
                             (
+                                # Change the UID/GID to jovyan/users
+                                "chown -R 1000:100 /pip-cache; "
                                 # Build
                                 f"buildah build -f {dockerfile_path} --layers=true "
                                 # Package managers caches.


fruttasecca (approved):
probably better to make it a `&&` so that a potential failure doesn't go unnoticed, aside from that, lgtm :+1:

fruttasecca:
What about the conda cache? Does it not lead to the same warning?

ricklamers:
I'd like to make sure we configure the filesystem in the right way where new files and folders will be permissively accessible for everything created in `/pip-cache`. There are various ways to achieve that depending on the filesystem assumptions we can safely make.

nhaghighat:
> Actually, this assumes a given user exists, might open up to edge cases, would it be better to set very relaxed write/read/execute etc. rules?

This is using the base image and we bake the base images with `jovyan` users in it, right? otherwise how would we mount it to `home/jovyan`?

> One thing to keep in mind is that users can also bring custom container images that have different user/group IDs

Can users bring their own base images? if not it is safe to assume the uid/gid always exist and all the files will have the same uid/gid

nhaghighat:
> Also, how does performance scale when the cache gets big?

I think it only has performance degradation the first time, after that the directory has proper permissions.

fruttasecca:
>This is using the base image and we bake the base images with jovyan users in it, right? otherwise how would we mount it to home/jovyan?

That's a very good point, I think this is some technical debt I left behind when we were pushing for the k8s release, definitively something we should aim to solve although not sure if now. The `conda` cache is not at the user home level, I wonder if there is a shared global cache for pip as well?

If fixing the aforementioned issue for custom user images I would be okay with temporarily mounting the pip cache only when we know it's one of our base images, although there are probably not too costly way to actually fix the issue altogheter.

nhaghighat:
I'm not fond of the idea of making it readable and writable for all users, I think we should explicitly specify the user that can use the directory and create the container to run as the uid/gid as we want, in this way we close the door for upcoming permission problems.

yannickperrenet:
> probably better to make it a `&&` (instead of `;`) so that a potential failure doesn't go unnoticed

I agree.

> Also, how does performance scale when the cache gets big? We moved away from permission adjustments on boot for that reason

I am also hesitant to just add this `chown -R` command here as it will run on every environment build (whereas I don't see why running it just once isn't enough). 

Similarly to (fixing permissions in the Dockerfile directly):
https://github.com/orchest/orchest/blob/945e815285a8fb876b730954bf7111835a0cdc92/services/base-images/base-kernel-py/Dockerfile#L40-L41
Why can't we fix the permissions of the volume mount directly? And why, in the first place, does the PVC not have the right permissions (does this have anything to do with the container-runtime given that we haven't seen it happening yet on minikube+docker)?

From the `buildah-build` manpage ([link](https://www.mankier.com/1/buildah-build)):
> Chowning Volume Mounts
>
> By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
>
> The :U suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.

yannickperrenet:
> Once the pvc is created it is empty and it does not have the directory, once we mount it to the builder pod, the directory will be created based on the default setting of the storage provider (default uid/gid and default mod). that's why I'm saying once we chown it for the first time, then for the next run the directory have proper permission. we don't even need to run it with -R.

That all makes a lot of sense. Thanks for clarifying :100: 

As long as we don't run `chown -R` on every build, then it would be fine with me. Back in the day we ran an optimized version of `chown -R` on the `userdir` and it became a huge bottleneck (hogging the CPU). So if it works to just make sure that the directory is created with the right permissions, then I'd say lets go for that!

nhaghighat:
Our discussion gave me another, probably safer for fixing the issue,
instead of mounting `pip-cache` (which is sub-directory of `builder-cache` pvc) to builder pod , mount the whole `builder-cache` to the builder pod, and create the `pip-cache` directory with correct permission if it does not exist.
So the directory will be created with write permission in the first place.

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -192,13 +193,8 @@ def _get_buildah_image_build_workflow_manifest(
...
                                 "name": "image-builder-cache-pvc",
-                                "subPath": "pip-cache",
-                                "mountPath": "/pip-cache",
-                            },
-                            {
-                                "name": "image-builder-cache-pvc",
-                                "subPath": "conda-cache",
-                                "mountPath": "/conda-cache",
+                                "subPath": "cache",
+                                "mountPath": "/builder-pvc/cache",
```
yannickperrenet, line 197: So I guess for existing users the sub-paths in the `image-builder-cache-pvc`, i.e. `pip-cache` and `conda-cache`, would stay lingering around in the PVC?

yannickperrenet: We could just delete the folders before the new `mkdir` command? As long as we add a comment like "To migrate existing users to a version >= v2022.07.2", I think that would be fine.

@fruttasecca What do you think?

nhaghighat: > So I guess for existing users the sub-paths in the image-builder-cache-pvc, i.e. pip-cache and conda-cache, would stay lingering around in the PVC

No, we have a cleanup builder cache task which will be created on startup and cleans the cache. right @jacobodeharo ?

yannickperrenet (approved):

services/orchest-api/app/app/core/image_utils.py
fruttasecca: >No, we have a cleanup builder cache task which will be created on startup and cleans the cache. right @fruttasecca ?

As @yannickperrenet mentioned it's run periodically but not on startup.

nhaghighat: A question for future, if we are `periodically` cleaning up the cache, what is the point of having the cache?

PR #1094, opened 2022-07-08
add gke and eks installation instructions

nhaghighat:
## Description

add gke and eks installation instructions.

Fixes: #issue

## Checklist


docs/source/_static/js/installer_widget.js
```diff
@@ -45,11 +53,12 @@ orchest install
 </pre>
 </div>
 
-<p>Now that Orchest is installed, it can be reached on the address returned by:</p>
+<p>Now that Orchest is installed, it can be reached on the IP address returned by:</p>
```
nhaghighat, line 56: I intentionally mentioned the address here, as the AWS does not return IP, it returns a DNS address.

PR #1096, opened 2022-07-08
Mount jupyterlab settings directory in image builder when building custom jupyter images

fruttasecca:
This PR also makes it so that the WORKDIR of custom jupyter images is the same as the non custom one, which is `/project-dir` (https://github.com/orchest/orchest/pull/1096/commits/4fee4c90b0df9acdf967c8b8457b1821376e2d09).


yannickperrenet (approved):
LGTM!

Feels like we should have caught these improvements earlier. Great job :smile:

Issue #1026, opened 2022-06-06
Auto-create step with new file


ncspost:
# Describe the problem this improvement solves

When I create a new file, I want to immediately add it to a step, so that there are fewer clicks in my workflow.

# Describe the solution you'd like

- Add a checkbox to the new file form
- Potentially remove new file dialog
- [Figma file](https://www.figma.com/file/BXk97lVzalwYVHvlSU7Y7N/?node-id=1324%3A10738)

# Screenshots

<img width="1895" alt="CleanShot 2022-06-06 at 12 31 40@2x" src="https://user-images.githubusercontent.com/6852479/172152685-8c762f7f-2bb4-4a34-a3da-4379ecca161b.png">

ricklamers:
I think the copy for the checkbox should be "Create a new step for this file"

Also, I think the checkbox can be auto-checked as I believe that's the common case, for when creating the file in the file manager integrated into the pipeline editor.

Files created in JupyterLab (or in VS Code/PyCharm with the SSH workspaces feature coming up, won't have this dialog naturally).

PR #1097, opened 2022-07-08
Create pipeline step on file creation

mausworks:
## Description

**I recommend reviewing this PR commit-by-commit**.

Closes: #1026

This PR adds the ability to automatically add a pipeline step to the project when you add a file, as described in #1026.
It's implemented as the first design implies: I haven't removed the dialog (I'll save that for next iteration when more of the new design is in place).

This is how it looks:

image

The new step is added to the pipeline viewport using the same conventions as "ADD STEP +" does, and I have extracted this into its own hook called `useAddStep`.

I have also replaced Formik with react-hook-form. The rationale is that we're only interested in using the `useFormik` hook to ease state-management, but using that hook is discouraged by the maintainers of the library:

> [Despite its name, it is not meant for the majority of use cases](https://formik.org/docs/api/useFormik)

react-hook-form is also about half the size of Formik, I personally prefer the API, and using react-hook-form actually reduced LOC and complexity overall. 

There are also some minor improvements, thus, I recommend reading the PR commit-by-commit.

## Checklist


services/orchest-webserver/client/src/pipeline-view/hooks/useCreateStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+      const [offsetX, offsetY] = pipelineCanvasState.pipelineOffset;
+
+      const position = {
+        x: -offsetX + clientWidth / 2 - STEP_WIDTH / 2,
+        y: -offsetY + clientHeight / 2 - STEP_HEIGHT / 2,
+      };
+
+      dispatch(createStepAction(environment, position, fileName));
+    } catch (error) {
+      setAlert("Failed to create step", String(error));
```
mausworks, line 40: I think it might be better to just let this hook throw (I'm not sure what exactly would do that, honestly).

Do you think we should to remove the `try`/`catch` @iannbing?

services/orchest-webserver/client/src/pipeline-view/file-manager/common.tsx
```diff
@@ -228,6 +228,16 @@ export const cleanFilePath = (filePath: string, replaceProjectDirWith = "") =>
     .replace(/^\/project-dir\:\//, replaceProjectDirWith)
     .replace(/^\/data\:\//, "/data/");
 
+/** Prettifies the root name for display purposes. */
+export const prettifyRoot = (root: string) => {
+  switch (root) {
+    case "/project-dir":
```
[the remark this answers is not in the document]
yannickperrenet: @ricklamers @astrojuanlu Any opinions from you guys?

I would say "Shared data" is a good one, it does cover the concept that `/data` is a directory that is shared across all projects. One nitpick would be that by not saying `/data` then users might not know how to write something to "Shared data", i.e. where does it live? How do I access the data in "Shared data"?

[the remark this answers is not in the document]
ricklamers: I'm not a fan of using "Shared data" as I find it a bit too "vague". I'd rather consistently call it `/data` everywhere (it's a pattern I've seen in multiple other tools) as it hints at the fact that we always have `/data` mounted and available in every execution context.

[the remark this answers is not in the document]
mausworks: Gotcha! I haven't worked that extensively in this industry, so I'm not aware of its common conventions.

I guess the thing that feels a little bit "off" to me, is that we prettify one, and not the other. But now I at least understand the rationale.

ncspost:
Design feedback:

services/orchest-webserver/client/src/pipeline-view/file-manager/ActionBar.tsx
```diff
@@ -23,55 +23,47 @@ const FileManagerActionButton = styled(IconButton)(({ theme }) => ({
...
-          setSelectedFiles([fullFilePath]);
+        isOpen={!pipelineIsReadOnly && creating === "file"}
+        onClose={() => setCreating(null)}
+        root={rootFolder}
+        onSuccess={(file) => {
+          setSelectedFiles([file.fullPath]);
           reload();
+
+          if (file.createStep) {
+            createStep(file.projectPath);
```
iannbing, line 58: After clicking "CONFIRM", you will see a step is created, but you would also see two failed requests:
- GET `/async/file-management/exists`
- POST `/async/pipelines/json/`

And then if you refresh the page, the step you just created disappeared from the canvas,  meaning that it failed to save the file path in the Pipeline JSON definition file (but the file was indeed created successfully).

Note that the file path of a step is relative to the pipeline file (i.e. the `.orchest` file). You would need `pipelineCwd` in `PipelineEditorContext` to get the relative path from `file.projectPath`.

services/orchest-webserver/client/src/pipeline-view/hooks/useCreateStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { pipelineCanvasState } = usePipelineCanvasContext();
+
+  return (fileName?: string) => {
+    if (!pipelineViewportRef.current) {
+      console.error("Failed to create step: pipeline viewport not set");
+      return;
+    }
+
+    // Use the first environment as the default:
+    // The user it can be changed later.
```
iannbing, line 23: Typo in the comment? 🙂

services/orchest-webserver/client/src/pipeline-view/file-manager/useCreateFile.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * The hook throws if there is no `projectUuid` in the route.
+ */
+export const useCreateFile = (root: string) => {
+  const { projectUuid } = useCustomRoute();
+
+  if (!projectUuid) {
+    throw new Error("A project UUID was not found in the route.");
+  }
+
+  return async (path: string) => {
```
iannbing, line 18: Similar to `useCreateStep`, an anonymous, un-memoized return function. 🙂

services/orchest-webserver/client/src/pipeline-view/file-manager/useCreateFile.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * The returned path will be an absolute project path, starting with `project-dir:/` or `data:/`.
+ *
+ * The hook throws if there is no `projectUuid` in the route.
+ */
+export const useCreateFile = (root: string) => {
+  const { projectUuid } = useCustomRoute();
+
+  if (!projectUuid) {
+    throw new Error("A project UUID was not found in the route.");
+  }
```
iannbing, line 16: In my experience, throwing an error that cause a blank page is not preferred. We could log it or send an error report in the background, but as long as other parts of the app still work, user is not 100% shut down (this would buy us some time to find and fix the bug).

I guess this `throw` could be useful during development. However, would `console.error` or `setAlert` be enough? I understand that console or setAlert might not sound alarming enough. However, this "break the whole app" approach seems rather radical and could have a bad influence at large (if it ever happened once on production).

iannbing (commented):
Nice work! You added a new feature with proper cleanup (especially replacing `formik` with `react-hook-form` 💯). There's a bug though when calling `createStep` after the file is created. Apart from that, nice job done!

services/orchest-webserver/client/src/pipeline-view/file-manager/CreateFileDialog.tsx
```diff
@@ -123,68 +121,58 @@ export const CreateFileDialog = ({
...
+                label="File path"
+                value={displayPath}
+                disabled
+                fullWidth
+                margin="normal"
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <InputLabel>
+                <Checkbox {...register("createStep")} checked={createStep} />
```
iannbing, line 171: User could create a file when there's no pipeline. Perhaps hide or disable this checkbox if there's no pipeline?

services/orchest-webserver/client/src/pipeline-view/contexts/PipelineEditorContext.tsx
```diff
@@ -32,7 +32,8 @@ export type PipelineEditorContextType = {
   eventVars: EventVars;
   dispatch: (value: EventVarsAction) => void;
   stepDomRefs: React.MutableRefObject<Record<string, HTMLDivElement | null>>;
-  pipelineCanvasRef: React.MutableRefObject<HTMLDivElement | null>;
+  pipelineCanvasRef: React.MutableRefObject<HTMLElement | null>;
+  pipelineViewportRef: React.RefObject<HTMLElement | null>;
```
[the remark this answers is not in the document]
mausworks: Nice catch, will fix! I'm so used to TypeScript screaming at me for type errors and not compiling, but our project doesn't seem to do that, am I missing something or does our project not just do that?

services/orchest-webserver/client/src/pipeline-view/file-manager/ActionBar.tsx
```diff
@@ -23,55 +23,47 @@ const FileManagerActionButton = styled(IconButton)(({ theme }) => ({
...
   rootFolder,
   setExpanded,
-}: {
-  rootFolder: FileManagementRoot;
-  uploadFiles: (files: File[] | FileList) => void;
-  setExpanded: (items: string[]) => void;
-}) {
-  const { projectUuid } = useCustomRoute();
+}: ActionBarProps) {
+  const [creating, setCreating] = React.useState<Creating | null>(null);
```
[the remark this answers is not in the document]
mausworks: Good idea, I'll revise the type.

services/orchest-webserver/client/src/pipeline-view/file-manager/CreateFileDialog.tsx
mausworks: I didn't know that this was a thing! Can we elaborate on this on monday?

services/orchest-webserver/client/src/pipeline-view/file-manager/useCreateFile.tsx
mausworks: I could require the `projectUuid` to be passed into the hook instead. Maybe that is to prefer?

services/orchest-webserver/client/src/pipeline-view/hooks/useCreateStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const useCreateStep = () => {
+  const { setAlert } = useAppContext();
+  const {
+    dispatch,
+    environments,
+    pipelineViewportRef,
+  } = usePipelineEditorContext();
+  const { pipelineCanvasState } = usePipelineCanvasContext();
+
+  return (fileName?: string) => {
```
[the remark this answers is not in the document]
mausworks: Indeed, good point. I thought I should memoize it, but then I thought YAGNI. But sounds like you've had performance issues with re-renders in the past, so I understand the concern, and I will memoize it. 👍

mausworks:
@ncspost the design (and screenshot) has been updated. 👍

yannickperrenet:
When a new step is created (on creation of a new file), then the wrong `File path` is used. It contains a leading slash `/` which shouldn't be there.

https://user-images.githubusercontent.com/26223174/178262149-c250f796-cda0-45bc-9c2d-c4e15c6dfd5c.mp4

mausworks:
OK @yannickperrenet, it's fixed now—together with the re-rendering bug. :)

services/orchest-webserver/client/src/.wdyr.ts
```diff
@@ -0,0 +1,7 @@
+import React from "react";
```
mausworks, line 1: I named this file `.wdyr.ts` since it needs to be the first import and we have the "order imports" plugin for prettier. :)

services/orchest-webserver/client/src/hooks/useCheckFileValidity.tsx
```diff
@@ -68,11 +66,11 @@ export const useCheckFileValidity = ({
   allowedExtensions,
   useProjectRoot = false,
 }: ValidateFileProps) => {
-  const isQueryArgsComplete =
+  const propsAreValid =
```
iannbing, line 69: Just out of curiosity, what's the reason this variable `propsAreValid` is not called `arePropsValid`? And another question: although `props` is plural in English, it could also be seen as one variable (i.e. a javascript object). If either case makes sense, I'd choose the singular one just make it "look" simpler. What's your rationale behind the naming?  I'm not proposing a change, just want to learn from different points of view .🙂

services/orchest-webserver/client/src/pipeline-view/file-manager/useCreateFile.tsx
```diff
@@ -0,0 +1,39 @@
...
+type FileCreator = (path: string) => Promise<string>;
+
+/**
+ * Returns a function which creates a file relative to the provided root.
+ * The returned path will be an absolute project path, starting with `project-dir:/` or `data:/`.
+ */
+export const useCreateFile = (root: string): FileCreator => {
+  const { projectUuid } = useCustomRoute();
+
+  const createFile = React.useMemo(
```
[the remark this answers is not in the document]
mausworks: Sure, I guess that makes sense. I guess "createFile" will at some point be considered a callback. 👍

services/orchest-webserver/client/src/hooks/useCheckFileValidity.tsx
mausworks: I choose `propsAreValid` instead of `arePropsValid`, simply because that's the way I would say it in english: _"if (the) props are valid, then …"_ I didn't think _too much_ about the rename.

Regarding whether you can say "isPropsValid", I guess you're right: props is just an object. I'll change it. :)

yannickperrenet:
#### Nitpick 1
(Not sure whether this was introduced by this PR. Probably not.)

After creating a file for the second time through the file-manager, you get two errors with exactly the same message:
> Error: File already exists.

This error dialog should only be shown once.

To reproduce:
1. Create file with name `x.ipynb` through icon in file-manager
2. Try to create file with name `x.ipynb` through icon in file-manager
3. Get the same error dialog twice.

#### Nitpick 2
image

How about hiding the "Create a new step ..." option altogether instead of blurring it out?

services/orchest-webserver/client/src/pipeline-view/hooks/useCreateStep.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { createStepAction } from "../action-helpers/eventVarsHelpers";
+import { usePipelineCanvasContext } from "../contexts/PipelineCanvasContext";
+import { usePipelineEditorContext } from "../contexts/PipelineEditorContext";
+import { STEP_HEIGHT, STEP_WIDTH } from "../PipelineStep";
+
+/**
+ * Removes leading slashes from file paths as the step API does not support them.
+ */
+const stepPath = (path?: string) =>
+  path?.startsWith("/") ? path.substring(1) : path;
```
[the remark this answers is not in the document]
mausworks: Good catch, I'll take a look. 👍

services/orchest-webserver/client/src/.wdyr.ts
iannbing: Smart setup! Heard of this library for so long and now it's time for me to try it out! 💯

services/orchest-webserver/client/src/pipeline-view/hooks/useCreateStep.tsx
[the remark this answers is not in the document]
mausworks: Fixed now. 👍

mausworks:
> This error dialog should only be shown once.

This is a regression, not sure why this is, taking a look. 👍 

> How about hiding the "Create a new step ..." option altogether instead of blurring it out?

I'm not sure, what's the rationale for it? I think it's good to show the available options, I also added a tooltip to explain _why_ it's disabled. I'll let @ncspost make the call here.

ncspost (approved):
Great work @mausworks - you're on 🔥

iannbing (approved):
Awesome work! All the edge cases are taken care of. 👍

PR #1099, opened 2022-07-11
Only use `image-puller` `initContainer` if `jupyter-server` is customized by user

nhaghighat:
## Description

This PR adds the image_puller manifest to jupyter-server and user-services if the image is not on global docker registries

## Checklist

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -149,6 +149,27 @@ def get_init_container_manifest(
     return init_container
 
 
+def add_image_puller_if_needed(
+    image_to_pull: str,
+    container_runtime: str,
+    container_runtime_image: str,
+    deployment_manifest: Dict[str, Any],
+):
+
+    global_domains = ["docker.io"]
+    domain, name = split_docker_domain(image_to_pull)
+
+    if domain not in global_domains:
+        image_puller_manifest = get_init_container_manifest(
+            f"{domain}/{name}",
+            container_runtime,
+            container_runtime_image,
+        )
+        deployment_manifest["spec"]["template"]["spec"]["initContainers"] = [
+            image_puller_manifest
+        ]
+
+
 # splitDockerDomain splits a repository name to domain and remotename
 # string. If no valid domain is found, the default domain is used.
 # Repository name needs to be already validated before.
--- a/lib/python/orchest-internals/_orchest/internals/utils.py
+++ b/lib/python/orchest-internals/_orchest/internals/utils.py
@@ -123,11 +123,11 @@ def get_step_and_kernel_volumes_and_volume_mounts(
 def get_init_container_manifest(
     image_to_pull: str,
     container_runtime: str,
-    container_runtime_image: str,
+    image_puller_image: str,
 ) -> Dict[str, Any]:
     init_container = {
         "name": "image-puller",
-        "image": container_runtime_image,
+        "image": image_puller_image,
         "env": [
             {
                 "name": "IMAGE_TO_PULL",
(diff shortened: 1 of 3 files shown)


lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -149,6 +149,27 @@ def get_init_container_manifest(
...
 
+def add_image_puller_if_needed(
+    image_to_pull: str,
+    container_runtime: str,
+    container_runtime_image: str,
+    deployment_manifest: Dict[str, Any],
+):
+
+    global_domains = ["docker.io"]
+    domain, name = split_docker_domain(image_to_pull)
```
fruttasecca, line 160: About the use of a `global_domains` variable, would this logic lead to the init container being used if the image comes from non-docker but public registries?

fruttasecca:
I really like this injection approach :+1: , question, is there a reason for not replacing calls to `get_init_container_manifest` with `add_image_puller_if_needed` to have a consistent approach and abstraction level?

lib/python/orchest-internals/_orchest/internals/utils.py
nhaghighat: We can also do it the other way around, if the domain is our local `docker-registry`, inject the init-container

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -149,6 +149,27 @@ def get_init_container_manifest(
...
+
+    if domain not in global_domains:
+        image_puller_manifest = get_init_container_manifest(
+            f"{domain}/{name}",
+            container_runtime,
+            container_runtime_image,
+        )
+        deployment_manifest["spec"]["template"]["spec"]["initContainers"] = [
+            image_puller_manifest
+        ]
```
fruttasecca, line 170: Another very minor thing, it's probably safer to check if the `initContainers` list exists already so that we extend instead of replace in that case.

nhaghighat:
> question, is there a reason for not replacing calls to get_init_container_manifest with add_image_puller_if_needed to have a consistent approach and abstraction level

Since pipeline steps and kernels, always use images from our registry, I thought there is no need to check if init-container is needed or not.
also pipeline manifest is of type `workflow`, so the path for injection of init-container is different than the deployment.
(in `deployment` it is in `.spec.template.spec.initContainers` and in `workflow` it is in  '.spec.template.initContainers')

and kernel manifest is of type `pod` which means a different path for injection.

lib/python/orchest-internals/_orchest/internals/utils.py
nhaghighat: it is indeed safer

fruttasecca:
>Since pipeline steps and kernels, always use images from our registry, I thought there is no need to check if init-container is needed or not.

Sounds good :+1:

lib/python/orchest-internals/_orchest/internals/utils.py
yannickperrenet: > Would it be easier (safer, maybe?) to invert the logic

+1 on this :)

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -149,6 +149,27 @@ def get_init_container_manifest(
...
+    container_runtime: str,
+    container_runtime_image: str,
+    deployment_manifest: Dict[str, Any],
+) -> None:
+
+    global_domains = ["docker.io"]
+    domain, name = split_docker_domain(image_to_pull)
+
+    if domain not in global_domains:
+        image_puller_manifest = get_init_container_manifest(
```
yannickperrenet, line 163: Seeing the name of `get_init_container_manifest` this time I feel like it would be good to be more explicit (about it being the initContainer for the pre-pulling of images), e.g. `get_image_puller_init_container_manifest`. Or would you say that is too verbose?

yannickperrenet:
Overall the logic looks good to me! :+1:

yannickperrenet (approved):
Haven't tested it but looks good to me :100:

PR #1102, opened 2022-07-11
Fix: a rebased branch cannot be re-deployed.

iannbing:
## Description

If the remote branch is rebased, `git` will fail to pull. Deleting the local branch and re-fetch-and-pull from remote
would solve the problem.

### DIFF
--- a/scripts/redeploy_branch.sh
+++ b/scripts/redeploy_branch.sh
@@ -31,12 +31,19 @@ set -e
 
 echo "Checking out branch ${1}..."
 
-if git show-ref --quiet refs/heads/$1 ; then
+if [ "${1}" != "dev" ] && git show-ref --quiet refs/heads/$1 ; then
     git checkout dev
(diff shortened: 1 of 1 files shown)


scripts/redeploy_branch.sh
```diff
@@ -30,6 +30,8 @@ fi
 set -e
 
 echo "Checking out branch ${1}..."
+git checkout dev
+git branch -D $1
```
fruttasecca, line 34: This will produce an error if the branch doesn't exist locally already.

scripts/redeploy_branch.sh
```diff
@@ -30,6 +30,8 @@ fi
...
 git fetch --all
 git checkout $1
 git pull origin $1
```
fruttasecca, line 37: I guess this is not needed anymore

fruttasecca (approved):
haven't manually tested it but lgtm

PR #1104, opened 2022-07-12
Release `v2022.07.2`

yannickperrenet:
## Description

New Orchest release that can run on Kubernetes clusters that use `containerd` as their container runtime, among other improvements and fixes :raised_hands: 

## Checklist


services/orchest-api/app/app/core/image_utils.py
```diff
@@ -269,6 +267,25 @@ def _get_buildah_image_build_workflow_manifest(
...
+                "subPath": ".orchest/user-configurations/jupyterlab/user-settings",
+            }
+        )
+        container["args"][0] = container["args"][0].replace(
+            "buildah build",
+            (
+                "buildah build -v "
+                "/jupyterlab-user-settings:/root/.jupyter/lab/user-settings "
+            ),
+        )
```
yannickperrenet, line 287: Looking at this now I feel like we would be running into an issue similar to the `/pip-cache` issue (https://github.com/orchest/orchest/pull/1093).

The `user-settings` subpath is not created on `orchest-api` initialization. Thus the storage provider would create the directory. https://github.com/orchest/orchest/blob/f9e4876254131266823663022ee861e7ab9daacb/services/orchest-api/app/app/__init__.py#L181-L189

I guess we could add the path to the linked code above (and use the new config var in the `subPath` as well). @fruttasecca @nhaghighat

yannickperrenet:
Found a bug. Steps to reproduce:

1. Create an environment called `X`
2. Configure a service (easiest is to just take the vscode template)
3. Set the image to environment `X`
4. Restart the session

See that the session never starts.

Some of the logs of the `orchest-api` that are of interest:
```
INFO:app.utils:_core.py - [12/Jul/2022 11:28:15] - Adding session sidecar to log user services.
ERROR:app:namespace_sessions.py - [12/Jul/2022 11:28:15] - 'initContainers' 
```

services/orchest-api/app/app/core/image_utils.py
nhaghighat: @yannickperrenet : is this `.orchest/user-configurations/jupyterlab/user-settings` created by us explicitly, or is it created by mounting?

yannickperrenet: I think it is created first here:

https://github.com/orchest/orchest/blob/919b27e6ed5845115227aaa464ce58f3455dd1da/services/orchest-api/app/app/core/sessions/_manifests.py#L95-L99

Thus through mounting.

nhaghighat: it is indeed created by mounting. I think it's better to create a ticket about it to check if we can treat it like `pip-cache` and skip changing this for this release.

yannickperrenet: Fixed in https://github.com/orchest/orchest/pull/1104/commits/4e4c0daf9123a866b906dbef5f422e721fe3991e

nhaghighat (approved):

Issue #1106, opened 2022-07-12
Webserver failure (file not found) when multiple .orchest in subdirectory


jerdna-regeiz:
**Describe the bug**
If there are multiple `.orchest` files inside a subdirectory, the webserver fails at various actions with an `FileNotFoundError`.

Cause is that `find_pipelines_in_dir` from `utilities.py` does try to apply a conversion to a relative path multiple times (for each file found). This breaks the path and further code then does not find the file (as the path is wrong).

I'll provide a fix in a minute.

**Expected behavior**
In case of multiple `.orchest` files, it should just find multiple pipelines without dying.

**To Reproduce**
1. Create a project with a subdirectoy
2. Create multiple pilelines (`.orchest`) in the subdirectory.
3. Reload the page
4. Cry, as no pipeline can be loaded anymore and your project is useless.

PR #1107, opened 2022-07-12
Fix: Allow multiple pipeline files (`.orchest` extension) to be present in a subdirectory

jerdna-regeiz:
## Description

utils tried to change the root directory of found .orchest files to a
relative path for each found file. This breaks if there are more than
one .orchest files in a subdirectoy.

When only done once per directoy it is fixed. So just moved the code a few lines up.

Fixes: #1106

## Checklist

### DIFF
--- a/services/orchest-webserver/app/app/utils.py
+++ b/services/orchest-webserver/app/app/utils.py
@@ -464,12 +464,13 @@ def find_pipelines_in_dir(path, relative_to=None):
 
             dirs[:] = [d for d in dirs if d not in ignore_dirs]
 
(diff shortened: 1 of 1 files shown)


fruttasecca:
Hi @jerdna-regeiz, thanks for reporting the issue and opening the PR, nice catch :). I'll take a look at it and come back in a few minutes, we would like to have this as part of the release we are about to do.

yannickperrenet (approved):
LGTM!

Otherwise we would indeed change `root` multiple times (as per `root = root[1:]`).

PR #1111, opened 2022-07-13
Fix project import race condition

fruttasecca:
## Description

Fixes a race condition in the flow to import a git project caused by the `git_clone_project` background task being considered as succeeded once the cloned project has been moved into the `/projects` directory. Example:
- project is moved in `/projects`, the background ask is succeeded
- an observer of said task assumes that it can find the project by querying `/async/projects`
- another request to `/async/projects` leads to a concurrent discovery of the project
- if this second request to `async/projects` is ahead of the first one it will cause the project to enter a state of `INITIALIZING` while it reads its content from the file system
- the projects returned to the first `/async/projects` request do not inclued said project since only `READY` projects are returned

The issue is fixed by having the background task query the `/async/projects` endpoint to trigger a discovery of the project. Only once the project has been discovered or has been renamed or delated by a third party the task is considered as succeded.

## Checklist


services/orchest-webserver/app/scripts/background_tasks.py
```diff
@@ -37,41 +39,56 @@ def git_clone_project(args):
...
+        # exist anymore at the desired path.). The project not being
+        # part of the projects returned by /async/projects is a rare
+        # occurrence that can happen if a concurrent request is leading
+        # to the discovery of the same project, which puts it in a
+        # INITIALIZING status.
+        for _ in range(5):
+            resp = requests.get("http://localhost/async/projects")
+            if resp.status_code != 200:
+                time.sleep(1)
+                continue
```
fruttasecca, line 76: Not so sure it's needed tbh :thinking:

services/orchest-webserver/app/scripts/background_tasks.py
```diff
@@ -37,41 +39,56 @@ def git_clone_project(args):
...
+
+        # The task is considered done once the project has been
+        # discovered or if the project has been renamed/deleted (doesn't
+        # exist anymore at the desired path.). The project not being
+        # part of the projects returned by /async/projects is a rare
+        # occurrence that can happen if a concurrent request is leading
+        # to the discovery of the same project, which puts it in a
+        # INITIALIZING status.
+        for _ in range(5):
+            resp = requests.get("http://localhost/async/projects")
```
yannickperrenet, line 73: Isn't it possible to query the DB instead of the calling the webserver API from within the webserver itself?

yannickperrenet: Excellent points. I think it warrants a comment to explain why we call the webserver from the webserver itself.

services/orchest-webserver/app/scripts/background_tasks.py
yannickperrenet: I like the "defensiveness" of it. So I would be in favor of leaving it in.

services/orchest-webserver/app/scripts/background_tasks.py
fruttasecca: See https://github.com/orchest/orchest/pull/1111/commits/172a45b49ffd8225f50530365abc28e405139275

yannickperrenet (approved):

PR #1112, opened 2022-07-13
Refactor SessionContext and AppContext.

iannbing:
## Description

The implementation of SessionContext and `requestBuild` in AppContext has become obsolete and starts to become the bottleneck for building new features. This PR cleans up the legacy code for better performance and readability.

The changes are as follows:
- Convert the implementation of `requestBuild` from callbacks-based to promise-based, so no need to put callbacks in state.buildRequest.
- Move the code for `requestBuild` from `AppContext` to `ProjectsContext` because requestBuild only makes sense within a project.
- Convert state.sessions from an array to an object for better performance with less code.
- Replace `toggleSession` with `startSession` and `stopSession`.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
+++ b/services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
@@ -1,9 +1,32 @@
 import type {
+  EnvironmentValidationData,
   PipelineMetaData,
   Project,
   ReducerActionWithCallback,
 } from "@/types";
+import { checkGate } from "@/utils/webserver-utils";
 import React from "react";
+
+export enum BUILD_IMAGE_SOLUTION_VIEW {
+  PIPELINE = "Pipeline",
+  JOBS = "Jobs",
+  JOB = "Job",
+  JUPYTER_LAB = "JupyterLab",
+}
+
+export type BuildRequest = {
+  projectUuid: string;
+  environmentValidationData: EnvironmentValidationData;
+  requestedFromView: string;
+  onComplete: () => void;
+  onCancel: () => void;
+};
+
+export type RequestBuildDispatcher = (
+  projectUuid: string,
+  environmentValidationData: EnvironmentValidationData,
+  requestedFromView: BUILD_IMAGE_SOLUTION_VIEW
+) => Promise<boolean>;
 
 const ProjectsContext = React.createContext<IProjectsContext>(
   {} as IProjectsContext
@@ -19,6 +42,7 @@ export type IProjectsContextState = {
   hasLoadedProjects: boolean;
   hasLoadedPipelinesInPipelineEditor: boolean;
   newPipelineUuid: string | undefined;
+  buildRequest?: BuildRequest;
 };
 
 export const useProjectsContext = () => React.useContext(ProjectsContext);
@@ -55,6 +79,10 @@ type Action =
   | {
       type: "UPDATE_PIPELINE";
       payload: { uuid: string } & Partial<PipelineMetaData>;
+    }
+  | {
+      type: "SET_BUILD_REQUEST";
+      payload: BuildRequest | undefined;
     };
 
 export type ProjectsContextAction = ReducerActionWithCallback<
@@ -65,6 +93,10 @@ export type ProjectsContextAction = ReducerActionWithCallback<
 export interface IProjectsContext {
   state: IProjectsContextState;
   dispatch: (value: ProjectsContextAction) => void;
+  ensureEnvironmentsAreBuilt: (
+    requestedFromView: BUILD_IMAGE_SOLUTION_VIEW
+  ) => Promise<boolean>;
+  requestBuild: RequestBuildDispatcher;
 }
 
 const reducer = (
@@ -164,6 +196,9 @@ const reducer = (
     }
     case "SET_PROJECTS":
       return { ...state, projects: action.payload, hasLoadedProjects: true };
+    case "SET_BUILD_REQUEST": {
+      return { ...state, buildRequest: action.payload };
+    }
     default: {
       console.log(action);
       return state;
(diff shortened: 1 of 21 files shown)


services/orchest-webserver/client/src/contexts/SessionsContext.tsx
```diff
@@ -1,98 +1,61 @@
...
-/* Fetchers
-  =========================================== */
+const launchSession = ({
+  projectUuid,
+  pipelineUuid,
+}: {
+  projectUuid: string;
+  pipelineUuid: string;
+}) => {
+  // NOTE: the trailing slash is required, otherwise the request will be picked up by other route.
```
mausworks, line 24: Would it be possible to fix this in the routing instead?

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
```diff
@@ -151,177 +112,132 @@ const reducer = (
...
-        } catch (error) {
-          setAlert("Error", "Failed to stop session.");
-          console.error(error);
-        }
-        return;
-      }
+  // NOTE: Make sure that your view component is added to useSessionsPoller's list.
+  const stopSession = React.useCallback(
+    async (pipelineUuid: string) => {
+      if (!projectUuid) return;
```
mausworks, line 219: I see this anti-pattern everywhere (and I do this myself as well), I think it makes the code appear defensive and unDRY.

Surely there should be some way for us to ensure that a `projectUuid` is available under some circumstances? For instance: if a component is _only_ available when there is a `project_uuid` in the URL (it's part of the route, so you can't access it otherwise), then I think it should be safe for us to make this assumption.

In a previous app I helped developed we had views that required the user to be authenticated and some views that didn't, we had one hook called `useUser` that could return `null`, and one called `useRequireUser` that returned the current user, or threw and error.

In this particular case, I think we can have a hook called `useActiveProject` that assumes that the current `project_uuid` is available in the route (for example)—although, that's just an example. I'm curious to hear your take on this.

Also, this is not a request for changes, just a general question/concern. :)

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
```diff
@@ -184,8 +219,81 @@ const initialState: IProjectsContextState = {
...
+          type: "SET_PIPELINE_IS_READONLY",
+          payload: false,
+        });
+      }
+      return hasBuilt;
+    },
+    [requestBuild, state.projectUuid]
+  );
+
+  const ensureEnvironmentsAreBuilt = React.useCallback(
```
mausworks, line 280: Why not simply call this function `buildEnvironments`? It makes sense that it is no-op if it's already built.

mausworks (commented):
Big improvement. 👍 Left some comments, but I'm not requesting changes, just some general questions/concerns and things we should investigate in the future.

So this LGTM! I haven't pulled the code and tested it, let me know if you want help with that!

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
iannbing: Good suggestion! I was planning to do something similar once we have a 404 page (see [this comment](https://github.com/orchest/orchest/blob/36f2085f9da5465731db680a2eb73d61ccc0a684/services/orchest-webserver/client/src/routingConfig.tsx#L177)). As I mentioned earlier in other PR, I think all error should be caught in order to prevent a blank page. In this case, an error in the URL (i.e. `project_uuid` is missing or something similar) we should catch it with a 404 page 😄.

mausworks: I agree that we should definitely take precautions to ensure that the user never gets a white page. I think that displaying a 404-page under these circumstances is a great idea. 👍

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
```diff
@@ -1,98 +1,61 @@
...
 
-const stopSession = ({ pipelineUuid, projectUuid }: IOrchestSessionUuid) =>
-  fetcher(`${ENDPOINT}${projectUuid}/${pipelineUuid}`, {
+const killSession = ({
+  projectUuid,
+  pipelineUuid,
+}: {
+  projectUuid: string;
+  pipelineUuid: string;
+}) => {
```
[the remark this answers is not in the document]
iannbing: Hmm, interesting. I think `pipelineUuid` itself is already an identifier of a pipeline. If I have to come up with a proper name, I probably would say `SessionIdentifier`. In other places I use `SessionKey` (i.e. `${projectUuid}|${pipelineUuid}`), but I like your proposal on this one. Will fix. 👍

Regarding your remark about inline/anonymous types, I agree that named types make beautiful code 😄, but I would try to be pragmatic about it. 

I see some benefits of inline typing in my development flow:
- reduce phantom productivity cost: if a type is never reused, why do we spend time on naming it? 
- avoiding naming collision: the naming convention of  React components and Types are the same. `PipelineStep` could be either one. The java/c# style typing annotation might help, e.g. `ISomeInterface` or `TSomeType`, but it reduces readability imo. In order to name a never-used type properly, we might have to double / triple the phantom cost.

And I'm curious, would you always named-typing a function param if it's an object? I think most of the time, it's more of a practical decision: an object is less prone to human mistakes (i.e. it's order-insensitive, you could use `?` freely, etc.). I'm not sure if it's realistic to enforce this rule fully.

services/orchest-webserver/client/src/contexts/ProjectsContext.tsx
iannbing: I have to admit that it's hard to name this function 😅. `buildEnvironments` implies a direct operation. In this case, it actually asks for user's approval for executing this operation. I think this could be misleading when we troubleshoot. That is, if the returned boolean `hasBuilt` is false, we might mistake it as "our BE failed to build the images" instead of "user doesn't want to build the images right now".

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
iannbing: Good idea 💯 . Pushed the fix.

services/orchest-webserver/client/src/contexts/SessionsContext.tsx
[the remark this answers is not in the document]
mausworks: I tend to prefer when most types are named, but I try to be pragmatic too, but maybe in a different way? 🤔 

I guess my decision making for when I want a type to have a name goes something like this:
- Is it a domain type? Then yes: in the case of Orchest, those are things like `Step` `OrchestSession`. These are important concept that require elaboration.
- Is this type used more than once? Then yes, I'll try to name it in most cases, [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) reigns supreme.
- Does it have more than three properties? Then yes, I'll usually name it `FooProps` (for react components) `FooParams`, `FooOptions` or similar—not a huge investment of time goes into naming these honestly. 😅 

Regarding naming collisions: I think that the `IFoo` convention makes sense in Java/C# since you often have the `IFoo` interface and the `Foo` class as a default implementation of it. In TypeScript on the other hand you usually don't have that problem, `Foo` is usually reserved for either a type _or_ a class: it's rarely both in my experience. So I think using [TypeScripts internal coding guidelines](https://github.com/microsoft/TypeScript/wiki/Coding-guidelines) and not using any prefix makes sense.

If you're in doubt you can usually throw "params", "options" or—heck—even "data" at the end of a type name, and it often makes sense.

I see types as first-class citizens in the code base: they are not only there to provide guards and convenience, they are primarily there to aid the reader in understanding intent and to encapsulate boundaries of the problem space they are working within. Anonymous types encapsulate meaning without labelling it or explaining "it's purpose"—this is left as an exercise for the reader—which I think can cause cognitive dissonance.

But while the above rant may make it seem like I might have some strong opinions about this: I really don't, this is just how I reason about it myself. 👍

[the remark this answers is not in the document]
mausworks: This sounds good to me. I guess we also agree on that `StepProps` are the props of the `Step` component. 👍

PR #1117, opened 2022-07-15
Improve `orchest-cli` loading spinner robustness

yannickperrenet:
## Description

The spinner displays the status of the Orchest Cluster at a specific moment in time. In order to correctly stop "spinning" the moment an end state is reached, there are some edge cases we need to handle, which are mainly introduced by the (variable) duration between cluster and the host invoking the CLI command. 

Summary of changes:
-  Emphasize that the spinner will be shown for at least 10s to allow for CLI commands where the desired end status is equal to the current status, e.g. for `orchest restart`.
- The first message by the spinner will no longer be a guess of the cluster's state and instead will be "Fetching Orchest Cluster status"
- The spinner for `orchest patch` is invoked differently so that it doesn't keep spinning when `orchest patch` is invoked whilst the cluster is not in a valid end state (i.e. `RUNNING` or `STOPPED`).
- Introduces `orchest status --wait=...`, e.g. `orchest status --wait=Stopped`, to wait for the cluster to reach the passed status. 

## Checklist


orchest-cli/orchestcli/cli.py
```diff
@@ -383,18 +383,31 @@ def version(json_flag: bool, latest_flag: bool, **common_options) -> None:
...
+) -> None:
     """Get Orchest Cluster status.
 
-    If invoked with `--json`, then failure to get Orchest Cluster status
-    will return an empty JSON Object, i.e. `{}`.
+    \b
+    If invoked with `--json`, then:
+        - failure to get Orchest Cluster status will return an empty
+          JSON Object, i.e. `{}`.
+        - `--wait` doesn't have any effect.
```
fruttasecca, line 403: Not sure `--json` and waiting should be mutually exclusive, feels like the fact that the spinner implements the waiting logic and that it can't be "silenced" is bubbling up the interface of the CLI

orchest-cli/orchestcli/cmds.py
```diff
@@ -907,9 +907,24 @@ def disable_telemetry() -> None:
...
+        # In case of STOPPED the spinner will run for 10s anyways.
+        _display_spinner(None, curr_status)
+    else:
+        if curr_status is None:
+            curr_status = ClusterStatus.UNKNOWN
+        echo(f"Orchest Cluster status is: {curr_status.value}")
+        echo(
+            "You might want to try running:"
+            f"\n\torchest status --wait={ClusterStatus.RUNNING.value}"
+        )
```
fruttasecca, line 927: Do you think it makes sense to have a summary of the commit message of https://github.com/orchest/orchest/pull/1117/commits/1576280cf3deac5772ca8e25bd099b92ea1b8a75 here as a comment?

fruttasecca: Not related to the PR but perhaps a `PATCHING` status could have been useful to avoid the `UPDATE` status being overloaded

fruttasecca:
What do you think of the fact that we have some commands with a `--watch` flag and `orchest status` will have a `--wait` flag, which sound similar? I think we are in a local optimum given that we can't change the `--watch` flags without breaking changes  and `orchest status --watch` doesn't make much sense when it comes to wording. We could add a `--wait` to the commands using `--watch` and deprecate `--watch` but imo not needed atm and I don't want to increase the scope of the PR for no reason

fruttasecca (approved):
lgtm, I've left some minor considerations. Not super stoked about https://github.com/orchest/orchest/pull/1117#discussion_r922171988 but I don't want the comment to increase the scope of the PR

orchest-cli/orchestcli/cmds.py
```diff
@@ -1448,6 +1448,11 @@ def echo(*args, **kwargs):
                 # command has finished.
                 if curr_status != end_status:
                     thread.wait()  # type: ignore
+
+        if prev_status == curr_status and curr_status == end_status:
+            echo("\r", nl=False)
```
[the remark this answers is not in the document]
yannickperrenet: From https://github.com/orchest/orchest/pull/1117/commits/7540a55285fa379bc7150ce826b4dee224820743:

```python
# In the case where the spinner is running just because of the
# minimal time limit, the loading spinner would be the last
# printed line. Thus we redraw the line to make sure the last
# status is indicated as finished/reached.
```

orchest-cli/orchestcli/cmds.py
yannickperrenet: Would indeed make sense to explain the reasoning better: https://github.com/orchest/orchest/pull/1117/commits/7540a55285fa379bc7150ce826b4dee224820743.

> perhaps a PATCHING status could have been useful to avoid the UPDATE status being overloaded

Would indeed be an improvement. Although it is not that important imo (and should happen at the Controller level).

yannickperrenet:
> What do you think of the fact that we have some commands with a --watch flag and orchest status will have a --wait flag, which sounds similar?

Good point. I see it more as:
* `--watch` : Watch cluster status. Which happens to be until the invoked command, e.g. `orchest update`, has succeeded/finished given that it invoked in the context of that command.
*  `--wait` : Wait until the given status is reached, a la `kubectl wait --for=condition...`. Thought it would be good to have so one can programmatically wait until the cluster is in a specific state.

In addition `--watch` is a boolean flag and `--wait` actually has to be passed a status. In an effort to make the difference clearer, I've improved the help message of `--watch` (https://github.com/orchest/orchest/pull/1117/commits/aa82bbd53817f3118c3eb921a946fc39ebf4d3b7).

I do agree it is only a slight difference, but to me it felt like introducing another verb (`--wait`) was better than overloading `--watch` because `--watch` is only used as a boolean flag.

orchest-cli/orchestcli/cli.py
yannickperrenet: Implemented the above in https://github.com/orchest/orchest/pull/1117/commits/331cebe06bc11969c2faacf7973cb23be0a58f7e.

If you don't like it, then we can just revert.

fruttasecca:
>Good point. I see it more as...
>I do agree it is only a slight difference, but to me it felt like introducing another verb (--wait) was better than overloading --watch because --watch is only used as a boolean flag.

I don't have much to add aside from the fact that I agree :)

orchest-cli/orchestcli/cli.py
fruttasecca: >I don't think it makes sense to keep redrawing [1] STDOUT in a JSON structure

Agree

>In addition, not outputting anything until STOPPED is reached also doesn't like an option because why would you want to have it in --json if you already now what the output will be (i.e. {"status": "Stopped"}).

I see your point, but I'd say that setup or CICD scripts can end up doing unexpected things to reach some goal. But the major point here would be consistency at the CLI level, imo. Also, if this CLI command was to expand in scope, having `--json` and --`wait` be mutually exclusive could have been annoying, e.g. (just a random example) what if we are to allow waiting but with a timeout? Here the potentially unexpected end status would surely be interesting

>The alternative could be (when invoking orchest status --json --wait=Stopped) to wait until Stopped was reached and then output a structure like:

Imo the old json without the addition would have been fine, but this is an improvement, so lgtm.

PR #1124, opened 2022-07-18
Release `v2022.07.3`

yannickperrenet:
## Description

New Orchest release: `v2022.07.3` :smile_cat: 

## Checklist

  - [x] Do sessions work as expected
  - [x] New feature: Renaming in the file-manager
  - [x] New feature: Creating a step on file creation
  - [x] CLI edge cases


yannickperrenet:
_During manual testing I ran into quite a few bugs/improvements/nitpicks. I am writing them all down here and we can choose which ones should get fixed before we release. Please refer to them explicitly in this thread (e.g. using the resp. number)._

---
## 1. Environments can't be deleted

To reproduce:
1. Create an environment
2. Try to delete the environment through the table in the `/environments` page.

@mausworks Could you fix?

## 2. Pipeline in read-only mode without hint to explain "why?"

When the pipeline is put in read-only mode due to environments being build, we show a dialog explaining why the pipeline is in read-only mode. However, when the pipeline is in read-only mode because JupyterLab is building then we don't show this dialog (which we should show).

https://user-images.githubusercontent.com/26223174/179678348-293d1607-3c97-47b6-938f-a67c1bc9ac7e.mp4

@iannbing @ricklamers Any specific reason this isn't happening yet? Otherwise I think it would be great if it can be added.

## 3. JupyterLab build not started when unsaved changes

See video. All dialogs are prompted, however, the build is actually never started.

https://user-images.githubusercontent.com/26223174/179678712-0f51c6e3-f5cb-43f2-850e-77412104a8f8.mp4

@iannbing How easy would this be to fix?

## 4. Select root folder in create-file dialog

When you have a specific folder selected in the file-manager, then pressing "create file" will try to create a file in the selected root folder. However, as a user, I might actually want to create the file elsewhere. To achieve that, I would have to cancel the file creation and select the correct folder to create the file in (why can't I just change it/type it in the dialog?). This is quite cumbersome in my opinion. 

@ncspost Is this picked up by the new design?

https://user-images.githubusercontent.com/26223174/179678856-c42091c7-738e-4cea-bb6f-271572730601.mp4

## 5. Step title doesn't get auto-focused

When creating a new step, the title is auto-focused. However, when creating a new step automatically when a new file is created, then we should also auto-focus the title (which is currently not the case) so that the user can enter it. This way the "new step" button and automatic step creation (on file creation) would work the same way. 

https://user-images.githubusercontent.com/26223174/179679511-e6b31a9a-f366-44bd-9a2f-9c1a0db8cc87.mp4

@mausworks Can you fix? Probably slipped through in https://github.com/orchest/orchest/pull/1097.

## 6. "Infinite" loop when canceling pipeline rename

When trying to rename the pipeline through the file-manager, then as a user I might want to cancel this rename (due to the required session restart). However, when canceling I am left in an infinite loop (see video).

https://user-images.githubusercontent.com/26223174/179679760-742856a4-2cec-4644-ad26-434f57b2b84c.mp4

@mausworks Can you fix? Probably slipped through in https://github.com/orchest/orchest/pull/1115.

mausworks:
> 6. "Infinite" loop when canceling pipeline rename

This is _sort of_ intended behavior. To cancel the rename, you hit <KBD>ESC</KBD>. Clicking outside of the rename field always saves the rename, thus triggering the dialog.

Do we want to always cancel the rename if the user hits cancel on _any_ dialog, or just this one?

iannbing:
> 3. JupyterLab build not started when unsaved changes
Screenshot 2022-07-19 at 12 07 00

It prompts a warning if user attempts to build JupyterLab while having active sessions. And if you click "Confirm", it stops sessions but stays in the same view with "BUILD" button disabled. I think you probably pressed ENTER to dismiss the first "unsaved warning" and also dismissed this warning at the same time (because CONFIRM in this dialog is auto-focused).

But, indeed, it's very strange that user stays in the same view without seeing any updates about stopping all sessions. Would it be better to navigate back to PipelineEditor, so user could see if all sessions are stoped?

yannickperrenet:
> I think you probably pressed ENTER to dismiss the first "unsaved warning" and also dismissed this warning at the same time (because CONFIRM in this dialog is auto-focused).

I actually used the mouse for both. I just reproduced the issue again and made sure not to press my keyboard.

> It prompts a warning if user attempts to build JupyterLab while having active sessions. And if you click "Confirm", it stops sessions but stays in the same view with "BUILD" button disabled.

The sessions are indeed stopped, but I find it strange that the `BUILD` button is disabled for two reasons:
* It never "snaps back"
* Because it is in a disabled state it appears as if the build was invoked (because when pressing the black build button it would become this "disabled" button), however, the build isn't actually started so as a user I would wait for a very long time (without anything ever happening). 

> Would it be better to navigate back to PipelineEditor, so user could see if all sessions are stoped?

I don't think it would be better to navigate back to the PipelineEditor, for two reasons:
* I am afraid we would (in special occasions) retrigger a session start
* From a UX perspective I don't like us suddenly redirecting to another page in this context

> But, indeed, it's very strange that user stays in the same view without seeing any updates about stopping all sessions.

**Why can't we just start the JupyterLab build the moment the sessions are stopped?** I am pretty sure this is the old behavior that we had.

iannbing:
> Why can't we just start the JupyterLab build the moment the sessions are stopped? I am pretty sure this is the old behavior that we had.

You're right. Fix this issue with showing a Snackbar while stopping all sessions. see  #1130.

yannickperrenet:
> Fix this issue with showing a Snackbar while stopping all sessions

Absolutely awesome fix! Thanks for picking that up so quickly :heart:

Issue #1092, opened 2022-07-07
On `orchest update` also update third-parties we ship (such as Argo Workflows)


yannickperrenet:
**Describe the problem this improvement solves**
Changes to the value files of a thirdparty, e.g. [orchest-values.yaml of argo-workflows](https://github.com/orchest/orchest/blob/5b74a78f9d01ca9e41694cb8d76ed1b6abb19854/services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml), are not picked up on `orchest update`. However, we of course want these changes to be picked up as well given that they interact with the Orchest application (and thus the deployments of the thirdparties needs to match the version of Orchest that is running).

(This issue was mentioned in https://github.com/orchest/orchest/pull/1048#issuecomment-1176079684 first).

**Describe the solution you'd like**
To make sure the `orchest-controller` also updates third-parties we can add the `orchest-values.yaml` of the third-parties to the respective CR Object (as `Managed Fields`). This way on `orchest update` the `orchest-controller` can change (or add) values in the CR Object to make sure the third-parties are updated as well.

By making them `Managed Fields` we also communicate to the user that they shouldn't be updating the values manually, but doesn't strictly prohibit them from doing so.

nhaghighat:
In #1126 it is implemented a little bit differently than this issue. The CR object is not updated by the default values in `orchest-values.yaml` files, once the user overwrites the default values or even the controller updated with different default values or new helm templates, the controller checks the deployed manifests of the `third-party` addon with the new manifests if the manifests are different, the controller deploys the new manifests.

PR #1126, opened 2022-07-19
Update third-parties on Orchest update

nhaghighat:
## Description

In this PR, the controller checks the deployed manifests of third parties with newly generated manifests and if they are not equal, it updates the helm releases. in this way even if we upgrade the helm templates of third parties, the controller still can update them.

Fixes: #1092 

## Checklist

### DIFF
--- a/services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
+++ b/services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
@@ -41,7 +41,7 @@ storage: filesystem
 
 tlsSecretName: registry-tls-secret
 secrets:
-  haSharedSecret: ""
+  haSharedSecret: "haSharedSecret"
   htpasswd: ""
 # Secrets for Azure
 #   azure:
--- a/services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
+++ b/services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
@@ -41,6 +41,10 @@ storage: filesystem
 
 tlsSecretName: registry-tls-secret
 secrets:
+  # If this value is not specified, the docker-registry helm chart (./helm/templates/secret.yaml) 
+  # randomly generates a new one, so each time the controller compares the deployed manifests with 
+  # the new manifest (to decide whether it needs an update or not), it thinks the manifest is 
+  # changed and re-deploy docker-registry 
   haSharedSecret: "haSharedSecret"
   htpasswd: ""
 # Secrets for Azure
--- a/services/orchest-controller/pkg/addons/addons.go
+++ b/services/orchest-controller/pkg/addons/addons.go
@@ -35,8 +35,8 @@ func NewDefaultAddonsConfig() AddonsConfig {
 }
 
 type Addon interface {
-	// Installs addon if the config is changed
-	Enable(ctx context.Context, namespace string, config *orchestv1alpha1.ApplicationConfig) error
+	// Enable the addon. preInstallHooks should be called, before enabling the addon
+	Enable(ctx context.Context, preInstallHooks []func() error, namespace string, config *orchestv1alpha1.ApplicationConfig) error
 
 	// Uninstall the addon
 	Uninstall(ctx context.Context, namespace string) error
@@ -104,7 +104,7 @@ func (m *AddonManager) EnableAddon(ctx context.Context, addonName, namespace str
 	go func(ctx context.Context, cancel context.CancelFunc) {
 		if addon, ok := m.addons[addonName]; ok {
 			// addon is registered with the addon manager
-			err := addon.Enable(ctx, namespace, nil)
+			err := addon.Enable(ctx, nil, namespace, nil)
 			if err != nil {
 				klog.Errorf("failed to deploy addon: %s, err: %s", addonName, err)
 				cancel()
--- a/services/orchest-controller/pkg/addons/helm_deployer.go
+++ b/services/orchest-controller/pkg/addons/helm_deployer.go
@@ -25,40 +25,57 @@ func NewHelmDeployer(name, deployDir string, valuesPath string) Addon {
 }
 
 // Installs deployer if the config is changed
-func (d *HelmDeployer) Enable(ctx context.Context, namespace string,
+func (d *HelmDeployer) Enable(ctx context.Context, preInstallHooks []func() error,
+	namespace string,
 	config *orchestv1alpha1.ApplicationConfig) error {
 
-	// First we need to check if there is already a release
-	_, err := helm.GetReleaseConfig(ctx, d.name, namespace)
-	if err == nil {
-		return nil
-	}
+	releaseName := fmt.Sprintf("%s-%s", namespace, d.name)
 
-	/*
-		// Transform the values struct to helm values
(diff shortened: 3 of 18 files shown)


services/orchest-controller/pkg/addons/helm_deployer.go
```diff
@@ -25,40 +25,57 @@ func NewHelmDeployer(name, deployDir string, valuesPath string) Addon {
...
+
+	// First we need to check if there is already a release
+	oldConfig, err := helm.GetReleaseConfig(ctx, releaseName, namespace)
+	if err == nil {
+		// check if update is required or not
+		newConfig, err := helm.RunCommand(ctx, deployArgs.WithTemplate().Build())
+		if err != nil {
+			// Failed to get new config, probably it is best to not update
+			return err
+		}
```
fruttasecca, line 61: Probably best to add some logging when an error happens here

services/orchest-controller/pkg/addons/helm_deployer.go
```diff
@@ -25,40 +25,57 @@ func NewHelmDeployer(name, deployDir string, valuesPath string) Addon {
...
+		// Unfortunately, the value returned from helm get manifest has 1 extra byte,
+		// so in order to compare it with the new config, we have to remove the tail of it.
+		if newConfig == oldConfig[:len(oldConfig)-1] {
+			// There is no need for update, return without err
+			return nil
+		}
+	}
```
fruttasecca, line 68: Is the byte always the same? Can we do trimming in a safer manner?

We are comparing string to string, is it a yaml or a json or anything easily parseable? I was wondering if it makes sense to actually parse it and do a deep equality check at a higher level of abstraction

services/orchest-controller/pkg/addons/addons.go
```diff
@@ -35,8 +35,8 @@ func NewDefaultAddonsConfig() AddonsConfig {
 }
 
 type Addon interface {
-	// Installs addon if the config is changed
-	Enable(ctx context.Context, namespace string, config *orchestv1alpha1.ApplicationConfig) error
+	// Enable the addon. preInstallHooks should be called, before enabling the addon
+	Enable(ctx context.Context, preInstallHooks []func() error, namespace string, config *orchestv1alpha1.ApplicationConfig) error
```
fruttasecca, line 39: I don't have the full context, would it make sense to have the implementors of this interface both have `RunPreinstallHooks` (or something like that) and `Enable`? So that the order of calling is not a concern of the implementor

services/orchest-controller/pkg/addons/helm_deployer.go
nhaghighat: If it returns error, the log will be printed here:
https://github.com/orchest/orchest/pull/1126/files#diff-ea00a1f5097a1bc613e94d7d1f0b13f2c785eab68b530ab5df8ed91f9b0b1773L596

services/orchest-controller/pkg/addons/helm_deployer.go
nhaghighat: I trimmed the string and it resolved the issue.

services/orchest-controller/pkg/helm/builder.go
```diff
@@ -18,17 +19,27 @@ func NewHelmArgBuilder() *HelmArgBuilder {
...
 func (builder *HelmArgBuilder) WithUpgradeInstall() *HelmArgBuilder {
-	builder.args = append(builder.args, "upgrade", "--install")
+	builder.command = []string{"upgrade", "--install"}
 	return builder
 }
 
 func (builder *HelmArgBuilder) WithUnInstall() *HelmArgBuilder {
-	builder.args = append(builder.args, "uninstall")
+	builder.command = []string{"uninstall", "--install"}
+	return builder
```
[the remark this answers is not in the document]
nhaghighat: good catch

services/orchest-controller/pkg/addons/addons.go
nhaghighat: Well, I want the hooks to be called if necessary, and the deployer has the information if the install/update is needed or not

yannickperrenet:
I just ran a test and it worked perfectly. What I did (for completeness):

1. Build Orchest on this branch with a version tag of `v2022.07.3`
2. Change `podLabels` in `argo-workflows/orchest-values.yaml` so I can check whether the argo controller indeed get's the newly added labels after the update
3. Follow [the docs](https://docs.orchest.io/en/stable/development/development_workflow.html#test-updating-orchest) to update to `v2022.07.4` with `--dev`
4. Check whether the update succeeded and whether the Argo controller has the newly added labels

And it did! The Argo controller got redeployed :smile_cat: 

Great job :fire:

services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
```diff
@@ -41,7 +41,7 @@ storage: filesystem
 
 tlsSecretName: registry-tls-secret
 secrets:
-  haSharedSecret: ""
+  haSharedSecret: "haSharedSecret"
```
yannickperrenet, line 44: Was this just for testing or why is this needed?

services/orchest-controller/pkg/addons/helm_deployer.go
```diff
@@ -25,40 +26,56 @@ func NewHelmDeployer(name, deployDir string, valuesPath string) Addon {
...
+		if err != nil {
+			// Failed to get new config, probably it is best to not update
+			return err
+		}
+		// Unfortunately, the value returned from helm get manifest has 1 extra byte,
+		if strings.TrimSpace(newConfig) == strings.TrimSpace(oldConfig) {
+			// There is no need for update, return without err
+			return nil
+		}
+	}
```
yannickperrenet, line 68: It seems to me that this is the "most important" addition of this PR. However, I don't completely understand how the `oldConfig` and `newConfig` are constructed. So my question is: How do we create the `oldConfig` and how do we create the `newConfig` in order to be able to compare the two to see whether there are changes?

Asking, because when updating Orchest we first update the `orchest-controller` and so the old container is no longer running.

services/orchest-controller/deploy/thirdparty/docker-registry/orchest-values.yaml
nhaghighat: This line deserves a comment in the code, if we don't provide this value, a new one will be randomly generated and every time we want to see if the docker registry is updated or not, since this value is different we think the docker-registry is updated.

yannickperrenet: Ahaa makes a ton of sense. A comment explaining exactly that would indeed be a good idea :100:

services/orchest-controller/pkg/addons/helm_deployer.go
nhaghighat: `helm` stores the manifests of the release in a `secret` inside k8s, and uses that to retrieve the deployed manifests.
regarding the new config, `helm` has an option to give us the manifest it is going to deploy without deploying it.

yannickperrenet: Great explanation, thank you! Can you find the appropriate place to add that as a comment? :)

PR #1130, opened 2022-07-19
fix: show a Snackbar when stopping all sessions.

iannbing:
## Description

Show a Snackbar if user chooses to stop all sessions in order to build JupyterLab. After all sessions are stopped, it will start to build JupyterLab automatically.

Screenshot 2022-07-19 at 12 42 27


yannickperrenet (approved):
:100:

PR #1135, opened 2022-07-20
Improve path abstractions

mausworks:
## Description

Me and @iannbing talked about doing this last week, and tonight I felt the itch to fix it.

The background of this PR is that while I was working on `FileTree` and friends, I had trouble finding the right functions for common path operations, such as joining, getting extensions, and so on. I was also unsure of what semantics they used, which made me sanity-check a lot of things manually. I also noticed that we had duplicate functionality, or near-similar functionality that could be consolidated into the same function.

This PR aims to remedy these things by …

- Making all (well most) path functions behave like those in NodeJS
- Renaming all path functions to their NodeJS equivalents
- Consolidates all common path functions into `@/utils/path`
- Use these path functions in all places that I could identify easily
- Adds unit tests for these functions

## Checklist


services/orchest-webserver/client/src/utils/__tests__/path.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+describe("relative", () => {
+  it("resolves the expected paths", () => {
+    expect(path.relative("/", "/")).toBe("");
+    expect(path.relative("/", "/a/b/c.py")).toBe("a/b/c.py");
+    expect(path.relative("/a/", "/a/b/c.py")).toBe("b/c.py");
+    expect(path.relative("/a/d/", "/a/b/c.py")).toBe("../b/c.py");
+    expect(path.relative("/a/b/d/e/f/", "/a/b/c.py")).toBe("../../../c.py");
+  });
+
+  it("returns an empty string for two inputs", () => {
```
mausworks, line 90:
```suggestion
  it("returns an empty string for two empty inputs", () => {
```

iannbing (approved):
Awesome work! Now we have a standardized way to handle file paths! 💯

PR #1140, opened 2022-07-20
Add Workflows Community Initiative metadata

astrojuanlu:
## Description

As per title. See https://workflows.community/systems/contribute

## Checklist

### DIFF
--- a/.wci.yml
+++ b/.wci.yml
@@ -1,0 +1,30 @@
+name: Orchest
+headline: "A GUI for developing, running and managing container workflows"
+description: >
+  Orchest is a tool for building data pipelines, made to be easy to use.
+  - Visually construct pipelines through our user-friendly UI
+  - Code in notebooks and scripts
+  - Run any subset of a pipelines directly or periodically
+  - Easily define your dependencies to run on any machine
+  - Spin up services whose lifetime spans across the entire pipeline run
+  - Version your projects using git
+
+language: Python, Bash, R, Julia, JavaScript
+
+documentation:
+  general: https://docs.orchest.io/en/stable/
+  installation: https://docs.orchest.io/en/stable/getting_started/installation.html
+  tutorial: https://docs.orchest.io/en/stable/getting_started/quickstart.html
+
+social:
+  twitter: orchestofficial
+
+execution_environment:
+  interfaces:
+    - Command line
+  #   - HTTP REST API
+  resource_managers:
+    - kubernetes
+    - <resource_manager_2>
+  # transfer_protocols:
+  #   - SSH
--- a/.wci.yml
+++ b/.wci.yml
@@ -22,9 +22,9 @@ social:
 execution_environment:
   interfaces:
     - Command line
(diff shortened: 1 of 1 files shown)


astrojuanlu:
The fields are not very well documented (or at least I could not find an explanation) so I guided myself from some existing examples, see:

https://sourcegraph.com/search?q=context:global+file:.wci.yml&patternType=literal

.wci.yml
```diff
@@ -0,0 +1,30 @@
...
+  installation: https://docs.orchest.io/en/stable/getting_started/installation.html
+  tutorial: https://docs.orchest.io/en/stable/getting_started/quickstart.html
+
+social:
+  twitter: orchestofficial
+
+execution_environment:
+  interfaces:
+    - Command line
+  #   - HTTP REST API
```
yannickperrenet, line 25: Wouldn't `GUI` also count as interface? We often times say "visual **interface**" so I guess it would.

.wci.yml
```diff
@@ -0,0 +1,30 @@
...
+  resource_managers:
+    - kubernetes
+    - <resource_manager_2>
```
yannickperrenet, line 28: I guess this line can be removed?

.wci.yml
```diff
@@ -0,0 +1,30 @@
+name: Orchest
+headline: "A GUI for developing, running and managing container workflows"
+description: >
+  Orchest is a tool for building data pipelines, made to be easy to use.
+  - Visually construct pipelines through our user-friendly UI
+  - Code in notebooks and scripts
+  - Run any subset of a pipelines directly or periodically
+  - Easily define your dependencies to run on any machine
+  - Spin up services whose lifetime spans across the entire pipeline run
+  - Version your projects using git
```
yannickperrenet, line 10: @ricklamers @astrojuanlu Do you think there would be more value in explaining "when to use Orchest" or "slam dunk use case" here instead of just listing some of the features?

Could possibly also be a change we can make to our README? Given the hits/views on our docs it doesn't seem like anyone is really pressing the links here (talking about the README).

.wci.yml
```diff
@@ -0,0 +1,30 @@
...
+
+language: Python, Bash, R, Julia, JavaScript
```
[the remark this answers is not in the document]
astrojuanlu: I have no idea to be honest, I wondered the same thing 🤷🏽

.wci.yml
astrojuanlu: We can add that and see how it goes 👍🏽

.wci.yml
astrojuanlu: Oops, yes

.wci.yml
astrojuanlu: We could definitely iterate on the README but I think it's useful to describe what the product does. And for a short description like this, I think listing a few features is probably the right thing to do.

astrojuanlu:
All comments addressed, merging.

PR #1141, opened 2022-07-20
Fix error on environments page

mausworks:
## Description

Removes the error dialog shown in this video:

https://www.loom.com/share/cff42bbb98e64fa88660c281268fcc43

## Checklist

- [x] I have manually tested my changes and I am happy with the result.
- [x] The documentation reflects the changes.
- [x] The PR branch is set up to merge into `dev` instead of `master`.

PR #1142, opened 2022-07-20
Fix attempting to start a session when environment build in progress issue.

iannbing:
## Description

When environment build is in progress, it shouldn't attempt to start a session, otherwise it will prompt an error, e.g. https://www.loom.com/share/cff42bbb98e64fa88660c281268fcc43.

This PR fixes this issue by setting read-only `true` as default and then set it to `false` if conditions met. So that it won't attempt to launch a session when view is just mounted.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/hooks/useIsReadOnly.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/hooks/useIsReadOnly.tsx
@@ -2,14 +2,12 @@ import {
   BUILD_IMAGE_SOLUTION_VIEW,
   useProjectsContext,
 } from "@/contexts/ProjectsContext";
-import { hasValue } from "@orchest/lib-utils";
+import { useCustomRoute } from "@/hooks/useCustomRoute";
 import React from "react";
 
-export const useIsReadOnly = (
-  projectUuid: string | undefined,
-  jobUuid: string | undefined,
-  runUuid: string | undefined
-) => {
+export const useIsReadOnly = () => {
+  const { jobUuid, runUuid: jobRunUuid } = useCustomRoute();
+
   const {
     dispatch,
     state: { pipelineIsReadOnly },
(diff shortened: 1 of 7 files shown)


services/orchest-webserver/client/src/pipeline-view/hooks/useIsReadOnly.tsx
```diff
@@ -26,17 +24,26 @@ export const useIsReadOnly = (
...
   React.useEffect(() => {
-    if (!hasActiveRun && hasValue(projectUuid)) {
-      ensureEnvironmentsAreBuilt(BUILD_IMAGE_SOLUTION_VIEW.PIPELINE);
+    // For inspecting a job run, it is ALWAYS read-only.
+    if (isJobRun) {
+      setIsReadOnly(true);
+    }
+    // For interactive runs, set it to read-only if any environment is not built.
+    if (!isJobRun) {
+      ensureEnvironmentsAreBuilt(
```
yannickperrenet, line 42: Given that this function uses the environment's gate it should be possible (given the implementation of the gate) to start a session when an environment build is in progress IF the environment has been build before.

I guess the issue referred to in this PR hits the case where a new environment was added and thus not yet build.

Just wanted to mention it to make you aware of it :)

iannbing: I see. So ideally FE should check if all the used environments have existing builds? Or would you suggest not warning user about the failed attempts on launching a session other than Building JupyterLab?

yannickperrenet: Just wanted to mention how the gate works for your understanding (which I thought would help thinking about this issue). Didn't think there would be any actionables for this PR though. Sorry for the confusion :grimacing: 

----

> So ideally FE should check if all the used environments have existing builds?

No, the FE should still just check the gate. The gate already checks whether all used environments have existing builds.

> When environment build is in progress, it shouldn't attempt to start a session

I would rather phrase this as: "When the gate check fails, the FE shouldn't attempt to start a session". Which I think is what would happen still in this PR, or not?

> Or would you suggest not warning user about the failed attempts on launching a session other than Building JupyterLab?

I am not advocating for any changes other than how we are currently doing things, so a warning if the gate checks fails (meaning that environments need to be build) or if there is an ongoing JupyterLab build.

yannickperrenet:
I guess this PR and https://github.com/orchest/orchest/pull/1141 are trying to fix the same thing? @iannbing @mausworks

services/orchest-webserver/client/src/pipeline-view/hooks/useIsReadOnly.tsx
iannbing: > "When the gate check fails, the FE shouldn't attempt to start a session". Which I think is what would happen still in this PR, or not?

It won't. `ensureEnvironmentsAreBuilt` involves a long chain of promises. It behaves like this:
- checkGate passes, return true.
- checkGate failed, ask user if he wants to build, if yes, wait until the build is successfully done, return true.
- checkGate failed and user chooses not to build, return false.
- checkGate failed and user chooses to build, but build failed, return false.

Therefore, the then callback can update `isReadOnly` accordingly.

> or if there is an ongoing JupyterLab build.

Good catch 👍. I set `isReadOnly` as true if use hit `JupyterEnvironmentBuildInProgress` error.

mausworks:
> I guess this PR and #1141 are trying to fix the same thing? 

Yup, they aim to solve the same thing. Mine is just the most simplest solution that I could think of … this solution is more elaborate, and I think addresses more of the core problem.

mausworks (approved):

PR #1143, opened 2022-07-20
Remove jQuery-based overflow listener

mausworks:
## Description

This PR removes our jQuery-based overflow listener. It seems to be mostly dead code. The only thing it seems to be used for right now is to add a drop shadow to the button container on the project settings page (see picture below). I never managed to get the page into a state where it actually displayed it however, so the picture below is from me adding the class in with dev tools.

image

## Checklist


services/orchest-webserver/client/src/pipeline-view/step-details/StepDetailsProperties.tsx
```diff
@@ -168,37 +168,43 @@ export const StepDetailsProperties = ({
     let oldConnectionIndex = 0;
     let newConnectionIndex = 0;
 
-    let numConnectionListItems = $(refManager.refs.connectionList).find(
-      ".connection-item"
-    ).length;
+    const connectionItems = [
+      ...refManager.refs.connectionList.querySelectorAll(".connection-item"),
```
iannbing, line 172: When opening the StepDetails panel, I got an error saying that `connectionList` is undefined. I looked up `refManager.refs` and there's only `titleTextField`.

services/orchest-webserver/client/src/App.tsx
```diff
@@ -16,24 +15,6 @@ import { useAppContext } from "./contexts/AppContext";
...
-
-    if (el.offsetHeight < el.scrollHeight || el.offsetWidth < el.scrollWidth) {
-      overflowed = true;
-    } else {
-      overflowed = false;
-    }
-  });
-
-  return overflowed;
-};
```
iannbing, line 35: I think this overflowing function was intended to make the element (i.e. the SAVE button) sticky to the bottom of the page, which could be achieved via pure CSS nowadays. This function is no longer working for a while, so I think it's safe to remove this. On top of this, we are deprecating this SAVE button and moving towards "auto saving". I agree that we could just remove this.

mausworks: Thanks for the explanation. :+1:

iannbing (approved):
Apart from the null checking, LGTM. Very happy to say goodbye to jQuery 👍

services/orchest-webserver/client/src/pipeline-view/step-details/StepDetailsProperties.tsx
mausworks: Great find, pushed a fix. :+1:

PR #1147, opened 2022-07-22
New UI design part 3: Pipeline Editor

iannbing:
## Description

- [x] Core functionality
- [ ] Read-only banner
- [ ] Step details panel
- [ ] Pipeline file name responsiveness
- [ ] Saving status chip


lib/javascript/utils/src/typed.ts
```diff
@@ -46,7 +46,13 @@ export function checkHeartbeat(url: string, retries = 250) {
 }
 
 // used in orchest-webserver only
-export function intersectRect(r1, r2) {
+type Rectangle = {
```
mausworks, line 49: Nit: This conflicts with the `Rectangle` component (although that's not really an issue). Let's rename this to `Rect`.

services/orchest-webserver/client/public/image/grid-dot.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 50 50">
```
mausworks, line 1: Update viewbox to `0 0 64 64`

services/orchest-webserver/client/src/components/HUD.tsx
```diff
@@ -0,0 +1,20 @@
+import { Stack } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+export const HUD = styled(Stack)`
```
mausworks, line 4: Rename to `Hud`

services/orchest-webserver/client/src/components/ResizableContainer.tsx
```diff
@@ -48,118 +44,149 @@ export const ResizableContainer = ({
...
+  ResizableContainerProps
+>(
+  (
+    {
+      children,
+      onResized,
+      initialWidth,
+      initialHeight,
+      minHeight,
+      maxHeight,
```
mausworks, line 60: We can probably assign `maxHeight` and `minHeight` to `Infinity` and `-Infinity` respectively here. That way we don't need to check `hasValue` later.

services/orchest-webserver/client/src/components/ResizableContainer.tsx
```diff
@@ -48,118 +44,149 @@ export const ResizableContainer = ({
...
-    [getNewWidth]
-  );
+    const getNewHeight = React.useCallback(
+      (position: React.MutableRefObject<ClientPosition>): number => {
+        const { top } = getOffset(localRef.current);
 
-  const onResizeHeight = React.useCallback(
-    (position: React.MutableRefObject<ClientPosition>) => {
-      const height = getNewHeight(position);
+        let newHeight = minHeight
```
mausworks, line 111: Nit: Use `const` here.

services/orchest-webserver/client/src/components/common/IconButton.tsx
```diff
@@ -2,9 +2,9 @@ import MuiIconButton, { IconButtonProps } from "@mui/material/IconButton";
 import Tooltip from "@mui/material/Tooltip";
 import React from "react";
 
-export const IconButton: React.FC<IconButtonProps> = (props) => {
-  return props.title ? (
-    <Tooltip title={props.title}>
+export const IconButton: React.FC<IconButtonProps> = ({ title, ...props }) => {
```
mausworks, line 5: Was this what fixed all the `Tooltip` errors after we changed eslint to build in dev-mode? If so, nice catch! :+1:

mausworks:
I've gone over the PR a few times, and done a lot of testing.

All I can find right now is a performance regression with the step details panel resize. I'm looking at that now. The idea is to replace that handler with the standardized `<ResizableContainer>` , but it currently doesn't work when it's anchored to the right, so I'm fixing that first.

ncspost:
Great work @iannbing - this is huge, so I'll start dropping what I find into this comment throughout the day and structure it as much as possible!

# File manager

### Icon button distribution

- [ ] On file manager resize, evenly distribute icon buttons across width rather than hang right

 
CleanShot 2022-08-02 at 09 43 52@2x

### selection style

- [ ] Could we change style to blue when selected?

CleanShot 2022-08-02 at 09 36 44@2x

### duplicate selection

- [ ] When you make a selection and then keyboard through file manager, you get a strange thing where there appear to be two selections. Could we just make the most recent keyboard or click interaction as a single selection?

CleanShot 2022-08-02 at 09 37 54@2x

### inconsistent cursor on pipeline links

- [ ] Make both finger pointers

CleanShot 2022-08-02 at 09 55 46

# Canvas

### Canvas background pattern

- [ ] Make the canvas bg pattern repeat forever without a border

CleanShot 2022-08-02 at 09 26 04@2x

### Alert padding

- [ ] Reduce vertical and horizontal padding for alert to 8px

CleanShot 2022-08-02 at 09 28 13@2x

# Toolbar

### Primary button text

- [ ] Change text size of primary button to ButtonSmall to match text buttons

### Button corners

- [ ] Fix strange button masking on right side

CleanShot 2022-08-02 at 09 33 35@2x

### Button spacing

- [ ] Increase from 8px to 16px horizontally

CleanShot 2022-08-02 at 09 58 56@2x

### Logs dialog

- [ ] Have a separate URL for Logs dialog than from pipeline UI


# Keyboarding order

- [ ] Remove additional tab item between logout button and new pipeline button

mausworks:
Thanks for all the feedback @ncspost. I'll go over this today. :+1:

ncspost:
> Thanks for all the feedback @ncspost. I'll go over this today. 👍

ncspost:
> Thanks for all the feedback @ncspost. I'll go over this today. 👍

np - there'll be more!

ncspost:
# Project page

- [ ] Add "rename" functionality to menu on Projects page

CleanShot 2022-08-03 at 08 25 02@2x

# Zoom button

- [ ] Text input is really dodgy when deleting and entering text because it immediately responds - maybe we shouldn't change zoom until hit submit
- [ ] Change zoom shortcut to `cmd +` and `cmd -`
- [ ] Change default zoom increments to: 2%, 3%, 6%, 13%, 25%, 50%, 100%, 200%
- [ ] Rename "centre view" to "zoom to fit" and move to top section
- [ ] Add "Zoom to 100%"

CleanShot 2022-08-03 at 08 30 45@2x

ncspost:
CleanShot 2022-08-03 at 08 46 11@2x

Docs link should link: https://docs.orchest.io/en/stable/fundamentals/pipelines.html#pipelines

ncspost:
When changing pipeline via sessions section the tooltip does a weird jump to the top of the file manager

CleanShot 2022-08-03 at 08 50 45

ncspost:
CleanShot 2022-08-03 at 08 56 30

Something weird happens when I try to stop pipeline sessions

mausworks:
@ncspost 
> Make the canvas bg pattern repeat forever without a border

We deemed this out of scope for this iteration, it's not as simple as it seems and a definite rabbit hole. It's not worth the time investment now.

ncspost:
Is it possible to display the status of environments built in the alerts?

CleanShot 2022-08-04 at 10 10 46@2x

mausworks:
@ncspost 

> Is it possible to display the status of environments built in the alerts?

Yes it is, but I don't want to scope creep this PR too much. I think it's more important to focus on getting the whole re-design shipped, and then we can focus on these types of features.

mausworks:
> When you make a selection and then keyboard through file manager, you get a strange thing where there appear to be two selections. Could we just make the most recent keyboard or click interaction as a single selection?

This is the default behavior of the MUI tree view. It is that way, because otherwise you wouldn't be able to select multiple files using the keyboard, the same behavior exists in VSCode:

https://user-images.githubusercontent.com/8259221/183058572-710b6d28-7a0d-4307-ac86-13a19e57e617.mp4

So it's pretty standard behavior, and I think we should conform to it. I also don't want to override standard behavior in MUI.

mausworks:
@ncspost 

> Button text: "Build environment(s)"

This doesn't quite make sense when the environments are being built. :thinking: 

[I think you can add your copywriting suggestions directly here in GitHub](https://egghead.io/lessons/github-add-suggestions-in-a-github-pr-review), it should also clear up any confusion about which states there are. :+1: 

I moved all the copy into the same switch-block, so you can see exactly. Here's the relevant section for it: https://github.com/orchest/orchest/pull/1147/files#diff-0417fe8b0e3c0b05c707c00c9ba655729990a300c911889013d3fe691042ce3cR62-R92

> Question: Does the environment build autostart when opening a pipeline? Or do you need to press the alert button?

The build should start immediately when you open the editor, but there could be an edge-case where that doesn't happen and the user has to trigger it.

mausworks:
@ncspost I have fixed all remarks now except for those noted below.

> Add numbers to incoming order so users know sort order

I had this in the first iteration, but we decided against it because it felt confusing
when reordering, [see this internal Slack thread](https://orchestinternal.slack.com/archives/C01RKD0EKGU/p1658586692276539?thread_ts=1658500464.354419&cid=C01RKD0EKGU).

> Can intercom button move and hang to the left of the step detail panel when open?
> Intercom button hangs over step details once I have clicked it

I think so, yes … but Intercom isn't displaying for me. 🤔
Will take a look once it does! :+1:

> Question: It doesn't look like log unread dots are implemented in the step detail panel. Will those be implemented later?

This is definitely a rabbit hole that I don't think is worth the time investment now.

> Add "rename" functionality to menu on Projects page

Out of scope since this PR Is for the pipeline editor. But definitey something we can do in the base branch/PR https://github.com/orchest/orchest/pull/1060

> Change zoom shortcut to cmd + and cmd -

The library we use for hotkeys seems to work inconsistently with <KBD>CTRL/CMD</KBD>+<KBD>+</KBD>, I spent a couple of hours trying to fix it, but got nowhere. Let's revisit this one.

> Disable an canvas panning on empty state

I disabled the selection rectangle, but I couldn't find a consistent way of disabling panning after spending a good amount of time on it. I don't think it's super important, but if @iannbing knows how to easily solve it, feel free to pick it up.

> Something weird happens when I try to stop pipeline sessions

I cannot reproduce this, so I'm guessing it was something on your machine, or something that has been fixed since then.

PR #1150, opened 2022-07-25
Use ctr to download images even in docker env

nhaghighat:
## Description

This PR uses ctr to download images even in the docker environment, ctr allows to skip SSL verification, in this PR, ctr downloads the image, tar it, and then load it to docker, it has an overhead for the first session, if the image is not already present.

## Checklist

### DIFF
--- a/utility-containers/image-puller/pull_image.sh
+++ b/utility-containers/image-puller/pull_image.sh
@@ -32,10 +32,10 @@ elif [ "$CONTAINER_RUNTIME" = docker ]; then
 
     log "attempting to pull image ${IMAGE_TO_PULL}}"
 
-    ctr -a=/var/run/containerd.sock i pull "${IMAGE_TO_PULL}" --skip-verify
-    ctr -a=/var/run/containerd.sock i export pulled-image.tar "${IMAGE_TO_PULL}"
-    docker -H unix:///var/run/docker.sock image import pulled-image.tar "${IMAGE_TO_PULL}"
-
+    ctr -a=/var/run/containerd.sock i pull "${IMAGE_TO_PULL}" --skip-verify && \
+    ctr -a=/var/run/containerd.sock i export pulled-image.tar "${IMAGE_TO_PULL}" && \
+    docker -H unix:///var/run/docker.sock load -i pulled-image.tar && \
+    rm -rf pulled-image.tar && ctr -a=/var/run/containerd.sock i rm "${IMAGE_TO_PULL}"
 
 else
     echo "Container runtime is not supported"
--- a/utility-containers/image-puller/pull_image.sh
+++ b/utility-containers/image-puller/pull_image.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -euo pipefail
 
 log() {
   timestamp=$(date +"[%m%d %H:%M:%S]")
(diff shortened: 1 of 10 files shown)


utility-containers/image-puller/pull_image.sh
```diff
@@ -32,10 +32,10 @@ elif [ "$CONTAINER_RUNTIME" = docker ]; then
...
     log "attempting to pull image ${IMAGE_TO_PULL}}"
 
-    ctr -a=/var/run/containerd.sock i pull "${IMAGE_TO_PULL}" --skip-verify
-    ctr -a=/var/run/containerd.sock i export pulled-image.tar "${IMAGE_TO_PULL}"
-    docker -H unix:///var/run/docker.sock image import pulled-image.tar "${IMAGE_TO_PULL}"
-
+    ctr -a=/var/run/containerd.sock i pull "${IMAGE_TO_PULL}" --skip-verify && \
+    ctr -a=/var/run/containerd.sock i export pulled-image.tar "${IMAGE_TO_PULL}" && \
+    docker -H unix:///var/run/docker.sock load -i pulled-image.tar && \
+    rm -rf pulled-image.tar && ctr -a=/var/run/containerd.sock i rm "${IMAGE_TO_PULL}"
```
yannickperrenet, line 38: Versus the existing 
```sh
    docker -H unix:///var/run/runtime.sock pull "${IMAGE_TO_PULL}" --disable-content-trust
```

these changes seem like they would incur significant overhead when running the `initContainer`, wouldn't they?

nhaghighat: They do, but only for the first time, because after that we check for the image existence,

yannickperrenet: > They do, but only for the first time

We would be hitting this case more often than you would initially think. Every time an Environment is build, this case is hit because it constitutes for a new image (i.e. `IMAGE_TO_PULL` here).

So I did a quick test (which ran on EKS with these changes):
1. Run the quickstart pipeline (`get-data.py` ran in `8s` as it wasn't the first time I ran the pipeline)
2. Rebuild the Environment with `echo "hi"` added to the setup script to ensure it would create a new image
3. Rerun the quickstart pipeline (`get-data.py` ran in `43s`)

That means the code we are discussing here has an overhead of `35s` in this case, which is not acceptable imo. We will be focusing on performance improvements in the next weeks anyways so we might as well make the change in this PR is performant.

yannickperrenet (changes requested):
As per https://github.com/orchest/orchest/pull/1150#discussion_r929159246, I don't think the performance hit is acceptable and should be fixed before we can merge.

Great find though and great that you could pick this up so quickly :muscle: 

(cc @ricklamers @fruttasecca )

nhaghighat:
@yannickperrenet : fixing the performance hit requires a significant amount of work, if we want to postpone this, then our EKS users cant use orchest for a good amount of time.

utility-containers/image-puller/pull_image.sh
yannickperrenet: And we would have to check whether kernels can timeout when taking such a long time to start.

ricklamers:
Not using this in favor of `buildah` in #1157

PR #1157, opened 2022-07-28
Feat/inject registry certificate

nhaghighat:
## Description

This PR fixes the problem of pulling in docker enviorment.

Here is the procedure:

- Controller finds a free IP address in the cluster and creates the certificate of the registry with that IP address
- Controller deploys the docker-registry with static Cluster-IP address
- NodeAgent uses life-cycle hooks to inject the certificate on startup and remove it on stop

Fixes: #issue

## Checklist


services/orchest-controller/pkg/certs/certgen.go
```diff
@@ -65,6 +65,9 @@ type Configuration struct {
 	// configuring Subject Alt Names on the certificates.
 	DNSName string
 
+	// IP address of the certificate
+	IP string
+
```
fruttasecca, line 70: Minor consideration, is the IP now required for generating certificates?

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -236,6 +244,92 @@ func getOrchestComponent(name, hash string,
...
+
+		// If next ip is not free
+		if bytes.Compare(nextIP, net.ParseIP(serviceList.Items[i+1].Spec.ClusterIP).To4()) == 0 {
+			continue
+		}
+
+		serviceIP = nextIP.To4().String()
+		break
+	}
+
```
fruttasecca, line 303: I guess we aren't just incrementing on the highest IP (i.e. the end of the sorted list of services) because some IPs in the range might be free right? If so, might be worth adding a comment here to remind us of that in the future.

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -236,6 +244,92 @@ func getOrchestComponent(name, hash string,
...
+	if serviceIP == "" {
+		serviceIP = nextIP(net.ParseIP(serviceList.Items[len(serviceList.Items)-1].
+			Spec.ClusterIP).To4(), 1).To4().String()
+	}
```
fruttasecca, line 307: This assumes that the last service has a `ClusterIP` and that it's not the end of the range, i.e. `nextIP` won't return `nil`, right?

What do we do in this case? I believe we can operate under the assumption that a service with a `ClusterIP` is there depending on the timeline of the installation (which I don't remember), e.g. if during the installation one of our services is set up before the registry. However, we still haven't fixed the case of `nextIP` returning `nil` right?

services/orchest-controller/pkg/controller/orchestcomponent/node_agent.go
```diff
@@ -107,6 +116,62 @@ func getNodeAgentDaemonset(metadata metav1.ObjectMeta,
...
+		certificateDirectory := fmt.Sprintf("/etc/docker/certs.d/%s/", registryIP)
+
+		template.Spec.Containers[0].Lifecycle = &corev1.Lifecycle{
+			PostStart: &corev1.LifecycleHandler{
+				Exec: &corev1.ExecAction{
+					Command: []string{
+						"/bin/sh",
+						"-c",
+						"mkdir " + certificateDirectory +
+							" && cp /tls-secret/ca.crt " + certificateDirectory,
```
fruttasecca, line 135: Guess `mkdir` needs a `-p`

fruttasecca:
From to the docs about `PreStop`:

>This hook is called immediately before a container is terminated due to an API request or management event such as a liveness/startup probe failure, preemption, resource contention and others. A call to the PreStop hook fails if the container is already in a terminated or completed state and the hook must complete before the TERM signal to stop the container can be sent. The Pod's termination grace period countdown begins before the PreStop hook is executed, so regardless of the outcome of the handler, the container will eventually terminate within the Pod's termination grace period. No parameters are passed to the handler.

Could it be that in case of an unforeseen failure of the `node-agent` we could end up with the certificate getting deleted? Can we avoid this? It would make the application go from a partial failure to a worse (if not total) failure, and lead to a difficult-to-debug behaviour

fruttasecca (commented):
lgtm but I'd like to make sure of some edge cases

services/orchest-controller/pkg/certs/certgen.go
nhaghighat: Hi, Yes, because the certificate the going to be generated for that IP

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
nhaghighat: good suggestion, will add the comment.

services/orchest-controller/pkg/controller/orchestcomponent/node_agent.go
nhaghighat: Fixed it

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
nhaghighat: What we do here, is that we sort all the service ips, and try to find the gap between them, if we could not, we get the next ip of the latest service

utility-containers/image-puller/pull_image.sh
```diff
@@ -1,10 +1,29 @@
...
+
+    set +e
     docker -H unix:///var/run/runtime.sock pull "${IMAGE_TO_PULL}" --disable-content-trust
+
+    if [ $? -ne 0 ]; then
+        set -e
+        echo "Docker pull failed, pulling with buildah."
+        buildah pull --tls-verify=false "${IMAGE_TO_PULL}"
+        echo "Pushing from buildah to docker-daemon."
+        # Expected by builda when docker-daemon is specified.
```
ricklamers, line 22: builda -> buildah in the comment

PR #1159, opened 2022-07-29
Fallback to buildah pull if docker pull fails

fruttasecca:
## Description

Fallbacks to using `buildah` to pull from the insecure internal registry and push to the docker daemon if pulling with docker fails, i.e. if we are in an environment or k8s flavor where docker won't pull from an insecure registry.


nhaghighat:
Sounds good, Thanks @fruttasecca

PR #1162, opened 2022-08-01
Use the IP of the existing registry service

nhaghighat:
## Description

This issue fixes the problem in updating to v2022.07.6.

In v2022.07.6 we introduced assigning the IP address of the registry service, but if the registry already exists, and the assigned IP by k8s differs from our IP, k8s fails to update the registry.

Fixes: #issue

## Checklist

### DIFF
--- a/services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
+++ b/services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
@@ -272,6 +272,13 @@ func setRegistryServiceIP(ctx context.Context, client kubernetes.Interface,
 		app.Config.Helm.Parameters = []orchestv1alpha1.HelmParameter{}
 	}
 
+	// We first check if there is an IP assigned by the controller for the service in Config
+	// then we check the actual IP of the service if exists, and if they are different the actual
+	// IP takes precedence and the configured IP will be changed to the actual one, (this is to
+	// fix the issue of the instances updated to v2022.07.6 because, controller in that version
+	// assigned the service IP, regardless of the IP of the present service.) If there was
+	// no service, we just find a free IP and set it in the configs
+
 	// We check the current controller assigned service IP first
 	serviceIpIndex := -1
 	for index, param := range app.Config.Helm.Parameters {
--- a/services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
+++ b/services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/orchest/orchest/services/orchest-controller/pkg/addons"
 	orchestv1alpha1 "github.com/orchest/orchest/services/orchest-controller/pkg/apis/orchest/v1alpha1"
 	"github.com/orchest/orchest/services/orchest-controller/pkg/certs"
 	orchestlisters "github.com/orchest/orchest/services/orchest-controller/pkg/client/listers/orchest/v1alpha1"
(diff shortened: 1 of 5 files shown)


services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -328,16 +349,6 @@ func setRegistryServiceIP(ctx context.Context, client kubernetes.Interface,
...
-	if app.Config.Helm.Parameters == nil {
-		app.Config.Helm.Parameters = []orchestv1alpha1.HelmParameter{}
-	}
-
-	app.Config.Helm.Parameters = append(app.Config.Helm.Parameters,
-		orchestv1alpha1.HelmParameter{
-			Name:  registryServiceIP,
-			Value: serviceIP,
-		})
-
```
fruttasecca, line 340: Guess this, or at least the second part, is actually needed to set the `serviceIP`if the logic got here?

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -266,12 +268,52 @@ func setRegistryServiceIP(ctx context.Context, client kubernetes.Interface,
...
+	} else {
+		currentIP = registryService.Spec.ClusterIP
+	}
+
+	// If the service is present
+	if currentIP != "" {
+		if serviceIpIndex >= 0 {
+			changed = app.Config.Helm.Parameters[serviceIpIndex].Value != currentIP && changed
+			app.Config.Helm.Parameters[serviceIpIndex].Value = currentIP
+			return changed, nil
```
fruttasecca, line 300: Should the `&& changed` be here?

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -349,6 +370,13 @@ func setRegistryServiceIP(ctx context.Context, client kubernetes.Interface,
 		serviceIP = nextIP(lastIP, 1).To4().String()
```
fruttasecca, line 370: Should the result be compared with `nil` for safety?

nhaghighat: nextIP always returns a result, it might be out of range but we don't know what is the range, I think there is not mush we can do here.

fruttasecca:
Is there any logic here that is strictly required just to "fix" `.7.6` instances? If so, might be worth putting a comment on it.

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
nhaghighat: No, line 298 already takes the current `changed` into account, but in line 298, it should be `|| changed`.
I will push it after testing it.

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
fruttasecca: I see, got confused by the 
```go
		nextIP := nextIP(net.ParseIP(serviceList.Items[i].Spec.ClusterIP).To4(), 1)
		if nextIP == nil {
			continue
		}
```
a few lines ahead, but we are already checking for `lastIP` for nil here

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
fruttasecca: You it be there at all, even as an OR? The only way `changed` is true at this point is `if app.Config.Helm == nil `, which implies that `if serviceIpIndex >= 0`, so this piece of logic wouldn't run

nhaghighat: I tend to keep the OR, we might add more logic later and there might be other cases where we can be here.
I think having it is more bulletproof for the future.

fruttasecca:
I've tested two updates, one going from a pre `.7.6` instance to this branch and then from this branch to itself, https://docs.orchest.io/en/stable/development/development_workflow.html#test-updating-orchest. I haven't seen suspicious logs in the controller and played around with Orchest a bit.

Nevertheless, it's better if you go through testing this multiple times and from different angles/existing states given the potential for issues

nhaghighat:
> Nevertheless, it's better if you go through testing this multiple times and from different angles/existing states given the potential for issues

Thanks, I'll test it a couple of more times with different branches, then If everything goes fine, will merge it.

Issue #1165, opened 2022-08-03
Image builder does not report any issue in case of unexpected errors


fruttasecca:
**Describe the bug**
When building an image, depending on the error type, the build logic might not recognize the error and skip directly to a build finished status. This is most likely due to the logic of the `ImageBuildSidecar` in `services/orchest-api/app/app/core/image_utils.py`.

**Expected behavior**
The builder should detect the error and a failure should be reported to the user.

**To Reproduce**
One way to reproduce the issue is to modify the code of `_get_buildah_image_build_workflow_manifest`  to point to a docker file that does not exist.

PR #1167, opened 2022-08-03
Get exit code to avoid race condition on failure

fruttasecca:
## Description
 
FIxes: #1165


ricklamers:
LGTM, but not familiar enough with k8s status cases to know whether this is comprehensive or not.

PR #1170, opened 2022-08-03
Add OpenReplay to Orchest Cloud

ricklamers:
## Description
This adds OpenReplay to Orchest Cloud users, with all privacy preserving settings enabled.

## Checklist


services/orchest-webserver/app/app/config.py
```diff
@@ -82,6 +82,8 @@ class Config:
     CLOUD = _config.CLOUD
 
     GPU_ENABLED_INSTANCE = _config.GPU_ENABLED_INSTANCE
+    OPENREPLAY_PROJECT_KEY = "yotEvKXSA22pOnv4vzSC"
+    OPENREPLAY_INGEST_POINT = "https://openreplay.orchest.io"
```
mausworks, line 86: I'm unsure of how our config is set up. Will this only be applied to cloud instances?

services/orchest-webserver/client/src/App.tsx
```diff
@@ -52,6 +53,16 @@ const App = () => {
         email: user_config.INTERCOM_USER_EMAIL,
         createdAt: config.INTERCOM_DEFAULT_SIGNUP_DATE,
       });
+
+      // Initialize OpenReplay
```
mausworks, line 57: Please remove this comment. I think the code is pretty self-descriptive. If you feel like extra detail is required, move this functionality to a function named `initializeOpenReplay`.

We just added a section in the [style guide Draft](https://www.notion.so/TypeScript-React-Conventions-Style-Guide-dc3172bc20c04da7900c6481e05525c3#586e5647ebbe4771909c387012f40392) about how we want comments to be used in the front-end. :+1:

mausworks (commented):
LGTM, except for the nit about the comment. :+1:

services/orchest-webserver/client/src/App.tsx
ricklamers: Haha agree

services/orchest-webserver/app/app/config.py
ricklamers: >I'm unsure of how our config is set up. Will this only be applied to cloud instances?

Yes.

mausworks (approved):

services/orchest-webserver/client/public/index.html
```diff
@@ -8,7 +8,7 @@
     <title>Orchest</title>
   </head>
   <body>
-    <link href="{{CSS_PATH}}" rel="stylesheet" />
+    {{CSS_BLOCK}}
```
mausworks, line 11: What do you think of doing this instead? `<style>{{CSS_CODE}}</style>`

That way you don't have to include the HTML in the CSS. :+1:

lib/javascript/build-frontend.sh
```diff
@@ -4,18 +4,37 @@ set -e
...
   JS_HASH_PLACEHOLDER="{{JS_HASH}}"
   JS_FILE_NAME="$1"
   CSS_FILE_NAME="$2"
 
-  cat public/index.html | sed "s/$JS_PATH_PLACEHOLDER/$JS_FILE_NAME/g" |
-    sed "s/$CSS_PATH_PLACEHOLER/$CSS_FILE_NAME/g" |
-    sed "s/$JS_HASH_PLACEHOLDER/$JS_FILE_NAME/g" >dist/index.html
+  CSS_BLOCK_FILE="dist/css-block.html"
+
+  if [ $3 == "inject-css" ]; then
```
mausworks, line 20: I don't think that we should support both inline CSS _and_ CSS via `<link>`, I think it's better if we pick just one strategy, and stick to it. Less work to support both, and a cleaner implementation.

mausworks (changes requested):
I think it's best we just go with the inline CSS solution. 🙂

lib/javascript/build-frontend.sh
ricklamers: The reason we have it like this is for the development reload loop. `esbuild`/`sass` build with watch supports writing to the file. I didn't want to implement a custom watch loop.

services/orchest-webserver/client/public/index.html
ricklamers: See other comment.

lib/javascript/build-frontend.sh
mausworks: Gotcha, didn't think of that. It makes sense. :+1: 

Let's hope we can move away from SASS entirely soon.

mausworks (approved):

iannbing (approved):

PR #1176, opened 2022-08-05
New UI design part 4: Environments

iannbing:
## Description

The new design of Environments view. This PR also introduces [zustand](https://github.com/pmndrs/zustand) for state management, and the new convention about `/api` folder structure. 

Note: this PR is based on #1147. The actual diff would be more visible after #1147 is merged.

Screenshot 2022-08-05 at 21 07 42


ncspost:
Great work @iannbing - this redesign is looking fresssssshhhhhhh

I'll be added more stuff to the list if I come across anything tomorrow.

# Env list

- [ ] When tab focus is on environments list, the next tab should focus on page content and not next environment in the list (ie tab once into list and navigate list with arrow keys)
- [ ] When tab focus is on environments list and enter is hit, keep tab focus on that environment in the environments list rather than moving it into the page content (ie I should be able to browse the env list with the keys, hit enter to view an environment and then browse the list some more)
- [ ] After deleting environment, move focus to the list item below rather than the top of the list  
- [ ] Change the alert build successful icon to the same as the env list build successful icon

CleanShot 2022-08-08 at 15 16 03@2x

- [ ] When list item selected, keep status icon, only show pencil when it is a draft

CleanShot 2022-08-08 at 17 24 28@2x

# Layout

- [ ] Do not persist accordion collapse/expand state across pages
- [ ] Show hover state (similar to icon button) on the accordion chevrons

# Alert

- [ ] When building an environment, show the blue alert. This should become sticky when scrolling so you have the ability to see build status and logs at the same time. ([see figma](https://www.figma.com/file/o9S8qtPAx3Vd1HIVtsQfYv/ORC-573-Navigation%2C-layout-%26-IA?node-id=1449%3A67924))

# Props

- [ ] Remove properties section title
- [ ] If name not edited and a new base image is selected, change name to match that base image type (e.g. Python 3 default changes to R when R is selected)
- [ ]  Reduce width of environment name text input to around half the current width
- [ ] Remove * in environment name text input
- [ ] Some funkiness when typing in the environment name text field (some letters randomly delete)
 
CleanShot 2022-08-08 at 17 37 19

# Base image

- [ ] Tooltip copy "Base images can be extended using the setup script below (see docs)"
- [ ] Change base image border for unselected items to 1px, keep selected as 2px
- [ ] When hovering over base image card, fill the card with the grey hover state colour
- [ ] Make selected base image bg colour primary/4p
- [ ] Base image box responsiveness
- [ ] Use 3 dot Julia logo instead of text logo 
- [ ] Show disabled cursor when hovering over unavailable image option

CleanShot 2022-08-08 at 15 18 39@2x

- [ ] When base image is in keyboard tab focus - don't make the "base image" section title blue

CleanShot 2022-08-08 at 17 20 24@2x

# Logs

- [ ] No way to exit out of the setup script with keyboard (maybe esc key?)
- [ ] Match colour of logs box to setup script box

CleanShot 2022-08-08 at 15 16 57@2x

- [ ] Remove top grey bar in logs box which says "logs"

CleanShot 2022-08-08 at 15 18 01@2x

mausworks (commented):
I'm done with around half the review, but I thought I might as well post and continue tomorrow.

services/orchest-webserver/client/src/pipeline-view/pipeline-canvas-header-bar/pipeline-operations/PipelineOperationButton.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const disabled = Boolean(pipelineReadOnlyReason) || hasNoStep;
+
+  const [buttonLabel, executeOperation] = React.useMemo(() => {
+    if (pipelineReadOnlyReason) return ["Run all", undefined];
+    if (displayedPipelineStatus === "RUNNING") return ["Cancel run", cancelRun];
+    if (displayedPipelineStatus === "CANCELING")
+      return ["Cancelling...", undefined];
+    if (shouldRunAll) return ["Run all", runAllSteps];
+    return ["Run selected", runSelectedSteps];
```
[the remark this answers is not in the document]
iannbing: Since in every case it already returns, the extra else-if becomes a noise to me. It drags me into thinking if I should continue reading the next code block. This naked(?) return is an important visual anchor for me; wrapping it with an else-if would double the time for me to troubleshoot 😅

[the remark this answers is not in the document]
iannbing: @mausworks Seems that GH eats up my reply. This file has been renamed in the previous PR. so I deleted this dead code. And I checked again in your revised version. I agree that it reads better (I would leave out the `else` though, but fine by me 🙂).

iannbing:
@ncspost Following are the items I would put as "won't fix" and the reason for that. Let me know if I misinterpreted you comment :-)

> Remove * in environment name text input
It's a standard implementation in MUI, showing that it's a required field. Any reason why we want to remove the sing?

> When base image is in keyboard tab focus - don't make the "base image" section title blue
It's also a standard implementation in favor of accessibility. The cards are actually radio boxes under the hood. When the radio boxes are focused, their label should be highlighted.

> Match colour of logs box to setup script box
It's implemented by convention. The theme we use for the script is called `dracula`, which is quite a popular code **editor** theme in the engineer community. Logs are commonly shown in the command-line style. These themes match their functionality very well imo.

ncspost:
> It's a standard implementation in MUI, showing that it's a required field. Any reason why we want to remove the sign?

True - but this is the only text field on the page. So it's sort of redundant because it doesn't differentiate from text fields that are optional.

> It's also a standard implementation in favor of accessibility. The cards are actually radio boxes under the hood. When the radio boxes are focused, their label should be highlighted.

But that's a section label rather than


> It's implemented by convention. The theme we use for the script is called dracula, which is quite a popular code editor theme in the engineer community. Logs are commonly shown in the command-line style. These themes match their functionality very well imo.

Fine by me for now

iannbing:
@ncspost 

> It's a standard implementation in MUI, showing that it's a required field. Any reason why we want to remove the sign? True - but this is the only text field on the page. So it's sort of redundant because it doesn't differentiate from text fields that are optional.

Okay. Just note that auto-save won't save any change if user intentionally leave this field blank. 🙂

>But that's a section label rather than the selected item, no?

Usually there's a label for a set of radio options to depict what these options are for. See [the example](here https://mui.com/material-ui/react-radio-button/#radio-group) from MUI. If you click on any of the options (Female, Male, Other), you'll see that `Gender` changes to the primary color.

We could opt for [standalone radio](https://mui.com/material-ui/react-radio-button/#standalone-radio-buttons), then we introduce an accessibility concern, meaning that from the HTML code, you only see that you're selecting something in Python, Julia, etc; the label `Base image` and the explanation tooltip are not attached to these radio options. I wouldn't recommend it.

yannickperrenet:
image

- [ ] In https://github.com/orchest/orchest/pull/1195 we actually removed the GPU listing in the Environments page and so it should not be present here either. (We will add it back in the future though once we have GPU support again.)

----

https://user-images.githubusercontent.com/26223174/185187022-abf67af8-4502-47e9-8e4b-84853c7fe94c.mp4

- [ ] Not sure what happened here, but changes to my setup script are not saved when I press "BUILD". Seems to be similar to the issue I had (which got fixed) when renaming a project that the rename would only become visible if I refresh the page. For the setup script, if I make a change, then the change is only reflected if I refresh the page before I click "BUILD"

----

https://user-images.githubusercontent.com/26223174/185188429-f7fa6625-9f3c-4786-a46b-a0ab5507400e.mp4

~~- [ ] Seems like there is something wrong with the session state with respect to environments. Not sure what is going on here.~~

Hmmm, I refreshed my page without cache and now everything seems to be correct again. Not sure what happened here.

----

https://user-images.githubusercontent.com/26223174/185190359-728d7604-8286-4e2e-a055-dbe322048e96.mp4

- [ ] Incorrect environment snackbar in pipeline editor. Seems to only be incorrect when environments are deleted.

----

https://user-images.githubusercontent.com/26223174/185191265-8bbf7b7d-a263-4329-bc6c-5f3000d6dc99.mp4

- [ ] I think in the FE we should prevent users entering an empty name for the Environment. Sure it is possible to use an empty name (from a BE perspective and from a pipeline definition perspective), but I think it leads to poor UX and so it is better to prevent it.

----

https://user-images.githubusercontent.com/26223174/185191824-cb01d881-df57-444a-b17c-38a5caec4382.mp4

- [ ] The Environment dropdown in the step properties for Python files doesn't seem to list available Environments.

----

image

- [ ] Moving indication of toggle to the left, so it is easier to discover the fact that you can click these rows. I read from left to right so after reading the title I stop reading and thus don't discover that I can click on the row. Thoughts @ncspost ?

ncspost:
> Moving indication of toggle to the left, so it is easier to discover the fact that you can click these rows. I read from left to right so after reading the title I stop reading and thus don't discover that I can click on the row. Thoughts @ncspost ?

I'm fine with that, if it's easy to achieve. I went with this pattern because it's standard MUI. Happy to align with the expand/collapse pattern in our file manager tree! i.e. Showing chevrons on left side of title.


CleanShot 2022-08-18 at 13 18 50@2x

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
```diff
@@ -0,0 +1,188 @@
+import { queryArgs } from "@/pipeline-view/file-manager/common";
+import {
+  Environment,
```
mausworks, line 3: No Action Needed (NAN): Maybe we should name this `EnvironmentData`, as per our style guide? This is the back-end data right?

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
```diff
@@ -0,0 +1,188 @@
...
+  try {
+    const response = await postValidate(projectUuid);
+    const envMap = new Map(existingEnvironments.map((env) => [env.uuid, env]));
+    let buildingEnvironments: string[] = [];
+    const environmentsToBeBuilt: string[] = [];
+    const updatedEnvironmentMap = produce(envMap, (draft) => {
+      response.actions.forEach((action, index) => {
+        const uuid = response.fail[index];
+        const environment = draft.get(uuid);
+        if (!environment) throw "refetch";
```
mausworks, line 70: Please avoid throwing strings. Even if it's an unreachable state.

I'm used to having this rule enabled: https://eslint.org/docs/latest/rules/no-throw-literal

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const validate = async (
+  projectUuid: string,
+  existingEnvironments: EnvironmentState[] = []
+): Promise<
+  | [EnvironmentState[], EnvironmentValidationData, boolean, string[], string[]]
+  | FetchError
+> => {
+  let hasActionChanged = false;
+  try {
```
mausworks, line 61: NAN: From this line to ~83 there's a farily complex state-management logic. I'm not implying that that complexity is uneeded, but the code _smells_ a little bit defensive.

Maybe there's a way to make it simpler? Maybe there are some assumptions that can be made that can simplify it? :)

services/orchest-webserver/client/src/api/environments/useEnvironmentsApi.ts
```diff
@@ -0,0 +1,361 @@
...
+import { FetchError } from "@orchest/lib-utils";
+import create from "zustand";
+
+export type EnvironmentBuildStatus =
+  | "allEnvironmentsBuilt"
+  | "environmentsNotYetBuilt"
+  | "environmentsFailedToBuild"
+  | "environmentsBuildInProgress";
+
+export type EnvironmentsApiState = {
```
mausworks, line 18: NAN: I think that this type describes the environment API itself, not the state of it, so I think it can be named just `EnvironmentsApi` :)

services/orchest-webserver/client/src/api/environments/useEnvironmentsApi.ts
```diff
@@ -0,0 +1,361 @@
...
+  const {
+    uuid,
+    name,
+    base_image,
+    gpu_support,
+    language,
+    project_uuid,
+    setup_script,
+  } = state;
+  return {
```
mausworks, line 79: You can use the new `omit` or `pick` helper here—whichever is most convenient! :)

```ts
return pick(
  state,     
  "uuid",
  "name",
  "base_image",
  "gpu_support",
  "language",
  "project_uuid",
  "setup_script");
```

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
```diff
@@ -0,0 +1,432 @@
...
+  OrchestConfig,
+  OrchestServerConfig,
+  OrchestUserConfig,
+  ReducerActionWithCallback,
+} from "@/types";
+import { ButtonProps } from "@mui/material/Button";
+import { fetcher, hasValue } from "@orchest/lib-utils";
+import React from "react";
+
+function parseLineBreak(lines: string) {
```
mausworks, line 12: NAN: This could be a component instead: `<PlainTextToHTML text={text} />`

services/orchest-webserver/client/src/environments-view/edit-environment/EditEnvironmentName.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { environmentUuid } = useCustomRoute();
+  const { environmentChanges, setEnvironmentChanges } = useEditEnvironment();
+  const [value = "", setValue] = React.useState<string>();
+  const [hasEdited, setHasEdited] = React.useState(false);
+
+  React.useEffect(() => {
+    if (
+      environmentUuid &&
+      environmentUuid === environmentChanges?.uuid &&
+      environmentChanges?.name
```
mausworks, line 17: Nit: please extract this large condition into a variable.

mausworks (approved):
Some nits and NANs (No action needed), I think it looks good overall.

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
iannbing: Changed it to

```
throw new Error("environment unavailable");
```

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
iannbing: See [commit](https://github.com/orchest/orchest/pull/1176/commits/474573a34af07c979aa24798e5f556196c88f813).

services/orchest-webserver/client/src/api/environments/useEnvironmentsApi.ts
iannbing: Changed. 👍

services/orchest-webserver/client/src/api/environments/useEnvironmentsApi.ts
iannbing: Nice! Very handy 👍.

services/orchest-webserver/client/src/environments-view/edit-environment/EditEnvironmentName.tsx
iannbing: Fixed :-).

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
iannbing: Indeed. Time to move update some old typings 🙂.

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
iannbing: Nice. This also helps me clean up something else. see [commit](https://github.com/orchest/orchest/pull/1176/commits/eb899fd2198bc2cf192561effb1089c3506dd09b).

yannickperrenet:
https://user-images.githubusercontent.com/26223174/185576883-e3b13c72-0009-445e-8c26-aad3292699a7.mp4

https://user-images.githubusercontent.com/26223174/185576893-b20efa37-a983-4759-872f-11f15d58177e.mp4

- [ ] Seeing the build logs of Environments requires a page refresh. This shouldn't be a requirement of course.

----

https://user-images.githubusercontent.com/26223174/185577941-fcdab0d1-a765-4fd4-bdab-9f901e2b7f8d.mp4

- [ ] The Environment snackbar in the editor still isn't correct for me after deleting an Environment and requires a page refresh.

----
image

- [ ] Not sure whether the correct icon is shown on Environment build failure.

----
image

Response from `/catch/api-proxy/api/validations/environments` of 
```
{"validation": "fail", "fail": ["103099ae-c8b5-4077-ac90-60a0508ec69c"], "actions": ["RETRY"], "pass": []}
```
isn't handled correctly.

To reproduce:
1. Create a new project
2. Immediately cancel the Environment build
3. So now now no Environments have been build (as the snackbar correctly indicates)
4. Add `exit 1` to the top of the setup script
5. Build the Environment
6. The build will fail
7. Go to the pipeline editor and see the incorrect snackbar.

ncspost:
Great work @iannbing - this is so clean

Few nitpicks:
- [ ] Can't tab out when logs box is in focus (use esc to exit focus)
- [ ] Currently no way to tab into env list (this should happen after new env button)
- [ ] Add accordion chevron to properties
- [ ] Reduce size of accordion chevrons to 16px
- [ ] Add side padding at smaller screen sizes
CleanShot 2022-08-19 at 10 24 00@2x

Great work!

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
```diff
@@ -0,0 +1,433 @@
...
+          return (
+            <>
+              {line}
+              <br />
+            </>
+          );
+        } else {
+          return line;
+        }
+      })}
```
mausworks, line 29: Don't forget those keys!

```suggestion
      {linesArr.map((line, index) => {
        if (index !== linesArr.length - 1) {
          return (
            <React.Fragment key={index}>
              {line}
              <br />
            </>
          );
        } else {
          return <React.Fragment key={index}>line</React.Fragment>;
        }
      })}
```

services/orchest-webserver/client/src/api/environments/environmentsApi.ts
mausworks: Much better, thanks! :)

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
iannbing: AFAIK, Fragments don't need a key, as they are not actually rendered. It also didn't throw warning without the keys. Could you verify?

iannbing:
> Currently no way to tab into env list (this should happen after new env button)

@ncspost  Normally tab order should go with the HTML elements structure (like z-index, we shouldn't change it unless there's no other way). Currently (in the latest commit), you can tab into all elements in this view. Note that when you tab into env list (it's right next to Creat Environment button), you need to press UP / DOWN to change the selected item.

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
mausworks: So—I'll have a look to verify, but AFAIK React will always throw an error for anything that doesn't have a key when you render an array, regardless if you have a `div` or `React.Fragment`. This is because ReactDOM will not be able to calculate what has to be re-rendered unless these keys are included—it's a performance optimization more than anything.

IMHO keys should always be included when rendering an array, it might just be pragmatism, I'll try to verify. :+1:

yannickperrenet (approved):
Haven't done a code review but I did test the application. Given that we will be extensively testing all changes on the big PR (`feat/ui-overhaul`), I would be fine to merge the PR instead of doing more tests on this PR alone.

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
iannbing: Added.

ncspost:
Great work @iannbing - the only thing small nitpick I have left is when tabbed into env list, the first item of the env list should be active (ie when I press down, it moves to second item in the list)

services/orchest-webserver/client/src/contexts/GlobalContext.tsx
mausworks: Just to verify, here's a minimal example using `React.Fragment`: 

https://codesandbox.io/s/intelligent-surf-8wwmkt?file=/src/App.js

PR #1182, opened 2022-08-08
Detect k8s distro and add ingress if needed

nhaghighat:
## Description

This PR detects the k8s distro and enables ingress if needed.

Fixes: #issue

## Checklist


services/orchest-controller/pkg/addons/addons.go
```diff
@@ -13,21 +14,42 @@ var (
...
 	ArgoWorkflow   = "argo-workflow"
 	DockerRegistry = "docker-registry"
+	IngressNginx   = "ingress-nginx"
 )
 
 type AddonsConfig struct {
 
 	// The list of addons to disable
-	DefaultAddons []string
+	Addons []string
```
yannickperrenet, line 23: I guess this comment should read `enable` instead of `disable`?

yannickperrenet: If it should be `disable` then I think this should be reflected in the name of options as otherwise it is pretty confusing ^_^

yannickperrenet (changes requested):
I thought that the following spec was to be implemented:

Seems like we can go with (similar to `--no-argo`) `--nginx/--no-nginx` where `--nginx` is the default: 

* `--nginx`: let orchest-controller detect the k8s distro and 
    * install if distro requires install AND nginx-ingress-controller not yet installed (as detected)
    * don't install if nginx-ingress-controller is already installed 
        * Note: Are we sure Orchest will work correctly if users have installed the nginx-ingress-controller? Are there certain config options that would prevent Orchest from working correctly?
    * don't install if distro requires us to not install
        * Note: we need to make sure that the installation widget correctly mentions to install the nginx-ingress-controller on k8s distros where we can't install it.

* `--no-nginx`: don't install nginx-ingress-controller

Which would mean that in k3s, GKE, EKS the default would be `--nginx`, and if already installed we can use that one.

However, I don't see anything of this in the code (and no CLI changes). Did we change the spec or is the PR simply not finished yet?

services/orchest-controller/pkg/addons/addons.go
nhaghighat: it should be enable indeed

nhaghighat:
> --nginx: let orchest-controller detect the k8s distro and ..

I'm not sure if we can follow the same Argo model, with attributes «--nginx/-no-nginx». 
First of all, these are the attributes of OrchestCluster CRD not the attributes of the whole k8s cluster because we cannot have several nginx ingress per cluster (well, we can have several Nginx installations, but then it gets more complex) but in Argo's case, we could have had several Argo installations per cluster. 
Secondly, for some platforms such as Minikube and Microk8s, the nginx ingress controller that we install is not accessible as they do not support the LoadBalancer service type.
So I think the best way to proceed is for the controller to detect if the installation of nginx is necessary per platform and if yes deploys it.

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -456,3 +456,44 @@ func detectContainerRuntime(ctx context.Context,
 
 	return runtime, runtimeSocket, nil
 }
+
+func isIngressIfRequired(ctx context.Context, client kubernetes.Interface) bool {
```
yannickperrenet, line 460:
```suggestion
func isIngressRequired(ctx context.Context, client kubernetes.Interface) bool {
```

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -456,3 +456,44 @@ func detectContainerRuntime(ctx context.Context,
...
+
+	if !isNginxIngressInstalled(ctx, client) {
+		return false
+	}
```
yannickperrenet, line 464: Shouldn't this be `if isNginxIngressInstalled` instead of `if !...`?

yannickperrenet:
I guess you are still planning to do the CLI changes in order to add the `--nginx/--no-nginx` flag?

docs/source/_static/js/installer_widget.js
```diff
@@ -55,6 +55,26 @@ your custom Environments to the MicroK8s node.
...
+        instructions: `
+<p>First, deploy <a href="https://k3s.io/">K3s</a> without having traefik installed by running
+the following command:</p>
+<div class="highlight">
+<pre>
+curl -sfL https://get.k3s.io | \\
+    INSTALL_K3S_EXEC="--no-deploy traefik" sh -s -
+</pre>
+</div>
+<p>Now the Orchest can be installed using the <code class="docutils literal notranslate"><span
```
yannickperrenet, line 68:
```suggestion
<p>Now Orchest can be installed using the <code class="docutils literal notranslate"><span
```

docs/source/_static/js/installer_widget.js
```diff
@@ -55,6 +55,26 @@ your custom Environments to the MicroK8s node.
...
+class="pre">orchest-cli</span></code>:</p>
+<div class="highlight">
+<pre>
+pip install --upgrade orchest-cli
+orchest install
+</pre>
+</div>
+<p>Now that Orchest is installed, it can be reached on K3s node IP address.</p>`,
```
yannickperrenet, line 76:
```suggestion
<p>Now that Orchest is installed, it can be reached on the IP address of the K3s node.</p>`,
```

In addition, let's also add the command to get the IP address (just like how we say `minikube ip` for the minikube installation instructions.

docs/source/_static/js/installer_widget.js
```diff
@@ -55,6 +55,26 @@ your custom Environments to the MicroK8s node.
 directly.</p>
           `,
       },
+      K3s: {
+        instructions: `
```
yannickperrenet, line 59: These `K3s` instruction should be added to `Windows` as well since they should work using WSL2.

docs/source/_static/js/installer_widget.js
```diff
@@ -55,6 +55,26 @@ your custom Environments to the MicroK8s node.
...
+<pre>
+pip install --upgrade orchest-cli
+orchest install
+</pre>
+</div>
+<p>Now that Orchest is installed, it can be reached on K3s node IP address.</p>`,
+      },
     },
     macOS: {
       minikube: {
```
yannickperrenet, line 80: Wasn't one of the main reasons to add the automatic ingress installation to make the installation producedure easier, i.e. users no longer having to install NGINX themselves anymore? With that being said, for EKS, GKE and such the lines like  _"Next, install the NGINX ingress controller using kubectl."_ should be removed from the installation instructions.

Or am I missing something?

docs/source/_static/js/installer_widget.js
```diff
@@ -55,6 +55,26 @@ your custom Environments to the MicroK8s node.
 directly.</p>
           `,
       },
+      K3s: {
+        instructions: `
+<p>First, deploy <a href="https://k3s.io/">K3s</a> without having traefik installed by running
```
yannickperrenet, line 60:
```suggestion
<p>First, install <a href="https://k3s.io/">K3s</a> without deploying traefik by running
```

Might be even better to make traefik into `traefik` as well (or a hyperlink).

nhaghighat:
> I guess you are still planning to do the CLI changes in order to add the `--nginx/--no-nginx` flag?

Added

docs/source/_static/js/installer_widget.js
nhaghighat: removed the extra parts from GKE and EKS installation docs

yannickperrenet:
I tried to install Orchest on Minikube to see whether that worked, but sadly it didn't:
```
❯ orchest install --dev

Installing the Orchest Controller to manage the Orchest Cluster...
Failed to install Orchest.
Could not create the required namespaced custom object.
```

If I remove:
```python
        applications.append(
            {
                "name": "ingress-nginx",
            }
        )
```
then it works again. So that needs to be fixed before we can merge.

orchest-cli/orchestcli/cmds.py
```diff
@@ -312,9 +312,18 @@ def install(
...
-        metadata["annotations"] = {"orchest.io/container-runtime-socket": socket_path}
+        metadata["annotations"]["orchest.io/container-runtime-socket"] = socket_path
+
+    if no_nginx:
+        echo(
+            "Disabling 'Nginx Ingress Controller' installation."
+            "\n\tMake sure 'Nginx Ingress Controller' is already installed "
+            "in your cluster"
+        )
+        metadata["annotations"]["controller.orchest.io/deploy-ingress"] = "false"
```
yannickperrenet, line 326: Why did you choose to go with the annotations approach instead of correcting the applications approach? Given that Argo and Nginx are both modeled as addons, I think they should both be modeled as "applications" as well. Feels a bit strange to model them differently imo.

What do you think?

nhaghighat: Well, With the annotation approach, the controller can decide whether to add the ingress to the CR applications or not. for example with the annotation approach, if the user installs the orchest with Nginx, but Nginx is not required, then the controller can ignore installing Nginx and successfully install the orchest, but if we use the application, then the orchest-controller has no way to install Nginx whether it is needed or not.

yannickperrenet: So basically what you are saying is:

* orchest-controller can't ignore an application (in the CR) if it is specified.

But in the implementation you just add a "switch" to the CR in the form of an annotation, i.e. `disable-ingress` instead of having to specify an `enable`. We can just change the applications section of the CRD to allow that as well, e.g. adding `enabled: boolean`. Similar to: 

https://github.com/orchest/orchest/blob/799df1c267e225f287f5dbd776896287ab7464e6/services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml#L272-L274

yannickperrenet: Hmm makes sense. So the logic becomes:
* `--no-nginx` --> CLI adds `deploy-ingress = false` annotation
* `--nginx` --> CLI doesn't add annotation --> orchest-controller detects whether it should install nginx --> adds appropriate annotation to CR

So I guess that it is more of a convention that the controller is allowed to change the annotations section (and status) section of the CR object, but not the spec?

nhaghighat: Indeed, it is more common that the controller makes changes to the annotation section.
our controller `still` makes changes to the spec, but it does not remove any thing, it changes the `spec` to make the CR more declarative, like adding the version if not specified.

PR #1184, opened 2022-08-09
Improve race condition handling in image builds errors

fruttasecca:
## Description

Improves the way errors are handled during image builds by playing around an observed race condition where the container/pod status does not reflect the failure state yet. Essentially we check another time after the pod has reached an end state.

## Checklist


ricklamers:
LGTM! Too bad the error checking code has to catch all these cases. Don’t see a way around that either, though.

fruttasecca:
>LGTM! Too bad the error checking code has to catch all these cases. Don’t see a way around that either, though.

Indeed not being able to rely on the container state after exiting the `watch` stream is quite a bummer :/

Issue #1181, opened 2022-08-08
Add "Run now" option to jobs with a recurring schedule


ricklamers:
**Describe the feature you'd love to have in Orchest**
Add a "Run now" button to jobs with a recurring schedule.

**Is your feature request related to a problem? Please describe.**
A job might be using a specific snapshot of the code and have a specific environment variable/parameter configuration. To make it easy to test if everything is working correctly (especially for idempotent pipelines that can just run multiple times) it would be great if you can just kick of a single pipeline run in an existing job in the job detail view.

Credit for this feature goes to the team at LinkThings who've had great feedback to date. Thanks guys!

PR #1186, opened 2022-08-09
Add endpoint to trigger a job run "manually"

fruttasecca:
## Description

Adds an endpoint (and later some FE changes) to allow triggering a job run (a batch of pipeline runs) for jobs that are `PENDING|STARTED|PAUSED`. 

Fixes: #1181 

## Checklist

@yannickperrenet @ricklamers 
The BE is not very strict about requirements to trigger a job run, but I'm not sure about exposing a button for one-off.  jobs that are STARTED. Maybe it should be shown only for cronjobs which are `STARTED|PAUSED` and `PENDING` one-off jobs (one-off jobs scheduled in the future?). Also happy to make the BE more picky about the required state

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_jobs.py
+++ b/services/orchest-api/app/app/apis/namespace_jobs.py
@@ -545,12 +545,12 @@ class JobRunTrigger(Resource):
     def post(self, job_uuid: str):
         """Triggers a batch of runs for a non end state job.
 
-        The job should be in a PENDING|STARTED|PAUSED state for a batch
-        of runs to be triggered. With batch of runs we mean a number of
-        runs equal to the "pipeline runs" that would be setup through
-        the job parameterization. For example, for a cronjob, this would
-        be as if the job run was performed because of its scheduled
-        time.
+        The job should either be a PENDING|STARTED cronjob or a PENDING
+        one-off job scheduled in the future for a batch of runs to be
+        triggered. With batch of runs we mean a number of runs equal to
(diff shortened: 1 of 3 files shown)


fruttasecca:
How it looks for a cronjob
image

And one-off job pending job

image

fruttasecca:
> Trigger schedules runs made me "scared" in the sense that I expected the recurring job to no longer execute on the next hit of the schedule.

I get a similar vibe from the text, i.e. that I'm about to run what I had scheduled *instead* of an extra job run

@ricklamers @ncspost

ricklamers:
I think I prefer `Trigger job` to `Trigger scheduled runs`. Semantically 'triggering a job' is also what's happening. You can trigger a job, and it will spawn a bunch of pipeline runs. You trigger _the job_, not the runs.

ricklamers:
I'm open to `Trigger job now` to make it clearer that it's a quick and dirty "one-off" action.

fruttasecca:
image
image

yannickperrenet (approved):
:+1: 

(cc @iannbing @mausworks -- wanted to make you two aware of these FE changes given the UX rework)

marrocksd:
hi, @fruttasecca are we able to call this endpoint externally (from Postman for example)? If so, how do we generate the token? It would open much more capability for integrations

ricklamers:
@marrocksd adding a public REST API is on our roadmap: https://github.com/orchest/orchest/issues/201

Good to know you'd find this a valuable feature!

PR #1203, opened 2022-08-15
Remove Argo Server

yannickperrenet:
## Description

Removes Argo Server (we have never used but never disabled it) and `userdir` from the top-level repo directory (we dynamically create the structure in the app now).

## Checklist


fruttasecca (approved):
:+1:

PR #1205, opened 2022-08-17
Allow using custom image for orchest components

nhaghighat:
## Description

This PR allows specifying a custom image for orchest components. if a user provides a custom image for a component, then the controller won't change the image back to the desired one and won't update it during an update if the version is changed in OrchestCluster CRD.

custom image means: it does not use the `registry` specified in `OrchestCluster` CRD or it uses a different name for components (for example for `orchest-api`, the controller expects the name to be `orchest/orchest-api`, or the version does not follow CalVer)

Fixes: #issue

## Checklist


services/orchest-controller/pkg/controller/orchestcluster/cluster_utils_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			result := isCalverVersion(test.version)
+			assert.Equal(t, result, test.result)
+		})
+	}
+}
+
+func TestPatseImageName(t *testing.T) {
```
yannickperrenet, line 49: Typo:
```suggestion
func TestParseImageName(t *testing.T) {
```

yannickperrenet (approved):
Works like a charm! Tested by making changes to the `orchest-webserver` and creating a new `OrchestCluster` that specifies a the custom image for the `orchestWebServer`; the custom image was picked up as expected.

I've also tested `orchest update` and it correctly updated all non-custom images whereas all custom images were left unchanged.

:fire:

PR #1212, opened 2022-08-22
New UI design part 5: Jobs

iannbing:
## Description

The new design of Jobs view.


services/orchest-webserver/client/src/types.d.ts
```diff
@@ -10,13 +10,7 @@ declare module "react" {
   }
 }
 
-export type Json =
-  | string
-  | number
-  | boolean
-  | null
```
mausworks, line 17: Why was `null` removed here? It's valid JSON. :thinking:

yannickperrenet:
image

- [ ] Empty state message/image is not centered correctly (ignore the blue bars, they mistakenly got into the screenshot).

----

- [ ] Similar to https://github.com/orchest/orchest/pull/1060#issuecomment-1217702785, could we add an arrow (in empty state) that points to the `+ new job` button? I get so drawn into the empty state that I completely miss the button in the top-left of my screen :grimacing: 

----

image

- [ ] Incorrect text on buttons when trying to delete a job.

----

image

- [ ] In my opinion, there is slightly to much space between the hint (i.e. "Every minute") and the Cron expression box. Given that it is just helper text, I think they can/should be closer together.

----

image

- [ ] I don't think it should be possible to not `Include` the parameterless run. Because then the job won't actually do anything. It won't run. 

~~How about even removing this `Pipeline runs` table in the case where no parameters are defined?~~ --> I guess this would increase scope since someone could also `load parameter file` which would then suddenly render the table.

----

image

- [ ] I think `pause scheduled job` is confusing, e.g.: Do I pause the job that is currently running? How can I even pause a job that still needs to be scheduled? Would the job afterwards run again? --> Instead we could just name it `pause job`.

----

https://user-images.githubusercontent.com/26223174/189628194-84cf5daa-29a3-43a3-af60-68e64469ed17.mp4

- [ ] Seems like the `auto-cleanup` toggle state is not remembered when `copy job configuration`.

----

image

- [ ] Incorrect icon when a job is paused. When the job is paused the same icon as `pause job` should be used instead of the icon of the action to restart it again.

----

https://user-images.githubusercontent.com/26223174/189629083-367051c1-89a4-4857-8160-95f245d9dad4.mp4

- [ ] Incorrect "unsaved changes" dialog when I just discarded my changes.

iannbing:
> Similar to https://github.com/orchest/orchest/pull/1060#issuecomment-1217702785, could we add an arrow (in empty state) that points to the + new job button? I get so drawn into the empty state that I completely miss the button in the top-left of my screen

@yannickperrenet @ncspost  This "drawing an arrow pointing at a certain button" is a deep rabbit hole when it comes to responsiveness. Most apps use an active Tooltip (i.e. triggered when conditions are met, for example, when there's no step / pipeline) to guide user. Is it something to consider?

ncspost:
> > Similar to [#1060 (comment)](https://github.com/orchest/orchest/pull/1060#issuecomment-1217702785), could we add an arrow (in empty state) that points to the + new job button? I get so drawn into the empty state that I completely miss the button in the top-left of my screen
> 
> @yannickperrenet @ncspost This "drawing an arrow pointing at a certain button" is a deep rabbit hole when it comes to responsiveness. Most apps use an active Tooltip (i.e. triggered when conditions are met, for example, when there's no step / pipeline) to guide user. Is it something to consider?

I'm happy to leave it for now and then implement a solution in https://linear.app/orchest/issue/ORC-620/adoption-tutorial-walkthrough-tooltips

iannbing:
After a short discussion, I'll merge this one directly to #1060 so that we can test and review in one PR.

Issue #855, opened 2022-04-01
Pre-pull the base environment image to the container builder PVC cache


ricklamers:
**Describe the problem this improvement solves**
The first environment build can be slow.

**Describe the solution you'd like**
Pre-pull the base environment image to the container builder PVC cache.

PR #1213, opened 2022-08-22
Change image distribution model

fruttasecca:
## Description

This is a bit of a WIP PR that sets up some needed elements to build upon.

Notable changes:
- new tables to keep track in which nodes jupyter and env images are stored
- when building a jupyter/env image the build process will keep track of the node of build
- improved filtering of active images to filter based on registry presence and node
- a new component of the `node-agent`, `image_pusher.py`, which takes care of pushing images to the registry when it finds an image on the node that is not on the registry
- both jupyter and env images are not labelled as inactive anymore for the time being since the logic requires the `digest` of the image, grabbing that is for another PR

Starting with this PR images are built directly on the node, the `image-pusher` then pushes new images to the registry. The way this works in the following:
- an image is built on the node. At this point, out of N nodes, only 1 has the image, and the registry doesn't.
- the image is pushed to the registry by the `node-pusher` when finding the image on the node
- the `image_puller` of the other N-1 nodes will pull the image from the registry

There is still the ongoing pain point that the split of jupyter and env images leads to the code being less DRY than desirable :/.

**Most of the new lines from the PR actually come from schema migrations, don't be scared :p**

Fixes: #855


services/orchest-api/app/app/models/_core.py
```diff
@@ -1122,3 +1122,47 @@ class ClusterNode(BaseModel):
 
     # https://kubernetes.io/docs/concepts/architecture/nodes/#node-name-uniqueness
     name = db.Column(db.String(), primary_key=True)
+
+
+class EnvironmentImageInNode(BaseModel):
+    """To track where are environment image is stored."""
```
yannickperrenet, line 1153:
```suggestion
    """To track where an environment image is stored."""
```

services/orchest-api/app/app/models/_core.py
```diff
@@ -1122,3 +1122,47 @@ class ClusterNode(BaseModel):
 
     # https://kubernetes.io/docs/concepts/architecture/nodes/#node-name-uniqueness
     name = db.Column(db.String(), primary_key=True)
+
+
+class EnvironmentImageInNode(BaseModel):
```
yannickperrenet, line 1152: I would argue that an `EnvironmentImage` is "on" a node, not "in".

services/orchest-api/app/app/core/environment_image_builds.py
```diff
@@ -363,28 +367,28 @@ def build_environment_image_task(
...
             raise e
         finally:
-            # We get here either because the task was successful or was
-            # aborted, in any case, delete the workflows.
-            # k8s_custom_obj_api.delete_namespaced_custom_object(
-            #     "argoproj.io",
-            #     "v1alpha1",
-            #     _config.ORCHEST_NAMESPACE,
-            #     "workflows",
-            #     f"image-cache-task-{task_uuid}",
```
yannickperrenet, line 386: Commenting here because GitHub doesn't allow me to comment on an unchanged line (but trying to refer to a few lines lower here).

Shouldn't the `f"image-build-task-{task_uuid}"` also use the newly introduced function from https://github.com/orchest/orchest/pull/1213/commits/ddff2fd64b18e2c5cc1f88bde17fc920f79c17ba#diff-db9b705b8165924d70ec6c98920cdd8dfc6c198a54428f8a20f4ae2cdbabffbcR98? (`def image_build_task_to_pod_name(task_uuid: str)`)

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
```diff
@@ -147,7 +152,7 @@ def put(self, project_uuid, environment_uuid, image_tag):
                 return
 
             if status_update["status"] == "SUCCESS":
-                digest = registry.get_manifest_digest(
+                _ = registry.get_manifest_digest(
```
yannickperrenet, line 155: This might as well be removed right?

yannickperrenet: And I guess `"SUCCESS"` now means: "build was successful, but image still needs to be pushed to registry"?

services/orchest-api/app/app/apis/namespace_environment_images.py
```diff
@@ -59,7 +61,14 @@ class ActiveEnvironmentImages(Resource):
...
         """Gets the list of environment images to keep on nodes."""
-        active_env_images = environments.get_active_environment_images()
+
+        active_env_images = environments.get_active_environment_images(
+            stored_in_registry=request.args.get(
+                "stored_in_registry", default=None, type=lambda v: v in ["True", "true"]
+            ),
+            in_node=request.args.get("in_node"),
+        )
+
```
yannickperrenet, line 71: Will probably be addressed in a future commit, but I didn't want to forget about the following case:

* The `node-agent` will poll this endpoint to determine active images, which will obviously not include newly build images. So the `node-agent` shouldn't end up deleting those images.

services/node-agent/app/container_runtime.py
```diff
@@ -183,7 +189,31 @@ async def delete_image(self, image_name: str) -> bool:
...
+                f"i push {image_name} --skip-verify "
+            )
+            success, stdout, stderr = await self.execute_cmd(cmd=cmd)
+            if not success:
+                raise ImagePushError(f"STDOUT: {stdout}\nSTDERR: {stderr}")
+        # ENV_PERF_TODO:
+        # Cover case of docker failing to push because of missing
+        # certificates , see the edge case in pull_image.sh. This will
+        # likely involve using a third party tool like buildah to push
+        # things separately.
```
yannickperrenet, line 218: Good one! :+1:

services/orchest-api/app/app/apis/namespace_environment_images.py
fruttasecca: Indeed, it's part of the todos

services/node-agent/app/image_pusher.py
```diff
@@ -0,0 +1,135 @@
...
+        logger.warning(
+            "Env variable CLUSTER_NODE is not set. Querying more images than necessary."
+        )
+    else:
+        endpoint += f"&in_node={node}"
+
+    async with session.get(endpoint) as response:
+        response_json = await response.json()
+        active_images = response_json["active_environment_images"]
+    logger.info(f"Found the following active env images after to push: {active_images}")
```
yannickperrenet, line 48:
```suggestion
    logger.info(f"Found the following active env images to push: {active_images}")
```

services/orchest-api/app/app/apis/namespace_environment_image_builds.py
fruttasecca: Indeed it can be removed, this PR has a bunch of WIP things that are to be covered when dealing with digests in the next PR

services/node-agent/app/image_pusher.py
```diff
@@ -0,0 +1,135 @@
...
+    pushed to the registry as images to push.
+    """
+    endpoint = (
+        "http://orchest-api/api/environment-images/active?stored_in_registry=false"
+    )
+    node = os.getenv("CLUSTER_NODE")
+    if node is None:
+        logger.warning(
+            "Env variable CLUSTER_NODE is not set. Querying more images than necessary."
+        )
```
yannickperrenet, line 41: If this case gets hit and we are running in a multi-node context, then the node-agent will try to push images that aren't actually on the node.

For example, `orchest-api` knows about nodes X and Y and Environment E1 is build on Y. Then the node-agent on X will try to push E1 even though it doesn't have it. 

Not saying the implementation should change, but might be worth it to slightly change the logging message (given that the environment variable `CLUSTER_NODE` should always be set).

services/orchest-api/app/app/apis/namespace_ctl.py
```diff
@@ -69,10 +69,11 @@ def get(self):
...
         pre_pull_orchest_images = [
             CONFIG_CLASS.IMAGE_BUILDER_IMAGE,
-            f"orchest/jupyter-enterprise-gateway:{CONFIG_CLASS.ORCHEST_VERSION}",
-            f"orchest/session-sidecar:{CONFIG_CLASS.ORCHEST_VERSION}",
-            f"docker.io/orchest/jupyter-server:{CONFIG_CLASS.ORCHEST_VERSION}",
             _config.CONTAINER_RUNTIME_IMAGE,
+            f"docker.io/orchest/jupyter-server:{CONFIG_CLASS.ORCHEST_VERSION}",
+            f"docker.io/orchest/base-kernel-py:{CONFIG_CLASS.ORCHEST_VERSION}",
+            f"docker.io/orchest/jupyter-enterprise-gateway:{CONFIG_CLASS.ORCHEST_VERSION}",  # noqa
+            f"docker.io/orchest/session-sidecar:{CONFIG_CLASS.ORCHEST_VERSION}",
```
yannickperrenet, line 76: Do we actually clean up these images in the `node-agent` if Orchest is updated?

services/node-agent/app/image_pusher.py
fruttasecca: Agree, now that I think about it, it might be worth it to actually fail if the `CLUSTER_NODE` env var is not there to avoid confusion, i.e. let's be explicit about it not being optional.

yannickperrenet:
So far looks good to me! Might be a good idea to put the PR in draft until the `digest` PR is merged into this branch. Moreover, can I find the todos somewhere (e.g. https://github.com/orchest/orchest/pull/1213#discussion_r952328140) so that I can be sure they are addressed? For example putting this in the PR description.

fruttasecca:
>So far looks good to me! Might be a good idea to put the PR in draft until the digest PR is merged into this branch

Sure, sounds good

>Moreover, can I find the todos somewhere (e.g. https://github.com/orchest/orchest/pull/1213#discussion_r952328140) so that I can be sure they are addressed? For example putting this in the PR description.

Those are currently in our internal docs at https://www.notion.so/Back-end-implementation-doc-dfda527e04b2457ba5e55a5e22744928, I'll put the ones related to the image distribution model in the PR description

services/orchest-api/app/app/apis/namespace_ctl.py
fruttasecca: Yes we do

fruttasecca:
Fixes: #855

PR #1214, opened 2022-08-22
Improve pipeline run performance on single-node Orchest deployments

yannickperrenet:
## Description

Improves the performance of pipeline runs by executing all containers (of every Step) in the same Kubernetes Pod. This way users don't have to wait for a Pod to be allocated for every Step (which can take a considerable amount of time, e.g. for the Orchest quickstart pipeline this takes 30s out of the total 60s to execute the pipeline on my machine). 

Note: These changes only apply to the single-node case, for multi-node deployments the previous implementation is(/should be) left unchanged with the exception of updating the Step status based on Kubernetes Pod status changes.

## How Argo `containerSet` works

> A container set actually starts all containers, and the Emissary only starts the main container process when the containers it depends on have completed." 

See: https://argoproj.github.io/argo-workflows/container-set-template/#resource-requests:

- I don't think resource requests are an issue on a single-node deployment given that the node already has certain hardware capabilities.

-> status of all containers gets set to Running when the pipeline run starts. Which we fix with some custom logic.

## How to test

When running using the setup of our development docs then you will be running on a **single-node** setup and thus you can test whether the single-node setup works.

For testing **multi-node**, you could update the `SINGLE_NODE` environment variable in the `celery-worker` deployment to make Orchest believe to run the pipelines in a multi-node way.

## Example executing the Quickstart pipeline on single-node

https://user-images.githubusercontent.com/26223174/185914570-15b5a1df-6baf-4f34-b3d9-98b97ca41b61.mp4

## Example executing the Quickstart pipeline on multi-node

https://user-images.githubusercontent.com/26223174/185931824-11b32950-5a60-43d1-80eb-2cd90688d40c.mp4

Note: Steps are now left in Pending (instead of Running) until the underlying Kubernetes Pod (that runs the Step container) is started.

## Checklist

  - [x] Job runs
  - [x] multi-node still works as expected?
  - [x] Cover requirement of the Emissary runne in the install docs (--no-argo), because then the user needs to run Argo with Emissary (on single-node) as well.
- [x] Can we get the `step_uuid` without having to use the `displayName` in the single-node setting to increase robustness? (as per [comment](https://github.com/orchest/orchest/pull/1214#discussion_r951517241)) --> I've asked the question in their Slack channel. --> We decided the PR can be released whilst we wait for an answer as the solution works fine for now.
- [x] Issues with image pull policy of argo init containers? (as per https://github.com/orchest/orchest/pull/1214#issuecomment-1223723360)
- [x] ~~Small rewrite instead of the `if` branching logic?~~ --> Given that a `containerSet` isn't allowed to link to other templates we can't really merge the `dag` and `containerSet` logic and thus would end up with this split case of single and multi-node anyways. Thus I don't really see the benefit.


services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
```diff
@@ -265,7 +265,11 @@ executor:
   # -- Resource limits and requests for the Workflow Executors
   resources: {}
   # -- Adds environment variables for the executor.
-  env: {}
+  env:
+    - name: RESOURCE_STATE_CHECK_INTERVAL
+      value: "1s"
+    - name: WAIT_CONTAINER_STATUS_CHECK_INTERVAL
+      value: "1s"
```
fruttasecca, line 272: How impactful is the default value here if users were to bring their own argo? Might be worth adding a note about this in the docs.

I think that "Environment variables are typically added to test out experimental features and should not be used by most users. Environment variables may be removed at any time." at [link](https://argoproj.github.io/argo-workflows/environment-variables/) is good to keep in mind, although not sure if we should add more overhead to our process for the time being. I guess we control the argo version anyway (I don't remember)

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -705,24 +707,45 @@ async def run_pipeline_workflow(
...
-                            step_uuid = param["value"]
-                            break
+                    if CONFIG_CLASS.SINGLE_NODE:
+                        if step.get("type", "") != "Container":
+                            continue
+
+                        # Argo doesn't allow to work with templates for
+                        # a containerSet. Thus we fall back to the name
+                        # we gave to steps, which includes its uuid.
+                        step_uuid = step.get("displayName")
```
fruttasecca: Let's make sure we won't meet some important limitations (wouldn't say it's important as of now, [example](https://github.com/argoproj/argo-workflows/issues/8937)) or argo fiddling with the `displayName`. I googled a bit and didn't find anything

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -155,9 +155,19 @@ def __init__(
         # Pipeline methods.
         self._children: List["PipelineStep"] = []
 
-        # Initial status is "PENDING".
+        # Used as a local copy for the state of the PipelineStep as part
+        # of a pipeline run, in order to correctly update step statuses
+        # in the DB.
         self._status: str = "PENDING"
```
fruttasecca, line 169: Was `_status` something that we forgot to cleanup during the migration to k8s?

I feel that by using the `Pipeline` class to store status we are crossing some sort of boundary where we go from using it as a definition to using it as a state, what's your opinion on that?

lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -42,7 +42,7 @@
 # Container Runtime configs.
 CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")
 CONTAINER_RUNTIME_IMAGE = os.environ.get("CONTAINER_RUNTIME_IMAGE")
-CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET")
+CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET", "")
```
fruttasecca, line 45: What's the reason for this? So far the codebase as treated the env var as something that must be there AFAIK

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -246,7 +256,8 @@ def to_dict(self) -> PipelineDefinition:
         description.update(self.properties)
         return description
 
-    # TODO: This is slow.
+    # TODO: This is slow, although reasonable for small/medium size
+    # pipelines.
```
fruttasecca, line 260: I guess that we can create a private lookup dict during initialization if we are worried about this (just trying to minimize the amount of TODOs across the codebase :p)

fruttasecca (approved):
lgtm aside from some minor considerations :)

I haven't manually tested it, will do so tomorrow morning

fruttasecca:
Might be worth it to slightly refactor the functions that create the pipeline workflow to avoid branching on `SINGLE_NODE` within their scope, i.e. splitting them into different functions

fruttasecca:
I see that there is a rather long start time of the pod/container when running in single node mode. There are some suspcious logs worth looking into
image

These pulls seem to always be happening

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
yannickperrenet: The thing is... the environment variables made a lot of sense to me from a documentation perspective ([link](https://argoproj.github.io/argo-workflows/environment-variables/)), however I didn't really see any impact on lowering the value to `1s` from the default value of `5s`. So I am a bit in doubt of what to do here :thinking:

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: Good point. I will take another look to see whether there is another way to get the step uuid.

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: I will change from `TODO` to `NOTE`. Don't think having a private lookup dict is going to make any difference here. :+1:

yannickperrenet: https://github.com/orchest/orchest/pull/1214/commits/ebad353ceb21e9180e83f8c8e6f7f596907d395c

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: > Was _status something that we forgot to cleanup during the migration to k8s?

Yes. So I ended up using it again.

> crossing some sort of boundary where we go from using it as a definition to using it as a state, what's your opinion on that?

Fair point. I didn't really see any way around doing so given that we want to check the status of parent steps (and thus need to know about the pipeline's structure/definition). I was first thinking about making a deepcopy so that the status wouldn't be changed by any other function on the same object, however, the class is only used in this module so I felt like it was okay not to do so. 

In the context of pipeline runs, pipeline steps can have a status imo. Although I do agree that currently the Pipeline class is more of a static definition. 

Not really sure what alternative there is besides renaming the classes slightly to better encapsulate what we use them for.

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
fruttasecca: >The thing is... the environment variables made a lot of sense to me from a documentation perspectiv

Indeed they do

>however I didn't really see any impact on lowering the value to 1s from the default value of 5s

What I meant is, what if a user deploys it's own argo where said env variables are at their default value, what's the impact? Will runs appear to be "slower" because of the longer interval?

services/orchest-api/app/app/core/pipelines.py
fruttasecca: Would using the `steps_to_finish` set combined with `step.parents` fix the issue? It's all state we already have

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: Wasn't able to find anything so I posted a question in their Slack channel. I'd say for now we can merge it as is and come back to it based on the reply.

services/orchest-api/app/app/core/pipelines.py
yannickperrenet: Smart! :nerd_face: -> https://github.com/orchest/orchest/pull/1214/commits/60d0d41210b9b3873b3ac8f4c543535d9e7132b1

services/orchest-api/app/app/core/pipelines.py
fruttasecca: Sounds good!

services/orchest-controller/deploy/thirdparty/argo-workflows/orchest-values.yaml
fruttasecca: Makes sense :)

yannickperrenet:
> These pulls seem to always be happening; this doesn't seem to occur on dev

Fixed in https://github.com/orchest/orchest/pull/1214/commits/6b1e9775042270f30ba0ab5e5dac0dc88130c9f0 (as per https://github.com/argoproj/argo-helm/blob/5112c7782423349707c1c13d3da973009c53d830/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml#L48). However, it does sound strange to me that this isn't happening on `dev` (but didn't feel the need to verify since this change will make sure it doesn't happen).

fruttasecca:
>Fixed in https://github.com/orchest/orchest/commit/6b1e9775042270f30ba0ab5e5dac0dc88130c9f0 (as per https://github.com/argoproj/argo-helm/blob/5112c7782423349707c1c13d3da973009c53d830/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml#L48). However, it does sound strange to me that this isn't happening on dev (but didn't feel the need to verify since this change will make sure it doesn't happen).

Great to hear. Indeed it not happening on `dev` is quite strange to me as well, I verified on a cloud instance and wasn't happening there either

yannickperrenet:
> Indeed it not happening on dev is quite strange to me as well, I verified on a cloud instance and wasn't happening there either

I think I found why. We do a `podSpecPatch` that sets the `imagePullPolicy` of the `"main"` container to `IfNotPresent`. This was only done in the multi-node case and thus explains why it didn't happen on `dev`.

fruttasecca (approved):
lgtm again, this is excluding the cpu resources changes that we have  reverted on `dev` and the related "We first have to better understand resource contention when running a containerSet."

Issue #450, opened 2021-08-30
Add types to Python codebase


ricklamers:
**Describe the solution you'd like**  
Add types to Python codebase

**What does your solution aim to solve?**  
Make it more explicit what arguments functions take and therefore make it easier to extend consume or refactor code.

ramgarg102:
Hi @ricklamers can I take up this issue? Could you also elaborate with just one example?

yannickperrenet:
Whilst working on this it is good to keep in mind the changes that are introduced in Python3.9, see [Pep 585](https://www.python.org/dev/peps/pep-0585/) (deprecating things like [typing.Iterable](https://docs.python.org/3/library/typing.html#typing.Iterable)).

For the time being we will stay on Python3.8 because of the base images we use for the `orchest-api`, `auth-server`, `orchest-webserver` and `update-server`: [tiangolo/meinheld-gunicorn-flask](https://hub.docker.com/r/tiangolo/meinheld-gunicorn-flask), which does not yet have a Python3.9 version available.

An example module where we already use type hints is [`analytics.py`](https://github.com/orchest/orchest/blob/master/services/orchest-webserver/app/app/analytics.py).

balotofi:
i still don't understand how you want this to be resolved

astrojuanlu:
Hi @balotofi! The idea is to extend gradual typing like this:

https://github.com/orchest/orchest/blob/b39ded5eca8e9ededf5e246de11e0c2581048beb/services/orchest-webserver/app/app/analytics.py#L131-L133

To the rest of the codebase.

One can start module by module, or package by package, rather than spending a lot of time working on a huge pull request.

sbarrios93:
@astrojuanlu I can help with this. Got a preference which part of the codebase should I focus on first?

Are we still sticking to typing as it was on Python 3.8 or should I use the standards of Python 3.9?

astrojuanlu:
Thanks @sbarrios93! We'll continue the conversation about #1216 on the PR itself.

About the Python version, I opened https://github.com/orchest/orchest/issues/1217.

PR #1216, opened 2022-08-23
Add type hints to auth server app

sbarrios93:
## Description

Added type hints to auth server app on modules `services/auth-server/app/app/views.py` and `services/auth-server/app/app/utils.py`

Related to: #450 

## Checklist


services/auth-server/app/app/views.py
```diff
@@ -85,43 +94,47 @@ def login_static(path):
         return serve_static_or_dev(path)
 
     @app.route("/login/admin", methods=["GET"])
-    def login_admin():
+    def login_admin() -> Union[Tuple[str, int], Response]:
```
astrojuanlu, line 97: We can do

```python
from __future__ import annotations
```

and

```suggestion
    def login_admin() -> Tuple[str, int] | Response:
```

Just checked locally:

```
$ docker run -it --rm python:3.8
Python 3.8.13 (default, Aug  2 2022, 11:41:59) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import Union
>>> def foo(a: Union[str, int]):
...   return a
... 
>>> def foo(a: str | int):
...   return a
... 
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'type'
>>> from __future__ import annotations
>>> def foo(a: str | int):
...   return a
... 
>>> foo(1)
1
```

astrojuanlu (commented):
Thanks for your contribution @sbarrios93! 🎉 I did not do an in-depth review, just left one comment about type union syntax, I think we can upgrade it even if Orchest runs (for now) on Python 3.8.

services/auth-server/app/app/views.py
yannickperrenet: Overall I think I like the explicit `Union` over the syntactic sugar `|`. But that is just my opinion  :man_shrugging:

yannickperrenet:
Thank you for the contribution @sbarrios93 :sunglasses:!

sbarrios93:
> Thank you for the contribution @sbarrios93 😎!

Happy to help! Looking forward to keep helping this project grow

services/auth-server/app/app/views.py
sbarrios93: I'm open to follow any convention the team decides. I think there's not much difference currently if we're sticking with Py3.8, because `typing` will still need to be used to import most of the library generics (i.e, `Dict`, `Tuple`, and so on...). I think the real value comes when we need to use `typing` for only a few modules like `Any` and `Literal`.

I think the team needs to also decide how to handle the import of the `typing` modules to keep the consistency in the code. Currently some modules are using `import typing as t` 

See: 
https://github.com/orchest/orchest/blob/a21bca9ac76401da2335827b7e4a7c88e1bcdec4/orchest-cli/orchestcli/cmds.py#L174-L176

astrojuanlu: We discussed a bit internally and decided for the new syntax, also because [it's recommended in the official docs](https://docs.python.org/3/library/typing.html#typing.Union):

> To define a union, use e.g. Union[int, str] or the shorthand int | str. Using that shorthand is recommended.

About the typing imports, I'm actually fine using the `t` alias. However, I propose we do that gradually in future PRs.

For now, @sbarrios93 please use the new union syntax in the files you already changed, and this PR will be good to go 🚀

astrojuanlu:
Opened an issue about verifying these static types as part of our CI https://github.com/orchest/orchest/issues/1222

astrojuanlu (approved):
LGTM 👍🏽

astrojuanlu:
Thanks for getting this started @sbarrios93! We took the liberty of finishing it :)

PR #1221, opened 2022-08-23
Change image digest management

fruttasecca:
## Description

Changes the way image digests are handled by Orchest. When building an image a new, last layer is created to ensure that the digest of the image is unique across all images built in Orchest. This allows us to simplify the way we handle images and their deletion since we don't need to get the digest of the image anymore, which would be an asynchronous operation given the fact that images are now built on the node and later pushed to the registry.

While we could drop the `JupyterImage.digest` column, we can't do that for `EnvironmentImage` in order to maintain backward compatibility and not break jobs in case an unused image and an image in use by a job have the same digest. I've tested a bit but would like to test more tomorrow given the changes to `_env_images_that_can_be_deleted`.
 
## Checklist


services/orchest-api/app/app/core/environment_image_builds.py
```diff
@@ -133,6 +144,10 @@ def write_environment_dockerfile(
         f"|| (echo {error_flag} && PRODUCE_AN_ERROR)"
     )
 
+    # Make it so that the digest of the produced image is unique.
+    statements.append(
+        f"RUN echo '{task_uuid}' | sudo tee /orchest/task_{task_uuid}.txt"
+    )
```
yannickperrenet, line 150: As to not create more files than necessary (e.g. getting confused by files when exec'ing into the container), isn't the `RUN echo '{task_uuid}'` enough to create a new layer?

yannickperrenet: From the Docker docs ([link](https://docs.docker.com/engine/reference/builder/#run)):

> The RUN instruction will execute any commands in a new layer on top of the current image and commit the results.

But probably a good idea to test it.

services/orchest-api/app/app/core/jupyter_image_builds.py
```diff
@@ -97,6 +108,10 @@ def write_jupyter_dockerfile(base_image, work_dir, bash_script, path):
     # on the process WORKDIR being the project directory.
     statements.append("WORKDIR /project-dir")
 
+    # Make it so that the digest of the produced image is unique.
+    statements.append(
+        f"RUN mkdir -p /orchest && echo '{task_uuid}' > /orchest/task_{task_uuid}.txt"
+    )
```
yannickperrenet, line 114: Same as https://github.com/orchest/orchest/pull/1221/commits/b043b7ad69b22d067fd423130bdb838c470881dd#r953662770

services/orchest-api/app/app/core/environments.py
```diff
@@ -335,14 +342,29 @@ def _env_images_that_can_be_deleted(
...
+        .join(models.JobInUseImage, models.EnvironmentImage.jobs_using_image)
+        .join(models.Job, models.Job.uuid == models.JobInUseImage.job_uuid)
+        .filter(models.Job.status.in_(["DRAFT", "PENDING", "STARTED", "PAUSED"]))
+    )
+    img_digests_in_use_by_jobs = (
         db.session.query(models.EnvironmentImage.digest)
+        # The digest is no longer necessary for images that have a
+        # guarantee of having a unique digest, for those images, we
+        # don't store their digest anymore.
+        .filter(models.EnvironmentImage.digest.is_not(None))
```
yannickperrenet, line 359: This is also a `# REMOVABLE_ON_BREAKING_CHANGE` right?

yannickperrenet (approved):
:+1: 

Great catch to prevent breaking existing jobs! :100: 

I guess for removal from registry will will continue using the pattern:
* Mark image for removal (using image name + tag)
* Run registry garbage collection
* Query registry for digest of image:tag
* Tell registry to delete that obtained digest

services/orchest-api/app/app/core/environment_image_builds.py
fruttasecca: I'll take a look, however we would have to make sure this runs across different container runtimes so that's why I went with the safest route

services/orchest-api/app/app/core/environments.py
fruttasecca: Indeed, haven't tagged it since any change removing the `Environment.digest` column would end up finding this line to remove, I'll tag it anyway

fruttasecca:
>I guess for removal from registry will will continue using the pattern:

Indeed

services/orchest-api/app/app/core/environment_image_builds.py
yannickperrenet: Good point. It is a total nitpick, so lets leave it as is so we can ship :+1:

PR #1224, opened 2022-08-24
Remove periodic builder cache cleanup task

fruttasecca:
## Description

Removes the periodic builder cache cleanup task since it won't be needed given the incoming changes to the way environments are built.

## Checklist


yannickperrenet (approved):
:mechanical_arm:

PR #1225, opened 2022-08-24
Remove builder cache pvc

fruttasecca:
## Description

Removes the builder cache PVC since it won't be used anymore after the changes to how we build environments.
I've tested both an install and an update, updates have been tested both with a `builder-cache-pvc-size` specified on install (which leads to the entry being in the CRD) and without.

## Checklist


orchest-cli/orchestcli/cli.py
```diff
@@ -205,9 +205,10 @@ def cli():
     "--builder-pvc-size",
     "builder_pvc_size",
     type=click.IntRange(min=5),
-    default=50,
+    default=None,
     show_default=True,
-    help="Size of the image builder volume claim in Gi.",
+    help="Deprecated and ignored. Size of the image builder volume claim in Gi.",
+    hidden=True,
```
yannickperrenet, line 211:
```suggestion
    help="(DEPRECATED) Size of the image builder volume claim in Gi.",
```

Two points why I think this suggestion makes more sense:
* `(DEPRECATED)` is similar to how `click` implements deprecation for `Commands` and will implement it for `Options` ([see PR](https://github.com/pallets/click/pull/2271)).
* I don't think we should hide the option given that it then will no longer show up in the `--help` menu. Otherwise the message is never shown that it the option is deprecated.

orchest-cli/orchestcli/cmds.py
```diff
@@ -174,13 +174,17 @@ def install(
...
+    builder_pvc_size: t.Optional[int],
     registry_pvc_size: int,
     **kwargs,
 ) -> None:
     """Installs Orchest."""
     ns, cluster_name = kwargs["namespace"], kwargs["cluster_name"]
 
+    if builder_pvc_size is not None:
+        # REMOVABLE_ON_BREAKING_CHANGE
+        echo("The 'builder_pvc_size' parameter is deprecated and ignored.")
```
yannickperrenet, line 186: How about moving this up to `cli.py`? That way the `builder_pvc_size` can be removed as an argument here. Deprecation of CLI arguments should be handled in `cli.py` imo.

fruttasecca: Will do

services/orchest-controller/pkg/apis/orchest/v1alpha1/types.go
```diff
@@ -76,14 +76,11 @@ type OrchestResourcesSpec struct {
...
-	ConfigDirVolumeSize string `json:"configDirVolumeSize,omitempty"`
-
-	// If specified, this components will be deployed provided image
-	BuilderCacheDirVolumeSize string `json:"builderCacheDirVolumeSize,omitempty"`
-
 	// The Storage class of user-dir/
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// Deprecated and ignored. TODO: remove it?
+	ConfigDirVolumeSize string `json:"configDirVolumeSize,omitempty"`
```
yannickperrenet, line 83: @nhaghighat How would we go about deprecating fields? As far as I understood, deleting a field is not possible in a backwards compatible way?

fruttasecca: I've actually removed `BuilderCacheDirVolumeSize` and it worked, what ended up happening is that, if the entry existed in the CRD, the entry got removed. I guess the reason is that this way one can update a CRD definition seamlessly?

yannickperrenet: Hmmm so you did something along the lines of:
1. Orchest is installed with `BuilderCacheDirVolumeSize` in the `OrchestCluster` CR Object.
2. Made these code changes where `BuilderCacheDirVolumeSize` was removed
3. Updated Orchest so that the new `orchest-controller` is used **and new CRD is applied**
4. The  `OrchestCluster` definition actually no longer lists `BuilderCacheDirVolumeSize` and Orchest continues to run smoothly

Pretty cool :thinking: (so a new CRD doesn't break existing CR Objects if an item is deleted).

yannickperrenet (approved):

services/orchest-controller/pkg/apis/orchest/v1alpha1/types.go
fruttasecca: Yeah, that's what I've experienced

orchest-cli/orchestcli/cmds.py
fruttasecca: See https://github.com/orchest/orchest/pull/1225/commits/96b0fad05d136f5b395046bd018baa5b72dc9408 and https://github.com/orchest/orchest/pull/1225/commits/e4f9769358741b27667e36aa5d7720549996053a

orchest-cli/orchestcli/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.1"
+__version__ = "0.6.1"
```
yannickperrenet, line 1: Shouldn't it be:
```suggestion
__version__ = "0.6.0"
```

PR #1227, opened 2022-08-24
Allow parallelism for environment and jupyter builds

fruttasecca:
## Description

The PR is pretty simple, although as discussed there is some tricky behaviour involved when it comes to garbage collecting the registry, I've optimistically implemented (but not tested) one of the proposals since it was simple enough.

## Checklist


yannickperrenet:
@fruttasecca We are still debating whether we want to go with this approach or a more "lock"ing like behavior, right?

fruttasecca:
Indeed, we are probably going with the locking behaviour

services/orchest-api/app/app/core/scheduler.py
```diff
@@ -1,7 +1,26 @@
...
+status. This is a form of technical debt but shouldn't lead to any
+issues since the scheduler is already gate keeping concurrent starts
+through the 'interval' value. One solution would be to not start a
+particular scheduler job if it's still 'RUNNING', which is something I
+haven't done because it can have very dangerous outcomes if an update to
+the status fails. Another alternative is to have a DB record for every
+instance of a scheduler job run, and to periodically cleanup the old
+ones. The introduction of the 'status' column was necessary because some
+functionality of the `orchest-api` requires knowing if one of these jobs
+(registry gc) is running or not to avoid a race condition.
```
fruttasecca, line 23: @yannickperrenet, keen to know your opinion, didn't went ahead with the change to not balloon the PR scope

docs/source/fundamentals/settings.md
```diff
@@ -30,7 +30,7 @@ Enables authentication. When enabled, Orchest will require a login. Create user
 
 : Integer between: `[1, 25]`.
 
-Controls the number of {term}`builds` that can be run in parallel. With builds we mean environment
```
yannickperrenet: Might be better to make it (I am slightly guessing syntax and references here): "the number of {term}`Environment` builds that can"

services/orchest-api/app/app/core/tasks.py
```diff
@@ -357,11 +357,18 @@ def registry_garbage_collection(self) -> None:
 
         if has_deleted_images or repositories_to_gc:
             registry.run_registry_garbage_collection(repositories_to_gc)
+        scheduler.notify_scheduled_job_done(
+            scheduler.SchedulerJobType.PROCESS_IMAGES_FOR_DELETION
+        )
```
yannickperrenet, line 404: I am slightly confused here, the `celery-worker` never accessed the DB before right? Why add it now?

If we let the `celery-worker` touch the DB then we might as well change how we do Pipeline Step updates.

services/orchest-api/app/app/core/scheduler.py
```diff
@@ -315,6 +335,8 @@ def schedule_job_runs(app) -> None:
             except Exception as e:
                 logger.error(e)
 
+        notify_scheduled_job_done(SchedulerJobType.SCHEDULE_JOB_RUNS)
```
yannickperrenet, line 338: We should probably add a note to the `Jobs` class or maybe to the module docstring that every `handle_func` should take care of `notify_scheduled_job_done` (either directly or indirectly) as otherwise we might end up with inconsistent state.

(Mentioning because we can indeed not "automatically" call it in the `_collateral` because some jobs run as Celery tasks)

yannickperrenet: Ghehe I see this is what you talk about in https://github.com/orchest/orchest/pull/1227/commits/cfc720dc5f49d15a6b6e675bc35347f22fc64264#r958537034 :+1: 

Resolving this conversation in favor of that one.

services/orchest-api/app/app/core/tasks.py
fruttasecca: The `celery-worker` does access the DB, it started doing that when we introduced notifications AFAIR, after a discussion about this pattern.

>If we let the celery-worker touch the DB then we might as well change how we do Pipeline Step updates.

That's indeed the case, it just something that we never caught up to, should wait to re-enable tests for it imo.

docs/source/fundamentals/settings.md
fruttasecca: Note that this particular snippet is outdated, this is the current state:

`Controls the number of image builds that can be run in parallel. With builds we mean environment
or custom jupyter server builds.`

services/orchest-api/app/app/core/scheduler.py
fruttasecca: >How about going with an alternative column (instead of status) called: running_tasks (or some other name to indicate the number of tasks/processes/jobs that are running the particular SchedulerJobType)? This way every job is allowed to increase the integer on start and decrease the integer on end and thus allowing certain actions (e.g. push to registry) when running_tasks==0.
>Like you alluded to in the comment, this would indeed introduce issues if requests to decrease the task counter fail. Putting the DB in a state where the certain actions (e.g. push to registry) would never run again. How far should we go operating under the assumption that requests don't go through? What do you think?

As you mentioned, going with a counter doesn't solve a request not going through leading to a bad state. I think we do have the option of resetting the counter or, if we go with a record for each run task, setting to `NOT_RUNNING` (or `FAILED`, whatever) during our cleanup phase so that a bad state can be fixed by restarting Orchest when discovered. We could also have a sort of deadline which fixes the state of these tasks if they have been "running" for too long, not sure it's feasible in a simple manner when it comes to counters.

Might be worth biting the bullet and going for a record for every instance of a task being run, with an "internal" scheduler task which takes care of only keeping the last N records (by task type) so that the storage use is not unbound, since the `deliveries` task runs quite often.

docs/source/fundamentals/settings.md
yannickperrenet: Simplified the language in https://github.com/orchest/orchest/pull/1227/commits/d49025334284585193a93d1329c7e0bbd4899b81.

services/orchest-api/app/app/core/tasks.py
yannickperrenet: > after a discussion about this pattern.

I guess I am getting old :stuck_out_tongue_closed_eyes: (now that you mention it I very vaguely remember that discussion).

> it just something that we never caught up to, should wait to re-enable tests for it imo.

Created: https://github.com/orchest/orchest/issues/1246

services/orchest-api/app/app/core/scheduler.py
yannickperrenet: > Might be worth biting the bullet and going for a record for every instance of a task being run

Alright, lets you with that approach then :+1: 

Could you also make the `deliveries` task log less often or change the logging level to debug? Now it is a bit hard to inspect the `orchest-api` logs.

> but I wanted to discuss the overall approach a bit before bumping scope

I like that decision :100:

services/orchest-api/app/app/core/scheduler.py
```diff
@@ -226,7 +254,48 @@ def _collateral(
...
+
+
+def cleanup_old_scheduler_job_records(app, task_uuid: str) -> None:
+    logger = logging.getLogger("cleanup_old_scheduler_job_records")
+
+    now = datetime.datetime.now(datetime.timezone.utc)
+
+    with app.app_context():
+
+        logger.info("Deleting old records")
```
yannickperrenet, line 272: Might be better to make this message more verbose, i.e. what old records?

yannickperrenet: Or is that clear because the `cleanup_old_scheduler_job_records` logger name is also in the log lines?

services/orchest-api/app/app/core/scheduler.py
```diff
@@ -226,7 +254,48 @@ def _collateral(
...
+                models.SchedulerJob.type == job_type.value,
+                models.SchedulerJob.uuid != task_uuid,
+            ).delete(synchronize_session="fetch")
+
+        logger.info("Fixing jobs that failed to report back their status.")
+        models.SchedulerJob.query.filter(
+            models.SchedulerJob.started_time < (now - datetime.timedelta(hours=1)),
+            models.SchedulerJob.status == "STARTED",
+        ).update({"status": "FAILED"})
+        db.session.commit()
```
yannickperrenet, line 296: Want to reduce this 1 hour to something like 15 minutes? Imo the smaller this timedelta the more robust our scheduler becomes, otherwise it could be that the we don't schedule jobs for 1 hour (to give an example).

What if the task is actually still running? Don't we need to issue a "kill" to the tasks that we say are "FAILED"? Otherwise we might still end up with concurrent tasks (which is exactly what we are looking to avoid here).

services/orchest-api/app/app/core/scheduler.py
fruttasecca: I think it's fine given the logger name, but I'm not against changes really

services/orchest-api/app/app/core/scheduler.py
fruttasecca: >Want to reduce this 1 hour to something like 15 minutes? Imo the smaller this timedelta the more robust our scheduler becomes, otherwise it could be that the we don't schedule jobs for 1 hour (to give an example).

Sure, will do

>What if the task is actually still running? Don't we need to issue a "kill" to the tasks that we say are "FAILED"? Otherwise we might still end up with concurrent tasks (which is exactly what we are looking to avoid here).

It would be another bump in the scope of the PR to make scheduler jobs abortable and I'm not sure it's worth it at the moment. To add some context, note that the logic of the scheduler in `_HandleRecurringSchedulerJob` schedules a job of the same type based on timestamps and interval values, ignoring the state of other jobs of the same type. If we don't have abortable scheduler jobs tasks we are left with two alternatives:
- have the risk of concurrent jobs running if one is too slow, which I think it's the same as on `dev`, i.e. we didn't introduce more risk. Note that critical code paths of jobs scheduled here already take care of being safe during concurrent calls, e.g. `schedule_job_runs` selects `with_for_update`
- have the risk of jobs not running because one db entry is stuck on `STARTED` because of some update failure, which is very improbable given the logic

services/orchest-api/app/app/core/scheduler.py
yannickperrenet: :shrug:

services/orchest-api/app/app/core/scheduler.py
yannickperrenet: I agree with what you said, let's not increase the scope of the PR :+1:

yannickperrenet (approved):

services/orchest-api/app/app/core/scheduler.py
fruttasecca: :+1:  I'll make a GH issue to keep track of this

PR #1232, opened 2022-08-26
fix: apply selectors properly to zustand hooks

iannbing:
## Description

The return value of the selector shouldn't be wrapped in an array or an
object, as they are new instances, which will always cause re-render.
Selector should return the value as-is for zustand to shallow compare.

Also extract Accordion from EnvironmentAccordion, as we'd also need to
use Accordion in other places.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/environments-view/edit-environment/EnvironmentImageBuildLogs.tsx
+++ b/services/orchest-webserver/client/src/environments-view/edit-environment/EnvironmentImageBuildLogs.tsx
@@ -11,13 +11,14 @@ import { useBuildEnvironmentImage } from "../hooks/useBuildEnvironmentImage";
 import { useEditEnvironment } from "../stores/useEditEnvironment";
 import {
   EnvironmentAccordion,
-  useEnvironmentAccordions,
+  useLogsAccordion,
 } from "./components/EnvironmentAccordion";
 
 export const EnvironmentImageBuildLogs = () => {
   const { projectUuid, environmentUuid } = useCustomRoute();
   const { config } = useGlobalContext();
-  const { isLogsOpen, setIsLogsOpen } = useEnvironmentAccordions();
+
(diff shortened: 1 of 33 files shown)


services/orchest-webserver/client/src/environments-view/edit-environment/EnvironmentImageBuildLogs.tsx
```diff
@@ -9,17 +10,23 @@ import React from "react";
...
+  EnvironmentAccordion,
+  useEnvironmentAccordions,
+} from "./components/EnvironmentAccordion";
 
 export const EnvironmentImageBuildLogs = () => {
   const { projectUuid, environmentUuid } = useCustomRoute();
   const { config } = useGlobalContext();
-  const { isLogsOpen, setIsLogsOpen } = useEnvironmentsUiStateStore();
-  const { environmentChanges } = useEditEnvironment();
+  const { isLogsOpen, setIsLogsOpen } = useEnvironmentAccordions();
```
mausworks, line 20: Nit: It's just one accordion, correct? So it should be `useEnvironmentAccordion`. :thinking:

mausworks (approved):
Great work, I also learned something. :+1: 

Minor nit, feel free to reject it and merge anyways, or fix it, then merge after. :+1:

services/orchest-webserver/client/src/environments-view/edit-environment/EnvironmentImageBuildLogs.tsx
iannbing: There are three accordions: Logs, SetupScript, and Properties. And I found that I didn't apply the same tactics to `useEnvironmentAccordions`, see [this commit](https://github.com/orchest/orchest/pull/1232/commits/7bf13ece96b9812a038d77c6536f248ee3076931).

So that when one accordion state changes won't cause re-render on other accordions.

Issue #1233, opened 2022-08-26
Pipeline params edition is flaky


astrojuanlu:
**Describe the bug**
While trying to edit the parameters of a pipeline I get flaky behavior: sometimes it works, sometimes it doesn't.

**Expected behavior**
Editing the pipeline parameters should always work.

**To Reproduce**

1. Create an empty project, a new pipeline, and an empty `test.py` file with a pipeline step
2. Add a simple pipeline parameter, for example `{"a": 1}`
3. Verify that the pipeline parameters were correctly saved
4. Try to add another pipeline parameter, for example `{"a": 1, "b": 2}`
5. Verify that it doesn't get saved ❌ 

**Screenshots**

https://user-images.githubusercontent.com/316517/186894489-23dd07d2-fc4b-48e0-a50f-48fee7b8f5a6.mp4

**Environment**

- OS (e.g. macOS): Linux Mint
- Browser (e.g. Chrome): Firefox
- Orchest's version (in the settings page): v2022.08.8

PR #1234, opened 2022-08-26
Fix unable to save pipeline parameters when clicking outside of the field.

iannbing:
## Description

`editableParameters` in `prettifyInputParameters` is not in sync with the latest value of editableParameters. Use setState callback to get the latest value. Also removed the unnecessary saving, as the value was already formatted and saved while typing. `onBlur` should only update the client-side state only.

Fixes: #1233 

## Checklist


mausworks (approved):
LGTM!

astrojuanlu:
I confirm that this seems to have solved the issue 👍🏽 Thanks!

PR #1240, opened 2022-08-29
Introduce `Snapshot` entity in models

fruttasecca:
## Description

This PR introduces `Snapshot`s as an `orchest-api` models entity. I've tried to minimize the exposure/knowledge of the client and the `orchest-webserver` to this new entity given that the intent was to support another feature (switching between pipelines for a `DRAFT` job) without closing doors for the future.

As it is, a `job` depends on a snapshot through a foreign key column `job.snapshoot_uuid`, existing jobs are migrated in this migration 44699afd40d8f5d9f72a765f6a6932857e30b50c. The reason is that inverting the dependency felt like a stronger decision and that currently, a snapshot is created before the job is, making the job effectively depend on the snapshot.

## Checklist

Open questions:
- the snapshot db record doesn't store anything related to the actual location of the snapshot, the application still uses the implicit path given by the `project_uuid` and `pipeline_uuid`. Do we want to change this? Likely not needed until/if we expand the scope of snapshots
- given the information within a snapshot like the `pipeline_definition` and the pipeline `path` we could, theoretically, slim down the job model, e.g. by dropping the pipeline definition, the run spec, etc. I'm not sure it's worth doing that while tests are still disabled since an error here could disrupt existing jobs


yannickperrenet:
Let me first reply to the open questions from the PR description:

> the snapshot db record doesn't store anything related to the actual location of the snapshot, the application still uses the implicit path given by the project_uuid and pipeline_uuid. Do we want to change this? Likely not needed until/if we expand the scope of snapshots

~~Actually... :thinking: I think that adding the location of the snapshot as a db record might make it easier to later make "snapshots" a broader concept in Orchest. For example, let's say we want to have `userdir/snapshots` (just like we have `userdir/jobs`) then getting to that world is easier if we store the location (I think).~~ We can always add the location as a DB record to snapshots later if anything changes in the future :+1: 

> given the information within a snapshot like the pipeline_definition and the pipeline path we could, theoretically, slim down the job model, e.g. by dropping the pipeline definition, the run spec, etc. I'm not sure it's worth doing that while tests are still disabled since an error here could disrupt existing jobs

Quite a nice idea! And agree with the caution of not doing it now. Could you open an issue for this (good idea to keep track of this technical debt).

services/orchest-api/app/migrations/versions/7beba3ba9e66_.py
```diff
@@ -0,0 +1,112 @@
...
+                job.pipeline_uuid: {
+                    "path": job.pipeline_run_spec["run_config"]["pipeline_path"],
+                    "definition": job.pipeline_definition,
+                }
+            },
+            # PR_DISCUSS: either this or an empty dict? The current
+            # choice allows to reconstruct things later if this
+            # information is needed.
+            project_env_variables=job.env_variables,
+            pipelines_env_variables={job.pipeline_uuid: job.env_variables},
```
yannickperrenet, line 102: Let's indeed just insert that information given that we have this.

The `pipelines=` option does not include all pipelines present in the snapshot, but I guess it is fine given that (the way we currently intend to use snapshots) the pipeline is only allowed to be changed of DRAFT jobs and it is unlikely that a user will have any DRAFT jobs when we migrate. Or do you think it makes sense to add more entries (so all pipelines present in the snapshot)?

services/orchest-api/app/app/apis/namespace_snapshots.py
```diff
@@ -0,0 +1,105 @@
...
+    def get(self, snapshot_uuid: str):
+        """Fetches a snapshot given its UUID."""
+        snapshot = models.Snapshot.query.filter_by(uuid=snapshot_uuid).one_or_none()
+        if snapshot is None:
+            abort(404, "Snapshot not found.")
+
+        return snapshot
+
+    def delete(self, snapshot_uuid: str):
+        """Deletes a snapshot given its UUID."""
```
yannickperrenet, line 70: Might be good to note in the docstring that it only deletes the DB entry but will also delete all jobs that depend on the snapshot.

services/orchest-api/app/app/apis/namespace_snapshots.py
```diff
@@ -0,0 +1,105 @@
...
+    @api.marshal_with(schema.snapshots)
+    def get(self):
+        """Fetches all snapshots."""
+
+        return {"snapshots": models.Snapshot.query.all()}
+
+    @api.doc("create_snapshot")
+    @api.expect(schema.snapshot_spec)
+    def post(self):
+        """Creates a snapshot entry in the db, doesn't alter the fs."""
```
yannickperrenet, line 29: Does this mean that the FE will have to start doing:
1. Create Snapshot
2. Create (DRAFT) Job

Was just thinking what the impact of this would be on #201 (although we might as well forget that for now).

yannickperrenet: Seems like the question was answered in the next commit: https://github.com/orchest/orchest/pull/1240/commits/304c1a2cf066f65887173e04c800c8d90db90409#diff-8094414017ecbbaa3bec8b61d57e9e7c605e70bcefde7b3ce3e7ad744b47af65R64-R66

So the BFF will take care of creating the snapshot when the FE does the Create Job request.

yannickperrenet (approved):

services/orchest-api/app/migrations/versions/7beba3ba9e66_.py
fruttasecca: >Or do you think it makes sense to add more entries (so all pipelines present in the snapshot)?

I think the current state is fine, even if users are going to migrate with a job in the `DRAFT` status, given that the content of a snapshot is detached from the current state of the project which implies that the FE should always base its decisions  (i.e. what pipelines to allowing switching to) on the content of the snapshot record. If it contains only a pipeline then that's a valid case as well. Moreover, accessing the file system and doing similar operations during a schema migration would be an added risk of breakage per se, without considering the fact that we don't want the `orchest-api` to access the file system.

fruttasecca:
>We can always add the location as a DB record to snapshots later if anything changes in the future +1

Indeed

PR #1243, opened 2022-08-29
Bust cache when manually checking for updates

yannickperrenet:
## Description

When clicking the "check for updates" in the settings, then we should not use the cached orchest update info JSON and instead query it again.

### DIFF
--- a/services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
+++ b/services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
@@ -41,15 +41,19 @@ const fetchOrchestVersion = () =>
     (response) => response.version
   );
 
-const fetchLatestVersion = () =>
-  fetcher<UpdateInfo>("/async/orchest-update-info").then(
+const fetchLatestVersion = (cache?: boolean) => {
+  const endpoint = `/async/orchest-update-info?cache=${
+    cache ? "true" : "false"
+  }`;
+  return fetcher<UpdateInfo>(endpoint).then(
     (response) => response.latest_version
   );
+};
 
-const requestToCheckVersions = async () => {
+const requestToCheckVersions = async (cache?: boolean) => {
   const [orchestVersion, latestVersion] = await Promise.all([
     fetchOrchestVersion(),
-    fetchLatestVersion(),
+    fetchLatestVersion(cache),
   ]);
 
   return [orchestVersion, latestVersion] as const;
@@ -178,7 +182,7 @@ export const useCheckUpdate = () => {
     // we want to be able to tell the user that no update is available
     // if this function is invoked.
     const [fetchedOrchestVersion, fetchedLatestVersion] = await makeCancelable(
-      requestToCheckVersions()
+      requestToCheckVersions(false)
     );
 
     if (fetchedOrchestVersion && fetchedLatestVersion) {
--- a/services/orchest-webserver/app/app/utils.py
+++ b/services/orchest-webserver/app/app/utils.py
@@ -16,6 +16,7 @@ from _orchest.internals import utils as _utils
 from _orchest.internals.utils import copytree, is_services_definition_valid, rmtree
 from app import error
 from app.config import CONFIG_CLASS as StaticConfig
+from app.core import scheduler
 from app.models import Environment, Pipeline, Project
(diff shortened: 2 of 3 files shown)


ricklamers (approved):

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -41,15 +41,19 @@ const fetchOrchestVersion = () =>
...
+  const endpoint = `/async/orchest-update-info?cache=${
+    cache ? "true" : "false"
+  }`;
+  return fetcher<UpdateInfo>(endpoint).then(
     (response) => response.latest_version
   );
+};
 
-const requestToCheckVersions = async () => {
+const requestToCheckVersions = async (cache?: boolean) => {
```
mausworks, line 53: I would prefer if this was instead some options object (passing `true/false` to a function tells you very little what that parameter does without looking into the function itself, or signature).

```suggestion
const requestToCheckVersions = async ({ cache = true }: { cache?: boolean }) => {
```

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -178,7 +182,7 @@ export const useCheckUpdate = () => {
     // we want to be able to tell the user that no update is available
     // if this function is invoked.
     const [fetchedOrchestVersion, fetchedLatestVersion] = await makeCancelable(
-      requestToCheckVersions()
+      requestToCheckVersions(false)
```
mausworks, line 185: Building on the previous comment:

```suggestion
      requestToCheckVersions({ cache: false })
```

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
```diff
@@ -41,15 +41,19 @@ const fetchOrchestVersion = () =>
     (response) => response.version
   );
 
-const fetchLatestVersion = () =>
-  fetcher<UpdateInfo>("/async/orchest-update-info").then(
+const fetchLatestVersion = (cache?: boolean) => {
+  const endpoint = `/async/orchest-update-info?cache=${
+    cache ? "true" : "false"
+  }`;
```
mausworks, line 47: Building on the previous comment: By providing default value for `cache` you can just interpolate it in the string directly:

```suggestion
  const endpoint = `/async/orchest-update-info?cache=${cache}`;
```

mausworks (approved):
I think the implementation looks good overall, but I have some remarks on code style. :+1:

services/orchest-webserver/client/src/hooks/useCheckUpdate.tsx
yannickperrenet: I don't see how passing an options object instead of arguments changes readability. They are exactly the same thing except when passing an options object you have to "unpack" the values (instead of accessing them as variables directly when passing as arguments). When passing an options object, then you still don't know what the option/parameter does.

Doesn't Typescript also allow to do `cache: boolean = true` (i.e. named parameters)? Seems to me that would better fit this use case.

mausworks: > I don't see how passing an options object instead of arguments changes readability. 

I think it improves readability a lot _at the callsite_! When I see a boolean parameter passed into a function, I always have to inspect that function signature in order to know what it does (either by "view definition", or hovering over the function call).

Comparing (A) `someFunction(false)` with (B) `someFunction({ cache: false })` we can see that in example A, we can't really make any assumptions about how `false` changes the behavior of the `someFunction`, but in example B, it explicitly tells me that it disables cache.

> When passing an options object, then you still don't know what the option/parameter does.

But surely you should be able to make the assumption that `{ cache: false }` will disable the cache.

> Doesn't Typescript also allow to do `cache: boolean = true` (i.e. named parameters)? Seems to me that would better fit this use case.

No, but I wish it did! :cry: That's why I'm suggesting this—it is how you commonly "name your parameters" in the JS world. The alternative is to do something like this (though I still think it's less expressive):

```ts
const cache = false;
await requestToCheckVersions(cache)
```

I am not—however—dying on this hill, I still approved the PR. :+1:

PR #1244, opened 2022-08-30
Dynamically update the pipeline of a job draft

fruttasecca:
## Description

Allows altering the pipeline of a job which is still in the `DRAFT` status. Environment variables are resolved using the snapshotted state contained in the snapshot in use by the job. See the docstring of `_resolve_environment_variables` for more details.

@iannbing 
The functionality is exposed to the client through a new `orchest-webserver` endpoint, `/catch/api-proxy/api/jobs/<job_uuid>/pipeline` (PUT), which requires a json of the form `{"pipeline_uuid": "<uuid>"}` which indicates
which pipeline the job should be using. From the client POV the endpoint is a proxy to the `orchest-api` endpoint `/jobs/<job_uuid>/pipeline`, see the `orchest-api` swagger docs for the dos and don'ts.

Note that the `orchest-webserver` now also exposes the endpoint `/catch/api-proxy/api/snapshots/<snapshot_uuid>`, which allows retrieving details about a snapshot. This information is what should be used when it comes to the pipelines (uuids, names, paths)  within the job: e.g.:
- get the job 
- the job has a `snapshot_uuid` field
- retrieve the snapshot through its uuid
- the snapshot contains information about what pipelines are contained within the job's snapshot, which are the pipelines between which the job can switch while in draft mode

## Checklist

### DIFF
--- a/services/orchest-api/app/app/apis/namespace_jobs.py
+++ b/services/orchest-api/app/app/apis/namespace_jobs.py
@@ -9,7 +9,7 @@ from croniter import croniter
 from flask import abort, current_app, request
 from flask_restx import Namespace, Resource, marshal, reqparse
 from sqlalchemy import desc, func
-from sqlalchemy.orm import joinedload, load_only, noload, undefer
+from sqlalchemy.orm import attributes, joinedload, load_only, noload, undefer
 
 import app.models as models
 from _orchest.internals import config as _config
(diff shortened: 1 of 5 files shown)


services/orchest-api/app/app/apis/namespace_jobs.py
```diff
@@ -158,12 +158,12 @@ def get(self, job_uuid):
 
         return job
 
-    @api.expect(schema.job_update)
+    @api.expect(schema.job_parameters_update)
     @api.doc("update_job")
```
yannickperrenet, line 162:
```suggestion
    @api.doc("update_job_parameters")
```

services/orchest-api/app/app/schema.py
```diff
@@ -495,20 +495,9 @@ def _session_base_url(s) -> str:
...
-            required=True,
-            description="Collection of pipeline definitions",
-        ),
-        "pipeline_run_ids": fields.List(
-            fields.Integer(
-                description=(
-                    "Pipeline index corresponding to respective "
-                    "list entries in pipeline_definitions."
-                )
-            ),
```
yannickperrenet, line 509: Don't we use these anymore? As far as I remember these where used so that the same ordering can be maintained when editing the pipeline runs of a job.

yannickperrenet (approved):
:100:

services/orchest-api/app/app/schema.py
fruttasecca: Not that I know of, this looks like how jobs operated long ago, i.e. by passing multiple pipeline definitions and taking care of injecting parameters at the client level, each non-interactive run index is now stored with each run record.

Issue #938, opened 2022-05-10
Document and clarify current JupyterLab terminal behavior


astrojuanlu:
_**Edit**: Repurposing this issue to document & clarify current behavior, while keeping #611 for the requested new functionality. See discussion below._

**Describe the feature you'd love to have in Orchest**
It would be nice to make the terminal behave like it is running inside the Orchest environment, with all its dependencies.

**Is your feature request related to a problem? Please describe.**
Some tasks need to be run from the command line using dependencies installed in the environment. Examples include:

- Management commands of meltano, dbt, and more
- Quick experiments running Python files from the command line to observe outputs
- Application of linters (black, isort etc)

However, at the moment when opening a Terminal, the user gets dropped to an environment that does not contain any of the dependencies installed in the environment, which live in Docker containers spawned by Jupyter Enterprise Gateway. This confuses users that are used to have a terminal sitting next to their files with access to the same Python virtual environment.

In some sense, the current setup is more similar to a remote development setting, where the Python runtime lives inside a Docker container, but with the limitation that users can't connect to such runtime from the terminal.

The only workaround is to use either a Jupyter notebook or a Jupyter console (both attached to a kernel), but there are some rough edges:

- Terminal commands need to be prefixed by an exclamation mark `!`
- Commands with interactive input block the kernel without possibility of typing such input
- Limited autocomplete support

**Additional context**
Example reaction from a user:

![image](https://user-images.githubusercontent.com/316517/167600397-d8868858-5d1c-4940-be15-1213abec2786.png)

lalligagger:
This is such a good explanation I'd recommend keeping open if there's a label for it. 😅 Thanks!

astrojuanlu:
To keep things tidy, let's keep #611 for "implementing terminal access to environments", and repurpose this issue to "document/clarify current terminal behavior".

Issue #611, opened 2021-12-22
Environment "peek with terminal" feature


ricklamers:
**Is your feature request related to a problem? Please describe.**  
The problem is that often you don't know the details of how things are organized in a container image. So if you're extending either `orchest/base-image-py` or a custom image. It often requires trial and error to get the `setup_script.sh` worked out nicely.

**Describe the solution you'd like**  
The idea is to add the option to jump into an interactive terminal that spawns in the environment image (pre or post build, e.g. base image or the "fine tuned" image) to be able to look around, run a few `bash` command (like `pip show pandas`) to get a sense of how things are organized in the image.

This can help you fine-tune the `setup_script.sh` in a much more time-effective way compared to re-running the `setup_script.sh` all the time.

**Describe alternatives you've considered**
What you can do now is run the "debug" commands as part of the `setup_script.sh` (by putting at the top, or uncommenting the "actual" command you want to run). This is pretty hacky and time-consuming.

astrojuanlu:
Notice that, beyond "peeking" or looking around the environment, this would be useful to run CLI-based tools like meltano, dbt, and the like, as discussed in #938.

astrojuanlu:
> Some tasks need to be run from the command line using dependencies installed in the environment. Examples include:
> 
> - Management commands of meltano, dbt, and more
> - Quick experiments running Python files from the command line to observe outputs
> - Application of linters (black, isort etc)
> 
> However, at the moment when opening a Terminal, the user gets dropped to an environment that does not contain any of the dependencies installed in the environment, which live in Docker containers spawned by Jupyter Enterprise Gateway. This confuses users that are used to have a terminal sitting next to their files with access to the same Python virtual environment.
> 
> In some sense, the current setup is more similar to a remote development setting, where the Python runtime lives inside a Docker container, but with the limitation that users can't connect to such runtime from the terminal.
> 
> The only workaround is to use either a Jupyter notebook or a Jupyter console (both attached to a kernel), but there are some rough edges:
> 
> - Terminal commands need to be prefixed by an exclamation mark `!`
> - Commands with interactive input block the kernel without possibility of typing such input
> - Limited autocomplete support

astrojuanlu:
Reference to the upstream IPython issue that prevents interactive commands usage: https://github.com/ipython/ipython/issues/10499

PR #1248, opened 2022-09-01
Environment shells

ricklamers:
## Description

This feature introduces the concept of "environment shells". It's an interactive shell that can be accessed from the JupyterLab web terminal that is based on one of the project defined environments.

Using this approach users can use the browser based development environment in a more similar style to local development. The terminal gives access to command line tools such as `git` but also data specific CLI programs like `dbt` or `meltano`. In addition, you can test runtime behavior of steps by invoking Python by running `python script.py` in a way that mirrors exactly how the steps will be invoked when part of the pipeline.

Open issue: how to handle step-specific environment variables needed for executing Orchest SDK code (.get_inputs() etc.).
Ideas: make SDK "intelligently" resolve step based on pipeline definition file and current file it's in (problem: file assigned to multiple steps).

Fixes: #938
Fixes: #611

Note, manually tested all base kernels (Julia, JavaScript, Python and R).

🎥 watch me talk through it: https://tella.video/environment-shells-feature-1-fx3w

## Checklist


services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -339,11 +463,46 @@ def _get_jupyter_server_deployment_service_manifest(
                             "imagePullPolicy": "IfNotPresent",
                             "volumeMounts": [
                                 volume_mounts_dict["project-dir"],
+                                # TODO: eval (Jacopo): why is the
```
ricklamers, line 466: @fruttasecca

docs/source/fundamentals/environments.md
```diff
@@ -103,6 +103,22 @@ Installing packages with `conda` is also supported but might take longer (due to
...
+All you need to do is open JupyterLab in a session and click on Terminal:
+
+```{image} ../img/environment-shells-jupyterlab.png
+:width: 800
+:alt: Terminal in JupyterLab
+:align: center
+```
+
+```{warning}
+If you make any changes to your environment, you'll need to restart your environment shell to see the changes. You can restart your environment shell by exiting out of it by typing Ctrl + D or typing 'exit'.
```
yannickperrenet, line 119:
```suggestion
If you make any changes to your environment, you'll need to restart your environment shell to see the changes. You can restart your environment shell by exiting out of it by pressing 'Ctrl + D' or typing 'exit'.
```

docs/source/fundamentals/environments.md
```diff
@@ -103,6 +103,22 @@ Installing packages with `conda` is also supported but might take longer (due to
 🚨 Do not install packages by running {code}`!pip install <package-name>` inside your Jupyter Notebook. This causes the package to be installed every time you run the Pipeline step. It is not saved in the environment as containers are stateless!
 ```
 
+## Environment shells
+
+JupyterLab has integrated support for shell sessions using a terminal emulator. To run your scripts and CLI commands while developing your data pipeline you can make use of shells that are based on the environments you have in your project. The environment container image is used such that you have access to all installed programs/dependencies.
```
yannickperrenet, line 108: We choose to capitalize all terms in Orchest, so `environment` should be `Environment` ([our implementation detail docs](https://docs.orchest.io/en/stable/development/implementation_best_practices.html#writing-docs)). 

```suggestion
JupyterLab has integrated support for shell sessions using a terminal emulator. To run your scripts and CLI commands while developing your data Pipeline you can make use of shells that are based on the Environments you have in your Project. The Environment container image is used such that you have access to all installed programs/dependencies.
```

docs/source/fundamentals/environments.md
```diff
@@ -103,6 +103,22 @@ Installing packages with `conda` is also supported but might take longer (due to
...
+
+All you need to do is open JupyterLab in a session and click on Terminal:
```
yannickperrenet, line 110:
```suggestion
All you need to do is open JupyterLab and click on Terminal:
```

Don't think `in a session` adds anything here (especially given that we start the Session automatically).

lib/python/orchest-internals/_orchest/internals/utils.py
```diff
@@ -128,6 +128,10 @@ def get_init_container_manifest(
     init_container = {
         "name": "image-puller",
         "image": image_puller_image,
+        "securityContext": {
+            "privileged": True,
+            "runAsUser": 0,
+        },
```
yannickperrenet, line 134: Why was this added in the context of Environment shells?

orchest-sdk/python/orchest/utils.py
```diff
@@ -8,6 +8,22 @@
 from orchest.pipeline import Pipeline
 
 
+def pretty_print_env_var_suggestion(pipeline: Pipeline):
```
yannickperrenet, line 11:
```suggestion
def pretty_print_env_var_suggestion(pipeline: Pipeline) -> None:
```

Common practice to explicitly return `None` if applicable.

orchest-sdk/python/orchest/utils.py
```diff
@@ -8,6 +8,22 @@
...
+        "Please specify a pipeline step UUID by setting the ORCHEST_STEP_UUID "
+        + "environment variable. Choose one of the following pipeline steps:"
+    )
+    for step in pipeline.steps:
+        print(
+            "ORCHEST_STEP_UUID="
+            + step.properties.get("uuid")
+            + " (Step %s %s)"
+            % (step.properties.get("title"), step.properties.get("file_path"))
+        )
```
yannickperrenet, line 22: Nitpick (feel free to choose whatever): in my opinion readability of fstrings is better here.
```suggestion
        print(
            f'ORCHEST_STEP_UUID={step.properties.get("uuid")} '
            f'(Step {step.properties.get("title")}, {step.properties.get("file_path")})'
        )
```

(Wanted to add it to show how fstring usage would look.)

orchest-sdk/python/orchest/utils.py
```diff
@@ -8,6 +8,22 @@
...
+def pretty_print_env_var_suggestion(pipeline: Pipeline):
+    print(
+        "Please specify a pipeline step UUID by setting the ORCHEST_STEP_UUID "
+        + "environment variable. Choose one of the following pipeline steps:"
+    )
+    for step in pipeline.steps:
+        print(
+            "ORCHEST_STEP_UUID="
+            + step.properties.get("uuid")
+            + " (Step %s %s)"
```
yannickperrenet, line 20:
```suggestion
            + " (Step: '%s' at '%s')"
```

orchest-sdk/python/orchest/utils.py
```diff
@@ -30,7 +46,11 @@ def get_step_uuid(pipeline: Pipeline) -> str:
     # Enterprise Gateway.
     kernel_id = os.environ.get("KERNEL_ID")
     if kernel_id is None:
-        raise StepUUIDResolveError('Environment variable "KERNEL_ID" not present.')
+        # Assuming since KERNEL_ID is missing that we're running in
+        # an environment shell. Print environment variable setter hint.
+        pretty_print_env_var_suggestion(pipeline)
+
+        raise StepUUIDResolveError("Could not resolve the pipeline step UUID.")
```
yannickperrenet, line 53: Given that the error is called `StepUUIDResolveError` I think it is already clear that the "pipeline step UUID couldn't be resolved". Therefore, I think this gives us the opportunity to use the message we can pass with the error to better explain what happened, e.g.:

```suggestion
        raise StepUUIDResolveError(
            'Environment variable "STEP_UUID" has to be explicitly passed when'
            ' executing Pipeline Steps in Environment shells.'
        )
```

services/base-images/base-kernel-javascript/Dockerfile
```diff
@@ -27,9 +27,12 @@ COPY ./runnable-shared/runner/requirements* /orchest/services/base-images/runnab
 COPY ./lib/python /orchest/lib/python
 COPY ./orchest-sdk /orchest/orchest-sdk
 RUN chown $NB_USER -R /orchest/orchest-sdk \
-    && chown $NB_USER -R /orchest/lib
+    && chown $NB_USER -R /orchest/lib && \
+    echo -e "jovyan\njovyan" | passwd jovyan
 
+# Exclude hostname from default prompt
 USER $NB_USER
+RUN echo "export PS1='\[\033[01;32m\]\u\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '" >> ~/.bashrc
```
yannickperrenet, line 35: To reduce the number of layers in the Dockerfile (and use consistent way of breaking `RUN` commands over multiple lines):
```suggestion
RUN chown $NB_USER -R /orchest/orchest-sdk \
    && chown $NB_USER -R /orchest/lib \
    && echo -e "jovyan\njovyan" | passwd jovyan \ 
    # Exclude hostname from default prompt
    && echo "export PS1='\[\033[01;32m\]\u\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '" >> /home/$NB_USER/.bashrc

USER $NB_USER
```

services/base-images/runnable-shared/bootscript.sh
```diff
@@ -10,6 +10,14 @@ umask 002
...
     /home/$NB_USER/venv/bin/python \
     /orchest/services/base-images/runnable-shared/runner/run.py  "$2" "$3"
+elif [ "$1" = "shell" ]; then
+    # Mark safe for git
+    git config --global --add safe.directory /project-dir
+    
+    env > /etc/environment
+    sudo service ssh start
+    sudo service ssh stop
+    sudo /usr/sbin/sshd -D
```
yannickperrenet, line 20: Why is the `ssh` service started and then immediately stopped?

services/jupyter-server/Dockerfile
```diff
@@ -41,6 +41,12 @@ RUN bash /patch-pip.sh
 
 COPY ./jupyter_server_config.py /root/.jupyter/jupyter_server_config.py
 
+COPY shellspawner/shellspawner /usr/bin/shellspawner
+
+# Add shellspawner to terminal profile
+# TODO: check if this doesn't get sourced for the main start.sh
```
yannickperrenet, line 47: I guess this `TODO` is no longer applicable?

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+            self.request_ready = True
+        except (urllib.error.HTTPError, urllib.error.URLError) as e:
+            print(e)
+
+
+def print_jupyter_warning():
+    print(
+        "\nUsing the JupyterLab terminal directly is not recommended "
+        + "read more about it in the docs\n"
+        + "➜ https://docs.orchest.io/en/stable/fundamentals/environments.html"
```
yannickperrenet, line 45: Nitpick: Using `+` is not needed actually.
```suggestion
        "\nUsing the JupyterLab terminal directly is not recommended"
        " read more about it in the docs\n"
        "➜ https://docs.orchest.io/en/stable/fundamentals/environments.html"
```

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    )
+
+
+def environment_shell_uuid_to_environment_uuid(environment_shell_uuid: str) -> str:
+    random_uuid_suffix_len = 6
```
yannickperrenet, line 50: Might be better to define a global constant given that the suffix is used in multiple functions

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    return environment_shell_uuid[: -(random_uuid_suffix_len + 1)]
+
+
+def environment_shell_uuid_to_name(
+    environment_shell_uuid: str, environments: List[Dict[str, Any]]
+) -> str:
+    random_uuid_suffix_len = 6
+
+    # Cut of last random UUID
+    environment_uuid = environment_shell_uuid[: -(random_uuid_suffix_len + 1)]
```
yannickperrenet, line 60: Why not use the dedicated function `environment_shell_uuid_to_environment_uuid` to do this?

yannickperrenet: Could even make it into `def split_environment_shell_uuid(...) -> Tuple[str, str]` which returns the `environment_uuid` and `shell_name`.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+
+
+def collect_choice(choice_count: int, prefix: str):
+    while True:
+        try:
+            choice_end = ("-" + str(choice_count - 1)) if choice_count > 1 else ""
+            print(f"{prefix} [0{choice_end}]: ", end="", flush=True)
+            user_input = input()
+            index = int(user_input)
+            assert index >= 0 and index <= choice_count - 1
```
yannickperrenet, line 83: I think it would be better to check with an if-statement and raise a `ValueError` given that it is user input we are testing. In addition assert statements (see [python docs](https://docs.python.org/3/reference/simple_stmts.html#assert)) are used as a debugging tool and can actually be disabled with a flag, e.g. `python -O script.py`. Given that our implementation is actually (in prod) dependent on the assert statement running, I think it is better to be explicit and make sure the program fails if we want it to.

```suggestion
            if not (index >= 0 and index <= choice_count - 1):
                raise ValueError("Invalid user input.")
```

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+        except KeyboardInterrupt:
+            return None
+        except Exception:
+            # Choose first option if only one option,
+            # still run rest of the logic
+            # to enable backing out using Ctrl+C KeyboardInterrupt
+            if choice_count == 1:
+                return 0
+
+            print("Invalid input.")
```
yannickperrenet, line 94:
```suggestion
            print(f"Invalid input. Enter integer between 0 and {choice_count - 1}.")
```

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+            continue
+
+
+def get_environments() -> List[Dict[str, Any]]:
+    req = request.Request(
+        ORCHEST_WEBSERVER_ADDRESS
+        + "/store/environments/"
+        + os.environ["ORCHEST_PROJECT_UUID"]
+    )
+    req.add_header("Content-Type", "application/json")
```
yannickperrenet, line 104: Given that Environment shells actually require Environments to have been built, wouldn't it be better to query the `orchest-api` instead? Otherwise we can run into the issue where a user requests an Environment shell but the Environment not being available yet.

(On top of my head, without actually testing it) a reproducible test case:
1. Start a session
2. Open jupyterlab
3. Open Orchest in another tab
4. Create a new Environment (but don't build it)
5. `orchest-websever` will now return an Environment that doesn't actually have an Environment image yet.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+            end="",
+            flush=True,
+        )
+        time.sleep(speed)
+        waited += speed
+        wait_index += 1
+
+    print("")
+    if waited >= timeout:
+        print("Warning! Loader timed out after %d seconds." % timeout)
```
yannickperrenet, line 173: Shouldn't this be an error instead? I guess if this case is hit then the Environment shell won't connect?

If that is the case then this print statement is a little misleading for users.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    json_resp = json.loads(content)
+
+    return json_resp
+
+
+def spawn_environment_shell(environment: Dict[str, Any]) -> Dict[str, Any]:
+    t = BackgroundRequest(lambda x: _spawn_environment_shell(x), [environment])
+    t.start()
+    show_loader(lambda x: x.request_ready, [t], "Starting shell")
+    return t.request_result
```
yannickperrenet, line 209: Ghehe nice! This pattern is actually very similar to using [futures](https://docs.python.org/3/library/asyncio-future.html) in Python: `await`ing the future and displaying a spinner in the meantime.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+
+
+def connect(environment_shell: str):
+    print("Connecting...", flush=True)
+
+    user = "jovyan"  # TODO: support more usernames
```
yannickperrenet, line 215: Wouldn't it be better to use the `$NB_USER` environment variable?

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    co = subprocess.run(command, shell=True)
+    if co.returncode == 255:
+        print("Could not connect to SSH in the environment. Try rebuilding ")
+        print("your environment to update to the latest base image.")
+
+    # Enter init loop
+    init()
+
+
+def choose_environment(environments: List[Dict[str, Any]]) -> Dict[str, Any]:
```
yannickperrenet, line 242: Strictly speaking the return type is actually: `Optional[Dict[str, Any]]`. Given that if `environment_choice` is `None`, then `None` is returned (which is possible).

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    elif len(environment_shells) == 1 and autoconnect:
+        # Autoconnect to shell if there's only one
+        connect(environment_shells[0])
+    else:
+        action = choose_action()
+
+    if action is None:
+        quit()
+
+    if action == 0:
```
yannickperrenet, line 307: This (using `0` hardcoded) closely couples this function with `choose_action`. Wouldn't it be possible to change the action into an `Enum`?

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+            stop_environment_shell(environment_shell["uuid"])
+
+        init()
+
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1 and sys.argv[1] == "--help":
+        print("This terminal supports connecting with Orchest environment shells.")
+        print("To restart, terminate, start or connect to a shell simply exit out of ")
+        print("your current shell by typing 'exit' or typing Ctrl + D.")
```
yannickperrenet, line 368:
```suggestion
        print("your current shell by typing 'exit' or pressing 'Ctrl + D'.")
```

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+        print("")
```
yannickperrenet, line 369: Total nitpick:
```suggestion
        print()
```

services/orchest-api/app/app/schema.py
```diff
@@ -239,6 +239,43 @@ def _session_base_url(s) -> str:
...
+        ),
+    },
+)
+
+environment_shell = Model(
+    "EnvironmentShell",
+    {
+        "uuid": fields.String(required=True, description="UUID of environment shell"),
+        "session_uuid": fields.String(required=True, description="UUID of session"),
+        "host": fields.String(required=True, description="host of k8s service"),
```
yannickperrenet, line 267: I am not sure I understand `"host of k8s service"`. What do you mean here?

services/orchest-api/app/app/core/sessions/_manifests.py
yannickperrenet: I believe this has to do with how Docker works. By mounting it directly, the pipeline file can be moved without breaking the mount (because Docker uses the inode). 

So it is still very much applicable. Could you change the `TODO` into a `NOTE` explaining the above?

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -339,11 +463,46 @@ def _get_jupyter_server_deployment_service_manifest(
...
+                                # to the absolute path of the mounted
+                                # pipeline file.
+                                # ORCHEST_SESSION_PIPELINE_PATH is used
+                                # to let the shellspawner pass the
+                                # relative pipeline path to the
+                                # environment-shells endpoint.
+                                {
+                                    "name": "ORCHEST_SESSION_PIPELINE_PATH",
+                                    "value": session_config["pipeline_path"],
+                                },
```
yannickperrenet, line 495: How about changing the name to `ORCHEST_SHELLSPAWNER_PIPELINE_PATH`? If you have to explain the name, then it is probably not a good one :P

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -23,7 +24,7 @@
 def _get_common_volumes_and_volume_mounts(
     userdir_pvc: str,
     project_dir: str,
-    pipeline_path: str,
+    pipeline_path: str = None,
```
yannickperrenet, line 27: If we set it to `None` by default then the type needs to be `Optional[str]`.

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
     }
 
 
+def get_orchest_sdk_vars(
```
yannickperrenet, line 282: Imo better to make this a private function, i.e. `_get_orchest_sdk_vars`

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
...
+    """
+    This manifest generation is in core/sessions
+    since the environment shell is part of the
+    session, but has a detached lifecycle in the sense
+    that environment shells can be started/stopped
+    independently from the session start/stop.
+
+    When a session is stopped all associated environment
+    shells should always be stopped.
+    """
```
yannickperrenet, line 316: Nitpick: Docstring doesn't seem to be wrapped at `72` chars.

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
...
+
+    metadata = {
+        "name": f"environment-shell-{environment_uuid}-{(str(uuid4()))[:6]}",
```
yannickperrenet, line 319: Can't we make this `6` a config option in the internal lib? I have seen this come by a few times now and I am scared we could forget updating it in one of its places.

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
...
+        "labels": {
+            "app": "environment-shell",
+            "project_uuid": project_uuid,
+            "session_uuid": session_uuid,
+        },
```
yannickperrenet, line 324: How about also adding `shell_uuid` here? The value can be the generated uuid4 from above (this way we also store the full uuid instead of just the trimmed (to 6 chars) version).

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
...
+        deployment_manifest,
+    )
+
+    service_manifest = {
+        "apiVersion": "v1",
+        "kind": "Service",
+        "metadata": metadata,
+        "spec": {
+            "type": "ClusterIP",
+            "selector": metadata["labels"],
```
yannickperrenet, line 407: Continuation on https://github.com/orchest/orchest/pull/1248/files#r968048749, I think (without specifying the shell_uuid in the labels) that this service would point to all Environment shells (given that all shells within the same session have the same uuid). Meaning that if a user starts many shells, then kubernetes will choose what shell to route traffic to.

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -351,6 +353,32 @@ def catch_api_proxy_sessions_post():
...
+            # Don't wait for the response
+            threading.Thread(
+                target=requests.post,
+                args=(
+                    "http://"
+                    + app.config["ORCHEST_API_ADDRESS"]
+                    + "/api/environment-shells/",
+                ),
+                kwargs={"json": json_data},
+            ).start()
```
yannickperrenet, line 380: This should probably use the `current_app.config["SCHEDULER"]` instead (to stay consistent with the rest of the codebase). Example: https://github.com/orchest/orchest/blob/377f596e915db5c03aa474f60dedb3fe5b5bf0c2/services/orchest-webserver/app/app/core/projects.py#L196-L201

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -351,6 +353,32 @@ def catch_api_proxy_sessions_post():
...
+        # shell this is a "client decision" hence it's not handled
+        # automatically by the orchest-api.
+        project_envs = get_environments(project_uuid)
+        if len(project_envs) > 0 and resp.status_code == 201:
+
+            json_data = {
+                "pipeline_uuid": pipeline_uuid,
+                "pipeline_path": pipeline_path,
+                "project_uuid": project_uuid,
+                "environment_uuid": project_envs[0].uuid,
```
yannickperrenet, line 366: Might be good explaining (in a comment) why environment `[0]` is used.

services/orchest-webserver/app/app/views/orchest_api.py
```diff
@@ -351,6 +353,32 @@ def catch_api_proxy_sessions_post():
             json=session_config,
         )
 
+        # Side effect of a session post is to create an environment
+        # shell this is a "client decision" hence it's not handled
+        # automatically by the orchest-api.
+        project_envs = get_environments(project_uuid)
```
yannickperrenet, line 359: I think this has the same problem as explained in https://github.com/orchest/orchest/pull/1248#discussion_r967153695.

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+        session_uuid,
+        environment_uuid,
+        project_uuid,
+        pipeline_uuid,
+        pipeline_path,
+        userdir_pvc,
+        project_dir,
+        environment_image,
+    )
+    environment_shell_service_manifest = serv
```
yannickperrenet, line 65: This is a redefinition of `environment_shell_service_manifest`, are you sure you didn't want to append the `serv` to the list?

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca: Indeed that's the reason

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+        return launch_environment_shell(
+            session_uuid,
+            environment_shell_config["environment_uuid"],
+            environment_shell_config["project_uuid"],
+            environment_shell_config["pipeline_uuid"],
+            environment_shell_config["pipeline_path"],
+            environment_shell_config["userdir_pvc"],
+            environment_shell_config["project_dir"],
+            environment_image_string,
+        )
```
yannickperrenet, line 108: In my opinion, this should be placed in the `_collateral` in order to split DB calls from infra changes (as is the idea behind the `TwoPhaseFunction`)

services/base-images/runnable-shared/bootscript.sh
fruttasecca: If this is intended it might be worth it to add a comment in there

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/env python3
+
+# Zero dependency shell spawner program
+# Connects JupyterLab terminal to an SSH daemon
+# running on an environment shell container.
```
fruttasecca, line 5: Might be preferable to write this as a module docstring

services/jupyter-server/shellspawner/shellspawner
fruttasecca: >I think it would be better to check with an if-statement and raise a ValueError given that it is user input we are testing

Agree

services/jupyter-server/shellspawner/shellspawner
fruttasecca: Agree with this, it could also be the case that an old environment image doesn't support shells given that it doesn't have the ssh server. The current surface of the orchest-api allows to know all this information. Also the orchest-webserver is not the right entity to ask for such information regardless, imo, this definitively feels like responsibility of the orchest-api.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+    print("[3] Terminate an existing shell ")
+
+    return collect_choice(4, "Choose option")
+
+
+def choose_existing_environment_shell(
+    environment_shells: List[Dict[str, Any]],
+    environments: List[Dict[str, Any]],
+    prompt_suffix: str,
+) -> Dict[str, Any]:
```
fruttasecca, line 267: This function can actually return `None`

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+
+@api.route("/<string:environment_shell_uuid>")
+class EnvironmentShell(Resource):
+    @api.doc("delete_environment_shell")
+    def delete(self, environment_shell_uuid):
+        """Stop environment shell for a given
+        session_uuid/environment_shell_uuid."""
+        try:
+            stop_environment_shell(environment_shell_uuid)
+            return {}, 201
```
fruttasecca, line 44:
```suggestion
            return {}, 200
```

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+        except Exception as e:
+            return {"message": "%s [%s]" % (e, type(e))}, 500
+
+
+@api.route("/")
+class EnvironmentShellList(Resource):
+    @api.doc("launch_environment_shell")
+    @api.expect(schema.environment_shell_config)
+    @api.marshal_with(schema.environment_shell, code=200)
```
fruttasecca, line 53:
```suggestion
    @api.marshal_with(schema.environment_shell, code=201)
```

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+    def post(self):
+        """Launches an environment shell."""
+        environment_shell_config = request.get_json()
+
+        isess = models.InteractiveSession.query.filter_by(
+            project_uuid=environment_shell_config["project_uuid"],
+            pipeline_uuid=environment_shell_config["pipeline_uuid"],
+        ).one_or_none()
+        if isess is None:
+            return {"message": "Session doesn't exist."}, 409
```
fruttasecca, line 63:
```suggestion
            return {"message": "Session doesn't exist."}, 404
```

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+
+        environment_shell = None
```
fruttasecca, line 65: Is this line needed?

services/orchest-api/app/app/apis/namespace_environment_shells.py
fruttasecca: Agree

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+                )
+        except Exception as e:
+            current_app.logger.error(e)
+            return {"message": str(e)}, 500
+
+        return environment_shell
+
+
+class CreateEnvironmentShell(TwoPhaseFunction):
+    def _transaction(self, environment_shell_config) -> Dict[str, Any]:
```
fruttasecca, line 79: In the transaction `environments.lock_environment_images_for_interactive_session` should be called to make sure the `orchest-api` knowns that the image is in use by the session

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
+from typing import Any, Dict
+
+from flask import request
+from flask.globals import current_app
```
fruttasecca, line 4: Imo a module-level docstring explaining the module scope, purpose, usage etc. will have a good ROI

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
+import time
+from typing import Any, Dict
+
+from _orchest.internals import config as _config
+from app import utils
+from app.connections import k8s_apps_api, k8s_core_api
+from app.core.sessions import _manifests
```
fruttasecca, line 7: Imo the fact that we need to access a "private" module of sessions is a sign that the taxonomy needs to be slightly altered. What I mean with this is that, as it is, environment shells are strictly encapsulated in interactive sessions, but not all code reflects that (like in this case). 

Imo we should reflect if environment shells strictly being part of an interactive session is desirable, meaning that we might choose to have environment-based shells without a backing interactive session in the future, and, if that's the case, the cost to implement that will become higher if the current implementation strongly assumes that env shells are part of an interactive session.

If env shells are part of interactive sessions then this module should probably be moved into the `app.core.sessions` directory

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+            name which is used for pretty-printing and restarts.
+        project_uuid: UUID of the project.
+        pipeline_uuid: UUID of the pipeline.
+        pipeline_path: Relative path (from project directory root) to
+            the pipeline file e.g. 'abc/pipeline.orchest'.
+        userdir_pvc: Name of the k8s PVC e.g. 'userdir-pvc'.
+        project_dir: Name of the project directory e.g. 'my-project'
+            note this is always a single path component.
+        environment_image: The full image specification that can be
+            given directly as the image string to the container runtime.
```
fruttasecca, line 37: The fact that an environment shell depends on a pipeline is a bit strange, I know that this information is needed somewhere but I feel like that the amount of information needed to start, essentially, a pod could be reduced. From a far-off (and perhaps naive) point of view one could say that only the environment, project and session uuids are really needed

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+      pods
+      service_accounts
+      role_bindings
+      roles
+
+    Will be cleaned up when the session is stopped.
+    """
+
+    environment_shell_service_k8s_deployment_manifests = []
+    environment_shell_service_manifest = []
```
fruttasecca, line 53: Are lists actually needed here?

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+        userdir_pvc,
+        project_dir,
+        environment_image,
+    )
+    environment_shell_service_manifest = serv
+    environment_shell_service_k8s_deployment_manifests.append(depl)
+
+    ns = _config.ORCHEST_NAMESPACE
+
+    logger.info("Creating environment shell services deployments.")
```
fruttasecca, line 70: Related to https://github.com/orchest/orchest/pull/1248/files#r968124248, this function only creates 1 deployment and service right?

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+    The resources created in k8s
+      deployments
+      services
+      ingresses
+      pods
+      service_accounts
+      role_bindings
+      roles
+
+    Will be cleaned up when the session is stopped.
```
fruttasecca, line 49: Imo the docstring should specify what's being returned

services/orchest-api/app/app/core/environment_shells.py
```diff
@@ -0,0 +1,151 @@
...
+    ].replace("environment-shell-", "")
+
+    return {
+        "host": environment_shell_service_manifest["metadata"]["name"],
+        "uuid": environment_shell_uuid,
+        "session_uuid": session_uuid,
+    }
+
+
+def get_environment_shells(session_uuid):
```
fruttasecca, line 107: Imo typing should be added for consistency with the rest of the module

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -246,27 +250,14 @@ def _get_session_sidecar_deployment_manifest(
...
-                                    "name": "ORCHEST_SESSION_TYPE",
-                                    "value": session_type,
-                                },
+                            "env": get_orchest_sdk_vars(
+                                project_uuid,
+                                pipeline_uuid,
+                                _config.PIPELINE_FILE,
+                                session_uuid,
+                                session_type,
+                            )
```
fruttasecca, line 259: Given that we are talking about the session sidecar, which doesn't use the orchest SDK, I'm not sure the name of the function fits

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
     }
 
 
+def get_orchest_sdk_vars(
+    project_uuid, pipeline_uuid, pipeline_file, session_uuid, session_type
+):
```
fruttasecca, line 284: Note that the `services/orchest-api/app/app/types.py` module has a `SessionType` enum

fruttasecca:
A bit meta but the PR description could/should have contained a small description of the architectural decisions behind the changes, e.g. "an env shell lifecycle can be different than a session", etc..

I think this particular feature deserves an entry in the "Implementation details" docs

fruttasecca:
A thing to keep in mind is that the current definition of orchest being idle in `services/orchest-api/app/app/apis/namespace_info.py` `is_orchest_idle` includes busy kernels from interactive sessions but doesn't include "busy" environment shells (not even sure we can detect that)

fruttasecca:
I think the cleanup logic in `services/orchest-api/app/app/__init__.py`, `cleanup()` needs to be incremented to include environment shells that are detached from interactive sessions

lib/python/orchest-internals/_orchest/internals/utils.py
ricklamers: Ah good catch, this was a left over of me experimenting with running the user containers (steps, shells, kernel) as UID 1000. The init container needed securityContext runAsUser 0/priviliged true to be able to still use the socket for its work. Since we're not changing to UID 1000 because of root owned disk state in production I can drop this as the parent container is still running as UID 0.

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca, line 316: Is this comment actually needed? It seems to suggest that some action needs to be taken to clean up things while the session shutdown logic will already take care of it AFAIK

lib/python/orchest-internals/_orchest/internals/utils.py
ricklamers: On second thought, I think it might be worth it leaving this in. We'd not be implicitly depending on the parent container of any container that includes the init container to be running as privileged/root. Wdyt?

orchest-sdk/python/orchest/utils.py
ricklamers: Haha still not convinced in this case :D

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +279,139 @@ def _get_session_sidecar_deployment_manifest(
...
+    userdir_pvc: str,
+    project_dir: str,
+    environment_image: str,
+) -> Tuple[Dict[str, Any], Dict[str, Any], Dict[str, Any]]:
+    """
+    This manifest generation is in core/sessions
+    since the environment shell is part of the
+    session, but has a detached lifecycle in the sense
+    that environment shells can be started/stopped
+    independently from the session start/stop.
```
fruttasecca, line 312: I was slightly confused by this comment, I initially interpreted it in a way that suggested that env shells could be started independently of interactive sessions, while it's actually that the lifetime of env shells is within the bounds of the interactive session, env shells are started independently within these bounds but are always stopped when the session is stopped.

services/jupyter-server/shellspawner/shellspawner
```diff
@@ -0,0 +1,371 @@
...
+        self.request_result = None
+
+        super().__init__()
+
+    def run(self):
+        try:
+            self.request_result = self.request_callable(*self.callable_args)
+            self.request_ready = True
+        except (urllib.error.HTTPError, urllib.error.URLError) as e:
+            print(e)
```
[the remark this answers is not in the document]
ricklamers: Makes sense, but in that case I think the entire try/except isn't necessary. Removed it

services/orchest-api/app/app/apis/namespace_environment_shells.py
```diff
@@ -0,0 +1,115 @@
...
+    launch_environment_shell,
+    stop_environment_shell,
+)
+from app.core.environments import get_env_uuids_to_image_mappings
+
+api = Namespace("environment-shells", description="Manage environment shells")
+api = schema.register_schema(api)
+
+
+@api.route("/<string:session_uuid>")
```
fruttasecca, line 23: Perhaps the API endpoints could be improved/made a bit more restful. Currently, we have the following:

- GET `/environment-shells/<uuid>` -> gets all shells of a session
- POST `/environment-shells` -> creates a new shell in a certain *session* which is implicitly inferred by the proj/ppl uuid
- DELETE `/environment-shells/<uuid>` -> deletes an env shell by uuid

So a uuid as the first value after `/environment-shells` is seen as a session uuid in GETs and as a shell uuid in DELETEs. This comment is related to https://github.com/orchest/orchest/pull/1248#discussion_r968114137, the taxonomy and the resulting API endpoints paths should be slightly adjusted imo, to better reflect the structure/architecture

services/jupyter-server/shellspawner/shellspawner
ricklamers: So `orchest-api` doesn't have the environment name information which is needed to show the list to the user.

The assumption we use (because of the gate) that no environment is unavailable if the user is able to start a session (that runs JupyterLab). This guarantee is not currently maintained in a multi-tab context anyway (irrespective of the environment shells feature).

As long as environment names aren't managed at the `orchest-api` level I think using the `orchest-webserver` for fetching the canonical list of environments I think makes sense.

I'm also not a fan of having "another gate" at the level of environment shells. I think that should be handled at the interactive session level which is a required thing to be running when you want to start an environment shell.

services/jupyter-server/shellspawner/shellspawner
ricklamers: I think throwing an error is the responsibility of the check_ready_callable not of the loader.

services/jupyter-server/shellspawner/shellspawner
ricklamers: :D

services/jupyter-server/shellspawner/shellspawner
ricklamers: I don't want to tightly couple the `jupyter-server` and the base images. $NB_USER doesn't currently exist in the context of `jupyter-server` to I think not introducing that is the best option.

The fact that this username is not generic is more of a MVP scope choice. Similar to how we currently don't have extensive docs/support for custom base images (although with enough work a user could technically bring their own custom images by mimicking what we're doing in our base images to the extent necessary - runner etc.).

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: I wanted to make it more explicit which environment variables are part of the SDK to not repeat that everywhere these are injected.

Maybe it's the wrong layer of abstraction. Do you have an idea that isn't repeating the list of environment variables needed for the SDK everywhere?

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: Good point, I'll make the comment more reflective of that fact.

services/orchest-api/app/app/apis/namespace_environment_shells.py
ricklamers: Isn't that handled by the interactive session that the environment shells always depend on?

services/orchest-api/app/app/core/environment_shells.py
ricklamers: The way I see it, currently we want the environment shell to be strictly dependent on interactive sessions but we may move out of that in the future. Hence I prefer "weakly" coupling them like I do now to make that migration in the future less disruptive (no need to "migrate" to new endpoints without the /sessions/  prefix for example).

The context in which I see us becoming independent of interactive sessions in the future is with remote IDE support.

services/orchest-api/app/app/core/environment_shells.py
ricklamers: This is an intended dependency, the environment shell is dependent on a specific pipeline in order to make the `orchest-sdk` function properly.

If we want to make environment shells independent of the pipeline path we'd first need to change how the `orchest-sdk` works which is out of scope imo.

services/orchest-api/app/app/core/environment_shells.py
ricklamers: Removed the return as part of the refactor that moves the launch_ call to collateral of the two phase function.

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: > given that all shells within the same session have the same uuid

They don't?

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: I don't see how the pipeline-file is even used in `jupyter-server`?

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: I think the environment variable should be named to reflect the content not the target program for which the environment variable is intended. It's the pipeline path for the current session. Since we have another similar name that is the actual path (on the system, the one used by the SDK to read the pipeline file at that location) I thought it would be good to describe the difference to avoid confusion.

services/orchest-api/app/app/schema.py
ricklamers: I've updated it to `hostname of environment shell in k8s cluster`. Does that make more sense?

I want to avoid baking "k8s concepts" into the domain model of the environment shell. `hostname` feels more generic for future use.

services/orchest-api/app/app/core/sessions/_manifests.py
yannickperrenet: I guess I was not clear the first time, what I meant to say is that the `metadata["labels"]` (with the current implementation) is not unique for different environment shells. This will lead to different environment shell k8s services will point to the same Pods as backends, meaning that the different services point to the same pods. This in turn means that as a user you might connect to a different shell than you are expected.

image

So the `metadata["labels"]` need to include the full `shell_uuid`.

ricklamers: This should now be fixed given we include a `shell_uuid` label

ricklamers:
> A thing to keep in mind is that the current definition of orchest being idle in `services/orchest-api/app/app/apis/namespace_info.py` `is_orchest_idle` includes busy kernels from interactive sessions but doesn't include "busy" environment shells (not even sure we can detect that)

Good point but out of scope imo. At least for now.

ricklamers:
> A bit meta but the PR description could/should have contained a small description of the architectural decisions behind the changes, e.g. "an env shell lifecycle can be different than a session", etc..
> 
> I think this particular feature deserves an entry in the "Implementation details" docs

Anything in particular you'd like to see documented there?

services/orchest-api/app/app/apis/namespace_environment_shells.py
fruttasecca: Not entirely because while the session is active you could build another image then start an env shell. See `/kernels/lock-image/` in `services/orchest-api/app/app/apis/namespace_sessions.py`. This is needed to allow users to update an environment image and get an up to date kernel/shell without restarting the session

services/orchest-api/app/app/core/sessions/_manifests.py
fruttasecca: Not sure tbh, let's leave it as is

services/jupyter-server/shellspawner/shellspawner
yannickperrenet: @ricklamers Don't see this committed yet? Also fine if you don't disagree with the change of course.

yannickperrenet:
Okay, now that we have the code review out of the way, I have some high level comments (that I ran into when testing the PR).

----

image

- [ ] I think the above message can be improved. As a new user I have no idea what Environment shells are and when seeing this message I will just `exit` instead of realizing I am already in an Environment shell. Because it reads "to ... connect to a shell simply exit out of your current shell". How about changing it to "connect to **another** shell" (without the bold of course).  

----

image

- [ ] On `Ctrl + C` (since we catch the error) there should be an empty line print to have the `Choose option` on a new line.
- [ ] How about catching the `ValueError` and instead just exit the choice-menu silently?
- [ ] I guess the only way for me to now get back into an Environment shell is to start another JupyterLab Terminal? This would be fine with me btw.

----

image

- [ ] Can we make it even clearer to users that they are in an Environment shell right now? How about doing something similar to activated virtual environment indications (what many people configure their shells to do, e.g. `oh-my-zsh`): 
   `jovyan:/project-dir (Environment shell: Python-3 223hdf) $ ll`

----

image

- [ ] Nitpick: Do we want another option to `Exit out of menu`? Otherwise only `Ctrl + C` gets a user to the JupyterLab terminal. Is it on purpose that it is so hard? I think another option in the menu works (given that the help message `exit` or `Ctrl + D` only get "Invalid input").

services/orchest-api/app/app/core/sessions/_manifests.py
```diff
@@ -288,6 +278,146 @@ def _get_session_sidecar_deployment_manifest(
...
+
+    metadata = {
+        "name": service_name,
+        "labels": {
+            "app": "environment-shell",
+            "project_uuid": project_uuid,
+            "session_uuid": session_uuid,
+            "shell_uuid": session_uuid,
+        },
+    }
```
yannickperrenet, line 331: :exclamation: `shell_uuid` should **not** map to `session_uuid` here. It has to point to a uuid unique to the shell, otherwise we run into the issue of 1 service pointing to multiple pods (which is still what happens currently).

services/jupyter-server/shellspawner/shellspawner
ricklamers: Typing is the generic term for entering keys to a system. Pressing implies a specific entry mechanism (one might use the mouse to 'type' Ctrl + D on some system). The quotes I think is more sensible for text because it's a literal string. Whereas Ctrl and D just refer to things, so I didn't think quoting them was necessary.

I'm :100: overthinking this.

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: Woop that's a misatek

services/orchest-api/app/app/core/sessions/_manifests.py
ricklamers: Not being used, removing.

ricklamers:
>  I guess the only way for me to now get back into an Environment shell is to start another JupyterLab Terminal? This would be fine with me btw.

I've added this to a print message. Technically they could also invoke `shellspawner` but I don't want to expose that detail to the user. See my remark below about not wanting to guide toward JupyterLab terminal.

> Nitpick: Do we want another option to Exit out of menu? Otherwise only Ctrl + C gets a user to the JupyterLab terminal. Is it on purpose that it is so hard? I think another option in the menu works (given that the help message exit or Ctrl + D only get "Invalid input").

I actually left that out on purpose. I want it to feel like "this is not what I'm supposed to do". You can get to the "underlying" JupyterLab shell, but everything is "broken" there because you have none of the environment installed dependencies available. I don't want to block it altogether so leaving it in behind a `Ctrl + C` keyboard interrupt seemed like the best trade-off option.

> Nitpick: How about adding an executable called help that users can invoke to re-print the initial greeting message? Whilst testing I something had to think "Hmmm what did I have to do again?". I feel like I am in the habit of always trying something like help in these types of environments, so I though it could be a good addition here.

That message is part of the `shellspawner` program which does not exist in the base images. A separation I'm keen on keeping (we would have to "ship" shellspawner to the base images just for the help message). In addition, the message is on the screen for every terminal you open in JupyterLab. Furthermore, we'd be overriding the built-in `help` of bash. Which some people may like referring to. So I'm not a fan of including our own `help` override.

Great feedback! What I ended up implementing is checkboxed. And I fixed the label bug.

ricklamers:
>I actually left that out on purpose. I want it to feel like "this is not what I'm supposed to do". You can get to the "underlying" JupyterLab shell, but everything is "broken" there because you have none of the environment installed dependencies available. I don't want to block it altogether so leaving it in behind a Ctrl + C keyboard interrupt seemed like the best trade-off option.

Since using `git` from a terminal actually works best from the built-in JupyterLab terminal until we implement `git` ssh key management I think we should leave the option in. This way people who've configured `git` through JupyterLab config (https://docs.orchest.io/en/stable/fundamentals/projects.html#using-git-inside-orchest-projects) can still make use of it through the terminal (the GUI will always work).

services/jupyter-server/shellspawner/shellspawner
yannickperrenet: > The assumption we use (because of the gate) that no environment is unavailable if the user is able to start a session (that runs JupyterLab).

Could you add that as a comment (@ricklamers)? I think it is a great explanation as to why the webserver is queried, not the orchest-api.

yannickperrenet: https://github.com/orchest/orchest/pull/1248/commits/e5476e0ddea924fce8545291707f15f8e5445701

yannickperrenet:
image

I think the current indication of the Environment shell is confusing (especially as a new user who doesn't know what Environment shells). How about: `(Environment: Python-2-typer)`?

We can just explain them they are running inside an Environment, doesn't even need to be `Environment shell` (who cares what we name it). Yes, this has the issue of showing the same indication for two different shells for same Environment, but I think that is just fine.

yannickperrenet:
image

I really like this message when exiting to the JupyterLab terminal! I think it should also be shown when option `[4] Exit to JupyterLab built-in shell` is chosen (currently it just exits and doesn't show this message).

yannickperrenet:
image

As a user it can be confusing that options can no longer be chosen, because we never indicated that the options are only inside Environment shells. Note that the `jovyan` prompt was never shown.

To reproduce:
1. Start session
2. Open JupyterLab terminal
3. Get auto connected to shell
4. `exit` 
5. Start a new shell
6. Wait until you are connected
7. Close the terminal
8. Start a new terminal
9. Note that the `Connecting...` etc. is never shown

ricklamers:
> As a user it can be confusing that options can no longer be chosen, because we never indicated that the options are only inside Environment shells. Note that the jovyan prompt was never shown.

The change that fixes this case is that we auto select the "connect" option for the case when multiple environment shells are active. That feels more consistent with the auto connect action we take when there's only a single environment shell.

> I think the current indication of the Environment shell is confusing (especially as a new user who doesn't know what Environment shells). How about: (Environment: Python-2-typer)?

After discussion decided not to change the prompt.

Added the warning message for option 4. Probably better to overcommunicate.

PR #1250, opened 2022-09-05
Tweak Orchest to better handle times of CPU contention

yannickperrenet:
## Description

This PR basically does two things 

* Set CPU resource requests for Orchest services, e.g. `orchest-api`
* Set DNS resolver `timeout` and `attempts` to `30` seconds and `2` respectively for Orchest services and (Pipeline Run) Step containers.

and recommends an additional thing for single-node deployments:

* Remove `readinessProbe` from `CoreDNS` pod. Given that Orchest doesn't control the pod, we can only hint users towards editing its manifest. Without removing this probe the DNS resolver timeout changes don't really take effect due to `kube-proxy` changing `iptables` rules.

~~I think we can then also merge #1214. What do you guys think?~~ @ricklamers @fruttasecca --> We can't because that would require changing the `initContainer` as it would have to pull in all the different Environment images of the Steps.

### How to test
```python
# Running the two steps below in parallel
# 1. Create 2 steps (CPU intensive and requests making)
# 2. Start run
# 3. Set `docker update minikube --cpus 0.1`
# 4. Wait some time so that the name resolution can fail (the PR should have made this less likely)
# 5. Set back to `docker update minikube --cpus 12` so you can inspect logs

# CPU intensive step
for i in range(int(1e12)):
    x = i**2

# Requests step
import socket
import time

family = socket.AF_INET

for i in range(20):
    *_, sockaddr = socket.getaddrinfo("example.org", 80, family, type=socket.SOCK_STREAM)[0]
    print(sockaddr)
    time.sleep(1)
```

## Checklist

- [x] I have manually tested my changes and I am happy with the result. --> Although rather hard to test.
- [x] Have `orchest-controller` update the `coredns` manifest by deleting the `readinessProbe` on `minikube` installations. ~~Possibly also adding an annotation so that users can disable it.~~
    - [x] Change installation docs accordingly


docs/source/getting_started/installation.md
```diff
@@ -183,6 +186,26 @@ sudo service nginx restart
...
+configure your Kubernetes cluster to scale with respect to the current load.
+
+During times of CPU resource contention, the [CoreDNS](https://coredns.io/) pod could start failing
+its `readinessProbe` leading to `kube-proxy` updating `iptables` rules to stop routing traffic to
+the pod ([k8s
+docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#types-of-probe)), for which
+it uses the `REJECT` target. This means that DNS queries will start failing immediately without the
+configured resolver timeout being respected (in Orchest we use a timeout of `30` seconds with `2`
+attempts). In order to respect the timeout instead of failing immediately, you can tweak the
+`readinessProbe` or simply remove it:
```
ricklamers, line 203: The language is slightly confusing, what does the command do exactly. It says "remove it". Looking at the command with a laymen understanding of `kubectl` it looks like it would remove `coredns` entirely. But from context I assume it would only remove the readinessProbe of `coredns`?

yannickperrenet: Slightly changed the language in https://github.com/orchest/orchest/pull/1250/commits/efdf6afd63349109505a1035470556c1a8fc74d8 :+1: 

In short, it lets the user edit the coredns deployment manifest so they can decide whether they want to tweak or remove the `readinessProbe`

services/orchest-api/app/app/core/pipelines.py
```diff
@@ -542,7 +542,19 @@ def _pipeline_to_workflow_manifest(
...
+            # installed (and then on the internet). This reduced the
+            # number of DNS queries, since we know that these are the
+            # only two valid options when doing pipeline runs.
             "dnsPolicy": "ClusterFirst",
+            "dnsConfig": {
+                "options": [
+                    {"name": "timeout", "value": "30"},  # 30 is max
+                    {"name": "attempts", "value": "2"},  # 5 is max
+                ],
+            },
```
fruttasecca, line 557: Would it make sense to specify this policy for kernel containers and session services as well?

fruttasecca:
Opinion on keeping track on which `k8s` version this has been tested? Maybe as part of the `cpu-contention-dns` section? Might help users in quickly concluding that it might be a `k8s` version issue if this solution breaks because of future `k8s` changes

fruttasecca (approved):
lgtm :)

fruttasecca:
Might also be good to keep track of why the `readinessProbe` is being altered and not the `liveness` one

fruttasecca:
Note that restarting `minikube` leads to the `coredns` deployment being back with the `readinessProbe`

yannickperrenet:
> Might also be good to keep track of why the readinessProbe is being altered and not the liveness one

Good one. Will do :+1: 

> Note that restarting minikube leads to the coredns deployment being back with the readinessProbe, might need to add this to the docs.

Great catch! I had totally missed that, hmmm... Will try to make the change persistent.

> Opinion on keeping track on which k8s version this has been tested? Maybe as part of the cpu-contention-dns section?

I think the assumption is that when something is in the docs for a specific version, then it is supported/tested. Given that we don't add such a note in other sections and are looking to keep supporting this (it is on `stable`), I don't think adding such a note is needed.

yannickperrenet:
> Great catch! I had totally missed that, hmmm... Will try to make the change persistent.

See https://github.com/orchest/orchest/pull/1250/commits/97a16b21978f8684db6be22eb8ae1b917eb057dd. Basically, not possible to make it persistent unless we provide a forked `kubeadm` with changed coredns configuration files that users can use (e.g. `minikube start --bootstrapper=customkubeadm`).

fruttasecca:
>See https://github.com/orchest/orchest/commit/97a16b21978f8684db6be22eb8ae1b917eb057dd. Basically, not possible to make it persistent unless we provide a forked kubeadm with changed coredns configuration files that users can use (in minikube you would have to substitute the shipped kubeadm binary with our compiled kubeadm).

noted

PR #1252, opened 2022-09-11
UI Overhaul Jobs: Job runs

mausworks:
## Description

This PR adds Job Runs to the each job page, and also introduces scoped stores. The lifecycle of scoped stores are automatically maintained. If a store is scoped to a job, all data within it is cleared out if the `job_uuid` is changed in the URL.

## Still TODO

- [x] "Prettified environment variables" when expanding a row
- [ ] Re-fetch the current job runs page when a new run has been started

## Checklist


services/orchest-webserver/client/src/api/job-runs/useJobRunsApi.ts
```diff
@@ -1,97 +1,83 @@
...
   };
 
   return {
+    page: undefined,
     runs: undefined,
-    fetch: async (jobUuid, runUuid) => {
-      const run = await jobRunsApi.fetch(jobUuid, runUuid);
-
-      if (!run) return;
+    fetchOne: choke(async (runUuid) => {
```
iannbing, line 61: While this seems a brilliant idea to preventing redundant requests, I'm wondering how we can solve other issues, such as:
if a user fired the first request, and for some reason the request never reaches the BE (e.g. unstable Internet), then user needs to wait until the request timed out (which could take up to 5 minutes, default in most browser), in order to proceed. Or user needs to refresh the the browser tab.

And this also makes it unpredictable, as the returned promise might not be the promise you intend to create in the first place. And one broken implementation would break all the other use of `fetchOne`.

Isn't this precaution too costly comparing to the merit it brings?

services/orchest-webserver/client/src/jobs-view/job-view/JobSummary.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const JobSummary = ({ job }: JobSummaryProps) => {
+  const { snapshot } = useSnapshot(job.snapshot_uuid);
+  const pipelineUrl = useRouteLink("pipeline", {
+    pipelineUuid: job.pipeline_uuid,
+  });
+
+  return (
+    <Stack direction="row" justifyContent="space-between">
+      <Field name="Schedule">{job.schedule ? "Scheduled" : "Run once"}</Field>
```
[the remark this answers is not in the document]
mausworks: Recurring sounds good. :+1:

services/orchest-webserver/client/src/jobs-view/job-view/JobRuns.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const JobRuns = () => {
+  const [pageNumber, setPageNumber] = React.useState(1);
+  const [pageSize, setPageSize] = React.useState(10);
+  const [fuzzyFilter, setFuzzyFilter] = React.useState("");
+  const [search, setSearch] = React.useState("");
+  const { page } = useJobRunsPage({
+    page: pageNumber,
+    pageSize,
+    fuzzyFilter: fuzzyFilter || undefined,
+  });
```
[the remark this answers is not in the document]
mausworks: I'm thinking polling for now. The most important part is that the run is added after the user clicks "Run job" manually.

services/orchest-webserver/client/src/api/job-runs/useJobRunsApi.ts
mausworks: Great feedback, thanks! :+1: 

> I'm wondering how we can solve other issues, such as […] timeouts

I think that is best be solved within the fetcher itself: long requests should reject after a reasonable time (let's say 1-5 seconds, depending on resource), in which case any affected cached promise will reject too, and the request can be retried.

> And this also makes it unpredictable […]

I'm not sure exactly what you mean. Yes, you might not create a new promise, but that's the idea, right? :thinking: Could you clarify what you mean with "one broken implementation", I'm not sure I understand, thanks! :+1:

> Isn't this precaution too costly comparing to the merit it brings?

I've tried to weigh the pros and cons. It seems like most elaborate HTTP client libraries try to solve request deduplication at the using cache-keys and hooks. Since we use more generic fetcher actions within the store itself—this is the closest analog I can think of: requesting something with the same parameters multiple times (while the first request is pending), should return the same result.

mausworks: To add: I this is just too convenient, and that convenience is what makes it worth it—we can request any data at any component (or hook) without having to worry about over-fetching, the result will be cached (within the scope) and it will be re-used wherever possible. Data-provider hooks become dead simple e.g. `usePipeline(pipelineUuid)` or `useJob(jobUuid)` and we can do both eager fetching and lazy fetching (in combination).

Sure, there may be some caveats, but I'm pretty sure we can overcome them with simple measures.

services/orchest-webserver/client/src/store/scoped.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Creates a zustand store that is bound to the current job.
+ * The store is reset to its initial state if the user switches between jobs.
+ *
+ * Calling the returned hook when a `job_uuid` is not available in the URL
+ * will throw an error, but calling
+ */
+export const createJobStore = <T extends object>(
+  stateCreator: ScopedStateCreator<T, "jobUuid">
+) => createScoped(["jobUuid"], stateCreator);
```
iannbing, line 43: The following flow threw an error.
1. Visit `/jobs` for the first time (created useJobRunsApi store). And scheduled a job.
2. Navigate away, e.g. to `PIPELINES`.
3. Navigate back to `JOBS`. An error is thrown by `useRequiredScope` because `job_uuid` is not yet available in the args. The hook `useSyncJobUuidWithQueryArgs` is responsible filling a valid `job_uuid` when user attempts to visit `/jobs` without a `job_uuid`.

iannbing: `/jobs` is valid without a `job_uuid` since user still can visit the view without any jobs.

yannickperrenet:
image

- [ ] The arrow on the left of the table doesn't seem to add any value, or does it?
- [ ] The `VIEW` text/button is too far right imo. Can't we make this the default action instead of the arrow thingy?
- [ ] The text under `started` isn't correctly vertically aligned.

----

https://user-images.githubusercontent.com/26223174/189633396-6e2aafc4-c0f0-4b5a-abd5-a7183b6f5bbd.mp4

- [ ] User should be taken to the new UI. In addition, pressing any actions/buttons in the view from the video resulting in the app crashing.

----

I have no idea what I did to cause it, but now everytime I go to `JOBS` my application crashes. Looking at the logs of the `orchest-webserver`:
```plain
 ERROR:app:app.py - [12/Sep/2022 10:43:01] - Exception on /catch/api-proxy/api/validations/environments [POST]
   File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
     response = self.full_dispatch_request()
   File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
     rv = self.handle_user_exception(e)
   File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 271, in error_router
     return original_handler(e)
   File "/usr/local/lib/python3.8/site-packages/flask_restful/__init__.py", line 271, in error_router
     return original_handler(e)
   File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
     reraise(exc_type, exc_value, tb)
   File "/usr/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
     raise value
   File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
     rv = self.dispatch_request()
   File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
     return self.view_functions[rule.endpoint](**req.view_args)
   File "/orchest/services/orchest-webserver/app/app/views/orchest_api.py", line 46, in catch_api_proxy_checks_gate
     environment.uuid for environment in get_environments(project_uuid)
   File "/orchest/services/orchest-webserver/app/app/utils.py", line 168, in get_environments
     project_dir = get_project_directory(project_uuid)
   File "/orchest/services/orchest-webserver/app/app/utils.py", line 96, in get_project_directory
     project_path = project_uuid_to_path(project_uuid)
   File "/orchest/services/orchest-webserver/app/app/utils.py", line 454, in project_uuid_to_path
     raise ValueError(f"project {project_uuid} not found.")
 ValueError: project 428a8c97-d63a-4d06-9ed7-1fddb05a7ed9 not found.
```
But that error message is not what I expected based on some recent actions I had taken:
1. Create new project
2. Now the default Environment will start building
3. Go to jobs and try to "create new job"

services/orchest-webserver/client/src/api/job-runs/useJobRunsApi.ts
iannbing: How about the boring solution: fetch only in one place? What is the impediment we are trying to remove?

> It seems like most elaborate HTTP client libraries try to solve request deduplication at the using cache-keys and hooks.

Take `swr` for example, it caches the URL and the response, but it doesn't cache the promise. 

If you're looking for a solution that is React agnostic, perhaps look into [rxjs](https://rxjs.dev/)? Although I'm not a big fan of it, it does solve your problem.

>I think that is best be solved within the fetcher itself: long requests should reject after a reasonable time (let's say 1-5 seconds, depending on resource), in which case any affected cached promise will reject too, and the request can be retried.

I'm not sure touching the default is a good idea. Wouldn't low Internet bandwidth directly result in rejections? Again, we probably don't benefit from implementing this ourselves before trying out all the solutions out there.

services/orchest-webserver/client/src/store/scoped.ts
mausworks: Nice catch! The store should only be initialized when a job has been selected, will fix. :+1:

services/orchest-webserver/client/src/api/job-runs/useJobRunsApi.ts
mausworks: > How about the boring solution: fetch only in one place? Isn't it solve the problem of over-fetching?

Sure—we can do that! :+1: But it's with the condition that this data must always be passed down to child components as `props`, otherwise there is quite a bit of `undefined`-checking that has to be done for each component. I'm absolutely fine with fetching the data at the top level, (i.e. the view) and then passing it down. 

I think that we should do that, regardless of which solution we end up with. :+1:

> I'm not sure touching the default is a good idea. Wouldn't low Internet bandwidth directly result in rejections? 

Yes, yes it would—that's the drawback with timeouts. We could make `choke` itself have a timeout (or max duration), after which it would create a new promise when fetching instead of reusing an old one. Would this resolve your concern? It is trivial to implement. :+1:

> Again, we probably don't benefit from implementing this ourselves before trying out all the solutions out there.

I have spent many nights trying to find something that is compatible with fetcher functions in zustand stores (as they are implemented now) and fetch-deduplication, and I have come up with basically nothing. I've tried both SWR and react-query (which are the two major players), and I've had a look at RxJS as well. They are solutions with a large footprint that either requires a large rewrite or learning a new framework.

I'm really am not trying to reinvent the wheel here—trust me on that one. :sweat_smile: I'm trying to come up with a bare-minimum solution that allows for flexible, robust & worry-free data-provider hooks, that are also compatible with what we have right now. 

Your concern about long/hung requests have been addressed above, and I would love to try this solution out unless you forsee any additional problems with it. I think that this is a simple solution that is worth testing at the very least, if it turns out that it's causing any problems when testing (or god forbid—in production), then I'll be the first person to remove it in favor of a different one.

services/orchest-webserver/client/src/utils/promise.ts
```diff
@@ -0,0 +1,74 @@
...
+  const pending: Record<string, Promise<T>> = Object.create(null);
+
+  return ((...args: Parameters<F>) => {
+    const key = args
+      .map((arg) =>
+        typeof arg === "bigint" ? arg.toString() : JSON.stringify(arg)
+      )
+      .join(", ");
+
+    return (pending[key] ??= fn(...args).finally(() => delete pending[key]));
```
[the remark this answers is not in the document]
mausworks: It's one of my favorite features—super powerful—but rarely seen. I understand your concern. I'll change this a bit. :+1:

services/orchest-webserver/client/src/hooks/useActiveJobRun.ts
```diff
@@ -0,0 +1,40 @@
...
+  }, [cancel, runUuid]);
+
+  React.useEffect(() => {
+    if (runUuid && !currentRun && status !== "PENDING") {
+      run(fetchRun(runUuid)).catch();
+    }
+  }, [fetchRun, currentRun, runUuid, run, status]);
+
+  return {
+    run,
```
iannbing, line 33: Probably you meant to expose `currentRun`?

services/orchest-webserver/client/src/api/job-runs/useJobRunsApi.ts
iannbing: >Sure—we can do that! 👍 But it's with the condition that this data must always be passed down to child components as props, otherwise there is quite a bit of undefined-checking that has to be done for each component. I'm absolutely fine with fetching the data at the top level, (i.e. the view) and then passing it down.

If we take the fetch-at-entry-point approach, I think I can live with having `choke` wrapper and observe how it plays out. 🙂

mausworks: So, I spent a little time reworking `choke` into something more versitile. :ok_hand: 

It is now now (more appropriately) conceptualized as "memoizing pending promises", which is something that have been written about around the internet quite extensively (just Google "promise memoization"). I have taken a look at the available frameworks (memoizee, et al), but they don't seem to cover some of the concerns that you had about "hung promises"; but they do indeed build on the same principles.

This means that `choke` has been renamed to `memoize`, and it now even supports bypassing of memoization entirely: just call `fetchAll.bypass()` instead of `fetchAll()`. This is useful if you want to manually refresh data in a component and want to make sure you don't get a memoized promise back.

Also (as promised) memoized promises can now have a duration set. Once the duration has elapsed, the pending promise is "forgotten", and a new promise will be created on the next call. You can also call `forget()` on the promise to manually delete it from memory.

Memoized promises are now also cancelable by default, so you can configure them to automatically be canceled once they expire, or even cancel them manually with `cancel()`. At this point it might be reasonable to conceptualize these functions as "data loaders" (or similar) since they have more features than just memoization, but that's not a discussion that we need to have right now—let's just see how this pans out first. :+1:

mausworks:
Opening a new PR towards the main overhaul branch.

PR #1253, opened 2022-09-12
Build images on the node directly

fruttasecca:
Makes it so that all image builds (environment and custom jupyter) are built directly on the node. This allows a faster build since the image doesn't need to be pushed to the registry to consider the build complete, and allows starting steps or other pods that depend on the new image right away.

To accomplish this, we distinguish between our two supported container runtimes (`docker` and `containerd`) and use a different builder for a different runtime, `buildx` and `buildkit` respectively. In the case of `buildkit` + `containerd`, bidirectional mounting is needed, and, given the introduced risk, the `buildkit` daemon is not run ephemerally as part of the builder pod but runs as a daemonset. This is to reduce the risk of issues when it comes to the daemonset having to remove temporary mounts etc.

When using `buildx`, we mount the container runtime socket into the builder pod, which will use the `docker buildx` CLI to communicate with the runtime and build. With `buildkit`, we mount the socket of the `buildkit` daemon into the builder pod; the `buildkit` daemon (and daemonset) will mount the socket of the container runtime.

This PR introduces 3 new Orchest images:
- `buildx-builder`
- `buildkit-builder`
- `buildkit-daemon`
 
 As a result of this change the way changes to base images are tested has been simplified.

### DIFF
--- a/lib/python/orchest-internals/_orchest/internals/config.py
+++ b/lib/python/orchest-internals/_orchest/internals/config.py
@@ -43,6 +43,7 @@ REGISTRY_FQDN = f"docker-registry.{ORCHEST_NAMESPACE}.svc.cluster.local"
 CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")
 CONTAINER_RUNTIME_IMAGE = os.environ.get("CONTAINER_RUNTIME_IMAGE")
 CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET")
+ARGO_EXECUTOR_IMAGE = "quay.io/argoproj/argoexec:v3.2.6"
 
 # Ingress configs
 INGRESS_CLASS = os.environ.get("INGRESS_CLASS", "nginx")
--- a/lib/python/orchest-internals/_orchest/internals/config.py
+++ b/lib/python/orchest-internals/_orchest/internals/config.py
@@ -42,7 +42,9 @@ REGISTRY_FQDN = f"docker-registry.{ORCHEST_NAMESPACE}.svc.cluster.local"
 # Container Runtime configs.
 CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")
 CONTAINER_RUNTIME_IMAGE = os.environ.get("CONTAINER_RUNTIME_IMAGE")
-CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET")
+CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET").replace(
+    "unix://", ""
+)
 
 # Ingress configs
 INGRESS_CLASS = os.environ.get("INGRESS_CLASS", "nginx")
--- a/lib/python/orchest-internals/_orchest/internals/config.py
+++ b/lib/python/orchest-internals/_orchest/internals/config.py
@@ -42,9 +42,7 @@ REGISTRY_FQDN = f"docker-registry.{ORCHEST_NAMESPACE}.svc.cluster.local"
 # Container Runtime configs.
 CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")
 CONTAINER_RUNTIME_IMAGE = os.environ.get("CONTAINER_RUNTIME_IMAGE")
-CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET").replace(
-    "unix://", ""
-)
+CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET")
 
 # Ingress configs
 INGRESS_CLASS = os.environ.get("INGRESS_CLASS", "nginx")
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -135,9 +135,14 @@ def _get_image_builder_manifest(
         },
     }
 
-    # For extensions that need access to the settings on install.
+    # Some jupyter extensions might require write access to settings
+    # during the setup script, e.g. on install. Since buildkit and
+    # buildx currently do not support writing to a "bind" volume we let
+    # the container write to its local filesystem then rsync the changes
+    # over.
     if _config.JUPYTER_IMAGE_NAME in image_name:
         container = manifest["spec"]["containers"][0]
+        # Needed to get the existing settings into the build context.
         container["volumeMounts"].append(
             {
                 "name": "userdir-pvc",
@@ -145,6 +150,29 @@ def _get_image_builder_manifest(
                 "subPath": ".orchest/user-configurations/jupyterlab/user-settings",
             }
         )
+        # Needed because the build takes place in the container runtime,
+        # which doesn't have access to the cluster dns.
+        container["env"] = container.get("env", []) + [
+            {
+                "name": "BUILDER_POD_IP",
+                "valueFrom": {"fieldRef": {"fieldPath": "status.podIP"}},
+            }
+        ]
+
+        # To pass the BUILDER_POD_IP env var.
+        container["command"] += ["-a"]
+
+        args = container["args"][0]
+        # Start the ssh server so that the build container can rsync.
+        args = f"/usr/sbin/sshd && {args}"
+        # Need to distinguish between buildkit and buildx args passing.
+        if _config.CONTAINER_RUNTIME == "containerd":
+            args = f"{args} --opt build-arg:BUILDER_POD_IP=$BUILDER_POD_IP"
+        elif _config.CONTAINER_RUNTIME == "docker":
+            args = f"{args} --build-arg=BUILDER_POD_IP=$BUILDER_POD_IP"
+        else:
+            raise ValueError()
+        container["args"] = [args]
 
     return manifest
 
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -8,6 +8,7 @@ from app import errors, utils
 from app.connections import k8s_core_api
 from config import CONFIG_CLASS
 
+
 def image_build_task_to_pod_name(task_uuid: str) -> str:
     return f"image-build-task-{task_uuid}"
 
@@ -133,11 +134,6 @@ def _get_image_builder_manifest(
             + _IMAGE_BUILDER_VOLUMES,
         },
     }
-
-    if CONFIG_CLASS.DEV_MODE:
-        # ENV_PERF_TODO: verify that it "just works" since the runtime
-        # socket is already mounted by default
-        ...
 
     # For extensions that need access to the settings on install.
     if _config.JUPYTER_IMAGE_NAME in image_name:
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -145,14 +145,6 @@ def _get_image_builder_manifest(
                 "subPath": ".orchest/user-configurations/jupyterlab/user-settings",
             }
         )
-        # ENV_PERF_TODO: settings should be copied by the dockerfile
-        # container["args"][0] = container["args"][0].replace(
-        #     "buildah build",
-        #     (
-        #         "buildah build -v "
-        #         "/jupyterlab-user-settings:/root/.jupyter/lab/user-settings "
-        #     ),
-        # )
 
     return manifest
 
--- a/services/orchest-api/app/app/core/image_utils.py
+++ b/services/orchest-api/app/app/core/image_utils.py
@@ -5,38 +5,61 @@ from kubernetes import watch
 from _orchest.internals import config as _config
 from _orchest.internals.utils import get_userdir_relpath
 from app import errors, utils
-from app.connections import k8s_core_api, k8s_custom_obj_api
+from app.connections import k8s_core_api
 from config import CONFIG_CLASS
-
-# This way the builder pod is always scheduled on the same node as the
-# registry to have quicker pushes. Moreover, the fact that the builder
-# pod always runs on the same node allow us to make use of on disk cache
-# for layers, which is mounted as a volume.
-_registry_pod_affinity = {
-    "podAffinity": {
-        "requiredDuringSchedulingIgnoredDuringExecution": [
-            {
-                "labelSelector": {
-                    "matchExpressions": [
-                        {
-                            "key": "app",
-                            "operator": "In",
-                            "values": ["docker-registry"],
-                        }
-                    ]
-                },
-                "topologyKey": "kubernetes.io/hostname",
-            }
-        ]
-    }
-}
-
 
 def image_build_task_to_pod_name(task_uuid: str) -> str:
     return f"image-build-task-{task_uuid}"
 
 
-def _get_image_builder_workflow_manifest(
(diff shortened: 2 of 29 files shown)


services/orchest-api/app/config.py
```diff
@@ -36,7 +36,13 @@ class Config:
     CLIENT_HEARTBEATS_IDLENESS_THRESHOLD = datetime.timedelta(minutes=30)
 
     # Image building.
-    IMAGE_BUILDER_IMAGE = "docker.io/orchest/buildah:1.26.0-patched"
+    _RUNTIME_TO_IMAGE_BUILDER = {
+        "docker": f"docker.io/orchest/image-builder-buildx:{ORCHEST_VERSION}"
+    }
+    IMAGE_BUILDER_IMAGE = _RUNTIME_TO_IMAGE_BUILDER[_config.CONTAINER_RUNTIME]
```
yannickperrenet, line 43: Nitpick: The `_config.CONTAINER_RUNTIME` could be `None`. 

We don't really have an answer for this currently anywhere in the app. When we read certain environment variables, e.g.:
    `CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")`
Do we then proceed if it doesn't exist? Or do we want the application to fail? 

Given that we are certain that the environment variable should exist, maybe we should add `assert` statements? Or just do `os.environ["name"]`?

utility-containers/image-builder-buildkit/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM moby/buildkit:latest
```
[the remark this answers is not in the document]
yannickperrenet: And do we not need to add `ARG` etc. to set the `ORHEST_VERSION`?

lib/python/orchest-internals/_orchest/internals/config.py
```diff
@@ -42,7 +42,9 @@
 # Container Runtime configs.
 CONTAINER_RUNTIME = os.environ.get("CONTAINER_RUNTIME")
 CONTAINER_RUNTIME_IMAGE = os.environ.get("CONTAINER_RUNTIME_IMAGE")
-CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET")
+CONTAINER_RUNTIME_SOCKET = os.environ.get("CONTAINER_RUNTIME_SOCKET").replace(
+    "unix://", ""
+)
```
yannickperrenet, line 47: How about adding the commit message as a comment here?

"Some k8s flavours place "unix://" in front of the path."

services/orchest-controller/pkg/controller/orchestcomponent/node_agent.go
```diff
@@ -75,6 +76,8 @@ func getNodeAgentDaemonset(registryIP string, metadata metav1.ObjectMeta,
 	var one int64 = 1
 
 	socketPath := utils.GetKeyFromEnvVar(component.Spec.Template.Env, "CONTAINER_RUNTIME_SOCKET")
+	// Apparently needed for old/existing clusters?
+	socketPath = strings.Replace(socketPath, "unix://", "", -1)
```
yannickperrenet, line 80: @nhaghighat?

Might also be a good idea to add the specific Orchest versions here, e.g.:
`// Apparently needed for old/existing clusters pre v2022.08.11?`

yannickperrenet: I have been seeing this `replace` statement in numerous places. Are you sure this is the best way to do this?

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -31,6 +31,7 @@ var (
 		controller.AuthServer,
 		controller.OrchestWebserver,
 		controller.NodeAgent,
+		controller.BuildKitDaemon,
```
yannickperrenet, line 34: How about swapping this order and made the `NodeAgent` last? 

In the case where a user can already connect to Orchest during the installation process (for some reason), then I think we want to make sure that the user can already build Environments (whereas the node-agent isn't vital to be up for Orchest to work).

fruttasecca: Sounds good

services/orchest-controller/pkg/controller/orchestcomponent/node_agent.go
fruttasecca: I think the one in the `orchest-api` can be removed, the ones at the controller level not so sure

services/orchest-controller/pkg/controller/orchestcomponent/buildkit_daemon.go
```diff
@@ -0,0 +1,206 @@
...
+							Name:      "containerd-socket",
+							MountPath: socketPath,
+						},
+						// TODO: allow customizing this for k8s
+						// flavours that may have containerd in non
+						// standard locations (k3s).
+						{
+							Name:      "run-containerd",
+							MountPath: "/run/containerd",
+						},
```
yannickperrenet, line 172: Not too familiar with this part, but can't it make use of:
```sh
orchest install --socket-path=/var/snap/microk8s/common/run/containerd.sock
```
which we use for the image-puller.

@nhaghighat Also a question for you I think :)

lib/python/orchest-internals/_orchest/internals/config.py
fruttasecca: I'm going to remove this `replace` (https://github.com/orchest/orchest/pull/1253#discussion_r969298034), but will add a comment at the controller level replace if it's not there already

utility-containers/image-builder-buildkit/Dockerfile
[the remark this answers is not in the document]
fruttasecca: Good catch about the `ORCHEST_VERSION`, will apply it to both the buildkit and buildx builders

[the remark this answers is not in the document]
fruttasecca: >latest

We can indeed move to something "safer" like `stable` or `10.` (https://hub.docker.com/r/moby/buildkit/tags)

services/orchest-api/app/config.py
fruttasecca: >Nitpick: The _config.CONTAINER_RUNTIME could be None.

Just realized that :/, will add an assertion since the issue will be clearer

>Do we then proceed if it doesn't exist? Or do we want the application to fail?

Imo we have to fail here

services/orchest-controller/pkg/controller/orchestcomponent/buildkit_daemon.go
fruttasecca: >Not too familiar with this part, but can't it make use of:

Generally speaking, that's probably a yes, but given the mount requirements that `buildkit` + `containerd` has it's not enough for the daemonset. I.e. custom mount paths need to be specified

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -104,60 +122,41 @@ def _get_image_builder_workflow_manifest(
...
-            {"name": "dockersock", "mountPath": "/var/run/docker.sock"}
-        )
-        container["args"][0] = container["args"][0].replace(
-            "buildah build",
-            # Check if there is a newer image, if so, pull it.
-            "buildah build --pull=true",
-        )
+        # ENV_PERF_TODO: verify that it "just works" since the runtime
+        # socket is already mounted by default
+        ...
```
yannickperrenet: Would indeed be great if we can easily build Environments when developing like this (should not forget to update the docs if that is the case).

services/orchest-api/app/app/core/image_utils.py
```diff
@@ -200,8 +205,12 @@ def __init__(
...
         self.user_logs_file_object = user_logs_file_object
         self.complete_logs_file_object = complete_logs_file_object
-        self.copying_regex = re.compile(r"^STEP\s+\d+\/\d+:\s+COPY.*")
-        self.userscript_begin_regex = re.compile(r"^STEP\s+\d+\/\d+:\s+RUN.*")
+        self.pulling_regex = re.compile(r"^\s*#\d+\s*sha256:")
+        self.copying_regex = re.compile(r"^\s*#\d+\s*\[\s*\d+\/\d+\s*\]\s*COPY.*")
+        self.userscript_begin_regex = re.compile(
+            r"^\s*#\d+\s*\[\s*\d+\/\d+\s*\]\s*RUN.*"
+        )
+        self.setup_script_cached_regex = re.compile(r"^\s*#\d+\s*CACHED")
```
yannickperrenet, line 201: Good lord... does there happen to be a guarantee on this (output) syntax given by the buildx/buildkit folks?

fruttasecca: I don't think such a guarantee is included, the only other option is to not do this kind of ""parsing"" and output raw logs. Altought `buildx` wants to have interface parity with `buildkit` so I don't see this changing soon. Pinning the image version should give us enough breathing room

services/orchest-api/app/app/core/image_utils.py
yannickperrenet: Fixed in: https://github.com/orchest/orchest/pull/1253/commits/54b412b84519c1c3fee74f6e5350a5762624dfca :100:

services/orchest-api/app/app/core/jupyter_image_builds.py
```diff
@@ -156,6 +163,17 @@ def prepare_build_context(task_uuid):
...
+    # Copy the settings into the context to make them available during
+    # build. It's done in this "simple" manner to be compatible with all
+    # container runtimes we use to build. This is done for extensions
+    # that need access to (READ) the settings on install. Writes are
+    # currently not supported since both buildkit and buildx do not
+    # support binding a writable directory.
+    copytree(
+        "/userdir/.orchest/user-configurations/jupyterlab/user-settings",
+        os.path.join(snapshot_path, "_orchest_configurations_jupyterlab_user_settings"),
+    )
```
yannickperrenet, line 175: So then the question is how do we get these build artifacts out of the container again, right? Which is what we are considering `rsync` for?

fruttasecca: Indeed, I'm implementing that right now in a separate PR

utility-containers/image-builder-buildkit/Dockerfile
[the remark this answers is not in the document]
yannickperrenet: > We can indeed move to something "safer"

I'd like to just move to a pinned version. Would be a shame if everything works during testing but then during the release process a new `stable` tag is released which breaks Orchest.

services/orchest-api/app/app/core/image_utils.py
yannickperrenet: > Pinning the image version should give us enough breathing room

That is indeed true! In that case I think its fine to leave it as is :+1:

yannickperrenet:
From my high-level understanding of what is being done in this PR, I think it looks good. The exact impact/requirement of the mounts I can only vaguely guess, but overall I don't see anything wrong with the approach. :smile_cat:

utility-containers/image-builder-buildkit/Dockerfile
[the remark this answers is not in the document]
fruttasecca: Indeed

services/orchest-api/app/config.py
fruttasecca: https://github.com/orchest/orchest/pull/1253/commits/60029d7bef0cb3cfb994a6362deaf0ca5ca4e55a

services/orchest-api/app/config.py
```diff
@@ -3,6 +3,10 @@
 
 from _orchest.internals import config as _config
 
+assert (
+    _config.CONTAINER_RUNTIME is not None
+), "CONTAINER_RUNTIME should be set for the orchest-api and celery-worker."
```
yannickperrenet, line 8: Maybe better to put this in the `_config` library itself? Feels like this is the wrong place

fruttasecca: I might be missing the context around which the `CONTAINER_RUNTIME` could be `None` then, I thought that there are some of our services that do have the internal lib but do not have the `CONTAINER_RUNTIME` environment variable set, and that's the reason for using a `get` instead of indexing the dictionary when retrieving it

lib/python/orchest-internals/_orchest/internals/config.py
fruttasecca: https://github.com/orchest/orchest/pull/1253/commits/52439ee09d1e2315c9c1854522a92b7a77d9876b

services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
fruttasecca: https://github.com/orchest/orchest/pull/1253/commits/1e8644bd89f621fe8449c6859cdcdfb4bb8fd09b

services/orchest-api/app/config.py
fruttasecca: These should the containers that have the internal lib:
```
    "base-kernel-py"
    "base-kernel-r"
    "base-kernel-julia"
    "base-kernel-javascript"
    "orchest-api"
    "orchest-webserver"
    "session-sidecar"
    "auth-server"
    "celery-worker"
    "jupyter-enterprise-gateway"
    "node-agent"
```

services/orchest-controller/pkg/controller/orchestcomponent/buildkit_daemon.go
```diff
@@ -0,0 +1,206 @@
...
+							Path: "/run/containerd",
+							Type: &hostPathDirectoryOrCreate,
+						},
+					},
+				},
+				{
+					Name: "var-lib-containerd",
+					VolumeSource: corev1.VolumeSource{
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/var/lib/containerd",
```
nhaghighat, line 97: Here we might be able to use the socket path provided by the `CLI`

services/orchest-controller/pkg/controller/orchestcomponent/node_agent.go
nhaghighat: If you fix the `CONTAINER_RUNTIME_SOCKET` in `cluster_controller.go` then I think there is no need to fix it also.

services/orchest-controller/pkg/controller/orchestcomponent/buildkit_daemon.go
fruttasecca: If the paths to the storage of `containerd` are by default non-standard then the mount paths can't remain as is for some k8s flavours like k3s unfortunately

nhaghighat: As discussed offline, this section for k3s could remain the same, only the volume part has to change, @fruttasecca: could you move the comment to the volumes section to avoid confusion in the future?

services/orchest-api/app/config.py
yannickperrenet: Woops... you are totally right, my bad :grimacing:

services/orchest-controller/pkg/controller/orchestcomponent/buildkit_daemon.go
fruttasecca: https://github.com/orchest/orchest/pull/1253/commits/92bcac10025122a3d51a69788acd7cf7f377bd91

nhaghighat (approved):

PR #1254, opened 2022-09-13
Adopt Orchest PVCs with no ownerReferences

nhaghighat:
## Description

This PR adopts the PVCs if the owner of PVC is `nil`, technically this should never happen. Still, it has been added because of the way https://velero.io/ creates a backup, the PVCs that Valero backups don't maintain the original ownerRefrences.

## Checklist


yannickperrenet:
@alexsanderp :angel:

yannickperrenet (approved):

PR #1256, opened 2022-09-13
Write to JupyterLab user settings during custom Jupyter server build

fruttasecca:
## Description

When using `buildah` to build images we could mount (RW) the jupyterlab settings from the builder to the container in which the build takes place, this is not possible anymore with `buildkit`. To circumvent that, the user settings are `rsync`'d from the container where the build takes place to the builder pod, which mounts the jupyter settings. 

To do that, a ssh server is started on the builder pod when building jupyter.


yannickperrenet (approved):

fruttasecca:
Tested with both container runtimes we support

PR #1260, opened 2022-09-13
UI Overhaul: Job runs

mausworks:
## Description

The old PR was #1252, this is pretty much the same.

## Checklist


services/orchest-webserver/client/src/hooks/useConfirm.ts
```diff
@@ -0,0 +1,53 @@
...
+      cancelLabel,
+      confirmButtonColor,
+      confirmLabel,
+      content,
+      setConfirm,
+      onCancel,
+      title,
+    ]
+  );
+};
```
iannbing, line 53: I'm not sure if the implementation is complete, as it made quite some compromises, and probably doesn't apply to all cases. Things to consider:

- action for `onCancel` is not exposed.
- fire-and-forget confirmation

```
const isDone = await setConfirm(
   "Fire and forget request", 
   "The dialog should not block user", 
   async (resolve) => { 
        resolve(true); // resolve the promise for the dialog without awaiting.
        await longRequest();
        return true;
    }
)

if(isDone) console.log("Long request is done!");

```

I think it's nice to wrap `setConfirm` per case to simplify feature-specific code, but this hook doesn't seem to be ready for generic purposes.

services/orchest-webserver/client/src/jobs-view/job-view/JobSummary.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const JobSummary = ({ job }: JobSummaryProps) => {
+  const { snapshot } = useSnapshot(job.snapshot_uuid);
+  const pipelineUrl = useRouteLink("pipeline", {
+    pipelineUuid: job.pipeline_uuid,
+  });
+
+  return (
+    <Stack direction="row" justifyContent="space-between">
+      <Field name="Schedule">{job.schedule ? "Recurring" : "Run once"}</Field>
```
iannbing, line 23: It's not in the design, but perhaps showing something like "Every minute" or "At 12:00 AM, only on Sunday" would be more useful than "Recurring"? 🙂

iannbing:
Some styling isue.
Screenshot 2022-09-13 at 21 01 15

iannbing (approved):
Left some comments. LGTM! Good job on implementing so much more than asked 💪 .

PR #1261, opened 2022-09-14
Automatically create a Pipeline on Project creation (if applicable)

yannickperrenet:
## Description

As per title.

Fixes: #948


astrojuanlu:
+1. Saves one action from the user, and I can't think of use cases of a project without pipelines.

Wondering how this affects feature discovery, e.g. if the user never has to create the first pipeline, maybe they're less likely to create a second one. But I'm only guessing, and in any case I believe this is a positive change.

PR #1262, opened 2022-09-15
Shift responsibility of kernels pod creation to `orchest-api`

fruttasecca:
## Description

Preliminary work needed for pod scheduling at the `orchest-api` level:
- have the `orchest-api` be notified of an image having been pulled to a node
- shift responsibility of kernels pod creation to orchest-api

### DIFF
--- a/services/node-agent/app/image_puller.py
+++ b/services/node-agent/app/image_puller.py
@@ -173,10 +173,11 @@ class ImagePuller(object):
                     self.logger.warning(f"Image '{image_name}' was not downloaded!")
                 except Exception as ex:
                     self.logger.warning(
-                        f"Attempt {retry} to pull image "
-                        f"'{image_name}' failed with "
-                        f"exception - retrying. Exception was: {ex}."
+                        f"Attempt {retry} to pull image '{image_name}' and notify the "
+                        "orchest-api failed with exception - retrying. "
+                        f"Exception was: {ex}."
                     )
+
             queue.task_done()
 
     async def run(self):
--- a/services/node-agent/app/image_puller.py
+++ b/services/node-agent/app/image_puller.py
@@ -87,7 +87,69 @@ class ImagePuller(object):
         # pulled are not pulled concurrently again.
         self._curr_pulling_imgs = set()
 
-    async def get_image_names(self, queue: asyncio.Queue):
+    async def _enqueue_pre_pull_orchest_images(
+        self, session: aiohttp.ClientSession, queue: asyncio.Queue
+    ):
+        endpoint = f"{self.orchest_api_host}/api/ctl/orchest-images-to-pre-pull"
+        async with session.get(endpoint) as response:
+            response_json = await response.json()
+            for image_name in response_json["pre_pull_images"]:
+                await queue.put((image_name, False))
+
+    async def _enqueue_active_environment_images(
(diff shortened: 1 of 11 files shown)


services/node-agent/app/image_puller.py
```diff
@@ -130,6 +165,10 @@ async def pull_image(self, queue: asyncio.Queue):
                 try:
                     self.logger.info(f"Pulling image '{image_name}'...")
                     if await self.container_runtime.download_image(image_name):
+                        async with aiohttp.ClientSession(trust_env=True) as session:
+                            await self.notify_orchest_api_of_image_pull(
+                                session, image_name
+                            )
```
yannickperrenet, line 171: Nitpick: Given that `notify_orchest_api_of_image_pull` or one of the two private function it uses could raise an Error, the logging statement in the `except` block needs to be updated. Because it could be that it actually managed to `pull` the image, it just wasn't able to notify the orchest-api of that.

services/orchest-api/app/app/apis/namespace_sessions.py
```diff
@@ -395,3 +391,134 @@ def _collateral(
...
+        )
+        _, env_uuid, _ = _utils.env_image_name_to_proj_uuid_env_uuid_tag(
+            kernel_spec["kernel_image"]
+        )
+        env_image = environments.lock_environment_images_for_interactive_session(
+            project_uuid, pipeline_uuid, set([env_uuid])
+        )[env_uuid]
+        registry_ip = k8s_core_api.read_namespaced_service(
+            _config.REGISTRY, _config.ORCHEST_NAMESPACE
+        ).spec.cluster_ip
```
yannickperrenet, line 413: We actually have a function for this: `get_registry_ip()` in `utils.py`.

yannickperrenet:
Don't really have anything to add here, besides saying that I like what I am seeing and that it "feels" (based on the code changes) like a good decision to move kernel starting responsibility to the `orchest-api`.

Great job! :fire:

services/orchest-api/app/app/apis/namespace_sessions.py
fruttasecca: Will fix

services/node-agent/app/image_puller.py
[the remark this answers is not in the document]
fruttasecca: You are correct, this is actually a fault in the PR, nice catch, will fix!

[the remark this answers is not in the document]
fruttasecca: Fixed the issue here: https://github.com/orchest/orchest/pull/1262/commits/6b4084e10e6db8963d4c39cd24513db45a6c7cad + https://github.com/orchest/orchest/pull/1262/commits/0fdd40ea1daf2b2648888b1e20fb6725e761df86. I didn't go for a batched approach since it (imo) made the logic more complex and puts more load on the DB given that we would end up needing to upsert records that already exist most of the time. Essentially the image puller will query for two sets of images; active images that are on the registry and active images that are on the registry but (according to the orchest-api) not in the node. All active images on the registry (first set) are pulled, while the orchest-api is only notified about images in the second set. This way images are pulled again if GC'd by k8s but we also notify the orchest-api correctly and will less overhead.

PR #1263, opened 2022-09-15
Alter pods scheduling based on image location across the cluster

fruttasecca:
## Description

This PR modifies the scheduling behaviour of all k8s resources that could involve environment images or custom jupyter images. It uses the fact that the `orchest-api` knows that a certain image is on a certain node and if it has been pushed to the registry already to schedule everything that is part of the "interactive" scope of Orchest (interactive pipeline runs, sessions, services, shells, etc.) to nodes that already have the image to avoid the user waiting for the image to be pulled. The "non-interactive" scope is less constrained.

For the time being the module in charge of injecting these changes in these resources manifests also injects the pre-pull init container in both interactive and non-interactive scopes, to account for the fact that k8s GC could remove an image from the node in case of disk pressure.

TODOS:
- [X] write a proper PR description
- [x] module docstring
- [x]  ~modify the pre-pull container to pull multiple images(?)~ postponed
- [x] test with multi-node
- [x] ~check for low-hanging fruits when it comes to getting rid of the pre-pull init container considering the fact that images could be GC by k8s in case of disk pressure~ postponed


services/orchest-api/app/app/core/pod_scheduling.py
```diff
@@ -0,0 +1,296 @@
...
+    return {
+        "nodeAffinity": {
+            "requiredDuringSchedulingIgnoredDuringExecution": {
+                "nodeSelectorTerms": [
+                    {
+                        "matchFields": [
+                            {
+                                "key": "metadata.name",
+                                "operator": "In",
+                                "values": [node_name],
```
yannickperrenet, line 66: The note/caveat was that this list is only allowed to contain 1 element, right? If that is the case, then adding that as a comment might be a good idea.

services/orchest-api/app/app/core/pod_scheduling.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def _is_image_of_interest(image: str) -> bool:
+    return "orchest-env" in image or _config.JUPYTER_IMAGE_NAME in image
+
+
+def _should_constrain_to_nodes(scope: str, image_is_in_registry: bool) -> bool:
+    # interactive -> always constrained to nodes which have the image.
+    # ron-interactive -> only constrained to nodes which have the image
+    # if the image is not in the registry.
```
yannickperrenet, line 83: typo: `ron-interactive` -> `non-interactive`

In addition, `non-interactive` runs don't get constrained whatsoever right? The current language is slightly confusing imo (`only` vs `always`?)

services/orchest-api/app/app/core/pod_scheduling.py
yannickperrenet: I see you added that comment to the function caller. 

What do you think about changing the `_get_node_name_hard_affinity(node_name: str)` function to:

```python
def _get_node_name_hard_affinity(node_names: List[str]) -> dict:
    node_name = random.choices(node_names)
    ...
```
Feels like the fact that we need to choose a random node is the responsibility of the function, not the caller. Because later we might not have to do a random pick, so it would be a change if we had to change all function invocations.

services/orchest-api/app/app/core/pod_scheduling.py
```diff
@@ -0,0 +1,296 @@
...
+
+def _modify_deployment_pod_scheduling_behaviour(
+    scope: str, manifest: Dict[str, Any]
+) -> None:
+    if manifest["kind"] != "Deployment":
+        raise ValueError("Expected a deployment manifest.")
+    spec = manifest["spec"]["template"]["spec"]
+    if len(spec["containers"]) > 1:
+        raise ValueError("Expected a single container in the deployment.")
+    image = spec["containers"][0]["image"]
```
yannickperrenet, line 179: With the merge of using the `containerSet` (of Argo Workflows) for interactive runs, all Steps will run within the same Pod. So I guess the `_get_required_affinity` has to be checked for all containers/steps actually?

Given that the `containerSet` is only used in the case of running single-node, it is not really an issue since the affinity doesn't have to be set in the first place. I will let you decide whether you want that to impact this code. Just wanted to bring it up :+1:

yannickperrenet: Nvm... I see you already thought about this. Just had to read on a bit further :P

services/orchest-api/app/app/core/pod_scheduling.py
```diff
@@ -0,0 +1,296 @@
...
+        pod_spec_patch_param = None
+        image = None
+        for param in parameters:
+            if param["name"] == "image":
+                image = param["value"]
+                break
+            if param["name"] == "pod_spec_patch":
+                pod_spec_patch_param = param
+                pod_spec_patch = json.loads(param["value"])
+                break
```
yannickperrenet, line 269: Inside the `if` there shouldn't be a `break`, because then we will always hit the case where either: `image` isn't set OR `pod_spec_patch` isn't set.

services/orchest-api/app/app/core/pod_scheduling.py
```diff
@@ -0,0 +1,296 @@
...
+
+        pod_spec_patch_param["value"] = json.dumps(pod_spec_patch)
+
+
+def modify_pipeline_scheduling_behaviour(scope: str, manifest: Dict[str, Any]) -> None:
+    if manifest["kind"] != "Workflow":
+        raise ValueError("Expected a workflow manifest.")
+    if CONFIG_CLASS.SINGLE_NODE:
+        _modify_pipeline_scheduling_behaviour_single_node(scope, manifest)
+        return
```
yannickperrenet, line 295: Total nitpick: might as well: `return _modify_p...` to reduce 1 line of code (serious business :rofl:)

services/orchest-api/app/app/core/pod_scheduling.py
fruttasecca: Makes, sense, will do

services/orchest-api/app/app/core/pod_scheduling.py
fruttasecca: >non-interactive runs don't get constrained whatsoever right

They can get constrained if the image is on a node but not in the registry, because otherwise the image puller would fail

fruttasecca: About the comment being confusing, I'll make a module docstring and improve comments

services/orchest-api/app/app/core/pod_scheduling.py
fruttasecca: Good catch, will fix

PR #1268, opened 2022-09-19
Fix/remove nginx if annotation is present

nhaghighat:
## Description

The controller has to remove the `Nginx` from the list of applications if present in `OrchestCluster` CRD. also the detection logic of whether the `Nginx` is required or not, is simplified a bit.

Fixes: #issue

## Checklist

### DIFF
--- a/services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
+++ b/services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
@@ -558,7 +558,7 @@ func (occ *OrchestClusterController) setDefaultIfNotSpecified(ctx context.Contex
 			copy.Spec.Applications = applications
 			changed = true
 		}
-	} else if isIngressAddonRequired(ctx, occ.Client()) {
+	} else if isIngressAddonRequired(ctx, occ.k8sDistro, occ.Client()) {
 		ingressEnabled := false
 		for _, app := range copy.Spec.Applications {
 			if app.Name == addons.IngressNginx {
--- a/services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
+++ b/services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
@@ -544,7 +544,21 @@ func (occ *OrchestClusterController) setDefaultIfNotSpecified(ctx context.Contex
 		copy.Spec.Applications = occ.config.DefaultApplications
 	}
 
-	if !isIngressDisabled(copy) && isIngressAddonRequired(ctx, occ.Client()) {
+	if isIngressDisabled(copy) {
+		ingressEnabled := false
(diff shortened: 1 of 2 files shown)


services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go
```diff
@@ -523,7 +518,7 @@ func isIngressAddonRequired(ctx context.Context, client kubernetes.Interface) bo
 	// In k3s ingress addon can be installed by us
 	switch k8sDistro {
 	case utils.K3s, utils.EKS, utils.GKE:
-		return true
+		return !isNginxIngressInstalled(ctx, client)
```
fruttasecca, line 521: >// In k3s ingress addon can be installed by us

Is the comment outdated? Can we install it also in `EKS` and `GKE`?

services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
```diff
@@ -544,7 +544,21 @@ func (occ *OrchestClusterController) setDefaultIfNotSpecified(ctx context.Contex
 		copy.Spec.Applications = occ.config.DefaultApplications
 	}
 
-	if !isIngressDisabled(copy) && isIngressAddonRequired(ctx, occ.Client()) {
+	if isIngressDisabled(copy) {
+		ingressEnabled := false
+		applications := make([]orchestv1alpha1.ApplicationSpec, 0, 0)
+		for _, app := range copy.Spec.Applications {
+			if app.Name != addons.IngressNginx {
+				ingressEnabled = true
```
fruttasecca, line 552: Should this be a `==`?

services/orchest-controller/pkg/controller/orchestcluster/cluster_controller.go
```diff
@@ -544,7 +544,21 @@ func (occ *OrchestClusterController) setDefaultIfNotSpecified(ctx context.Contex
...
+				continue
+			}
+			applications = append(applications, app)
+		}
+		if ingressEnabled {
+			copy.Spec.Applications = applications
+			changed = true
+		}
```
fruttasecca, line 560: Since we assign to `Applications` instead of appending to it, could we run the risk of overwriting some existing content?

nhaghighat: We need to assign the applications, because we want to remove the `nginx`. k8s makes sure we are using the latest version the object, otherwise rejects the update.

fruttasecca: I see, makes sense. Could you add a comment in there about the reason so that this knowledge doesn't get lost in time?

fruttasecca (approved):
Haven't run the code myself but lgtm :+1:

PR #1273, opened 2022-09-21
Update file tree when pipeline is deleted

mausworks:
## Description

I wasn't able to reproduce this issue before, because I didn't know that you could delete a pipeline from this location.
This PR ensures that the file tree is updated when the pipeline is deleted from the "more actions" context menu.

## Checklist


iannbing (approved):
LGTM!

PR #1276, opened 2022-09-22
Fix/disable "Pause Job" and "Trigger a new Run now" for one-off jobs

iannbing:
## Description

Hide "Pause Job" and "Trigger a new Run now" options for a recurring job, as they are recurring-jobs-only features. This PR also set default `max_retained_pipeline_runs` to 250.

### DIFF
--- a/services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobActionMenu.tsx
+++ b/services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobActionMenu.tsx
@@ -34,18 +34,22 @@ export const useJobActionMenu = () => {
               action: startEditingActiveCronJob,
             }
           : undefined,
-        {
-          label: hasPaused ? "Resume job" : "Pause job",
-          icon: hasPaused ? "resume" : "pause",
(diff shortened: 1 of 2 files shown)


services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobActionMenu.tsx
```diff
@@ -34,18 +34,22 @@ export const useJobActionMenu = () => {
...
-          label: "Trigger job now",
-          icon: "run",
-          disabled: !isAllowedToTriggerScheduledRuns,
-          action: triggerJobNow,
-        },
+        isCronJob
+          ? {
+              label: hasPaused ? "Resume job" : "Pause job",
+              icon: hasPaused ? "resume" : "pause",
+              disabled: !isCronJob || !isRunning,
```
mausworks, line 41: Nit: `isCronJob` was already checked on line 37 so it's enough to just check `!isRunning`

mausworks (approved):
LGTM! Just a small nit.

PR #1279, opened 2022-09-22
Fix: You get navigated back into JupyterLab on load

mausworks:
This PR fixes:
- The user gets "sucked back" into the JupyterLab view if the reload the page but quickly navigate away from it.
- "Setting up JupyterLab" flashes for a brief second if opening a file when JupyterLab has already been loaded

Additionally:
- Break out common functionality (like getting `_orchest_app`,  `_orchest_docmanager` and widgets) into shared functions
- Removes some unecessary `try`/`catch` blocks and instead doing using `?.` for dot-operations which they protected against.

## Checklist

## Tested:

Before JupyterLab has been initialized (first page load):
- [x] Opening JupyterLab by double-clicking steps, the right tab opens.
- [x] Opening JupyterLab by double-clicking files in the file manager, the right tab opens.
- [x] Opening the JupyterLab from the menu resumes the last viewed tab.
- [x] Refreshing the page in the JupyterLab view resumes the last viewed tab.

After JupyterLab has been initialized:
- [x] Opening JupyterLab by double-clicking steps, the right tab opens.
- [x] Opening JupyterLab by double-clicking files in the file manager, the right tab opens.
- [x] Opening the JupyterLab from the menu resumes the last viewed tab

### DIFF
--- a/services/orchest-webserver/client/src/jupyter/Jupyter.tsx
+++ b/services/orchest-webserver/client/src/jupyter/Jupyter.tsx
@@ -246,7 +246,7 @@ class Jupyter {
   }
 
   hasRenderingProblem() {
-    const shellNode = this.iframe?.contentWindow?._orchest_app.shell.node;
+    const shellNode = this.getApp()?.shell?.node;
     const contentPanel = shellNode?.querySelector("#jp-main-content-panel");
 
     if (!contentPanel) return true;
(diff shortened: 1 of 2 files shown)


yannickperrenet (approved):
> - The user gets "sucked back" into the JupyterLab view if the reload the page but quickly navigate away from it.
> - "Setting up JupyterLab" flashes for a brief second if opening a file when JupyterLab has already been loaded

Tested the PR and these cases have indeed been fixed for me.

Haven't reviewed the code changes though.

PR #1289, opened 2022-09-26
fix: Ensure websocket connections are closed correctly.

iannbing:
## Description

This PR adds `@socketio.on("disconnect", ...)` event handlers to `orchest-webserver` backend, and also creates a zustand store for socketio's to manage connections globally.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/components/ImageBuildLog.tsx
+++ b/services/orchest-webserver/client/src/components/ImageBuildLog.tsx
@@ -40,43 +40,36 @@ export const ImageBuildLog = ({
   }, [fitAddon, xtermRef]);
 
   const socket = useSocketIO(socketIONamespace);
-  const hasRegisteredSocketIO = React.useRef(false);
-
-  React.useEffect(() => {
-    hasRegisteredSocketIO.current = false;
-  }, [socket]);
-
-  React.useEffect(() => {
-    if (!hasRegisteredSocketIO.current && !ignoreIncomingLogs) {
-      hasRegisteredSocketIO.current = true;
-      socket.on(
-        "sio_streamed_task_data",
-        (data: { action: string; identity: string; output?: string }) => {
-          // ignore terminal outputs from other builds
-          if (data.identity === streamIdentity) {
-            if (
-              data.action === "sio_streamed_task_output" &&
-              !ignoreIncomingLogs
-            ) {
-              let lines = (data.output || "").split("\n");
-              for (let x = 0; x < lines.length; x++) {
-                if (x == lines.length - 1) {
-                  xtermRef.current?.terminal.write(lines[x]);
-                } else {
-                  xtermRef.current?.terminal.write(lines[x] + "\n\r");
-                }
-              }
-            } else if (data["action"] == "sio_streamed_task_started") {
-              // This blocking mechanism makes sure old build logs are
-              // not displayed after the user has started a build
-              // during an ongoing build.
-              xtermRef.current?.terminal.reset();
+  const socketEventListener = React.useCallback(
+    (data: { action: string; identity: string; output?: string }) => {
+      if (!streamIdentity) return;
+      // ignore terminal outputs from other builds
+      if (data.identity === streamIdentity) {
+        if (data.action === "sio_streamed_task_output") {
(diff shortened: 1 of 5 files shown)


services/orchest-webserver/client/src/components/ImageBuildLog.tsx
```diff
@@ -40,43 +40,36 @@ export const ImageBuildLog = ({
...
+      if (!streamIdentity) return;
+      // ignore terminal outputs from other builds
+      if (data.identity === streamIdentity) {
+        if (data.action === "sio_streamed_task_output") {
+          let lines = (data.output || "").split("\n");
+          for (let x = 0; x < lines.length; x++) {
+            if (x === lines.length - 1) {
+              xtermRef.current?.terminal.write(lines[x]);
+            } else {
+              xtermRef.current?.terminal.write(lines[x] + "\n\r");
```
yannickperrenet, line 54: I guess it would be faster (performance wise) to do:
```ts
xtermRef.current?.terminal.write(lines[x]);
xtermRef.current?.terminal.write("\n\r");
```
given the immutability of strings.

iannbing: Nice feedback! Pushed the changes 👍.

services/orchest-webserver/client/src/pipeline-view/hooks/useSocketIO.tsx
```diff
@@ -1,17 +1,53 @@
...
-    [namespace]
+type SocketIO = {
+  sockets: Record<string, Socket>;
+  init: (namespace: string | undefined) => void;
+  disconnect: (namespace: string | undefined) => void;
+};
+
+const useSocketIOStore = create<SocketIO>((set) => ({
+  sockets: {},
+  init: (namespace) => {
```
mausworks, line 13: Nit: Since the other function is `disconnect`, maybe this one should be `connect`? :thinking:

mausworks (approved):
Minor nit, but otherwise: awesome cleanup! :tada:

services/orchest-webserver/client/src/pipeline-view/hooks/useSocketIO.tsx
iannbing: Socket, by default, will attempt to connect automatically upon creation. So, this `init` function actually created a Socket instance and connect under the hood 🙂. I added [some comments](https://github.com/orchest/orchest/pull/1289/commits/2bff5e986fbaac54523765696df8e96f68036460) in the code for clarification.

services/orchest-webserver/client/src/components/ImageBuildLog.tsx
ricklamers: Likely .write is more expensive than the string concat. Potato potato :P

services/orchest-webserver/client/src/pipeline-view/hooks/useSocketIO.tsx
mausworks: I guess that makes sense!👌

It's just that when I think of websockets/sockets, my intent (when I want to create a connection) is to look for "connect" _("it's what I want to do")_.

Even if that function could be no-op (because I'm already connected), or if it does the connection under the hood upon initialization, I'm still thinking "connect".

Init + disconnect, just looks a bit awkward, compared to "Connect + disconnect" (being opposites).

I don't care that much, I'll just CTRL+Space and see what functions are available in VSCode, but it's my few cents. ✌️

iannbing: Actually it's quite opposite. 

To use this hook, user needs a bit of understanding about `socket.io-client`, as it exposes a `socket` object. It's not exposing a generic function (otherwise, I would've named it `connect` in the first place). And the `socket` object also has a `connect` function, which confuses future maintainers even further. I'd stick with the design of `socket.io-client`, that is, `io(...)` is to instantiate the `socket` object, and `autoConnect` is a configurable param which you could change to `false` if you like.

It may feel awkward from a language perspective, but it makes it hard to comprehend in the context of code. I'd weigh more on the later, not former. So, I find it hard to agree `connect` is better than `init`.

PR #1290, opened 2022-09-26
Fix - k8s for docker for desktop support

fruttasecca:
## Description

Makes it so that the `node-agent` will be able to push and pull images when the container runtime is docker and the runtime refuses to interact with the internally deployed registry, which currently happens only in k8s for docker for desktop.

### DIFF
--- a/services/node-agent/Dockerfile
+++ b/services/node-agent/Dockerfile
@@ -8,12 +8,20 @@ WORKDIR /tmp/download
 ARG CONTAINERD_VERSION="1.6.4"
 ENV CONTAINERD_DOWNLOAD_URL="https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION-linux-amd64.tar.gz"
 
-# install containerd
+# Install Buildah. This is needed to pull and push images without ssl
+# verification for container runtimes which don't allow interacting with
+# insecure registries by default and for which the ssl certificates
+# injection did not fix the issue, like in kubernetes for docker for
+# desktop.
+RUN apt-get update && apt-get install -y buildah && buildah version && \
+    apt-get clean
+
+# Install containerd.
 RUN curl -L $CONTAINERD_DOWNLOAD_URL --output containerd-$CONTAINERD_VERSION-linux-amd64.tar.gz \
     && tar zxvf containerd-$CONTAINERD_VERSION-linux-amd64.tar.gz -C /tmp/download \
     && mv /tmp/download/bin/ctr /usr/local/bin
 
-# install crictl
+# Install crictl.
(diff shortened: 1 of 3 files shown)


services/node-agent/app/container_runtime.py
```diff
@@ -106,6 +110,30 @@ async def image_exists(self, image_name: str) -> bool:
...
+        Note that this is going to be much slower and will temporarily
+        use storage.
+
+        Note that this has the open issue of leaving dangling storage in
+        cases where the logic would be interrupted before being able to
+        cleanup the temporarily stored image. For example if the
+        node-agent pod restarts during that moment, and the image is not
+        considered for pulling anymore or if, somehow, this function is
+        not called because docker managed to pull.
+        """
```
fruttasecca, line 126: @yannickperrenet  opinion on this?

services/node-agent/Dockerfile
```diff
@@ -8,12 +8,20 @@ WORKDIR /tmp/download
...
 ENV CONTAINERD_DOWNLOAD_URL="https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION-linux-amd64.tar.gz"
 
-# install containerd
+# Install Buildah. This is needed to pull and push images without ssl
+# verification for container runtimes which don't allow interacting with
+# insecure registries by default and for which the ssl certificates
+# injection did not fix the issue, like in kubernetes for docker for
+# desktop.
+RUN apt-get update && apt-get install -y buildah && buildah version && \
+    apt-get clean
```
yannickperrenet, line 17: I believe the following would shave off a few more MBs:
```
RUN apt-get update \
    && apt-get install -yq --no-install-recommends buildah \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*
```

services/node-agent/app/container_runtime.py
```diff
@@ -106,6 +110,30 @@ async def image_exists(self, image_name: str) -> bool:
         )
         return result
 
+    async def _pull_image_for_docker_with_buildah(self, image_name: str) -> bool:
+        """Pulls and image in the docker runtime using buildah.
```
yannickperrenet, line 114:
```suggestion
        """Pulls an image in the docker runtime using buildah.
```

services/node-agent/app/container_runtime.py
yannickperrenet: I think it is fine. Don't expect that many people to use k8s for docker for desktop, so if we learn people actually do and run into this issue than we can always dedicate time to it later :+1:

ricklamers:
Until improved support and more extensive testing we'll ship the required changes but won't mention installation support in the docs.

PR #1291, opened 2022-09-26
fix: Hide job status alert when isEditing is true

iannbing:
## Description

Job status banner should be hidden when user is editing the job. User shouldn't be allowed to perform other operations while editing a recurring job.

## Checklist


yannickperrenet (approved):
:+1:

Issue #94, opened 2020-12-17
Dynamic canvas spawning in the pipeline editor


yannickperrenet:
If you are in the pipeline editor, you notice (when dragging the canvas or zooming out) that to the left and top there is no canvas being drawn. When a component of the pipeline is placed here it should dynamically add canvas in that position.

The idea is basically an implementation of the dynamic canvas spawning seen in https://draw.io/.

yannickperrenet:
@mausworks Aren't you working on this in #1060?

PR #1292, opened 2022-09-26
Expand the pipeline canvas

mausworks:
## Description

This PR expands the canvas from `2000px` to `12000px`. Not _quite_ infinite of course, but you now really have to put your back into it in order to find the edges of it. :wink: 

After a lot of trial and error with trying to find a strategy that doesn't require too many recalculations of positions, I decided to go with this approach:
- The pipeline canvas size is expanded by the padding (times two) so `2000 + 5000 * 2` pixels.
- The _content_ on the canvas is offset by the padding in the X and Y axis using `transform: translate(5000px, 5000px)`.
- The initial position of the canvas is set to the negative padding in both the X and Y direction: `[-5000, -5000]`

The simple solution of setting `{ padding: CANVAS_PADDING }` causes more inconsistencies than the above solution.

As an added bonus: While making the selection rectangle compatible with this solution, I also managed to optimize it further, so it's now even snappier. :tada:

Fixes: #94

## Checklist

## Tested

- [x] A newly imported "quickstart pipeline" is located in the expected position
- [x] Connections are positioned correctly
- [x] Interactive connections (visible when steps are moved) are positioned correctly
- [x] New connection lines are positioned correctly  
- [x] Padding is _not_ added to moved steps in the pipeline definition metadata
- [x] Dropping a pipeline step on the canvas positions it under the cursor
- [x] Hitting <kbd>+ NEW STEP</kbd> adds the step to the center of the canvas 
- [x] Zooming is performed towards the cursor

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/PipelineEditor.tsx
@@ -15,6 +15,7 @@ import { useOpenFile } from "./hooks/useOpenFile";
 import { useSavePipelineJson } from "./hooks/useSavePipelineJson";
 import { PipelineCanvasHeaderBar } from "./pipeline-canvas-header-bar/PipelineCanvasHeaderBar";
 import { PipelineConnection } from "./pipeline-connection/PipelineConnection";
+import { CANVAS_PADDING } from "./pipeline-viewport/common";
 import { PipelineViewport } from "./pipeline-viewport/PipelineViewport";
 import { PipelineEditorRoot } from "./PipelineEditorRoot";
 import { PipelineStep, STEP_HEIGHT, STEP_WIDTH } from "./PipelineStep";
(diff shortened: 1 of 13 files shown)


yannickperrenet:
Nitpick: When zooming out then it becomes noticeable that certain dots have a lower transparency. Not sure whether this was already the case before this PR though.

image

The red squares indication blocks that have edges (not all its edges) where the dots have a lower transparency.

Anyways.... I think it works great for the near future and don't see any reason (from a manual testing perspective) to not merge this PR :+1:

mausworks:
> Nitpick: When zooming out then it becomes noticeable that certain dots have a lower transparency. Not sure whether this was already the case before this PR though.
> 
> image
> 
> The red squares indication blocks that have edges (not all its edges) where the dots have a lower transparency.
> 
> Anyways.... I think it works great for the near future and don't see any reason (from a manual testing perspective) to not merge this PR 👍

This is indeed the case since before as well, although it might be more noticeable now when the canvas is larger.

It might be reasonable to hide the dots after a certain zoom level (as it tends to look quirky in one way or another).

ricklamers:
A new step is created here for me: 
image

ricklamers:
Other than that, LGTM!

iannbing (approved):
LGTM!

mausworks:
Closes in favor of #1300

PR #1295, opened 2022-09-27
Display the "no steps" empty state when no steps have been created but there are scripts available

mausworks:
Display "no scripts or notebooks" when:
- there are no scripts or notebooks in the project
- AND no steps have been created

Display "no steps" when:
- there are scripts or notebooks in the project
- AND no steps have been created

Display "no pipelines" when:
- no pipelines have been created

## Checklist


ricklamers:
Should these behaviors "re-appear" when you delete stuff? E.g. if I delete my only pipeline, should we show the "no pipelines" screen again? I don't think that's working at the moment.

ricklamers:
>I don't think that's working at the moment.

This does appear for the "no steps" and "no scripts" empty states.

ricklamers:
LGTM now! :+1:

ricklamers (approved):

services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
```diff
@@ -245,3 +281,13 @@ export const PipelineViewport = React.forwardRef<
     </PipelineViewportContextMenuProvider>
   );
 });
+
+const hasSomeScriptFile = (node: FileTree | undefined | null) => {
+  if (node?.path && node.type === "file") {
+    return hasExtension(node.path, ...ALLOWED_STEP_EXTENSIONS);
+  } else if (node?.children) {
+    return node.children.some(hasSomeScriptFile);
```
yannickperrenet, line 289: I guess this would go through the entire FileTree. How about we cap the depth at like 1 (so one recursive call)? For performance reasons.

What do you think?

mausworks: Even with thousands of files, it should still be fairly quick. The result is also memoized until the file tree changes (which happens when you add, remove of move files). So it is not on every render cycle.

There's also a fast bailout in the memo: if there are any steps, we don't even bother to check the file tree. But I can add a max-depth: for sanities sake. :+1:

yannickperrenet: I see. I will leave the decision up to you (whether or not to use a max depth). Sounds like you've thought about it :+1:

yannickperrenet (approved):
Works like a charm! :100:

PR #1296, opened 2022-09-27
fix: PipelineLogs should show all steps.

iannbing:
## Description

This PR fixes the issue that some steps are sometimes missing in Pipeline > Logs.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/pipeline-logs-dialog/PipelineLogs.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/pipeline-logs-dialog/PipelineLogs.tsx
@@ -15,9 +15,9 @@ import ListSubheader from "@mui/material/ListSubheader";
 import Stack from "@mui/material/Stack";
 import Typography from "@mui/material/Typography";
 import { hasValue } from "@orchest/lib-utils";
-import cloneDeep from "lodash.clonedeep";
 import React from "react";
 import { usePipelineDataContext } from "../contexts/PipelineDataContext";
+import { usePipelineUiStateContext } from "../contexts/PipelineUiStateContext";
 import { topologicalSort } from "./common";
 import { LogViewerPlaceHolder } from "./LogViewerPlaceHolder";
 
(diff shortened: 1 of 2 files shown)


services/orchest-webserver/client/src/pipeline-view/pipeline-logs-dialog/common.tsx
```diff
@@ -41,9 +39,9 @@ export const topologicalSort = (
 
   // Find roots
   for (const step of Object.values(steps)) {
-    if (step.incoming_connections.length == 0) {
-      addSelfAndChildren(step);
-    }
+    // if (step.incoming_connections.length == 0) {
```
ricklamers, line 42: Should this comment remain?

ricklamers:
Fixing this should still ensure we preserve the sorting behavior. @iannbing did you check for this?

The idea is to traverse the pipeline. Roots (steps without incoming connections) can appear in any order, but then each step with children will have a "breadth first" listing (all direct children appear next to eachother in the list).

services/orchest-webserver/client/src/pipeline-view/pipeline-logs-dialog/common.tsx
iannbing: It should be uncommented. Fixed.

services/orchest-webserver/client/src/pipeline-view/pipeline-logs-dialog/PipelineLogs.tsx
```diff
@@ -39,9 +42,8 @@ export const PipelineLogs = () => {
   const isQueryArgsComplete = hasValue(pipelineUuid) && hasValue(projectUuid);
 
   const sortedSteps = React.useMemo(() => {
-    if (!pipelineJson) return undefined;
-    return topologicalSort(cloneDeep(pipelineJson.steps));
-  }, [pipelineJson]);
```
iannbing, line 44: @ricklamers  The steps are still sorted. The only change is that the source should be processed steps (from `usePipelineUiStateContext()` instead of the raw `pipelineJson`.

iannbing:
@ricklamers As I mentioned in the code, the sorting is still there, and the algorithm is not changed.

ricklamers:
With the uncommenting I think the sorting is restored. We can write unit test for this in the next cycle :) LGTM as of now.

yannickperrenet (approved):

PR #1300, opened 2022-09-27
Expanded canvas: the simple solution.

mausworks:
## Description

For some reason, this solution just escaped me yesterday. :facepalm: 

Either way: it's not intrusive in any way, it just moves the background to its own element and then offsets it.
Since the canvas overflow isn't hidden&mdash;it just works. :shrug: 

Bonus: the background now also gets faded (using a [smoothstep function](https://en.wikipedia.org/wiki/Smoothstep)) when you zoom out: reducing artifacts.

https://user-images.githubusercontent.com/8259221/192518097-5e36de1b-4fac-456c-addc-c398c6089115.mp4

Fixes: #94 

## Checklist


yannickperrenet (approved):
Works for me! :100: 

I like simple approaches ;)

ricklamers:
Tested, LGTM!

iannbing:
@mausworks  Just noticed that AutoLayout doesn't center the canvas anymore.

mausworks:
>Just noticed that AutoLayout doesn't center the canvas anymore.

It doesn't do that on dev either for me. It just organizes the steps.

iannbing:
> It doesn't do that on dev either for me. It just organizes the steps.

It centers the canvas before the UI-overhaul. I'll leave the decision up to @ricklamers.

mausworks:
> It centers the canvas before the UI-overhaul. I'll leave the decision up to @ricklamers.

That may very well be the case, but it seems like an issue unrelated to this PR. 

I think we can both update the "Zoom to fit" and "Auto-layout" functionality. Zoom to fit currently just moves to `-1, -1` and sets zoom to 100%). It should instead (probably) zoom so that all steps are in the center and to a zoom level that fits all steps (if they don't fit at 100%).

ricklamers:
I think I prefer to not center on auto-layout. I always felt a bit disoriented, all I wanted to do was change the pipeline layout. Not reset my pan/zoom.

PR #1301, opened 2022-09-27
fix: polling job runs at the right time

iannbing:
## Description

Currently the interval calculation for polling job runs is incorrect. This PR makes it so that: polling job runs precisely on 15 seconds pass a full minute.

## Checklist


mausworks (approved):

yannickperrenet:
Uhm for me the job runs table doesn't show any entries anymore whatsoever.

PR #1302, opened 2022-09-27
fix: File Manager is not refreshed when switching to a project that doesn't have a pipeline.

iannbing:
## Description

This PR makes it so that File Manager would attempt to fetch the file tree when project_uuid is changed.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/file-manager/FileManager.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/file-manager/FileManager.tsx
@@ -1,4 +1,5 @@
 import { useDebounce } from "@/hooks/useDebounce";
+import { useHasChanged } from "@/hooks/useHasChanged";
 import { useUploader } from "@/hooks/useUploader";
 import { queryArgs } from "@/utils/text";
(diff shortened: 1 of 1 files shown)


yannickperrenet (approved):
LGTM!

services/orchest-webserver/client/src/pipeline-view/file-manager/FileManager.tsx
```diff
@@ -188,11 +189,12 @@ export function FileManager() {
...
    */
 
+  const hasChangedProject = useHasChanged(projectUuid);
   React.useEffect(() => {
     reload();
-    // Only load once when on mount.
+    // Only load when projectUuid is updated.
     // Put `reload` in the deps would trigger unwanted requests.
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [hasChangedProject]); // eslint-disable-line react-hooks/exhaustive-deps
```
mausworks, line 197: Why not just use `projectUuid` here in the hook? The effect will run once the value changes, no need to use `useHasChanged`. :)

PR #1305, opened 2022-09-27
fix: Enable open in new tab for some links.

iannbing:


ricklamers:
Tested, LGTM!

mausworks (approved):
Looks good for now. Future-wise, we should probably just try to use normal links.

ricklamers:
>Future-wise, we should probably just try to use normal links.

I think that was difficult to pair with the MUI component used for the tabs if I recall @iannbing's explanation correctly.

PR #1307, opened 2022-09-27
Hotfix UTC dates without tz

mausworks:
## Description

Some dates were displayed as UTC. This is because the server doesn't add "+00:00" to the end of the date string.

This is a quick hotfix in the front-end. Will investigate what can be done in the back-end.


ricklamers:
LGTM!

PR #1311, opened 2022-09-27
Fix: draft job schedule is empty after switching to another pipeline

iannbing:
## Description

If user doesn't change schedule after switching to another pipeline, the submitted schedule is empty (i.e. both `schedule` and `next_scheduled_time` are `null`). This PR ensures that the schedule is re-assigned right after switching pipelines of a draft job. This PR also remove the unnecessary sorting when creating a draft job.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobScheduleOption.tsx
+++ b/services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobScheduleOption.tsx
@@ -28,13 +28,19 @@ export const useJobScheduleOption = () => {
       : "one-off";
   });
 
-  const hasInitialized = React.useRef(false);
+  const jobPipelineUuid = useEditJob(
+    (state) => state.jobChanges?.pipeline_uuid
+  );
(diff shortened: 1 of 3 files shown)


services/orchest-webserver/client/src/jobs-view/job-view/hooks/useJobScheduleOption.tsx
```diff
@@ -38,20 +39,37 @@ export const useJobScheduleOption = () => {
...
 
+  const setSchedule = React.useCallback(
+    (value: ScheduleOption) => {
+      if (value === "one-off") {
+        setJobChanges({
+          next_scheduled_time: toUtcDateTimeString(nextScheduledTime),
+          schedule: null,
+        });
+      }
+      if (value === "recurring") {
```
mausworks, line 50: Nit: this feels like it should be an else-if, since value is either "one-off" or "recurring", but never both. Or maybe just `if/else` since there are two options. We could say that one is default.

mausworks (approved):
LGTM! :)

PR #1320, opened 2022-09-29
Sticky header shadows and unify layout components

mausworks:
_Since a lot of files have been moved I have tried to make it as easy as possible to review this PR commit-by-commit!_

<div align="center">

**:point_right: [Start your review here](https://github.com/orchest/orchest/pull/1320/commits/557c32f7ad04b506ceafb84d5e843ed7105e1773)** :point_left: 

</div>

## Description

This PR adds a soft shadow under the sticky headers for each view. I also added a backdrop blur to each header, I personally kinda like it, but it should be easy to revert if not desired. (Edit: @ncspost decided that it can stay—at least for now)

Here's how it looks! :tada: 

https://user-images.githubusercontent.com/8259221/193034672-1fcb53ed-b2c1-43e3-a24f-9ddc643d59f0.mp4

(The video is a bit outdated since the tabs are now also included in the header, but the effect is largely the same)

Since I had to break the header out of the "MainContainer" I also had to mess around with a lot of the layout components, so I made an effort to unify those, and clean up a bit here as well! This has the effect that the header on the Projects page is now sticky as well (as seen in the video above).

This PR also prepares for [ORC-909](https://linear.app/orchest/issue/ORC-909/sticky-header-shrinking) by capturing the state of the sticky header in child components. See: 0a6229e

## Checklist


iannbing (approved):
Nice job! I added to commits:
- get rid of the legacy grouped export: `@components/layout/index.ts`
- remove `transition` property in `StickyHeader.tsx` because of the latency it causes.

PR #1322, opened 2022-09-29
Improve step selector performance

mausworks:
## Description

I did this optimization in #1292, but it was then closed for the simpler solution in #1300 . This is that fix, but stand-alone, as I think it has some merit on its own. :+1: 

Basically the improvement is: by using `canvasPointAtPointer` (which is really cheap) we don't have to pass around the canvas offset to update the state, and can instead just pass in the point, and thus, we don't have to do any extra calculations.

## Checklist


iannbing (approved):
Did some profiling. Significant improvements 💯 !

mausworks:
> Did some profiling. Significant improvements :100: !

Awesome that you confimed it too! 

I did some quick profiling too (after 98d5423), I think this summary speaks the loudest …

**Before:**
image

**After:**
image

From ~530ms (over ~4.2s) blocking time, to ~290ms. (over ~4.4s) :+1:

Array allocation in hot paths should definitely be avoided if possible. :laughing:

PR #1328, opened 2022-10-03
Fix/node agent push pull docker desktop

fruttasecca:
## Description

Fixes an issue that can occur on k8s for docker for desktop on linux due to the fact that the container runtime can't
reach the internally deployed registry. Todos:
- [ ] test on windows
- [ ] test on macos

### DIFF
--- a/services/node-agent/app/container_runtime.py
+++ b/services/node-agent/app/container_runtime.py
@@ -5,7 +5,7 @@ import os
 import time
 from asyncio import subprocess
 from enum import Enum
-from typing import List, Optional, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import aiodocker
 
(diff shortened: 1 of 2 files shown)


yannickperrenet (approved):
LGTM!

Quick question though: In the single-node case (which I guess is the majority case given that we are talking about k8s for Docker Desktop here), the overhead of pushing and pulling from the registry isn't be noticeable by the user right? Unless k8s GC cleaned up one of the images.

fruttasecca:
>Quick question though: In the single-node case (which I guess is the majority case given that we are talking about k8s for Docker Desktop here), the overhead of pushing and pulling from the registry isn't be noticeable by the user right? Unless k8s GC cleaned up one of the images.

When it comes to waiting for some pod to start you are correct, the user shouldn't notice unless k8s GC kicks in and a pod is started before the image is pulled again from the registry.

PR #1333, opened 2022-10-04
Fix race condition between job scheduler and job cancellation

fruttasecca:
Fixes a race condition that could occur between the job scheduler and the cancellation of job.

### DIFF
--- a/services/orchest-api/app/app/core/scheduler.py
+++ b/services/orchest-api/app/app/core/scheduler.py
@@ -336,9 +336,7 @@ def schedule_job_runs(app, task_uuid: str) -> None:
                     "schedule",
                     "next_scheduled_time",
                 )
-            )
-            # Ignore drafts.
-            .filter(models.Job.status != "DRAFT")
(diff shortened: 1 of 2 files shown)


services/orchest-api/app/app/core/scheduler.py
```diff
@@ -336,9 +336,7 @@ def schedule_job_runs(app, task_uuid: str) -> None:
                     "schedule",
                     "next_scheduled_time",
                 )
-            )
-            # Ignore drafts.
-            .filter(models.Job.status != "DRAFT")
+            ).filter(models.Job.status.in_(["PENDING", "STARTED"]))
```
ricklamers, line 339: Shouldn't you still ignore drafts too?

fruttasecca: That already happens due to the filter right? :thinking:

yannickperrenet (approved):

services/orchest-api/app/app/core/scheduler.py
ricklamers: Ah nvm

PR #1337, opened 2022-10-05
Move log-streamer to its own container in webserver pod

fruttasecca:
Fixes an issue which would lead to having multiple `log_streamer` instances running thus duplicated logs. This can happen if the child of the PID1 gunicorn process dies, leaving the `log_streamer` process to be inherited by PID1. PID1 will start another instance of the application, leading to multiple `log_streamer`s.

Considered alternatives:
- using `prctl`, `subprocess.Popen(["sleep","10000"], preexec_fn=lambda : libc.prctl(1, signal.SIGKILL))`. 2nd favoured solution, I went with a separate container because it feels more native to k8s and decreases complexity, also considering that a webserver and `log-streamer` split will likely have to happen eventually
- checking in the `log_streamer` loop if the parent PID is 1, if so exit: felt like a breach of abstraction

Handwavy diagram
```
-> = parent/child process relationship

# normal state
gunicorn PID1 -> gunicorn C1 -> log_streamer
                             -> webserver threads

# C1 dies
gunicorn PID1 -> log_streamer

# PID1 restarts C1
              -> log_streamer
gunicorn PID1 -> gunicorn C1 -> log_streamer
                             -> webserver threads
 ```

### DIFF
--- a/services/orchest-controller/pkg/controller/orchestcomponent/orchest_webserver.go
+++ b/services/orchest-controller/pkg/controller/orchestcomponent/orchest_webserver.go
@@ -163,6 +163,17 @@ func getOrchestWebserverDeployment(metadata metav1.ObjectMeta,
 					},
 					VolumeMounts: volumeMounts,
 				},
+				{
+					Name:            "log-streamer",
+					Command:         []string{"python"},
+					Args:            []string{"/orchest/services/orchest-webserver/app/scripts/log_streamer.py"},
+					Image:           image,
+					ImagePullPolicy: corev1.PullIfNotPresent,
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("100m")},
+					},
+					VolumeMounts: volumeMounts,
+				},
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: services/orchest-controller/pkg/controller/orchestcomponent/orchest_webserver.go [hunk 1]
```diff
@@ -163,6 +163,17 @@ func getOrchestWebserverDeployment(metadata metav1.ObjectMeta,
...
 				},
+				{
+					Name:            "log-streamer",
+					Command:         []string{"python"},
+					Args:            []string{"/orchest/services/orchest-webserver/app/scripts/log_streamer.py"},
+					Image:           image,
+					ImagePullPolicy: corev1.PullIfNotPresent,
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("100m")},
+					},
```
yannickperrenet, line 174: Probably want to set this to a lower value. Otherwise the `orchest-webserver` Pod will request `200m`. 

`20m`?

ricklamers:
@fruttasecca this is very nice! :clap:

PR #1340, opened 2022-10-06
Add retries to k8s python sdk

fruttasecca:
## Description

Adds retries to the k8s python SDK to reduce failures under system load.

To test, run this script after having killed the `kube-apiserver` container.
```
from kubernetes import client as k8s_client
from kubernetes import config
config.load_config()
from kubernetes.client.api_client import ApiClient
from kubernetes.client.configuration import Configuration
from urllib3.util.retry import Retry

def _get_k8s_api_client() -> ApiClient:
    configuration = Configuration.get_default_copy()
    _retry_strategy = Retry(
        total=5,
        backoff_factor=1,
    )
    # See urllib3 poolmanager.py usage of "retries".
    configuration.retries = _retry_strategy
    a= ApiClient(configuration=configuration)
    return a


k8s_core_api = k8s_client.CoreV1Api(api_client=_get_k8s_api_client())

a = k8s_core_api.list_namespace()
print(len(a.items))

```

### DIFF
--- a/orchest-cli/orchestcli/cmds.py
+++ b/orchest-cli/orchestcli/cmds.py
@@ -69,11 +69,11 @@ def _get_k8s_api_client() -> ApiClient:
 
 
 API_CLIENT = _get_k8s_api_client()
-APPS_API = client.AppsV1Api(api_client=API_CLIENT)
-CORE_API = client.CoreV1Api(api_client=API_CLIENT)
-CUSTOM_OBJECT_API = client.CustomObjectsApi(api_client=API_CLIENT)
-EXT_API = client.ApiextensionsV1Api(api_client=API_CLIENT)
-RBAC_API = client.RbacAuthorizationV1Api(api_client=API_CLIENT)
+APPS_API = client.AppsV1Api(api_client=_get_k8s_api_client())
+CORE_API = client.CoreV1Api(api_client=_get_k8s_api_client())
+CUSTOM_OBJECT_API = client.CustomObjectsApi(api_client=_get_k8s_api_client())
+EXT_API = client.ApiextensionsV1Api(api_client=_get_k8s_api_client())
+RBAC_API = client.RbacAuthorizationV1Api(api_client=_get_k8s_api_client())
 
 get_namespaced_custom_object = partial(
     CUSTOM_OBJECT_API.get_namespaced_custom_object,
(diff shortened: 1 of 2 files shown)


services/orchest-api/app/app/connections.py
```diff
@@ -1,19 +1,36 @@
...
-k8s_core_api = k8s_client.CoreV1Api()
-k8s_apps_api = k8s_client.AppsV1Api()
-k8s_networking_api = k8s_client.NetworkingV1Api()
-k8s_custom_obj_api = k8s_client.CustomObjectsApi()
-k8s_rbac_api = k8s_client.RbacAuthorizationV1Api()
+k8s_core_api = k8s_client.CoreV1Api(api_client=_get_k8s_api_client())
+k8s_apps_api = k8s_client.AppsV1Api(api_client=_get_k8s_api_client())
+k8s_networking_api = k8s_client.NetworkingV1Api(api_client=_get_k8s_api_client())
+k8s_custom_obj_api = k8s_client.CustomObjectsApi(api_client=_get_k8s_api_client())
+k8s_rbac_api = k8s_client.RbacAuthorizationV1Api(api_client=_get_k8s_api_client())
```
yannickperrenet, line 36: Why not just create on `API_CLIENT` here as well?

orchest-cli/orchestcli/cmds.py
```diff
@@ -50,12 +53,27 @@
...
+        total=5,
+        backoff_factor=1,
+    )
+    # See urllib3 poolmanager.py usage of "retries".
+    configuration.retries = _retry_strategy
+    a = ApiClient(configuration=configuration)
+    return a
+
+
+API_CLIENT = _get_k8s_api_client()
```
yannickperrenet, line 71: I think that works, but just to be sure, have you tested a clean `orchest install`? In some places we use the `API_CLIENT` directly.

services/orchest-api/app/app/connections.py
fruttasecca: Now that I think of it I think it might be better to pass a different API client to each k8s API object in the CLI as well

orchest-cli/orchestcli/cmds.py
fruttasecca: Indeed installation goes through, same for other commands

services/orchest-api/app/app/connections.py
yannickperrenet: Why? Can't/Shouldn't the same api client not be reused?

fruttasecca: The behaviour we had before this PR (which is also the default) is to not share API clients, I simply want to minimize the risk of any issue down the line since sharing the client has no tangible reward atm

PR #1341, opened 2022-10-06
Don't concurrently push or pull images

fruttasecca:
## Description

Makes it so that the `ContainerRuntime` class does not allow concurrent pushes or pulls of images. This has been done to reduce overhead and avoid some edge cases that have been observed, for example, multiple pushes through `buildah`.

I tried to go for the simplest solution, alternatives related to fiddling with the asyncio queue that holds images to pull or push (or other queues) felt a bit too risky for my tastes.

Also fixes a low impact bug where an image would be pulled multiple times in a row after being pulled successfully


yannickperrenet (approved):
LGTM!

PR #1342, opened 2022-10-06
Fix: Environment build logs sometimes are failed to render.

iannbing:
## Description

Environment build logs sometimes failed to render because socket.io failed to reconnect immediately after being disconnected. This connect and disconnect is actually not necessary, as FE could just continue using the same connection. 

This PR made the following changes to remove the unnecessary disconnect:
- BE doesn't send logs on connect.
- FE sends a request to get logs when needed.
- FE doesn't disconnect until the browser tab is closed.

## Checklist

### DIFF
--- a/services/orchest-webserver/app/app/socketio_server.py
+++ b/services/orchest-webserver/app/app/socketio_server.py
@@ -87,6 +87,19 @@ def register_build_listener(namespace, socketio):
                     include_self=False,
                     namespace=namespace,
                 )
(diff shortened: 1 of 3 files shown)


ricklamers:
> FE doesn't disconnect until the browser tab is closed.

Note there are limits to the total number of concurrent websocket connections one can have. As long as we hold at most one connection per namespace we should be in the clear.

mausworks (approved):
LGTM! :)

PR #1343, opened 2022-10-07
Docker desktop installation docs

fruttasecca:


docs/source/_static/js/installer_widget.js
```diff
@@ -80,6 +80,30 @@ orchest install --socket-path=/run/k3s/containerd/containerd.sock
...
+</div>
+<p>Now Orchest can be installed using the <code class="docutils literal notranslate"><span
+class="pre">orchest-cli</span></code>:</p>
+<div class="highlight">
+<pre>
+pip install --upgrade orchest-cli
+orchest install
+</pre>
+</div>
+<p>Now that Orchest is installed, it can be reached on localhost, port 80.</p>`,
```
yannickperrenet, line 105: How about making `localhost:80` a hyperlink so that users can just click it?

fruttasecca: will do

docs/source/_static/js/installer_widget.js
```diff
@@ -80,6 +80,30 @@ orchest install --socket-path=/run/k3s/containerd/containerd.sock
...
+Docker Desktop </a> on linux.</p>
+<p>Once installed, go to its settings (top right) and if you wish to do so, change the
+allocated resources. At least 2 cpus, 8GB of memory and 15GB of disk are required.</p>
+<p><a href="https://docs.docker.com/desktop/kubernetes/"> Enable kubernetes in Docker Desktop</a></p>
+<p>Install the nginx ingress (or any other ingress of preference)</p>
+<div class="highlight">
+<pre>
+kubectl apply -f \
+https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.2.1/deploy/static/provider/cloud/deploy.yaml
+</pre>
```
yannickperrenet, line 95: Ideally this should not be required as the `orchest-controller` can detect running on KDD and set up ingress for the user automatically. 

@fruttasecca Could you share the annotations and labels of a KDD node? (`kubectl describe nodes` output). Or just make the changes following:

To have the `orchest-controller` do this automatically all you need to do is c2fc44743103c18564eb79e1762dabecceb25e12 and add KDD to https://github.com/orchest/orchest/blob/f94b14075f7f9f3eda3709692e0c4294818a237b/services/orchest-controller/pkg/controller/orchestcluster/cluster_utils.go#L513-L528.

fruttasecca: Mmh unfortunately there is no label or annotation key that can hint the k8s distribution but we can use the hostname value for that. I think it's safe enough for the time being and adding back the nginx instructions is quite easy.
image

fruttasecca: https://github.com/orchest/orchest/pull/1343/commits/42ec26730ade43d50885550c3c8b93d2f4e988ed

nhaghighat: @fruttasecca: recently we added an annotation for k8s distribution.

the key is `controller.orchest.io/k8s`

yannickperrenet (approved):

docs/source/_static/js/installer_widget.js
nhaghighat: It seems it is safe enough.

PR #1348, opened 2022-10-07
Indicate that the setup script can't be edited on build

mausworks:
## Description

image

Fixes [ORC-931](https://linear.app/orchest/issue/ORC-931)

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/components/common/CodeMirror.tsx
+++ b/services/orchest-webserver/client/src/components/common/CodeMirror.tsx
@@ -1,8 +1,6 @@
 import { useEscapeToBlur } from "@/hooks/useEscapeToBlur";
-import LockOutlined from "@mui/icons-material/LockOutlined";
 import { Theme } from "@mui/material";
 import Box from "@mui/material/Box";
-import { grey } from "@mui/material/colors";
 import React from "react";
 import { Controlled, IControlledCodeMirror } from "react-codemirror2";
 
@@ -13,7 +11,6 @@ type CodeMirrorProps = Omit<
   borderColor?: string;
   /** Makes the editor read-only, but also fades the text out and displays a lock icon. */
   locked?: boolean;
-  error?: boolean;
   onFocus?: () => void;
   onBlur?: () => void;
 };
@@ -25,8 +22,6 @@ export const CodeMirror = React.forwardRef<HTMLDivElement, CodeMirrorProps>(
       onFocus,
       error,
       onBlur,
-      locked,
-      options,
       ...props
     }: CodeMirrorProps,
     ref
@@ -47,7 +42,7 @@ export const CodeMirror = React.forwardRef<HTMLDivElement, CodeMirrorProps>(
     const getBorderColor = (theme: Theme) =>
       error
         ? theme.palette.error.main
-        : locked
+        : props?.options?.readOnly
         ? theme.palette.grey[200]
         : isFocused
         ? theme.palette.primary.main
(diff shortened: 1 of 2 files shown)


services/orchest-webserver/client/src/components/common/CodeMirror.tsx
```diff
@@ -31,23 +44,51 @@ export const CodeMirror = React.forwardRef<HTMLDivElement, CodeMirrorProps>(
...
+          ".CodeMirror-lines": {
+            cursor: locked ? "not-allowed" : undefined,
           },
         }}
-        ref={ref}
       >
-        <Controlled {...props} onFocus={handleFocus} onBlur={handleBlur} />
+        <Controlled
+          {...props}
+          options={{ ...options, readOnly: locked || options?.readOnly }}
```
iannbing, line 75: What is the reason of adding a prop `locked` and then actually take `locked || options?.readOnly` in the component? Wouldn't it be simpler just stick with the API of CodeMirror?

iannbing: And the `<LockOutlined />` is only rendered when `locked` is true. Is there any case that a CodeMirror is `locked` but not `readOnly`?

iannbing: Note that there are many instances of read-only `<CodeMirror />` (approximately 5?) across the whole codebase. But only this one applied `locked`. Is it intended?

mausworks: It is intended. I didn't want to enforce the styling of "locked" editors on all editors that are read-only, since the ticket only covers the environments setup script. There may also be a good reason for one of them not to have the "locked styling", but still be read-only.

Conversely it may be useful to use the locked styling in other views/components. So now you can pick. 👌

So, essentially: `readOnly` makes the editor functionally read-only, and `locked` makes the editor visually locked + read-only.

> And the <LockOutlined /> is only rendered when locked is true. Is there any case that a CodeMirror is locked but not readOnly?

No, it wouldn't really make sense to visually indicate that the editor is locked, but that you can also edit it.

iannbing: I think it's worth it checking with Nick and the team about this. There might be a good chance to tackle this in one go, and save the time on discussing the API design of this component. To me, it's more of a visual design question than a tech question.

iannbing: And if the conclusion was that we only want to show this icon here, I'd suggest just adding a wrapper to show the icon in this very occasion rather than making it part of this generic component. As you can see, the API is confusing.

iannbing: Seems like a fair solution for now 👍

iannbing (approved):

services/orchest-webserver/client/src/components/common/CodeMirror.tsx
mausworks: Since I didn't get any clear response, I just moved it to the EnvironmentSetupScript in 
af47666.

No big change in general, so I'm merging. :+1:

mausworks: Added back the error prop that I accidentally removed in `dev`: https://github.com/orchest/orchest/commit/47dffddeec88eba4c14f7c620dbeaf1087041acc :sweat_smile:

PR #1359, opened 2022-10-11
fix: Hide Intercom when the side panel or a Snackbar appears.

iannbing:
## Description

Intercom should be hidden when the side panel in Pipeline Editor or a bottom-right Snackbar appears.

## Checklist


mausworks (approved):
LGTM, nice work.

PR #1360, opened 2022-10-11
Fix "session already exists"-error

mausworks:
## Description

This can can happen when the request to start the session is fired before the session has is fetched.
This error can safely be ignored, but as a follow-up, the root-cause should also be investigated.

## Checklist


iannbing (approved):
LGTM! Good call on ignoring the false alarm. 👍

Issue #1365, opened 2022-10-12
Empty-states in `Jobs` and `Environments` should NOT show non-empty-state actions


yannickperrenet:
**Describe the bug**
The bug is best described using screenshots:

![image (2)](https://user-images.githubusercontent.com/26223174/195323230-97709499-6bd8-4b5a-80be-230a0d72d080.png)
![image (3)](https://user-images.githubusercontent.com/26223174/195323233-41e38961-2756-45a5-af24-690c37ebcabc.png)

**Expected behavior**

These actions should not be present.

PR #1366, opened 2022-10-12
fix: Header should be hidden if no items available in Environments and Jobs

iannbing:
## Description

The header should be hidden if there's no item in the views of Environments and Jobs.

Fix: #1365 

## Checklist


mausworks (approved):
Great catch! :+1:

PR #1369, opened 2022-10-13
Fix: Orchest version should be refreshed after being updated.

iannbing:
## Description

Currently FE only fetches Orchest version once, on launch. However, it will no longer be correct after updating Orchest to the latest version, unless user refresh the page. This is causing some other issues, e.g. the latest Environment image is seen as a custom image. This PR makes it so that it will refetch Orchest version right after updating succeeded.

Note: it's quite tricky to test it locally. Because running in dev mode doesn't really allow you to update to a _real_ version, which always results in the same version. Nevertheless, this PR should fix the said issue. 😅

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/contexts/AppContext.tsx
+++ b/services/orchest-webserver/client/src/contexts/AppContext.tsx
@@ -9,6 +9,7 @@ import { useAutoFetchPipelines } from "./useAutoFetchPipelines";
 
 export type AppContextType = {
   orchestVersion: string | null | undefined;
+  fetchOrchestVersion: () => Promise<string | void | null | undefined>;
   checkUpdate: () => Promise<void>;
   webhooks: NotificationWebhookSubscriber[];
   setWebhooks: StateDispatcher<NotificationWebhookSubscriber[]>;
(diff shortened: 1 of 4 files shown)


services/orchest-webserver/client/src/contexts/AppContext.tsx
```diff
@@ -9,6 +9,7 @@ import { useAutoFetchPipelines } from "./useAutoFetchPipelines";
 
 export type AppContextType = {
   orchestVersion: string | null | undefined;
+  fetchOrchestVersion: () => Promise<string | void | null | undefined>;
```
mausworks, line 12: Not: Is there any way that this beast of a return type can be simplified to something like `string | undefined`?

mausworks (approved):
LGTM! Just a small nit. 🔥

services/orchest-webserver/client/src/contexts/AppContext.tsx
iannbing: Good catch. [Removed some legacy bad typing](https://github.com/orchest/orchest/pull/1369/commits/c43b29886d48b61d69c43f4bc4134575a2d3e8e6).

PR #1373, opened 2022-10-13
Offset pipeline starting position if read-only banner is showing

mausworks:
## Description

If you open a pipeline editor where the steps are located close to `[0, 0]` in the canvas, the read-only banner covers them, which is undesired:

CleanShot 2022-10-13 at 10 39 29@2x

This PR fixes that so that the starting position is offset when the banner is showing. If the user hasn't moved away from the starting position when the banner goes away, the offset is also automatically removed.

image (1)

This offset triggers on any read-only state in the editor, so it automatically supports all banners.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/pipeline-viewport/PipelineViewport.tsx
@@ -1,4 +1,3 @@
-import { useEnvironmentsApi } from "@/api/environments/useEnvironmentsApi";
 import { FileTree } from "@/types";
 import {
   isSamePoint,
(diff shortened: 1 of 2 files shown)


yannickperrenet:
I don't really like this "jumping behavior" that happens when the banner is no longer showing:

https://user-images.githubusercontent.com/26223174/195604803-ec5da4f9-18dc-4a51-8b1c-dab7214031a2.mp4

And an "edge case" where the pipeline wasn't completely in the viewport to begin with:

https://user-images.githubusercontent.com/26223174/195605633-f05c5f82-16cc-46c6-a909-e4e25dfffa92.mp4

----

Feels like this is "edge case" territory to come up with the perfect way to show a banner without it going over a step. The goal of: "we just want to prevent the alert from overlapping when the page initially loads." is reasonably hard to achieve perfectly. Might be worth coming up with a slightly more nuanced goal, e.g.: "we just... initially loads if the pipeline is auto-aligned". Although depending on the goal the implementation might not be worth it in the first place.

mausworks:
@yannickperrenet I think the last pushes should fix both issues! Jumping behavior is gone, and the "adjustment" should be applied now even if you moved the pipeline before navigating between views.

yannickperrenet:
> I think the last pushes should fix both issues!

I must be doing something wrong then, because I am still experiencing both of them :sweat_smile:

mausworks:
> I must be doing something wrong then, because I am still experiencing both of them 

Maybe the pull failed for some reason? I can't reproduce either issue on my machine, but maybe I'm taking the wrong steps.

https://user-images.githubusercontent.com/8259221/195622830-6c30aa12-99c4-4d7f-84e2-bf9cbd2be39e.mp4

yannickperrenet:
After a full rebuild and re-install it indeed works for me now and I quite like it! :muscle: Nice one.

yannickperrenet (approved):

ncspost (approved):
LGTM!

iannbing (approved):
Small details require lots of hard work 👍

PR #1389, opened 2022-10-19
Enable GPU in Kubenetes

nhaghighat:
## Description

TBD.

Fixes: #issue

## Checklist


services/orchest-controller/pkg/apis/orchest/v1alpha1/types.go
```diff
@@ -233,7 +233,7 @@ type OrchestClusterSpec struct {
 
 	RabbitMq OrchestComponentTemplate `json:"rabbitMq,omitempty"`
 
-	Applications []ApplicationSpec `json:"applications,omitempty"`
+	Applications map[string]ApplicationSpec `json:"applications,omitempty"`
```
yannickperrenet, line 236: Would this be a breaking change for existing `OrchestClusters` or will applying the new CRD (which we do on update) correctly migrate it?

nhaghighat: It should correctly migrate it, but this is still a WIP PR, so a lot of things will change here, I suggest to not spend a lot of time on it.

PR #1397, opened 2022-10-25
Set empty string as default file path when creating new step

yannickperrenet:
## Description

Using `/` instead would make users believe they need to enter an absolute path, which actually won't work.


mausworks (approved):
LGTM. This will be updated in the new file picker as well. I'll try to have that PR up today as well. ✌️

PR #1399, opened 2022-10-25
Fail Project import if repo couldn't be `clone`d

yannickperrenet:
## Description

Properly handle `git clone` (e.g. specifying a private repo URL) failures during Project import:

https://user-images.githubusercontent.com/26223174/197776080-2245b982-d7c3-474d-b7d1-1c42b590733d.mp4

### DIFF
--- a/services/orchest-webserver/app/scripts/background_tasks.py
+++ b/services/orchest-webserver/app/scripts/background_tasks.py
@@ -33,7 +33,9 @@ def git_clone_project(args) -> Tuple[int, str]:
 
         # This way we clone in a directory with the name of the repo if
         # the project_name is not provided.
-        git_command = f"git clone {args.url}"
+        # - GIT_TERMINAL_PROMPT makes the `git` command fail if it has
+        #   to prompt the user for input, e.g. for their username.
+        git_command = f"GIT_TERMINAL_PROMPT=0 git clone {args.url}"
 
         # args.path contains the desired project name, if the user
         # specified it.
--- a/services/orchest-webserver/client/src/projects-view/ImportDialog.tsx
+++ b/services/orchest-webserver/client/src/projects-view/ImportDialog.tsx
@@ -53,6 +53,8 @@ import React from "react";
 import { useImportGitRepo, validProjectName } from "./hooks/useImportGitRepo";
 
 const ERROR_MAPPING: Record<CreateProjectError, string> = {
+  "git clone failed":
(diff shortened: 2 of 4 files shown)


services/orchest-webserver/app/scripts/background_tasks.py
```diff
@@ -33,7 +33,9 @@ def git_clone_project(args) -> Tuple[int, str]:
 
         # This way we clone in a directory with the name of the repo if
         # the project_name is not provided.
-        git_command = f"git clone {args.url}"
+        # - GIT_TERMINAL_PROMPT makes the `git` command fail if it has
```
mausworks, line 36: TIL! :fire:

services/orchest-webserver/app/scripts/background_tasks.py
```diff
@@ -33,7 +33,9 @@ def git_clone_project(args) -> Tuple[int, str]:
...
+        #   to prompt the user for input, e.g. for their username.
+        git_command = f"GIT_TERMINAL_PROMPT=0 git clone {args.url}"
```
yannickperrenet, line 38: Note that we just fail whenever a prompt is issued by `git`. This does not distinguish between cases as to why the `git clone` failed. Felt unnecessary for now, it is already a good improvement of the current implementation.

services/orchest-webserver/client/src/projects-view/ImportDialog.tsx
```diff
@@ -53,6 +53,8 @@ import React from "react";
 import { useImportGitRepo, validProjectName } from "./hooks/useImportGitRepo";
 
 const ERROR_MAPPING: Record<CreateProjectError, string> = {
+  "git clone failed":
+    "failed to clone repo. Only public repos can be imported.",
```
mausworks, line 57: Nit: Not sure if intended but it looks weird when the F isn't capitalized in "Failed", and then when O is capitalized in "Only". :thinking:

mausworks (approved):
One small nit, otherwise LGTM. :+1:

services/orchest-webserver/client/src/projects-view/ImportDialog.tsx
yannickperrenet: The text is actually displayed as: `Import failed: failed to clone repo. Orchest can only import public repos.`

And not capitalizing is in line with the other errors that we display.

PR #1401, opened 2022-10-26
New file picker, and various file state fixes

mausworks:
## Description

**Note:** I still have some testing to do.

This PR adds a new file picker to the project that looks like this:

image

This should make it easier to create new files for steps if you use the "create step" workflow.

Most work has gone into figuring out how the state between the file picker and file browser (and beyond) can safely be shared without causing weird side effects (see fixes below). 

Features:
- New design based on @ncspost work :tada:
- "New file" directly in the file picker :+1: 
- Great keyboard support for those who prefer typing: The dialog is searchable and fully reactive to changes in the path, and you can use up/down + enter to navigate too! :ok_hand: 

Fixes:
- When reloading the file tree (previously `fetchFileTrees`, now `init`), it now loads up to the currently deepest loaded directory (not what's currently expanded in the UI). This solves multiple issues with the shared state between the file picker and file browser.
- Loading a tree node that has already been loaded no longer prunes its grandchildren
- Duplicate fetch requests when the file browser is initialized are removed
- The whole file tree is no longer refreshed when a file is uploaded, a pipeline is deleted, or when a file is created or moved: just the affected directory is.

In order to facilitate easier merging and management of the file tree, the internal state representation for the file tree has been flattened: it's just a `Record<string, FileMetadata>` instead of a tree structure. This makes checks like "is this file in the tree `O(1)`", searching and merging nodes is also much easier. Parents, children, etc ... can be derived with basic path operations. It also means that all paths are unique, bypassing any need for deduplication.

## Checklist

## Tests

- [x] Swapping between the pipeline view and a job run view shows the right files
- [x] Swapping between a job view and a pipeline view shows the right files
- [x] Swapping between a job snapshot view and the pipeline view shows the right files
- [x] Creating a new file in a deep directory does not collapse it
- [x] Creating a new folder in a deep directory does not collapse it
- [x] Expanding a directory in the file browser always re/loads it
- [x] Expanding a directory in the file picker always re/loads it
- [x] Hitting "Reload" in the file browser does not unload any deep subdirectories


ricklamers:
A work of art :clap:

Tried it locally, haven't found any issues so far.

ricklamers:
Minor issues as mentioned during the call:
- Missing /data /Project files radio option
- Opening a step seems (click to open pane) quite slow. Worth profiling and spotting if there's low hanging fruit here

yannickperrenet:
https://user-images.githubusercontent.com/26223174/198243493-b6b74d4e-112a-4724-9e3a-ac5fa9ff14e5.mp4

- [ ] Regression, https://github.com/orchest/orchest/pull/1397 should still be apply but currently it doesn't.

----

https://user-images.githubusercontent.com/26223174/198244555-73c37ca9-769e-4c99-83cc-ed9449f0a03e.mp4

- [ ] Error on new file creation. In addition, creating a folder isn't even possible and throws multiple errors.

----

https://user-images.githubusercontent.com/26223174/198244965-191fc7e6-d295-46e4-a843-24613f290a96.mp4

- [ ] If there are many project files the file picker should probably use a scrollbar instead of filling the entire screen height.

----

- [ ] Nitpick: Would it be possible to only show matching file names in the file picker when typing? This way the number of options are essentially filtered by what the user types. Now it jumps to the first matching word which is also fine imo but not as good as filtering. If its easy to add then filtering would have my preference, otherwise we can leave it as is :+1:

ncspost:
Great work @mausworks! I like your solution for the breadcrumb collapsing. 

- [ ] Add dropdown icon to filepath box similar to select boxes below
- [ ] When there are lots of files/folders, the dropdown menu obscures the select box component... making it difficult to see validation or text input. Can we reduce box size and introduce scrolling quicker to keep the select box in view?

CleanShot 2022-10-27 at 10 02 13@2x

- [ ] I'm not sure the "project files /" text in the select box is super useful. It has the potential to mask longer filepaths. Maybe we should remove and rely just on menu breadcrumb to indicate it's part of project files for now?
- [ ] If we do keep the "project files /" text, when I click on it, it should open the select box menu... currently it does nothing
- [ ] When a user has input file path text "example" into select box that doesn't exist, could we change the button to [Create "example" file]?
- [ ] Change new folder dialog for consistency...
- New folder
- No buttons icons
- "Save folder" primary button

CleanShot 2022-10-27 at 10 18 31@2x

- [ ] Pass "file not found" status to step box instead of "ready" status - don't worry about changing colour yet
- [ ] When a file path is pasted into step detail before name, could we pre-fill name based on filepath? e.g. blah.ipynb → blah name
- [ ] Stretch button in menu full width
CleanShot 2022-10-27 at 10 36 23@2x

mausworks:
> Regression, https://github.com/orchest/orchest/pull/1397 should still apply but currently it doesn't.

Great catch! This actually revealed a much deeper issue (see 72f1967). :ok_hand:

mausworks:
> I'm not sure the "project files /" text in the select box is super useful. It has the potential to mask longer filepaths. Maybe we should remove and rely just on menu breadcrumb to indicate it's part of project files for now?

I agree. I have removed it now. It is only visible when the radio buttons for "root" aren't visible. This component is also used for "loading parameter files" in the jobs view, and you're locked to the "Project files" folder there.

yannickperrenet (approved):
Couldn't find anything anymore. Really well done!! :fire:

mausworks:
> Couldn't find anything anymore. Really well done!! :fire: 

@yannickperrenet awesome! Always great to have your seal of approval. :heavy_check_mark:  :seal:

Just a minor issue left with directories collapsing when they shouldn't, then it's good to go! :+1:

services/orchest-webserver/client/src/components/FilePicker.tsx
```diff
@@ -1,372 +1,326 @@
...
-            {options?.map((childNode, index) => {
-              const nodeName = childNode.name;
-              return (
-                <MenuItem
-                  key={childNode.name}
-                  onClick={(e) => onSelectListItem(e, childNode)}
-                  ref={isRootNode && index === 0 ? menuFirstItemRef : undefined}
-                >
-                  {childNode.type === "directory" && (
+      <Popover
```
iannbing, line 239: Quite some efforts on implementing this dropdown. Some expected behavior is missing: when pressing KeyUp or KeyDown, the inner menu doesn't scroll, i.e. the focused item is outside of the visible area). I'm wondering if there are other MUI components that we could use here 🧐

iannbing (commented):
Nice improvements! The only thing that I have questions is the custom dropdown (mostly about keyboard navigation). It would be best if we could have something from MUI out-of-box, so that we don't need to directly manipulate DOM elements in our implementation.

services/orchest-webserver/client/src/components/FilePicker.tsx
mausworks: Good feedback on the component. I'll look to see if there's anything more elaborate in MUI. :ok_hand:

mausworks: So, there's one key difference between the project picker and the file picker. You want to be able to type _while_ using the keyboard, for example, if I have:

- abc.py
- cde.py
- foo1.py
- foo2.py

If I want to select "foo2.py", I should be able to type "f" hit <kbd>&darr;</kbd>, and then <kbd>Enter</kbd> to select it. In the project picker, if I hit "key down", the search field loses focus, and I have to refocus it to continue typing.

I also need to make sure that the keyboard navigation starts from the current "best match", and not from the top. In the end, the solution was super simple (see 892242e), so I don't mind this "one-off" custom logic. It should be fairly isolated to this component. :+1:

iannbing (approved):

Issue #1211, opened 2022-08-22
Guess ports when user fills the image of a service


astrojuanlu:
**Describe the problem this improvement solves**
When I add a new service, I have to have a look at the ports (by default 8080) and usually change them.

**Describe the solution you'd like**
Since many common applications have standard ports, it would be nice if the UI filled them for me. We already have part of that logic here:

https://github.com/orchest/orchest/blob/a21bca9ac76401da2335827b7e4a7c88e1bcdec4/lib/python/orchest-internals/_orchest/internals/compat.py#L82-L102

cacrespo:
Hello! 
I can start working on this topic from October 15. If no one is in charge of this, please assign it to me. I may have to use some spanglish but I think it's not a problem for you. :smile:

astrojuanlu:
Thanks @cacrespo! English is preferred in the repo, we can work on that through other channels :)

I'm guessing you're already setting up a local development environment. Keep us posted about your progress.

PR #1403, opened 2022-10-30
add postgres as a default option when adding services

cacrespo:
## Description

PostgreSQL maintains a huge popularity so I think it is convenient that it is available as a default option when adding services.

I took the postgres defaults values from [here](https://github.com/ricklamers/orchest-redis-postgres) and for the icon I re-used [this code](https://github.com/orchest/orchest/blob/21f899c177ef9db652d926745d7a08a2fbef6499/services/orchest-webserver/client/src/icons/index.tsx). I'm not sure what else should be added by default. 

image
image

Fixes: maybe related to #1211 

## Checklist


astrojuanlu (changes requested):
Hello @cacrespo, thanks a lot for your PR! Unfortunately, we already decided to hide stateful services like databases https://github.com/orchest/orchest/pull/1005

Since Orchest services are bound to the Session, all the data would be wiped out when it stops, unless properly configured. As such, they were somewhat error prone unless you know what you're doing. I'll wait a couple of days in case any of my colleagues want to chime in, but I'd be inclined to not merge this.

cacrespo:
Oh I understand. No problem!
Happy to try to help.

Thank you very much and keep up with this excellent project! :muscle::muscle::muscle:

Issue #1404, opened 2022-10-31
When opening the snapshot of a Job, then the file-manager shows the state of the project directory instead of snapshot directory


yannickperrenet:
**Describe the bug**
As per title.

**Expected behavior**
Job snapshot should be reflected in file-manager

**To Reproduce**

1. Open the quickstart project
2. Create a job
3. Remove files from the quickstart project
4. Open the snapshot of the job
5. See that the file-manager no longer shows the deleted files, even though they are present in the snapshot

**Screenshots**
![image](https://user-images.githubusercontent.com/26223174/198988440-00d58caa-eb96-4fba-bd86-02022f07aaae.png)

![image](https://user-images.githubusercontent.com/26223174/198988096-409f5a62-007e-493e-a0ad-5f5df27bfb64.png)

PR #1405, opened 2022-10-31
fix: Load the correct file tree from the Snapshot of a Job.

iannbing:
## Description

Currently, when inspecting a snapshot, File Manager loads the current project file tree incorrectly. This PR makes it so that it loads the file tree from the Snapshot, and also prevent unintentional requests while inspecting a snapshot.

Fixes: #1404 

## Checklist

### DIFF
--- a/services/orchest-webserver/app/app/utils.py
+++ b/services/orchest-webserver/app/app/utils.py
@@ -5,7 +5,7 @@ import os
 import re
 import uuid
 from datetime import datetime
-from typing import Dict, List, Optional
+from typing import Dict, List, Literal, Optional, Union
 
 import requests
 from flask import current_app, safe_join
@@ -88,7 +88,7 @@ def get_project_directory(
     project_uuid: Optional[str] = None,
     pipeline_uuid: Optional[str] = None,
     job_uuid: Optional[str] = None,
-    run_uuid_or_snapshot: Optional[str] = None,
+    run_uuid_or_snapshot: Optional[Union[Literal["snapshot"], str]] = None,
 ):
     if project_uuid is None:
         raise ValueError("project_uuid is required.")


services/orchest-webserver/app/app/core/filemanager.py
```diff
@@ -30,30 +30,28 @@ def _construct_root_dir_path(
...
-        pipeline_uuid=pipeline_uuid,
-        job_uuid=job_uuid,
-        run_uuid=run_uuid,
-    )
+    else:
+        return get_project_directory(
+            project_uuid=project_uuid,
+            pipeline_uuid=pipeline_uuid,
+            job_uuid=job_uuid,
+            run_uuid_or_snapshot=run_uuid_or_snapshot,
```
mausworks, line 53: Would it be possible to pass these individually instead? :thinking: 

```suggestion
            run_uuid=run_uuid,
            snapshot_uuid=snapshot_uuid,
```

iannbing: `run_uuid_or_snapshot` doesn't contain anything about `snapshot_uuid`. This variable is either a run UUID or a string literal "snapshot".  🙂

mausworks: Oh! Haha. :facepalm:

It can be left as-is then. :+1:

services/orchest-webserver/app/app/utils.py
```diff
@@ -85,27 +85,28 @@ def get_pipeline_directory(
 
 
 def get_project_directory(
-    project_uuid: str,
+    project_uuid: Optional[str] = None,
     pipeline_uuid: Optional[str] = None,
     job_uuid: Optional[str] = None,
-    run_uuid: Optional[str] = None,
+    run_uuid_or_snapshot: Optional[str] = None,
```
yannickperrenet, line 91: Although rather verbose, it might be good to make the type more explicit:

`Optional[Union[Literal["snapshot"], str]]`

Now it wasn't immediately clear to me that it was "snapshot" instead of a snapshot uuid.

yannickperrenet (approved):
Haven't tested it, but looks good to me. Really appreciate you touching the Python code :muscle:

Issue #679, opened 2022-01-29
Add ability to hide (secret) environment variables


ricklamers:
**Describe the feature you'd love to have in Orchest**
Add ability to hide (secret) environment variables

**Is your feature request related to a problem? Please describe.**
Sometimes you’d rather not display env vars on screen as they can contain sensitive data.

**Additional context**
We should make marking environment variables as sensitive optional or make it the default. If it can be toggled we’d need to understand where to store that state and at what level (e.g. same env var name have different behavior across projects?)

yannickperrenet:
> We should make marking environment variables as sensitive optional or make it the default.

I am in favor of hiding environment variable values by default and add an :eye_speech_bubble: icon to show the value.

PR #1406, opened 2022-11-02
Make environment variables "secret" by default

mausworks:
## Description

This PR makes all environment variables "secret" by default, meaning you have to click a "reveal" button to make them visible. However, if the value of the variable is empty; the input is automatically revealed as this makes it more convenient to enter a value.

https://user-images.githubusercontent.com/8259221/199490076-a3b83b01-f0cc-4cb9-91bc-2cd41d3f38c7.mp4

I decided to create a new component called `EnvVarInput` to more easily encapsulate the "revealed" state and some logic around that.

Fixes: #679

## Checklist

## Tests

- [x] Ensure it works as intended in ...
  - [x] The project settings
  - [x] The pipeline settings
  - [x] The job settings

### DIFF
--- a/services/orchest-webserver/client/src/components/EnvVarInput.tsx
+++ b/services/orchest-webserver/client/src/components/EnvVarInput.tsx
@@ -1,0 +1,82 @@
+import { isValidEnvironmentVariableName } from "@/utils/webserver-utils";
+import { VisibilityOffOutlined, VisibilityOutlined } from "@mui/icons-material";
+import DeleteOutline from "@mui/icons-material/DeleteOutline";
+import { InputAdornment } from "@mui/material";
+import IconButton from "@mui/material/IconButton";
+import Stack from "@mui/material/Stack";
+import TextField from "@mui/material/TextField";
+import Tooltip from "@mui/material/Tooltip";
+import React from "react";
+
+export type EnvVarInputProps = {
+  /** The current environment variable name. */
+  name: string;
+  /** The current environment variable value.  */
+  value: string;
+  /** Disables both inputs. */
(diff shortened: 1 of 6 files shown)


yannickperrenet (approved):
LGTM!

Nitpick: Would it be possible to prevent the browser from thinking I have entered a password when entering an environment variable?
image

mausworks:
> Nitpick: Would it be possible to prevent the browser from thinking I have entered a password when entering an environment variable?

Not a nitpick at all, that's really annoying! :+1: But I thought I fixed that with `autoComplete="off"`, with https://github.com/orchest/orchest/pull/1406/commits/c3354d11b224722573ab06bfaa29756d8a9c2854. Trying again! :+1:

mausworks:
Removing the "save-password"-dialog actually seems pretty hard: https://stackoverflow.com/a/35244490/4697655

Workarounds are possible, but they are not great. Gonna see if I can find one.

yannickperrenet:
If it requires to much research/implementation, then I think it is fine to just merge. Whenever a user presses "never", it won't pop-up again anyways.

iannbing (approved):
LGTM! 💯

mausworks:
> If it requires to much research/implementation, then I think it is fine to just merge. Whenever a user presses "never", it won't pop-up again anyways.

Indeed, can't find a workaround that feels good either. It seems like the right way to solve it is to make our own text-input styling to be masked, either using a custom font which always displays `•` for every character, blur the text or mask it in some other way.

Password inputs seem to be strictly reserved for just that: passwords: It's not "just a display thing".

PR #1409, opened 2022-11-04
Replace "Unable to fetch sessions" alert with unintrusive snackbar message

mausworks:
## Description

This PR removes the "Unable to fetch sessions" error dialog, in favor of a snackbar message that is less intrusive. This snackbar is only displayed after a few initial attempts to fetch sessions have failed and is only displayed in the "Pipelines" and "JupyterLab" views, where it is relevant.

https://user-images.githubusercontent.com/8259221/200020531-8bc768ec-0f76-498c-978d-890954281614.mp4

The logic for which the snackbar is displayed is this:
- Nothing is displayed if it succeeds in getting the session status under five tries (~5 seconds).
- We start displaying the spinner if it fails more than five times
- We display the error message if it fails more than 180 times (~3 minutes).
  - The user likely has to do some intervention here. There is probably something wrong with the instance.
- If any previous message was displayed, and if at any point the connection succeeds, "Session connected!" is displayed for 2.5 seconds.

The aforementioned "error message":
image

## Checklist

- [x] I have manually tested my changes and am happy with the result.

### DIFF
--- a/services/orchest-webserver/client/src/header-bar/HeaderBar.tsx
+++ b/services/orchest-webserver/client/src/header-bar/HeaderBar.tsx
@@ -1,6 +1,5 @@
 import { IconButton } from "@/components/common/IconButton";
 import { useGlobalContext } from "@/contexts/GlobalContext";
-import { useSessionsPoller } from "@/hooks/useSessionsPoller";
 import LogoutIcon from "@mui/icons-material/Logout";
 import AppBar from "@mui/material/AppBar";
 import Stack from "@mui/material/Stack";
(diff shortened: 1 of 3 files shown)


iannbing (approved):
The hard part is to find a good UX solution! Great job! LGTM 👍

yannickperrenet:
https://user-images.githubusercontent.com/26223174/200289819-c3ab4a78-3207-4834-98df-e6b3b8c6d7da.mp4

How about reducing the threshold for when the snackbar is shown so that snackbar won't disappear and reappear again? As long it can't connect to the session then we might as well show the "Trying to connect to session" snackbar imo.

Other than that, really really nice! Huge improvement :fire:

mausworks:
> How about reducing the threshold for when the snackbar is shown so that snackbar won't disappear and reappear again? As long it can't connect to the session then we might as well show the "Trying to connect to session" snackbar imo.

Agree! The reason it happens is that I need to keep track of whether "there ever was an error" (`hadIssue`), since I don't want to show the "session connected" message unless there were. However, I was using this for another purpose as well: to clear the message altogether.

I have changed this now, so there are two different flags and two different timers. The "grace period" to clear the `hadIssue` flag is now also increased to 10s instead.

PR #1411, opened 2022-11-07
Move all Snackbar message to bottom center.

mausworks:
## Description

For #1409 @ncspost mentioned that "the best position for Snackbars is the center for Orchest". So this PR changes that.
The `Snackbar` component will now (by default) be displayed in the bottom center of the app.

This also means that we don't have to hide the Intercom-button in the general case.

## Checklist

- [x] I have manually tested my changes and am happy with the result.


iannbing (approved):
LGTM!

PR #1418, opened 2022-11-09
Fix/enable type checking in CI

iannbing:
## Description

Because of some deprecated design system code, type checking was switched off for quite a while. This PR enable type checking again in CI, to prevent regressions caused by refactoring.

## Checklist


mausworks (approved):
Fantastic! :tada:

LGTM :+1:

Issue #853, opened 2022-04-01
Allow downloading of the root project folder


ricklamers:
**Describe the problem this improvement solves**
Sometimes you want to share a project with someone by downloading all project files and sending them the zip instead of going through VCS systems like `git`.

**Describe the solution you'd like**
Allow download root project folder as zip like we allow any subdirectory to currently be downloaded.

Up for debate: allow users to download the other root folder (`/data`) too. It's usually probably pretty big.

yannickperrenet:
Isn't it good enough that you can easily select all files in the project root and then download them?

ricklamers:
> Isn't it good enough that you can easily select all files in the project root and then download them?

I would argue that's not good enough as it feels quite error prone. Feels like this is a small change that would make it simpler

yannickperrenet:
@ricklamers Do you think it makes sense to update the issue description to include downloading a selection of files as a `zip` archive instead of downloading all the selected step individually?

![image](https://user-images.githubusercontent.com/26223174/184159425-0676d744-dfd4-4598-8b1f-2298d534f2af.png)

Doing this (and pressing download) would actually lead to three separate download requests instead of just one (and putting all files in a zip).

PR #1421, opened 2022-11-10
Support download project & data folder

mausworks:
## Description

This PR adds support for downloading both the project & data folder and also fixes some minor issues with the context menu.

- Menu items no longer disappear when closing
- Fixes "fragments are not valid children of Menu"
- Disable download when read-only (we always download the latest version)
- Consistent disabled items in read-only mode

Right-clicking a root folder:
image

Right-clicking a file:
image

Right-clicking a file in read-only mode:
image

Fixes: #853 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/pipeline-view/file-manager/FileManagerContextMenu.tsx
+++ b/services/orchest-webserver/client/src/pipeline-view/file-manager/FileManagerContextMenu.tsx
@@ -5,7 +5,7 @@ import { useCustomRoute } from "@/hooks/useCustomRoute";
 import { siteMap } from "@/routingConfig";
 import { unpackPath } from "@/utils/file";
 import { Point2D } from "@/utils/geometry";
-import { join } from "@/utils/path";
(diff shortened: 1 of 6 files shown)


services/orchest-webserver/client/src/pipeline-view/file-manager/FileManagerContextMenu.tsx
```diff
@@ -127,92 +117,71 @@ export const FileManagerContextMenu: React.FC<{
...
-              top: metadata?.origin[1] ?? 0,
-              left: metadata?.origin[0] ?? 0,
-            }
-          : undefined
-      }
+      anchorPosition={{ left, top }}
     >
-      {menuItems}
-      {children}
+      {(!hasPath || isRoot) && (
```
mausworks, line 133: There's quite some duplication here, but the MUI `Menu` component doesn't support fragments (it causes a `console.error`). I could do an array (like before), but I think I actually prefer this as it is very flexible & easy to reason about.

iannbing (approved):
Works like a charm 💯  LGTM!

PR #1424, opened 2022-11-11
Fix broken command palette "Page:" shortcuts

mausworks:
## Description

Also removes both `/logs` entries from the sitemap, since those are no longer available.

These are the remaining ones:

image

## Checklist


iannbing (approved):
Nice catch and clean-up!

Issue #1270, opened 2022-09-20
jsonschema - unexptected caching


HeKR96:
**Describe the bug**
When I create a schema file for a step, the initial content of the schema file appears in the UI. However, when I try to change anything, nothing happens. I noticed several layers of the issue:

- Changing the content of the file does not change the form
- When I delete the schema file, the form is still used for the step.
- When I delete the schema and the step file and then create a new step file with the same name, the form is still applied to the step, even though there is not schema file.

**Expected behaviour**
I expect the following behaviour:

- When I change the schema file, the changes should reflect in the form
- When I delete the schema file, the step should have no form and offer me to create a new schema file
- When I delete the schema and the step file and then recreate the step file, the step should have no form and offer me to create a new schema file

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a pipeline
2. Create a new step
3. Put the step in the pipeline
4. Create a schema file
5. Review the generated form
6. Change the schema file 
7. Review the generated form
8. Delete schema file
9. Review the generated form

**Screenshots**
![This is an image](https://i.ibb.co/09kdFhN/orchest-jsonschema.png)

Notice that there is a step form even though there is no schema file.

**Environment**

- OS: openSUSE Leap 15.3
- Python: 3.10.5
- Minikube version: v1.25.2
- Browser: Chromium Version 104.0.5112.101 (openSUSE Build) stable (64-bit), Caching disabled
- Orchest's version: v2022.09.1
- Installation: convenience_install.sh

astrojuanlu:
Hello @HeKR96, thanks for reporting this issue!

I observed a similar problem a few days ago but forgot to open a public issue about it, and I'm quite sure that @iannbing already addressed it. We are about to ship a big refactor (see https://github.com/orchest/orchest/pull/1269) so probably the problem is already fixed there. I will make some tests locally and leave a comment here.

If you want to try it out yourself, you can either follow [our development installation procedure](https://docs.orchest.io/en/latest/development/development_workflow.html), or wait for us to make a new release (very soon!)

astrojuanlu:
I made some experiments with the latest development branch and still observed some weird behavior, so we'll look into this soon.

PR #1436, opened 2022-11-17
fix: the response of FM read shouldn't be cached.

iannbing:
## Description

Currently `/async/file-management/read` allows the client to cache. This is causing issues that client doesn't get prompt updates about the file changes.

Fixes: #1270 

## Checklist


iannbing:
@yannickperrenet  Also applied this to `/download`, in case user changed something and download the same file again. The rest doesn't seem to apply (POST, DELETE, etc.)

yannickperrenet (approved):

PR #1439, opened 2022-11-18
Fix JupyterLab on mobile

mausworks:
## Description

JupyterLab is currently unusable on devices with a screen size smaller than 500 pixels.
This is because our current "did JupyterLab render OK?"-check is a bit naive.

This PR fixes that. :slightly_smiling_face: 

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/jupyter/Jupyter.tsx
+++ b/services/orchest-webserver/client/src/jupyter/Jupyter.tsx
@@ -127,14 +127,11 @@ class Jupyter {
       if (this.iframeHasLoaded) {
         this._unhide();
 
-        if (this.hasRendered() && this.hasRenderingProblem()) {
-          console.log("Reloading iframe because JupyterLab failed to render");
-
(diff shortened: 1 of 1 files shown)


services/orchest-webserver/client/src/jupyter/Jupyter.tsx
```diff
@@ -245,16 +242,17 @@ class Jupyter {
...
+      ?.clientWidth;
 
-    if (!contentPanel) return true;
+    if (!contentWidth || !iframeWidth) return false;
 
-    return (
-      contentPanel.clientWidth !== shellNode.clientWidth ||
-      contentPanel.clientWidth <= 500
-    );
+    const tolerance = 2;
```
mausworks, line 253: I'm a bit paranoid about fractional device pixel ratios (i.e. 1.5x), so I added some tolerance here. :sweat_smile:

ricklamers (approved):
LGTM!

PR #1448, opened 2022-11-24
Jupyter notebook preview export embed images

ricklamers:
## Description

Images referred to in markdown need to be rendered in the HTML preview. This enables the embed config option of `nbconvert` to resolve the path loading issues if images are not embedded.

## Checklist

- [ ] I haven't introduced breaking changes that would disrupt existing jobs, i.e. backwards compatibility is maintained.


ricklamers:
@fruttasecca we need to go to Flask>=2 to meet the Jinja>=3 requirement for the `nbconvert` version we need to support embedding images.

We depend on some Flask 1.X only features for `flask_script` which causes issues when upgrading to Flask>=2.

```
  File "/orchest/services/orchest-webserver/app/migration_manager.py", line 2, in <module>                                             
    from flask_script import Manager                                                                                                   
  File "/usr/local/lib/python3.9/site-packages/flask_script/__init__.py", line 15, in <module>                                         
    from flask._compat import text_type                                                                                                
ModuleNotFoundError: No module named 'flask._compat'
```

Can we refactor `flask_script` / migration invocation to be able to use Flask>=2?

I'm OK with only doing this for `orchest-webserver` for the time being. If you'd prefer to also upgrade `orchest-api` and `auth-server` to Flask>=2 that would also work for me.

mausworks (approved):
LGTM! :)

PR #1452, opened 2022-11-28
New project state management + "more" menu in project drawer

mausworks:
## Description

This PR replaces the current project state management with a zustand store and new hooks related to that.

## Checklist


iannbing (commented):
Nice improvement!

This PR also fixes a bug that, when visiting Project Settings, the Project Selector is correctly loaded. There's one thing can be improved in this PR or in another PR:

- Visit Project settings with an invalid UUID, e.g. `/project-settings?project_uuid=foo`, then the page hangs. I would expect it to redirect back to `/projects`. 

Several regressions that I found:

- Rename project dialog doesn't appear at all.
- Visit a URL with an invalid `project_uuid`, although the app loads a project successfully, but the project_uuid in the URL doesn't reflect on the loaded project. This is gonna be a problem when user share the project URL with others. Or, for example, if user bookmarks the URL.
- In the past, if the said case occurs, there will be an alert "Project not found". It's removed from this PR.
- The Project Selector should be reset when user is viewing `/projects`. 
Screenshot 2022-11-29 at 12 59 42

services/orchest-webserver/client/src/hooks/useFallbackProject.ts
```diff
@@ -0,0 +1,31 @@
...
+        : projects?.[0],
+    [projects, storedUuid]
+  );
+
+  const setFallback = React.useCallback(
+    (uuid: string | undefined) => setStoredUuid(uuid ?? ""),
+    [setStoredUuid]
+  );
+
+  console.log(fallback?.path);
```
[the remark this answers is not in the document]
mausworks: Actually just forgot to push. 😬

mausworks:
I converted the "Project not found alert" into a Snackbar instead since it was just as easy:

image

It's also only shown if the UUID in the URL is valid, so if you enter `project_uuid=foo` you only get redirected to `/projects`.

iannbing:
There's an edge case:
Open two browser tabs, both viewing the same pipeline in the same project A. In the first tab, go to `/projects` and delete project A; go to the other browser tab, you'll see PipelineEditor in a confusing, blank state.

I guess some of the logic should be kicked when the browser tab regains focus.

ncspost:
> I converted the "Project not found alert" into a Snackbar instead since it was just as easy

Could we just have "<Project> not found"?... "it may have been deleted" is redundant.

Please also remove the "okay" button and have it disappear after some time.

ncspost:
Great work @mausworks !! Nitpicks...

- [ ] When clicking the ellipses button, it currently disappears when menu opens - keep it visible
- [ ] Right-click should also access same menu
- [ ] "Rename Project" menu item → "Rename"
- [ ] "Delete Project" menu item → "Delete"
 CleanShot 2022-12-19 at 13 16 22@2x

   - [ ] Upper case "P" in Project
   - [ ] Change field title from "Project name*" to just "Name"
   - [ ] Change error message to "Project name already exists" 
   - [ ] "Save" submit button → "Rename"
   - [ ] Is it possible to not display this message when renaming the Project that already has that name
- [ ] When I navigate to Project settings and then use the menu to navigate between projects, it takes me to each project's respective settings page. Clicking on the Project should always take you to the Pipeline view.
- [ ] When keyboarding through Project drawer, show the menu on the currently focused item
- [ ] Remove quotation marks from project in title 
CleanShot 2022-12-19 at 13 28 42@2x

iannbing (approved):
Nice job on catching all the edge cases while implementing this feature. While the code looks solid, I had quite some trouble reviewing the code. And I was not able to capture all the issues (if any) by reading the code, as all the regressions are caught by manual testing.

mausworks:
I have fixed all issues pointed out by both @iannbing and @ncspost, great catches!

I'm going to leave this open while I work on `feat/new-home` to see if I can catch any additional bugs. But so far it's looking solid. :ok_hand:

services/orchest-webserver/client/src/hooks/useBackgroundTask.ts
```diff
@@ -0,0 +1,41 @@
...
+  status === "FAILURE" || status === "SUCCESS";
+
+/**
+ * Polls a background task with the specified UUID until it finishes.
+ * @param taskUuid The UUID of the task to poll, or `undefined` if a UUID is not yet available.
+ * @returns The background task if it exists, or `undefined`.
+ */
+export const usePollBackgroundTask = (taskUuid: string | undefined) => {
+  const [backgroundTask, setBackgroundTask] = React.useState<BackgroundTask>();
+  const { run } = useAsync<BackgroundTask>();
```
iannbing, line 20: `run` is only used in the dependency array. This can be removed.

services/orchest-webserver/client/src/hooks/useBackgroundTask.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const hasFinished = (status: BackgroundTask["status"]) =>
+  status === "FAILURE" || status === "SUCCESS";
+
+/**
+ * Polls a background task with the specified UUID until it finishes.
+ * @param taskUuid The UUID of the task to poll, or `undefined` if a UUID is not yet available.
+ * @returns The background task if it exists, or `undefined`.
+ */
+export const usePollBackgroundTask = (taskUuid: string | undefined) => {
```
iannbing, line 18: Function name `usePollBackgroundTask` doesn't match the file name `useBackgroundTask`. And overall the naming logic doesn't seem consistent to me.

In some places, `useNoun` was preferred, for example, `useGitImport`, whereas other places `useVerbNoun` is preferred, for example, `useFetchProjects`. This kind of renaming happens almost in every PR.

Could we stick with one convention / preference? Maybe a verb-first convention?
- `useVerbNoun` or `useVerb` or `useVerb-er` when possible, such as `useImportGit`, `usePollBackgroundTask`, `usePoller`, etc.
- choose `useNoun` if it makes more sense, e.g. when it's a compound state, and should be used just like `React.useState`, e.g. `useProjectName`.

Curious to know what you think @mausworks.

services/orchest-webserver/client/src/hooks/useBackgroundTask.ts
mausworks: It should have been used in the handle actually. Nice catch. :+1:

services/orchest-webserver/client/src/hooks/useBackgroundTask.ts
mausworks: I guess it depends on what the hook does internally, not all hooks are made equal. :thinking: 

I try to use my judgment when naming things (hooks included), I think I abide by the rules in your bullet points pretty well, and I think that some confusion may lie in that `GitImport` sounds a bit like a verb, but is, in reality, a noun: _"A git import"_ which "describes the state of the current import process". You can call `start` on this state, sure, _but it isn't the hook itself that has this control_. That's why it's named `useNoun`, and not `useVerbNoun`, since the hook (when mounted) doesn't really do anything, or have any side effects. 

This is in direct contrast to hooks named `useFetchFoo` since these hooks will indeed attempt to fetch "foo" as soon as they are mounted: they both "do something" _and_ have effects, so I use `useVerbNoun` (or `useVerb`) to indicate that. Both `useValidateName(name)` and `useFetchProjects()` are examples of these types of hooks, and their primary purpose is "action".

For hooks that return "some data", but don't (really) have side effects, I go with `useNoun` instead, this is the category for `useGitImport` and `useActiveProject`. Their primary purpose is to provide state.

To elaborate on your "what makes more sense for `useNoun`", I can summarize it as this:

1. If the primary purpose of the hook is to perform some _action_ (pollers, fetchers, timers, and side-effects in general) name it `useVerb` or `useVerbNoun`
2. If the purpose of your hook is to just provide a state, name it `useNoun`

Some hooks may do a little bit of both (e.g. after calling `GitImport.start`, it will have side effects, but not until then). At this point, I think it's fine to use your best judgment. 

What are your thoughts on this @iannbing?

iannbing: I like the take on `useNoun`, such as `useActiveProject`.

I think what confused me was using `useNoun` when this hook has a purpose of action, i.e. import a Git repo. In such cases, I'd suggest keeping it simple: if one could "verb" it, then use verb (it's most readable from language perspective), and then use common sense. The verb here is to make the intention clear. It by no means implies any implementation details.

`useDoSomething` means this hook is to execute an action of "do something", be it a side effect that returns nothing, or returning a `doSomething` function. For example, `useDeleteProject` and `useFetchProject` both feel natural, even though `useDeleteProject` returns a function and `useFetchProject` has a side effect. 

`useGitImport` is a very good example that shows the value of the verb-first convention: is it just a state that holds the `gitImport` value, or a hook that is meant for importing a Git repo? If later, why not just `useImportGit`?

PR #1455, opened 2022-12-02
New in-editor file preview

mausworks:
## Description

This PR adds and changes some behavior regarding the file preview:

https://user-images.githubusercontent.com/8259221/206498687-67455e46-fabd-45a9-aca9-ae6d6cbb84ed.mp4

- The file preview now opens in place of the pipeline viewport directly in the pipeline editor
- Files (that are assigned to steps) are now previewed with a single-click
- Pipelines are now also opened with a single click to make it convenient to switch between files and pipelines
- The file preview has gotten an improved toolbar, that now also displays related pipelines

## Checklist

## Tested

- [x] Opening any file works (both in the data and project directory)
- [x] Files are fetched from the correct project
- [x] Files are fetched from the job snapshot if you're previewing that
- [x] Files are fetched from the job run if you're viewing that


services/orchest-webserver/client/src/hooks/useProjectPipelineJsons.ts
```diff
@@ -0,0 +1,35 @@
+import { usePipelineJsonApi } from "@/api/pipeline-json/usePipelineJsonApi";
+import { useProjectsContext } from "@/contexts/ProjectsContext";
+import React from "react";
+import { useAsync } from "./useAsync";
+
+/** Returns all pipeline definitions within the active project. */
+export const useProjectPipelineJsons = () => {
```
mausworks, line 7: I now realize that this doesn't take snapshots into account (the snapshot API needs to be used for that).

Need to fix that.

services/orchest-webserver/client/src/api/pipeline-json/usePipelineJsonApi.ts
```diff
@@ -0,0 +1,25 @@
+import { PipelineState } from "@/types";
+import { memoizeFor, MemoizePending } from "@/utils/promise";
+import create from "zustand";
+import { FetchPipelineJsonParams, pipelineJsonApi } from "./pipelineJsonApi";
+
+export type DefinitionMap = { [pipelineUuid: string]: PipelineState };
```
mausworks, line 6: Saving the pipelines by UUID proved beneficial in many cases, I think it's a pattern we should use in other stores as well.

Accessing entities by UUID is a very common operation, and you can still iterate using `Object.values`.

services/orchest-webserver/client/src/hooks/useCustomRoute.ts
```diff
@@ -212,3 +212,53 @@ export const useRouteLink = (
...
+ * the previous page are persisted when navigating between routes.
+ */
+export const useNavigate = () => {
+  const { navigateTo, ...currentQuery } = useCustomRoute();
+
+  return React.useCallback(
+    ({
+      route,
+      query = {},
+      sticky = true,
```
mausworks, line 245: I know this helper might seem weird at first, but this `sticky` parameter helps remove _a lot_ of logic around "forwarding" parameters. E.g. you're in the pipeline view, and now you want to go to the `/file-preview` view with the same query parameters, but just with a `step_uuid` appended. 

Previously, we would always have to get them from `useCustomRoute` just to forward them on to `navigateTo`, but with `useNavigate` we no longer have to do that: the current parameters are "sticky".

It also has some convenience changes (such as typing+autocomplete of query-parameters) and I don't want to introduce breaking changes to `navigateTo`.

ricklamers:
Some feedback `5db45832ced5bc581e6a1bb0316f13ea6cd5d49f`:
- Can't use right-click context menu in the file manager (e.g. on file)
- Can't use this context-menu
image
- I can't single click open files that are not in the pipeline
- If the file is not in the pipeline definition the header section related to step navigation should be hidden entirely
image
- Coordinating with Nick about a button or direct way to go back to the "active pipeline" (the one selected in the file preview, or otherwise just the active pipeline of the project).

ncspost:
Great work @mausworks - some small nitpicks:
- When viewing a file and navigating to a pipeline via a pipeline sessions link, the file selector should change to reflect pipeline move. I also get this when trying to navigate to files via the DAG traversing toolbar. I also get this when navigating back to a pipeline after I clicked "edit in JL" on a file.
- I don't see the key down tooltip warning users that it's just a file preview in `ipynb` files
- When I click on a file that is in a pipeline but not connected to other steps, the DAG toolbar should be hidden
- Remove `properties` option in step context menu and reorder menu as follows: CleanShot 2022-12-09 at 10 19 37@2x
- Change animation when hiding/showing connections to something like this: CleanShot 2022-12-09 at 10 26 38
- I don't see the option to switch between pipelines using the same file yet in the DAG toolbar yet
- DAG toolbar should list step name here, not file name: CleanShot 2022-12-09 at 10 30 17@2x
- Show/hide connections toggle choice should be persistent across files
- Could we remove grey bg and border on numbers column and a 8px padding more between numbers column and code:  CleanShot 2022-12-09 at 10 33 43@2x
- There's a short loading lag between screens... maybe we should use a spinner orskeleton loader? CleanShot 2022-12-09 at 10 35 26
-

PR #1477, opened 2022-12-15
Fix double-click issues in the file browser

mausworks:
## Description

It currently isn't possible to open files by double-clicking in the file browser. In short, this is because our `onSelect` and `onDoubleClick` handlers are "fighting each other", since they both cause a navigation.

This PR changes so that the "click to preview" reacts to `onClick` instead of `onNodeSelect`, and sets up a more lenient double-click handler alongside the default double-click handler to capture more clicks and provide more granular control over click behavior.

## Checklist


ricklamers:
This looks like a reasonable approach to me. Select/click/double click looks like a complex interacting set of events. In the future keyboard actions might be added, in which case replacing select with click starts to become problematic. But that's something we can address when we add keyboard file manager navigation support :+1:

PR #1479, opened 2022-12-16
Fix: losing environment build status when switching browser tabs

iannbing:
## Description

In /environments, when the browser tab regains focus, `fetchAll` is called to keep the environment list up-to-date, but `fetchAll` also washes away the current UI state (i.e. `lastBuild`), and thus the build state is then only re-stored in the next round of polling. This PR persists `lastBuild` so the UI state is persisted.

## Checklist

### DIFF
--- a/services/orchest-webserver/client/src/utils/record.ts
+++ b/services/orchest-webserver/client/src/utils/record.ts
@@ -61,3 +61,7 @@ export const equalsShallow = <T extends AnyRecord, A extends T>(
   actual: A,
   props = propsOf(expected)
 ) => props.every((prop) => actual[prop] === expected[prop]);
+
+export const toMap = <T>(arr: T[], key = "uuid") => {
+  return new Map<typeof key, T>(arr.map((env) => [env[key], env]));
+};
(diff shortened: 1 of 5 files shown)


services/orchest-webserver/client/src/utils/record.ts
```diff
@@ -61,3 +61,7 @@ export const equalsShallow = <T extends AnyRecord, A extends T>(
   actual: A,
   props = propsOf(expected)
 ) => props.every((prop) => actual[prop] === expected[prop]);
+
+export const toMap = <T>(arr: T[], key = "uuid") => {
+  return new Map<typeof key, T>(arr.map((env) => [env[key], env]));
```
mausworks, line 66: I actually need this exact abstraction in my branch as well, but I need it as a record.

Could we make it `mapRecord` instead and have it return `Record<typeof keyof T, T>`, i.e.
`Map` doesn't work in my use-case since it has to be JSON serializable.

```suggestion
  return Object.fromEntries(arr.map((env) => [env[key], env])) as Record<typeof keyof T, T>;
```

mausworks (approved):
LGTM, but see the comment about shared/needed abstraction! :)

services/orchest-webserver/client/src/utils/record.ts
iannbing: Changed 👍

mausworks (approved):

PR #1485, opened 2022-12-21
New home view

mausworks:
## Description

This removes the `/projects` view in favor of a new more flexible "home" view, where other things than Projects can be displayed, such as all Job and Interactive Runs.

## Features

- [x] Remove the `/projects` page and fix any inbound links
- [x] Redirect `/projects` to `/?tab=projects`
- [x] Move projects to the new home page and align components with new design
- [x] Update the Project selector menu with the new design
- [x] Implement interactive job runs under "all runs"
- [x] Implement all job runs under "all runs"
  - [x] Make the back-end support querying of multiple (or all) projects
  - [x] Implement filtering, pagination, etc... in the new front-end based on the new API.

## Checklist


services/orchest-webserver/client/src/api/jobs/jobsApi.ts
```diff
@@ -65,64 +68,57 @@ const put = async ({ uuid, schedule, ...changes }: JobChangesData) => {
...
+  fetcher<JobData>(join(BASE_URL, "duplicate"), {
     method: "POST",
     headers: HEADER.JSON,
     body: JSON.stringify({ job_uuid: jobUuid }),
   });
 
 const triggerScheduledRuns = (jobUuid: string) =>
   fetcher<{ next_scheduled_time: string }>(
-    `/catch/api-proxy/api/jobs/${jobUuid}/runs/trigger`,
+    join(BASE_URL, jobUuid, "runs/trigger"),
```
iannbing, line 102: [nit] any reason use both `join` and a `slash` literal?

services/orchest-webserver/client/src/api/jobs/useJobsApi.ts
```diff
@@ -1,217 +1,39 @@
...
-            return job;
-          }),
-        };
-      });
-    },
-    fetchAll: async (projectUuid) => {
-      set({ projectUuid });
-      const jobs = await jobsApi.fetchAll(projectUuid);
+    jobs: undefined,
+    fetchOne: memoizeFor(500, async (jobUuid) => {
```
iannbing, line 20: Could you parameterize  this magic number or at least document it? 
There are 200, 500, 900, 1000 across the whole codebase. 
And also, was it used as a precaution (under the assumption that the same fetcher hook can be freely used in any component)?
If a fetcher hook is carefully placed only once in a view, do I still have to apply this precaution regardless?

services/orchest-webserver/client/src/hooks/useHydrate.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Note: Ensure that the hydrate function is properly memoized.
+ * @param hydrate A function that loads data and updates a store.
+ */
+export const useHydrate = <H extends () => Promise<unknown>>(
+  hydrate: H
+): HydrationState => {
+  const { run, error, status } = useAsync();
+
+  const reload = React.useCallback(async () => {
+    await run(hydrate().catch());
```
iannbing, line 26: If the promise returned by `hydrate` function is caught like this, `useAsync` will never capture the error, and thus `error` will always be `undefined`.

services/orchest-webserver/client/src/hooks/useHydrate.ts
```diff
@@ -0,0 +1,41 @@
...
+  isLoading: boolean;
+  /** True if the (re)hydration has completed. */
+  isLoaded: boolean;
+  /** Calls the hydrate function and tracks its promise. */
+  reload: () => Promise<void>;
+};
+
+/**
+ * Calls the hydrate function once and keeps track of its async state.
+ * Note: Ensure that the hydrate function is properly memoized.
```
iannbing, line 17: As I asked in another comment, do you suggest memoize `fetch` at all times? (By reading this JSDoc, it's an _yes_?). If it's the case, why not apply memoization in this hook?

ricklamers:
image
I'm hitting a snag with `orchest-api` not coming up. Also doesn't produce log. Retried as clean installation, no luck. @fruttasecca did you experience this at some point?

ricklamers:
Creating a new project in the project drawer gives me:

image

image

The project does create.

ricklamers:
image

Seems like there is no polling, because the view doesn't update for me. Should it be auto-updating? If not, let's add a refresh button?

ricklamers:
Is showing 5 per page the default? Seems a bit low. Maybe go with 10?

ricklamers:
Overall: super impressive work! Big kudos honestly @mausworks

ricklamers:
For performance, do we limit runs to the last "N" days where 30 feels sane? Otherwise I'm afraid it will choke on instances with a lot of activity.

ricklamers:
Just a screenshot because it's so beautiful:

image

services/orchest-webserver/client/src/hooks/useToggle.ts
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+export const useToggle = (initialState = false) => {
+  const [toggled, setToggled] = React.useState(initialState);
+
+  const toggle = React.useCallback((newState?: boolean) => {
+    if (newState) setToggled(newState);
```
iannbing, line 7: I think `newState` should be wrapped by `hasValue()` here.

iannbing (commented):
New home looks amazing and very well implemented 💯. Quite a big PR to review even per commit. Apart from the issues I commented, there are definitely some work to be done regarding the inconsistency between the old contexts and the new zustand stores.

services/orchest-webserver/client/src/hooks/useHydrate.ts
iannbing: I think you probably meant memoizing the function with `React.useCallback`, but readers could confuse it with `memoizeFor`. Could you be more explicit in the comments?

services/orchest-webserver/client/src/api/pipeline-runs/usePipelineRunsApi.ts
```diff
@@ -0,0 +1,35 @@
...
+  cancel: MemoizePending<(runUuid: string) => Promise<void>>;
+  /** Fetches all running (STARTED/PENDING) pipeline runs for all projects. */
+  fetchRunning: MemoizePending<() => Promise<PipelineRun[]>>;
+};
+
+/** State container for interactive pipeline runs. */
+export const usePipelineRunsApi = create<PipelineRunsApi>((set) => {
+  return {
+    running: [],
+    cancel: memoizeFor(500, async (runUuid: string) => {
```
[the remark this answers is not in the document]
mausworks: It shouldn't, nice catch!

services/orchest-webserver/client/src/hooks/useHydrate.ts
mausworks: Great catch! Fixed!

services/orchest-webserver/client/src/hooks/useHydrate.ts
mausworks: Nice catch! Clarified in the `Note:` a bit in 7ff382caf9f4ba172e0fe483bf50aaf9aa308f60

Yep, upstream consumers of `reload` expect it to only change because it fetches a new item (e.g. with new parameters).
So the recommendation I'd say is to use `useCallback` or another "stable" fetcher function directly (e.g. from a zustand store).

mausworks:
> Seems like there is no polling, because the view doesn't update for me

@ricklamers haha, no you're right. :facepalm: :joy: 

Fixed: https://github.com/orchest/orchest/pull/1485/commits/f8c2263ead5b2922e29cec3bcde22bd7cea96325

fruttasecca:
> image I'm hitting a snag with `orchest-api` not coming up. Also doesn't produce log. Retried as clean installation, no luck. @fruttasecca did you experience this at some point?
> 
> Update: I expect this to be a migration conflict related to Alembic. Was able to resolve when I made sure I started without state and installing directly from dev only after building all container images.

I believe the cause is/are branching migrations between what's on `dev` and this branch, easily fixed/verified

mausworks:
> Is showing 5 per page the default? Seems a bit low. Maybe go with 10?

@ricklamers updated to 8 now! Do you still think 10?

mausworks:
> Creating a new project in the project drawer gives me:

@ricklamers this is also fixed now :+1:

mausworks:
@iannbing 
> Quite a big PR to review even per commit. 

Yeah, I'm sorry. The PR touches on pretty much every entity in some way (examples, projects, jobs, pipelines, job runs), making it naturally quite large. :grimacing:

> there are definitely some work to be done regarding the inconsistency between the old contexts and the new zustand stores.

I have now removed `projectUuid`, `project`, `pipeline`, and `pipelines` from the project context entirely. `useActiveProject`, `useActivePipeline`, `useProjectPipelines` can be used instead. `projectUuid` is available in `useCurrentQuery`.

services/orchest-webserver/client/src/hooks/useActivePipeline.ts
```diff
@@ -0,0 +1,14 @@
...
+import React from "react";
+import { useCurrentQuery } from "./useCustomRoute";
+
+export const useActivePipeline = () => {
+  const { pipelineUuid, projectUuid } = useCurrentQuery();
+  const find = usePipelinesApi((api) => api.find);
+  const pipeline = React.useMemo(() => {
+    if (!projectUuid || !pipelineUuid) return undefined;
+    else return find(projectUuid, pipelineUuid);
+  }, [find, projectUuid, pipelineUuid]);
```
iannbing, line 11: [nit] the code block in `React.useMemo` could be moved into the zustand selector function. If it's intentional, I'm curious to know the reason 🙂.

```
const pipeline = usePipelinesApi((api) => {
    if (!projectUuid || !pipelineUuid) return undefined;
    else return api.find(projectUuid, pipelineUuid);
});
```

services/orchest-webserver/client/src/hooks/useProjectPipelines.ts
```diff
@@ -0,0 +1,12 @@
...
+import React from "react";
+
+export const useProjectPipelines = (projectUuid: string | undefined) => {
+  const pipelines = usePipelinesApi((api) => api.pipelines);
+
+  return React.useMemo(() => {
+    return projectUuid
+      ? pipelines?.filter(({ project_uuid }) => project_uuid === projectUuid)
+      : [];
+  }, [pipelines, projectUuid]);
```
iannbing, line 11: I see that you tend to use `React.useMemo` separately instead of putting the logic in the selector function. I think `zustand` already takes care of memoization, and it seems to work very well. Curious to know the reason behind your implementation 🙂.

services/orchest-webserver/client/src/hooks/useActivePipeline.ts
mausworks: There is a very large difference actually. While zustand may take care of the memorization, it will still run the selector on every render.

Using the following code:

```ts
export const useActivePipeline = () => {
  const { pipelineUuid, projectUuid } = useCurrentQuery();
  const find = usePipelinesApi((api) => api.find);
  usePipelinesApi((api) => {
    console.count("Selector!");

    if (!projectUuid || !pipelineUuid) return undefined;
    else return api.find(projectUuid, pipelineUuid);
  });
  const pipeline = React.useMemo(() => {
    console.count("Memo!");

    if (!projectUuid || !pipelineUuid) return undefined;
    else return find(projectUuid, pipelineUuid);
  }, [find, projectUuid, pipelineUuid]);

  return pipeline;
};
```

Shows "Selector!" being executed 300+ times, while "Memo!" is only executed 10-15 times. (`useActivePipeline` is used in `FileTreeRow`, that's why it's executed that many times). But maybe memorization isn't needed here at all, the pipeline array is probably pretty small. :shrug:

iannbing: Nice explanation! Yet, this brings up another question: `useActiveProject` and `useActivePipeline` both find the active entity per usage. And because they are used everywhere, i.e. the same operation is executed N times (O(n)). Following your example, the `find` operation for setting `activePipeline` will only be done once, rather than N times, plus this also saves the memoization in this `useActivePipeline` hook. The old implementation was done in such a manner (e.g. `SET_PIPELINE` action). Any reason you opt for this `useActiveSomethin` approach?

iannbing: I don't have a strong option on this. Just curious about the thinking behind the implementation. 10-15 times seems manageable, considering the average amount of projects and pipelines of our users.

mausworks: So, I have changed most stores to use a `Record<string, EntityData>` type internally. This means that most `useActiveEntity` hooks are simply just: `useEntityApi((api) => api.entities?.[entityUuid]);`. No memoization is needed, and the lookup is always O(1). In truth, I just did this for convenience because looking up a job name from a job run is much easier when jobs are indexed by their UUID.

It is actually the `usePipelinesApi` that is the odd one out. :sweat_smile: I was trying to decide between using a composite id of `projectUuid` + `pipelineUuid`, or an array. In the end, I went with an Array because it was the most straightforward approach that required the least prior knowledge. I think that the overhead of the `React.useMemo` + dependency array is OK in this case (simply because it is an edge case that you can't index pipelines by their UUIDs).

I think finding a good place to call `useFetchSomeEntity()` can be hard enough (especially if you want to handle empty, loading, and error states), and introducing something such as `useSetActiveEntity()` creates a similar inconvenience. So I think it's better to spend a couple of CPU cycles (commonly just on the first render) for the convenience of it working without any setup. :ok_hand:

services/orchest-webserver/client/src/api/jobs/jobsApi.ts
mausworks: No reason, just copy and paste probably. Fixed now :)

services/orchest-webserver/client/src/api/jobs/useJobsApi.ts
mausworks: You are right, this looks pretty smelly. I'll see what I can do. :+1:

mausworks: I cleaned it up. Thanks for the feedback. :tada:

ricklamers:
When I load a project the main pipeline is not loaded:

image

I just created a new project, added a step to the pipeline and clicked on JupyterLab.

E.g. clicking the pipeline in the pipeline sessions element in the lower left corner doesn't open the pipeline either.

ricklamers:
LGTM! @mausworks for pres

services/orchest-webserver/client/src/hooks/useActivePipeline.ts
iannbing: Thanks for the explanation. 

> I think finding a good place to call useFetchSomeEntity() can be hard enough (especially if you want to handle empty, loading, and error states), and introducing something such as useSetActiveEntity() creates a similar inconvenience.

Indeed. Yet, from a troubleshooting friendliness perspective, is it better? (e.g. `useFetchProjects` appears in 10 files). It took me 30 minutes to find the side effect in `useActiveProject`. Not sure if it's a good thing moving toward this direction: convenient to develop; hard to debug. You've mentioned: "worry-free" many times, but it is quite opposite to me, as I'm not sure if I could spot issues with common sense anymore.

So, if I understand correctly, `useActiveSomething` usually can mean:
- it retrieves something from the store
- fetch data if the store is empty (under the protection of `memoizeFor` and `useAsync`).
- some side effects, e.g. redirect when the asked item is unavailable.

I'll keep it in mind. 💯

iannbing (approved):

services/orchest-webserver/client/src/hooks/useActivePipeline.ts
mausworks: I'm starting to lean more into trying to fetch things at a "high level" now rather than "as needed". The reason is mostly to be able to handle error & empty states, so I too think it's best to do it that way to do it! Though I still think it's good to have a safety such as `memoize` just in case you render two components that fetch something at the same time.

In general: `useActiveFoo` will not fetch and just return the object itself (if loaded) (although I'm sure there's some exception) and `useFetch(Active)Foo` will fetch something and return it (together with fetch status).

PR #1491, opened 2022-12-29
Remove background tasks from orchest webserver

fruttasecca:
Removes the concept of a background task from the `orchest-webserver` and thus the client, in favor of having the `orchest-webserver` use the dedicated git import endpoints from the `orchest-api`.


fruttasecca:
Not exactly sure if `services/orchest-webserver/client/src/api/git-imports/useGitImportsApi.ts` is needed

PR #1493, opened 2023-01-03
Support Git configurations

iannbing:
## Description

This PR adds the FE implementation to support Git configurations.

## Checklist


fruttasecca:
The GIT & ssh setting tab is appearing while I'm not in auth mode

fruttasecca:
There doesn't seem a way to save the git confguration, I think I'm missing something, see screenshoot
image

fruttasecca:
The git import panel needs to support ssh git urls
image

fruttasecca:
Something I forgot to mention (my bad) when handing over is that the git import GET endpoint is a bit more detailed when it comes to error state(s), for example:

```
{
  "uuid": "e9c2917b-1f26-4568-b850-ad741b44927d",
  "url": "https://github.com/orchest/cloud-dashboard-backend",
  "requested_name": "cloud-dashboard-backend-714d2b60-8388-46c7-8e2d-6383464656db",
  "status": "FAILURE",
  "project_uuid": null,
  "result": {
    "error": "GitCloneFailed"
  }
}
```

The content of result -> error can be used to be more explicative to the user. Here are the possible values (from the swagger docs): `ProjectNotDiscoveredByWebServer`, `GitCloneFailed`, `NoAccessRightsOrRepoDoesNotExists`, `ProjectWithSameNameExists`. Imo the FE can show the same generic message for `ProjectNotDiscoveredByWebServer`, `GitCloneFailed`, since `ProjectNotDiscoveredByWebServer` is a bit of an internal detail.

iannbing:
> The GIT & ssh setting tab is appearing while I'm not in auth mode

Yeah, I noticed. The fact is that, if user logged in while still in auth mode, user is actually still logged in even auth mode is off, (`auth_user_uuid` is still present in the cookie, this is the single source of truth for webserver client to know if user is logged in). To achieve the desired behavior, I think auth-server needs to log out all users when restarting. @fruttasecca What do you think?

Counting on the value of `AUTH_ENABLED` could lead to a very confusing state on the FE side because this setting requires restarting to take effect (e.g. let's say, setting `AUTH_ENABLED` as false and not yet clicking "restart", should FE show the UI or not?).

> There doesn't seem a way to save the git confguration, I think I'm missing something, see screenshoot

In the new design, it's auto-saved. There are some inconsistency between the old settings views and this new view.

> The git import panel needs to support ssh git urls

Will do. Wasn't aware of this 🙂

fruttasecca:
>Counting on the value of AUTH_ENABLED could lead to a very confusing state on the FE side because this setting requires restarting to take effect (e.g. let's say, setting AUTH_ENABLED as false and not yet clicking "restart", should FE show the UI or not?).

Agree

>Yeah, I noticed. The fact is that, if user logged in while still in auth mode, user is actually still logged in even auth mode is off, (auth_user_uuid is still present in the cookie, this is the single source of truth for webserver client to know if user is logged in). To achieve the desired behavior, I think auth-server needs to log out all users when restarting. @fruttasecca What do you think?

I see, that's indeed a bit annoying, I'll look for some low-hanging fruit in the auth server

services/orchest-webserver/client/src/hooks/useOpenDialog.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+/** A simple hook that is useful when maintaining the `isOpen` state of a dialog. */
+export const useOpenDialog = () => {
```
mausworks, line 4: You can use `useToggle` here instead. It does pretty much the same thing. :+1:

services/orchest-webserver/client/src/hooks/useTextField.ts
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+type StringPredicate = (value: string) => boolean;
+
+export const useTextField = (predicate: StringPredicate = () => true) => {
```
mausworks, line 5: I think a lot of functionality can be covered by [react-hook-form instead: dirty checking, validation, etc](https://react-hook-form.com/api/useform/formstate) …
Even if the form only has one input, I think it makes sense to use this instead.

What do you think?

services/orchest-webserver/client/src/settings-view/SettingsView.tsx
```diff
@@ -130,230 +110,125 @@ const SettingsView: React.FC = () => {
...
-          <div className="column">
-            <p>Version information.</p>
-          </div>
-          <div className="column">
-            {orchestVersion ? (
-              <p>{orchestVersion}</p>
-            ) : (
-              <LinearProgress className="push-down" />
+    <SettingsViewLayout header={<Typography variant="h5">General</Typography>}>
+      <Stack direction="column" spacing={3}>
```
mausworks, line 115: Awesome work cleaning this up! :tada:

fruttasecca:
@iannbing I've made the required changes in https://github.com/orchest/orchest/pull/1493/commits/72f153abdecb1f9a542af9e21a7b34310076374c, it's a bit of a hack that I'm not sure can be avoided given how the auth server and k8s currently interact if we want cookie clearing to be solely the responsibility of the auth server. I think for now this will suffice but I'm definitively open to other ideas

services/orchest-webserver/client/src/projects-view/ImportDialog.tsx
```diff
@@ -264,7 +256,8 @@ export const ImportDialog = ({
     setNewProjectUuid(undefined);
     setTempProjectName(undefined);
     resetUploader();
-  }, [setImportUrl, resetUploader]);
+    gitImport.reset();
```
iannbing, line 259: @mausworks FYI, gitImport needs to be reset after complete. Otherwise user could get stuck if:
1.  import a project, click "CONTINUE BROWSING" after complete
2. try to import another project without navigating away.
3. the import dialog stays in "Import complete" state, and user probably has no idea to get away with this.

services/orchest-webserver/client/src/hooks/useOpenDialog.tsx
iannbing: @mausworks I couldn't find this hook in the codebase. Could you check if it's the right name?

services/orchest-webserver/client/src/hooks/useTextField.ts
iannbing: It might be useful to apply this library. But I probably wouldn't apply it in this PR at this point.

iannbing: Actually, in most of the new design, the changes of the forms are auto-saved. And this means that "press enter to validate the whole form and submit form if no error" behavior occurs in very few occasions. In most cases, validation occurs per fields in order to facilitate auto-saving.
I guess that's why I didn't consider libraries like formik or react-hook-form in the first place. We could, however, still apply this library to streamline these few forms. Just that the ROI doesn't seem as much as intended.

services/orchest-webserver/client/src/hooks/useOpenDialog.tsx
mausworks: Oh damn. 🤦 I thought I added it in `feat/system-status`, but I added it in `feat/new-home` which is not yet merged.

mausworks (approved):
LGTM 🚀

ricklamers:
I think the git config Username and Email are not yet being updated on the settings page.

PR #1504, opened 2023-01-10
Auto save user config

iannbing:
## Description

Currently, the user config requires the user to click "SAVE" to save the config.
This PR enables auto-save to streamline the UX.

## Checklist


services/orchest-webserver/client/src/utils/record.ts
```diff
@@ -11,21 +11,24 @@ const entryHasValue: EntryPredicate = ([, value]) => hasValue(value);
  * Returns the properties of the record T as a typed array.
  * Only call this when you are sure that T does not include any extra properties.
  */
-export const propsOf = <T extends AnyRecord>(record: T): PropsOf<T> =>
+const propsOf = <T extends AnyRecord>(record: T): PropsOf<T> =>
   Object.keys(record);
 
 /** Returns a shallow copy of the record T with the properties P excluded. */
 export const omit = <T extends AnyRecord, P extends PropsOf<T>>(
```
mausworks, line 18: I understand the intent here, but this will realistically make the function slower in the common case. This is because memory allocation (creating the `keyRecord`) is "quite expensive" (and creates garbage that will have to be cleaned up later), and `Array.includes` is actually insanely fast (especially since `keys` is often tiny, with 1-5 elements).

For small arrays, it is faster to do a linear search than to create a hash set (or similar) and perform lookups on it. I was curious about this and created a [small benchmark](https://jsben.ch/WH1lI) to verify it. For large objects (200 properties) the linear search through keys can sometimes be slower, but in the common case (5-20 properties on the object, and 1-5 keys to omit), linear search outperforms the lookup variant by around 30-40% on my machine.

Counterintuitively `O(n^2)` can sometimes be faster than `O(n)`, because not all `n`'s are created equal. Some `n`'s take longer time than others, and big-O notation measures worst-case performance, which is seldom realistic. I watched [this video](https://www.youtube.com/watch?v=U16RnpV48KQ) yesterday where he mentions the "linear search" vs "lookup" problem, and why it is sometimes (counterintuitively) faster to do the former. I think it was a great watch in overall! :+1:

With all this in mind, I think it makes sense to revert to the old solution. We often use omit with 1-2 keys, and our objects seldom have more than 10 properties.

services/orchest-webserver/client/src/utils/__tests__/record.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { omit } from "../record";
+
+describe("omit", () => {
+  it("removes the specified properties without mutating the original object.", () => {
```
mausworks, line 4: Awesome! :tada:

services/orchest-webserver/client/src/settings-view/hooks/useOrchestStatus.tsx
```diff
@@ -1,16 +1,82 @@
...
+          console.error(error);
+          resolve(false);
+          setAlert("Error", "Could not trigger restart. Please try again.");
+          return false;
+        }
+      }
+    );
+  }, [resetRequireRestart, setAlert, setConfirm, setStatus]);
+
+  return [status, restartOrchest] as const;
```
mausworks, line 59: When using this hook it's not immediately obvious what the second item does here, I would expect it to be a `setState` function, but in this case, it very much isn't.

mausworks (changes requested):
Great work and fantastic clean up! :tada: 

I left some comments on the `omit` optimization, but otherwise, it looks great.

PR #1511, opened 2023-01-20
Release - v2023.01.8

mausworks:


mausworks:
Looks like two integration tests are failing (relating to accounts). I don't see any reason for why they should stop the release, but I will do a follow-up PR with a fix.

